PR #42, opened 2020-07-27
feat: el-divider

Caaalabash:
el-divider


HerringtonDarkholme (changes requested):
Please change the test case to vue-test-utils' method

jw-foss (approved):

PR #140, opened 2020-08-17
feat(popper): popper component

### LINKED ISSUE #141: ### Popper
Fundamental component `Popper` for `Tooltip` and `Select`, etc.

jw-foss:
- [x] Component Migration

This is a pre-quest component for e.g. `Tooltip` `Selector`

### DIFF
--- a/packages/popper/src/index.vue
+++ b/packages/popper/src/index.vue
@@ -1,0 +1,410 @@
+<template>
+  <transition :name="transition">
+    <div
+      v-show="visible"
+      :id="tooltipId"
+      ref="popperRef"
+      role="tooltip"
+      :aria-hidden="visible ? 'false' : 'true'"
+      :class="['el-tooltip__popper', 'is-' + effect, popperClass]"
+      @mouseenter="_show"
+      @mouseleave="_hide"
+    >
+      <!-- mark this parent container un-indexable to disable it from being indexed -->
+      <slot>
+        <div>{{ content }}</div>
+      </slot>
+
+      <div
+        v-if="showArrow"
+        ref="arrowRef"
+        class="el-tooltip__arrow"
+        data-popper-arrow
+      ></div>
+    </div>
+  </transition>
+</template>
+
+<script lang="ts">
+import {
+  computed,
+  defineComponent,
+  ref,
+  onBeforeUnmount,
+  onMounted,
+  onUnmounted,
+  watch,
+} from 'vue'
+import { isArray } from '@vue/shared'
+import { debounce } from 'lodash-es'
+import { createPopper } from '@popperjs/core'
+
+import { generateId } from '@element-plus/utils/util'
+import { on, off } from '@element-plus/utils/dom'
+
+import useModifer from './useModifier'
+import { PropType, Ref } from 'vue'
+import type { Placement, Instance as PopperInstance, PositioningStrategy } from '@popperjs/core'
+
+type Effect = 'dark' | 'light'
+
+const stop = e => e.stopPropagation()
+
+export default defineComponent({
+  name: 'ElPopper',
+  props: {
+    arrowOffset: {
+      type: Number,
+      default: 15,
+    },
+    appendToBody: {
+      type: Boolean,
+      default: true,
+    },
+    boundariesPadding: {
+      type: Number,
+      default: 0,
+    },
+    content: {
+      type: String,
+      default: '',
+    },
+    closeDelay: {
+      type: Number,
+      default: 200,
+    },
+    cutoff: {
+      type: Boolean,
+      default: false,
+    },
+    disabled: {
+      type: Boolean,
+      defualt: false,
+    },
+    effect: {
+      type: String as PropType<Effect>,
+      default: 'dark' as Effect,
+    },
+    enterable: {
+      type: Boolean,
+      default: true,
+    },
+    flip: {
+      type: Boolean,
+      default: true,
+    },
+    hideAfter: {
+      type: Number,
+      default: 0,
+    },
+    manualMode: {
+      type: Boolean,
+      default: false,
+    },
+    openDelay: {
+      type: Number,
+      default: 0,
+    },
+    offset: {
+      type: [Number, Array] as PropType<[number, number] | number>,
+      default: [0, 12] as [number, number],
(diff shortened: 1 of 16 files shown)


packages/popper/src/useModifier.ts
```diff
@@ -0,0 +1,44 @@
...
+  boundariesPadding: number
+  cutoff: boolean
+  offset: [number, number] | number
+  showArrow: boolean
+  fallbackOptions: Record<string, unknown>
+}
+
+export default (options: IBuildModifierOptions) => {
+  const offset = isArray(options.offset) ? options.offset.slice(0, 2) : [0, options.offset]
+  const modifiers = [
```
hanx316, line 15: can we make this configs outside the function or it will be initialized every time when useModifier is called

packages/popper/src/index.vue
```diff
@@ -0,0 +1,409 @@
...
+      referenceElement.setAttribute('tabindex', props.tabIndex)
+      on(referenceElement, 'mouseenter', _show)
+      on(referenceElement, 'mouseleave', _hide)
+      on(referenceElement, 'focus', handleFocus)
+      on(referenceElement, 'blur', handleBlur)
+      on(popperRef.value, 'click', stop)
+    }
+
+    watch(() => props.referrer, (val, prev) => {
+      if (val !== null) {
```
hanx316, line 274: `if (val === null) return` maybe better

packages/popper/src/index.vue
```diff
@@ -0,0 +1,409 @@
...
+      clearTimer(timeout)
+      if (props.openDelay === 0) {
+        show.value = true
+      } else {
+        timeout.value = setTimeout(() => {
+          show.value = true
+        }, props.openDelay)
+      }
+
+      if (props.hideAfter > 0) {
```
hanx316, line 194: Here is a side case, if we both have openDelay AND hideAfter, or even the value of hideAfter < openDelay, looks like there would be a trouble.

JeremyWuuuuu: Correct, I'll fix that

packages/popper/src/index.vue
```diff
@@ -0,0 +1,409 @@
...
+      }
+      show.value = false
+      if (props.disabled) {
+        doDestroy(true)
+      }
+    }
+
+    const debouncedClose = debounce(() => {
+      closePopper()
+    }, props.closeDelay)
```
zazzaz, line 215: const debouncedClose = debounce(closePopper, props.closeDelay)

JeremyWuuuuu: Indeed😂

packages/popper/src/index.vue
```diff
@@ -0,0 +1,466 @@
...
+      trigger.value = null
+    }
+
+    const showPopper = () => {
+      if (!exceptionState.value || props.manualMode || props.disabled) return
+      clearTimer(timeout)
+      if (props.openDelay === 0) {
+        show.value = true
+        if (props.hideAfter > 0) {
+          timeoutPending.value = setTimeout(() => {
```
hanx316, line 232: it is better to extract this part of code

packages/popper/src/index.vue
```diff
@@ -0,0 +1,466 @@
...
+      off(_trigger, 'mouseleave', _hide)
+      off(_trigger, 'focus', handleFocus)
+      off(_trigger, 'blur', handleBlur)
+      trigger.value = null
+    }
+
+    const showPopper = () => {
+      if (!exceptionState.value || props.manualMode || props.disabled) return
+      clearTimer(timeout)
+      if (props.openDelay === 0) {
```
hanx316, line 229: openDelay is more like the pair of closeDelay rather than hideAfter. showAfter or openAfter maybe better in this context

packages/popper/src/index.vue
```diff
@@ -0,0 +1,466 @@
...
+  Placement,
+  Instance as PopperInstance,
+  PositioningStrategy,
+} from '@popperjs/core'
+import useModifier from './useModifier'
+
+type Effect = 'dark' | 'light';
+
+const stop = (e) => e.stopPropagation()
+const throwError = (e: string) => {
```
hanx316, line 60: can we put this function into utils?

JeremyWuuuuu: Maybe we can make it a custom error for ElementPlus

packages/popper/src/index.vue
```diff
@@ -0,0 +1,466 @@
...
+    const popperInstance = ref(null as PopperInstance)
+    const exceptionState = ref(false)
+    const timeout = ref(null as NodeJS.Timeout)
+    const timeoutPending = ref(null as NodeJS.Timeout)
+    const show = ref(false)
+    const popperId = ref(`el-tooltip-${generateId()}`)
+    const trigger = ref(null as Nullable<HTMLElement>)
+
+    const visible = computed(() => {
+      return props.manualMode ? props.value : !props.disabled && show.value
```
zouhangwithsweet, line 177: why not `modelValue`?

packages/popper/src/index.vue
```diff
@@ -0,0 +1,466 @@
...
+  setup(props, { slots }) {
+    const popperRef = ref(null as Nullable<HTMLElement>)
+    const arrowRef = ref(null as Nullable<HTMLElement>)
+    const popperInstance = ref(null as PopperInstance)
+    const exceptionState = ref(false)
+    const timeout = ref(null as NodeJS.Timeout)
+    const timeoutPending = ref(null as NodeJS.Timeout)
+    const show = ref(false)
+    const popperId = ref(`el-tooltip-${generateId()}`)
+    const trigger = ref(null as Nullable<HTMLElement>)
```
zouhangwithsweet, line 174: Maybe use `ref<Nullable<HTMLElement>>()`  is better

packages/popper/src/index.vue
JeremyWuuuuu: Actually, this one is a underlaying component for components like `tooltip` `popover`, so I don't think that we need to implement the exact same API as `tooltip` did

JeremyWuuuuu:
@hanx316 @zouhangwithsweet All done

hanx316 (approved):
LGTM

PR #158, opened 2020-08-19
feat(tabs): add tabs

### LINKED ISSUE #134: ### Tabs
Dependencies:
- not dependent on other components

bastarder:
- [x] Component Migration

Fix #134

### DIFF
--- a/packages/tabs/__tests__/tabs.spec.ts
+++ b/packages/tabs/__tests__/tabs.spec.ts
@@ -1,0 +1,591 @@
+import { mount } from '@vue/test-utils'
+import { nextTick } from 'vue'
+import Tabs from '../src/tabs.vue'
+import TabPane from '../src/tab-pane.vue'
+import TabNav from '../src/tab-nav.vue'
+
+describe('Tabs.vue', () => {
+  test('create', async () => {
+    const wrapper = mount({
+      components: {
+        'el-tabs': Tabs,
+        'el-tab-pane': TabPane,
+      },
+      template: `
+        <el-tabs>
+          <el-tab-pane label="label-1">A</el-tab-pane>
+          <el-tab-pane label="label-2">B</el-tab-pane>
+          <el-tab-pane label="label-3" ref="pane-click">C</el-tab-pane>
+          <el-tab-pane label="label-4">D</el-tab-pane>
+        </el-tabs>
+      `,
+    })
+
+    const tabsWrapper = wrapper.findComponent(Tabs)
+    const navWrapper = wrapper.findComponent(TabNav)
+    const panesWrapper = wrapper.findAllComponents(TabPane)
+    await nextTick()
+
+    const navItemsWrapper = navWrapper.findAll('.el-tabs__item')
+
+    expect(navItemsWrapper[0].classes('is-active')).toBe(true)
+    expect(panesWrapper[0].classes('el-tab-pane')).toBe(true)
+    expect(panesWrapper[0].attributes('id')).toBe('pane-0')
+    expect(panesWrapper[0].attributes('aria-hidden')).toEqual('false')
+    expect(tabsWrapper.vm.currentName).toEqual('0')
+
+    await navItemsWrapper[2].trigger('click')
+    expect(navItemsWrapper[0].classes('is-active')).toBe(false)
+    expect(panesWrapper[0].attributes('aria-hidden')).toEqual('true')
+    expect(navItemsWrapper[2].classes('is-active')).toBe(true)
+    expect(panesWrapper[2].attributes('aria-hidden')).toEqual('false')
+    expect(tabsWrapper.vm.currentName).toEqual('2')
+  })
+
+  test('active-name', async () => {
+    const wrapper = mount({
+      components: {
+        'el-tabs': Tabs,
+        'el-tab-pane': TabPane,
+      },
+      data() {
+        return {
+          activeName: 'b',
+        }
+      },
+      methods: {
+        handleClick(tab) {
+          this.activeName = tab.setupState.paneName
+        },
+      },
+      template: `
+        <el-tabs :active-name="activeName" @tab-click="handleClick">
+          <el-tab-pane name="a" label="label-1">A</el-tab-pane>
+          <el-tab-pane name="b" label="label-2">B</el-tab-pane>
+          <el-tab-pane name="c" label="label-3" ref="pane-click">C</el-tab-pane>
+          <el-tab-pane name="d" label="label-4">D</el-tab-pane>
+        </el-tabs>
+      `,
+    })
+
+    const tabsWrapper = wrapper.findComponent(Tabs)
+    const navWrapper = wrapper.findComponent(TabNav)
+    const panesWrapper = wrapper.findAllComponents(TabPane)
+    await nextTick()
+
+    const navItemsWrapper = navWrapper.findAll('.el-tabs__item')
+    expect(navItemsWrapper[1].classes('is-active')).toBe(true)
+    expect(panesWrapper[1].classes('el-tab-pane')).toBe(true)
+    expect(panesWrapper[1].attributes('id')).toBe('pane-b')
+    expect(panesWrapper[1].attributes('aria-hidden')).toEqual('false')
+    expect(tabsWrapper.vm.currentName).toEqual('b')
+
+    await navItemsWrapper[2].trigger('click')
+    expect(navItemsWrapper[1].classes('is-active')).toBe(false)
+    expect(panesWrapper[1].attributes('aria-hidden')).toEqual('true')
+    expect(navItemsWrapper[2].classes('is-active')).toBe(true)
+    expect(panesWrapper[2].attributes('aria-hidden')).toEqual('false')
+    expect(tabsWrapper.vm.currentName).toEqual('c')
+  })
+
+  test('card', async () => {
+    const wrapper = mount({
+      components: {
+        'el-tabs': Tabs,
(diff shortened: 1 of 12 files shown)


packages/tabs/__tests__/tabs.spec.ts
```diff
@@ -0,0 +1,591 @@
...
+describe('Tabs.vue', () => {
+  test('create', async () => {
+    const wrapper = mount({
+      components: {
+        'el-tabs': Tabs,
+        'el-tab-pane': TabPane,
+      },
+      template: `
+        <el-tabs>
+          <el-tab-pane label="用户管理">A</el-tab-pane>
```
zouhangwithsweet, line 16: How about  use en ?

packages/tabs/src/tab-bar.vue
```diff
@@ -0,0 +1,76 @@
...
+      type: Array,
+      default: () => [],
+    },
+  },
+  setup(props) {
+    const rootTabs = inject('rootTabs', { props: {} })
+    const instance = getCurrentInstance()
+
+    const getBarStyle = () => {
+      let style = {}
```
zouhangwithsweet, line 25: it should have a interface

packages/tabs/src/tab-nav.vue
```diff
@@ -0,0 +1,376 @@
...
+    const navOffset = ref(0)
+    const isFocus = ref(false)
+    const focusable = ref(true)
+
+    const navScroll$ = ref(null)
+    const nav$ = ref(null)
+    const el$ = ref(null)
+
+    const sizeName = computed(() => {
+      return ['top', 'bottom'].indexOf(rootTabs.props.tabPosition) !== -1 ? 'width' : 'height'
```
zouhangwithsweet, line 50: Maybe `includs` is better for Array

packages/tabs/src/tab-nav.vue
```diff
@@ -0,0 +1,376 @@
...
+    const rootTabs = inject('rootTabs', { props })
+
+    const scrollable = ref(false)
+    const navOffset = ref(0)
+    const isFocus = ref(false)
+    const focusable = ref(true)
+
+    const navScroll$ = ref(null)
+    const nav$ = ref(null)
+    const el$ = ref(null)
```
zouhangwithsweet, line 47: These should have a expected interface, `ref<null | HTMLElement>`

packages/tabs/src/tab-nav.vue
```diff
@@ -0,0 +1,376 @@
...
+      update,
+      changeTab,
+      setFocus,
+      removeFocus,
+      visibilityChangeHandler,
+      windowBlurHandler,
+      windowFocusHandler,
+    }
+  },
+  render() {
```
zouhangwithsweet, line 251: why not use `setup` for render 
```ts
  setup() {  return /* render */ () => {}}
```

packages/utils/util.ts
```diff
@@ -147,3 +147,7 @@ export function rafThrottle(fn: (args: Record<string, unknown>) => unknown): (..
 export const objToArray = castArray
 
 export { isVNode } from 'vue'
+
+export function firstUpperCase(str: string): string {
```
zouhangwithsweet, line 151: this util can be found in `@vue/shared`, please check it

zouhangwithsweet (changes requested):

packages/tabs/src/tab-bar.vue
```diff
@@ -0,0 +1,76 @@
+<template>
+  <div
+    class="el-tabs__active-bar"
+    :class="`is-${ rootTabs.props.tabPosition }`"
```
JeremyWuuuuu, line 4: Combine them?

packages/tabs/src/tab-bar.vue
```diff
@@ -0,0 +1,76 @@
...
+<script lang='ts'>
+import { defineComponent, inject, getCurrentInstance, watch, nextTick, ref } from 'vue'
+import { firstUpperCase } from '@element-plus/utils/util'
+
+export default defineComponent({
+  name: 'ElTabBar',
+  props: {
+    tabs: {
+      type: Array,
+      default: () => [],
```
JeremyWuuuuu, line 17: Specify the type here using 
```suggestion
type: Array as PropType<TypeAsItIs[]>
default: () => [] as TypeAsItIs[]
```

packages/tabs/src/tab-bar.vue
```diff
@@ -0,0 +1,76 @@
...
+    },
+  },
+  setup(props) {
+    const rootTabs = inject('rootTabs', { props: {} })
```
JeremyWuuuuu, line 21: I do think when tabs were used individually is some kind of error, which should be threw out

packages/tabs/src/tab-bar.vue
```diff
@@ -0,0 +1,76 @@
...
+          tabSize = $el[`client${firstUpperCase(sizeName)}`]
+
+          const tabStyles = window.getComputedStyle($el)
+
+          if (sizeName === 'width' && props.tabs.length > 1) {
+            tabSize -= parseFloat(tabStyles.paddingLeft) + parseFloat(tabStyles.paddingRight)
+          }
+          if (sizeName === 'width') {
+            offset += parseFloat(tabStyles.paddingLeft)
+          }
```
JeremyWuuuuu, line 48:
```suggestion
if (sizeName === 'width') {
  if (props.tabs.length > 1) {
    tabSize -= parseFloat(tabStyles.paddingLeft) + parseFloat(tabStyles.paddingRight)
  }
 offset += parseFloat(tabStyles.paddingLeft)
```

packages/tabs/src/tab-bar.vue
```diff
@@ -0,0 +1,76 @@
...
+          return false
+        }
+      })
+
+      const transform = `translate${firstUpperCase(sizeDir)}(${offset}px)`
+      style[sizeName] = tabSize + 'px'
```
JeremyWuuuuu, line 54: Let's keep this consistent `${tabSize}px`

packages/tabs/src/tab-bar.vue
```diff
@@ -0,0 +1,76 @@
...
+      style.transform = transform
+      style.msTransform = transform
+      style.webkitTransform = transform
```
JeremyWuuuuu, line 57: Can this be general method?

packages/tabs/src/tab-pane.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+  <div
+    v-if="(!lazy || loaded) || active"
```
JeremyWuuuuu, line 3: Maybe combine them as `computed`

packages/tabs/src/tab-pane.vue
```diff
@@ -0,0 +1,65 @@
...
+    v-show="active"
+    :id="`pane-${paneName}`"
+    class="el-tab-pane"
+    role="tabpanel"
```
JeremyWuuuuu, line 7: typo `tabpane`

packages/tabs/src/tab-nav.vue
JeremyWuuuuu:
```ts
type RefElement = Nullable<HTMLElement>
const el$ = ref<RefElement>(null)
```

packages/tabs/src/tab-nav.vue
```diff
@@ -0,0 +1,376 @@
...
+
+export default defineComponent({
+  name: 'ElTabNav',
+  components: {
+    TabBar,
+  },
+  props: {
+    panes: {
+      type: Array,
+      default: () => ([]),
```
JeremyWuuuuu, line 16: As I mentioned above, specify the type could help typing system to identify error easier

packages/tabs/src/tab-nav.vue
```diff
@@ -0,0 +1,376 @@
...
+    },
+    currentName: {
+      type: String,
+      default: '',
+    },
+    editable: Boolean,
+    onTabClick: {
+      type: Function,
+      default: NOOP,
```
JeremyWuuuuu, line 25: Common mistake here, default should've been
```ts
default: () => Noop
```
if you choose to use this like
```ts
default: NOOP
```
and the user did not specify `onTabClick` prop, then when you try to invoke `onTabClick` it will throw error because `onTabClick` is resolved as value `undefined` or `void` because `NOOP` were evaluated and returned nothing

packages/tabs/src/tab-nav.vue
```diff
@@ -0,0 +1,376 @@
...
+    },
+    onTabRemove: {
+      type: Function,
+      default: NOOP,
```
JeremyWuuuuu, line 29: Same here

jw-foss (changes requested):

packages/tabs/src/tab-nav.vue
bastarder: because i can't get refs in setup render  if my setup return `render function` not { refName }

packages/tabs/__tests__/tabs.spec.ts
bastarder: sure！

packages/tabs/src/tab-nav.vue
bastarder: i am wrong, i should use { ref: nameRef }

bastarder: if i use `render` in `setup`, i can't get `const a = ref(null)` such as this, in component instance.

jw-foss (changes requested):

PR #159, opened 2020-08-19
feat(directives): directive clickoutside helper for popper and modal

### LINKED ISSUE #157: ### Clickoutside
Directive clickoutside

jw-foss:
- [x] Component Migration

Fix #157


packages/directives/click-outside/index.ts
```diff
@@ -0,0 +1,73 @@
...
+
+type FlushList = Map<
+  HTMLElement,
+  {
+    documentHandler: DocumentHandler
+    bindingFn: (...args: unknown[]) => unknown
+  }
+>;
+
+const nodeList: FlushList = new Map()
```
zouhangwithsweet, line 16: how about `weakMap` for auto GC

JeremyWuuuuu: WeakMap is not iterable

PR #173, opened 2020-08-21
Fix: complete unit test for radio-button

TingYinHelen:
Fix #


packages/radio/__tests__/radio.spec.ts
```diff
@@ -290,11 +290,20 @@ describe('Radio Button', () => {
...
     const radio1 = wrapper.findComponent({ ref: 'radio1' })
     const radio2 = wrapper.findComponent({ ref: 'radio2' })
+    const radio3 = wrapper.findComponent({ ref: 'radio3' })
     const vm = wrapper.vm as any
     expect(vm.radio).toEqual(6)
     radio2.trigger('keydown.left')
     expect(vm.radio).toEqual(3)
     radio1.trigger('keydown.left')
     expect(vm.radio).toEqual(9)
+    await nextTick()
```
zazzaz, line 300: wow! why do we need this nextTick to make it passed?

PR #175, opened 2020-08-22
refactor(backtop): use toRefs open props

liaoliaojun:


Ryan2128 (changes requested):
I think do not need make props be Ref, why need?

PR #187, opened 2020-08-24
feat(alert): add alert prop types

Leecason:


packages/alert/src/index.vue
```diff
@@ -63,13 +77,13 @@ export default defineComponent({
...
     effect: {
-      type: String,
+      type: String as PropType<AlertEffect>,
       default: 'light',
       validator: (value: string): boolean => ['light', 'dark'].indexOf(value) > -1,
     },
   },
   emits: ['click'],
-  setup(props, ctx) {
+  setup(props: IAlertProps, ctx) {
```
zouhangwithsweet, line 86: useless interface for setup Props

zouhangwithsweet (changes requested):

packages/alert/src/index.vue
JeremyWuuuuu: Typescript automatically infers props' type for user if you have adapted `defineComponent`

packages/alert/src/index.vue
```diff
@@ -31,6 +31,20 @@
...
+interface IAlertProps {
+  title: string
+  description: string
+  type: AlertType
+  closable: boolean
+  closeText: string
+  showIcon: boolean
+  center: boolean
+  effect: AlertEffect
+}
```
JeremyWuuuuu, line 46: Useless interface here

jw-foss (changes requested):
First of all, this is not a feat, this is either a fix or a chore in this case.
Secondly, almost all types defined in the props will automatically inferred by typescript, refer to https://v3.vuejs.org/api/composition-api.html#setup

packages/alert/src/index.vue
```diff
@@ -31,6 +31,20 @@
 <script lang='ts'>
 import { defineComponent, computed, ref, PropType } from 'vue'
 
+type AlertType = 'success' | 'info' | 'error' | 'warning';
+type AlertEffect = 'light' | 'dark';
```
JeremyWuuuuu, line 35: Also, these comma needs to be removed

jw-foss (changes requested):

packages/alert/src/index.vue
Leecason: I'm a little confused now 🤔, because some component files had the prop interface such as `packages/divider/src/index.vue` so i did this.

PR #189, opened 2020-08-24
avoid use css scoped in component

caozhong1996:
- [ ] Component Migration

Fix #


JeremyWuuuuu:
Hey @caozhong1996 , currently we only uses css from v2. this is not out concern yet, if you are interested of setting up `scss` for this repo, please feel free to do so.

PR #213, opened 2020-08-27
fix(popper): should check nested Fragment

HeftyKoo:
- [ ] Component Migration

`ElPopper` may be use to implement other component,  such as:

```
<el-popper>
 <template v-slot:trigger>
   <slot name="trigger"></slot>
</template>
</el-popper>
```

So, when get the trigger element, should check nested Fragment。

### DIFF
--- a/packages/popper/src/usePopper.ts
+++ b/packages/popper/src/usePopper.ts
@@ -27,17 +27,18 @@ const getTrigger = () => {
   // SubTree is formed by <slot name="trigger"/><popper />
   // So that the trigger element is within the slot, we need to take it out of the slot in order to attach
   // events on top of it
-  const targetSlot = children[0]
+  let targetSlot = children[0]
   if (targetSlot.length > 1) {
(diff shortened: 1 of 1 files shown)


JeremyWuuuuu:
You can go checkout how `Tooltip` component implemented this with `Popper` refer to https://github.com/element-plus/element-plus/blob/dev/packages/tooltip/src/index.vue

JeremyWuuuuu:
I do think this is a good idea though

jw-foss (approved):

yeyuqiudeng:
> You can go checkout how `Tooltip` component implemented this with `Popper` refer to https://github.com/element-plus/element-plus/blob/dev/packages/tooltip/src/index.vue

Yes, This is a good solution. But i prefer use template rather than render function in complex dom structure。And template sometimes can have some benefit in compile time.

JeremyWuuuuu:
@yeyuqiudeng , Agreed, the way `Tooltip` implemented is not a general way of handling nested templates.

PR #214, opened 2020-08-27
Feat/implementing trigger for popper

jw-foss:
- Add `trigger` API for `ElPopper`
- Add arguments for `ClickOutside`

### DIFF
--- a/packages/directives/click-outside/index.ts
+++ b/packages/directives/click-outside/index.ts
@@ -47,7 +47,7 @@ function createDocumentHandler(
     const isTargetExists = !mouseUpTarget || !mouseDownTarget
     const isContainedByEl = el.contains(mouseUpTarget) || el.contains(mouseDownTarget)
     const isSelf = el === mouseUpTarget
-    console.log(excludes[0], mouseUpTarget)
+
     const isTargetExcluded =
       ( excludes.length &&
         excludes.some(item => item?.contains(mouseUpTarget))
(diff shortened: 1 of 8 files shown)


packages/tooltip/src/index.vue
```diff
@@ -87,6 +87,10 @@ export default defineComponent({
       type: String,
       default: 'el-fade-in-linear',
     },
+    trigger: {
+      type: [String, Array] as PropType<string | string[]>,
+      default: () => ['hover'],
+    },
```
JeremyWuuuuu, line 93: I think I should set a specific type here

packages/directives/click-outside/index.ts
```diff
@@ -30,21 +30,44 @@ function createDocumentHandler(
   el: HTMLElement,
   binding: DirectiveBinding,
 ): DocumentHandler {
+  let excludes: HTMLElement[] = []
+  if (Array.isArray(binding.arg)) {
+    excludes = binding.arg
+  } else {
+    // due to current implementation on binding type is wrong the type casting is necessary here
+    excludes.push(binding.arg as unknown as HTMLElement)
```
zouhangwithsweet, line 38: `undefined` cannot be pushed in `excludes`

packages/popper/src/index.vue
```diff
@@ -153,31 +153,7 @@ export default defineComponent({
...
-      popperId,
-      doDestroy,
-      onShow,
-      onHide,
-      popperRef,
-      popperInstance,
-      initializePopper,
-      visible,
-    }
+    return usePopper(props as IPopperOptions)
```
zouhangwithsweet, line 156: The type of the return value of this function should be correctly deduced

zouhangwithsweet (approved):

packages/directives/click-outside/index.ts
JeremyWuuuuu: Right, I will add a type check on this

PR #215, opened 2020-08-27
fix(popper): distinguish different trigger type

Chersquwn:
- [ ] Component Migration

The `trigger` prop is not used，and some events like `click` and `mouseenter` will conflict. So, should distinguish trigger type.


JeremyWuuuuu:
Modifications have been made in #214 , please leave your comments

JeremyWuuuuu:
Closing this PR since there is no further response from @Chersquwn

PR #224, opened 2020-08-29
docs(layout): fix invalid style syntax

ntnyq:


bastarder:
`/doc/*.vue` just used for test. we will remove them after document website is done

bastarder (approved):

PR #233, opened 2020-08-31
refactor(checkbox):  optimize code logic

### LINKED ISSUE #223: Checkbox: label type and modelValue type
**label:**
When I was refactoring `Transfer` component which depend on `Checkbox` and `CheckboxGroup` component, I met the warn below.

image

Then I checked the API of Element2's `Checkbox` components and found that the type of `label` is `string / number / boolean`.  But now it receives the type is `[Object, Boolean, String]`, is there missing `Number`?

```
label: {
  type: [Object, Boolean, String],
  default: '  ',
},
```

**modelValue:**
I also found that the type  of `modelValue` is `[Object, Boolean]`, why the type of `modelValue` can be `Object`? Shouldn't it always be a `Boolean`?

In my test, when use `Checkbox` component individually, the `modelValue` will change to a `Boolean` value if pass a object to `modelValue` and change the status of it.

```
modelValue: {
  type: [Object, Boolean],
  default: () => undefined,
},
```

zouhangwithsweet:
`checkbox` and `checkbutton` have many same logic code. 
So they need to be integrated

fix: #223

### DIFF
--- a/packages/checkbox/src/checkbox-button.vue
+++ b/packages/checkbox/src/checkbox-button.vue
@@ -51,12 +51,10 @@
 <script lang='ts'>
 import {
   defineComponent,
-  ref,
   computed,
   PropType,
-  watch,
 } from 'vue'
-import { useCheckboxGroup } from './useCheckbox'
+import { useCheckbox, useCheckboxGroup } from './useCheckbox'
 
 export default defineComponent({
   name: 'ElCheckboxButton',
@@ -85,55 +83,9 @@ export default defineComponent({
     },
   },
   emits: ['update:modelValue', 'change'],
-  setup(props, { emit }) {
-    let selfModel = false
-    const { elForm, isGroup, checkboxGroup, elFormItemSize, elFormItem, ELEMENT } = useCheckboxGroup()
-    const focus = ref(false)
-    const isLimitExceeded = ref(false)
-    const store = computed(() => checkboxGroup ? checkboxGroup.modelValue.value : props.modelValue)
-    const model = computed({
-      get() {
-        return isGroup.value ? store.value : props.modelValue !== undefined ? props.modelValue : selfModel
-      },
-
-      set(val: any) {
-        if (isGroup.value) {
-          isLimitExceeded.value = false
-
-          if (checkboxGroup.min !== undefined && val.length < checkboxGroup.min) {
-            isLimitExceeded.value = true
-          }
-          if (checkboxGroup.max !== undefined && val.length > checkboxGroup.max) {
-            isLimitExceeded.value = true
-          }
-
-          isLimitExceeded.value === false && checkboxGroup.changeEvent?.(val)
-        } else {
-          emit('update:modelValue', val)
-          selfModel = val
-        }
-      },
-    })
-    const isChecked = computed(() => {
-      if (Object.prototype.toString.call(model.value) === '[object Boolean]') {
-        return model.value
-      } else if (Array.isArray(model.value)) {
-        return model.value.includes(props.label)
-      } else if (model.value !== null && model.value !== undefined) {
-        return model.value === props.trueLabel
-      }
-    })
-    const isLimitDisabled = computed(() => {
-      const max = checkboxGroup.max
-      const min = checkboxGroup.min
-      return !!(max || min) && (model.value.length >= max && !isChecked.value) ||
-          (model.value.length <= min && isChecked.value)
-    })
-    const isDisabled = computed(() => {
-      return isGroup.value
-        ? checkboxGroup.disabled || props.disabled ||  (elForm as any || {} as any).disabled || isLimitDisabled.value
-        : props.disabled || (elForm as any || {} as any).disabled
-    })
+  setup(props) {
+    const { focus, isChecked, isDisabled, size, model, handleChange } = useCheckbox(props)
+    const { checkboxGroup } = useCheckboxGroup()
 
     const activeStyle = computed(() => {
       return {
@@ -143,37 +95,6 @@ export default defineComponent({
         'box-shadow': '-1px 0 0 0 ' + checkboxGroup.fill,
       }
     })
-
-    const size = computed(() => checkboxGroup.checkboxGroupSize || elFormItemSize || (ELEMENT || {}).size)
-
-    function addToStore() {
-      if (
-        Array.isArray(model.value) &&
-        !model.value.includes(props.label)
-      ) {
-        model.value.push(props.label)
-      } else {
-        model.value = props.trueLabel || true
-      }
-    }
-
-    function handleChange(e: UIEvent) {
-      if (isLimitExceeded.value) return
-      let value = ref(undefined)
-      if ((e.target as HTMLInputElement).checked) {
-        value.value = props.trueLabel === undefined ? true : props.trueLabel
-      } else {
-        value.value = props.falseLabel === undefined ? false : props.falseLabel
-      }
-
-      emit('change', value.value, e)
-    }
-
-    watch(() => props.modelValue, val => {
-      elFormItem.changeEvent?.(val)
-    })
-
-    props.checked && addToStore()
 
 
     return {
--- a/packages/checkbox/src/checkbox-button.vue
+++ b/packages/checkbox/src/checkbox-button.vue
@@ -92,7 +92,7 @@ export default defineComponent({
         backgroundColor: checkboxGroup.fill || '',
         borderColor: checkboxGroup.fill || '',
         color: checkboxGroup.textColor || '',
-        'box-shadow': '-1px 0 0 0 ' + checkboxGroup.fill,
+        boxShadow: '-1px 0 0 0 ' + checkboxGroup.fill,
       }
     })
 
--- a/packages/checkbox/src/checkbox-button.vue
+++ b/packages/checkbox/src/checkbox-button.vue
@@ -56,17 +56,17 @@ import {
   PropType,
   watch,
 } from 'vue'
-import { useCheckbox } from './useCheckbox'
+import { useCheckboxGroup } from './useCheckbox'
 
 export default defineComponent({
   name: 'ElCheckboxButton',
   props: {
     modelValue: {
-      type: [Object, Boolean, String, Number] as PropType<Record<string, unknown> | boolean | number>,
+      type: [Boolean, Number, String] as PropType<boolean | number | string>,
       default: () => undefined,
     },
     label: {
-      type: [Object, Boolean, String] as PropType<Record<string, unknown> | boolean | string>,
+      type: [Boolean, Number, String] as PropType<boolean | number | string>,
     },
     indeterminate: Boolean,
     disabled: Boolean,
@@ -86,11 +86,11 @@ export default defineComponent({
   },
   emits: ['update:modelValue', 'change'],
   setup(props, { emit }) {
-    const { elForm, isGroup, _checkboxGroup, _elFormItemSize, elFormItem, ELEMENT } = useCheckbox()
-    const selfModel = ref(false)
+    let selfModel = false
+    const { elForm, isGroup, checkboxGroup, elFormItemSize, elFormItem, ELEMENT } = useCheckboxGroup()
     const focus = ref(false)
     const isLimitExceeded = ref(false)
-    const store = computed(() => _checkboxGroup ? _checkboxGroup.modelValue.value : props.modelValue)
+    const store = computed(() => checkboxGroup ? checkboxGroup.modelValue.value : props.modelValue)
     const model = computed({
       get() {
         return isGroup.value ? store.value : props.modelValue !== undefined ? props.modelValue : selfModel
@@ -100,17 +100,17 @@ export default defineComponent({
         if (isGroup.value) {
           isLimitExceeded.value = false
 
-          if (_checkboxGroup.min !== undefined && val.length < _checkboxGroup.min) {
+          if (checkboxGroup.min !== undefined && val.length < checkboxGroup.min) {
             isLimitExceeded.value = true
           }
-          if (_checkboxGroup.max !== undefined && val.length > _checkboxGroup.max) {
+          if (checkboxGroup.max !== undefined && val.length > checkboxGroup.max) {
             isLimitExceeded.value = true
           }
 
-          isLimitExceeded.value === false && _checkboxGroup.changeEvent?.(val)
+          isLimitExceeded.value === false && checkboxGroup.changeEvent?.(val)
         } else {
           emit('update:modelValue', val)
-          selfModel.value = val
+          selfModel = val
         }
       },
     })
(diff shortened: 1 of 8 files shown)


packages/checkbox/src/useCheckbox.ts
```diff
@@ -1,23 +1,183 @@
...
+  const { elForm, isGroup, checkboxGroup } = useCheckboxGroup()
+  const isLimitDisabled = computed(() => {
+    const max = checkboxGroup.max
+    const min = checkboxGroup.min
+    return !!(max || min) && (model.value.length >= max && !isChecked.value) ||
+      (model.value.length <= min && isChecked.value)
+  })
+  const isDisabled = computed(() => {
+    return isGroup.value
+      ? checkboxGroup.disabled || props.disabled ||  (elForm as any || {} as any).disabled || isLimitDisabled.value
```
Ryan2128, line 111: Is `elForm` can asigned a type below? And also `elFormItem`.
```
type ElForm = {
  size?: string
  disabled?: boolean
}
```

I think this way is better than `as any`.

Ryan2128 (changes requested):

packages/checkbox/src/useCheckbox.ts
zouhangwithsweet: This type will be provided by the `el-form`.

packages/checkbox/src/useCheckbox.ts
```diff
@@ -1,23 +1,183 @@
...
+    return isGroup.value
+      ? checkboxGroup.checkboxGroupSize || temCheckboxSize
+      : temCheckboxSize
+  })
+
+  return {
+    isChecked,
+    focus,
+    size,
+    checkboxSize,
```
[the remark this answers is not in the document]
zouhangwithsweet: `size` is used in `el-checkbox-button`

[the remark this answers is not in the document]
Ryan2128: OK, I didn't notice it when I was watching.

packages/checkbox/src/checkbox-group.vue
```diff
@@ -71,7 +71,7 @@ export default defineComponent({
       fill: props.fill,
```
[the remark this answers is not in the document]
zouhangwithsweet: Actually, I didn’t want child component to get dynamic values at first. I will think about this question. :)

packages/checkbox/src/useCheckbox.ts
Ryan2128: YES, I also think so, you can wait for ElForm's update.

packages/checkbox/src/checkbox-group.vue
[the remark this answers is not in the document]
Ryan2128: If you do this, when the user has a variable that manages the global size, changing this variable will not trigger the size change, the other is the same.

I think these props listed will have scenarios that change their values.

If don't use reactive, its behavior is different from element2, it should be point out when release.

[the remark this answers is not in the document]
bastarder: maybe `toRef` is better

packages/checkbox/src/checkbox.d.ts
```diff
@@ -0,0 +1,30 @@
...
+  name?: string
+  trueLabel?: string| number
+  falseLabel?: string| number
+  id?: string
+  controls?: string
+  border?: boolean
+  size?: string
+}
+
+export type IArgs<T extends (...args: any) => any> = Partial<ReturnType<T>>
```
HerringtonDarkholme, line 30:
```suggestion
export type PartialReturnType<T extends (...args: unknown[]) =>  unknown> = Partial<ReturnType<T>>
```

packages/checkbox/src/useCheckbox.ts
```diff
@@ -1,23 +1,181 @@
...
+
+const useModel = (props: ICheckboxProps) => {
+  let selfModel = false
+  const { emit } = getCurrentInstance()
+  const { isGroup, checkboxGroup } = useCheckboxGroup()
+  const isLimitExceeded = ref(false)
+  const store = computed(() => checkboxGroup ? checkboxGroup.modelValue?.value : props.modelValue)
+  const model = computed({
+    get() {
+      return isGroup.value ? store.value : props.modelValue !== undefined ? props.modelValue : selfModel
```
HerringtonDarkholme, line 41:
```suggestion
      return isGroup.value 
        ? store.value 
        : props.modelValue !== undefined ? props.modelValue : selfModel.value
```

packages/checkbox/src/useCheckbox.ts
```diff
@@ -1,23 +1,181 @@
...
   //todo: ELEMENT
   const ELEMENT = null
   const elForm = inject('elForm', {})
   const elFormItem = inject('elFormItem', {}) as any
-  const _checkboxGroup = inject('CheckboxGroup', {}) as any
-  const focus = ref(false)
-  const isGroup = computed(() => _checkboxGroup && _checkboxGroup.name === 'ElCheckboxGroup')
-  const _elFormItemSize = computed(() => {
+  const checkboxGroup = inject<ICheckboxGroupInstance>('CheckboxGroup', {})
+  const isGroup = computed(() => checkboxGroup && checkboxGroup.name === 'ElCheckboxGroup')
```
HerringtonDarkholme, line 19: I wonder when `checkboxGroup` will be falsy.

zouhangwithsweet: It's a question -_-

packages/checkbox/src/useCheckbox.ts
```diff
@@ -1,23 +1,181 @@
...
+  watch(() => props.modelValue, val => {
+    elFormItem.changeEvent?.(val)
+  })
+
+  return {
+    handleChange,
+  }
+}
+
+export const setAria = (props: ICheckboxProps) => {
```
HerringtonDarkholme, line 157: Maybe `useSetAria` is better since this uses `getCurrentInstance` hook?

packages/checkbox/src/useCheckbox.ts
```diff
@@ -1,23 +1,181 @@
...
+  return {
+    model,
+    isLimitExceeded,
+  }
+}
+
+const useCheckboxStatus = (props: ICheckboxProps, { model }: IArgs<typeof useModel>) => {
+  const { isGroup, checkboxGroup, elFormItemSize, ELEMENT } = useCheckboxGroup()
+  const focus = ref(false)
+  const size = computed<string|undefined>(() => checkboxGroup.checkboxGroupSize?.value || elFormItemSize || (ELEMENT || {}).size)
```
HerringtonDarkholme, line 72: If I read code right, `elFormItemSize` is a `computed` and its truth value will be always true?

packages/checkbox/src/checkbox-button.vue
```diff
@@ -51,22 +51,21 @@
...
-import { useCheckbox } from './useCheckbox'
+import { UPDATE_MODEL_EVENT } from '@element-plus/utils/constants'
+import { useCheckbox, useCheckboxGroup } from './useCheckbox'
 
 export default defineComponent({
   name: 'ElCheckboxButton',
   props: {
     modelValue: {
-      type: [Object, Boolean, String, Number] as PropType<Record<string, unknown> | boolean | number>,
+      type: [Boolean, Number, String] as PropType<boolean | number | string>,
```
HerringtonDarkholme, line 64: I wonder if Vue can correctly identify the prop type without annotation.

packages/checkbox/src/checkbox-group.vue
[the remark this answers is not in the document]
Ryan2128: Emmmm, yes, nothing is computed here.

packages/checkbox/src/useCheckbox.ts
```diff
@@ -1,23 +1,183 @@
...
   const elForm = inject('elForm', {})
   const elFormItem = inject('elFormItem', {}) as any
-  const _checkboxGroup = inject('CheckboxGroup', {}) as any
-  const focus = ref(false)
-  const isGroup = computed(() => _checkboxGroup && _checkboxGroup.name === 'ElCheckboxGroup')
-  const _elFormItemSize = computed(() => {
+  const checkboxGroup = inject<ICheckboxGroupInstance>('CheckboxGroup', {})
+  const isGroup = computed(() => checkboxGroup && checkboxGroup.name === 'ElCheckboxGroup')
+  const elFormItemSize = computed(() => {
     return (elFormItem || {} as any).elFormItemSize
```
JeremyWuuuuu, line 21: We might need to specify the `ElForm` interface since I saw lots of code like this just casting it to any is not quite what we need right?
Can you put it into a new file for formItems? Maybe just for temporary, when form gets finished, we could just replace it with this

JeremyWuuuuu: Also, can you use the `?.` syntax to access `elFormItemSize`?

packages/checkbox/src/useCheckbox.ts
```diff
@@ -1,23 +1,183 @@
...
+  const { emit } = getCurrentInstance()
+  const { isGroup, checkboxGroup } = useCheckboxGroup()
+  const isLimitExceeded = ref(false)
+  const store = computed(() => checkboxGroup ? checkboxGroup.modelValue?.value : props.modelValue)
+  const model = computed({
+    get() {
+      return isGroup.value
+        ? store.value
+        : props.modelValue ?? selfModel
+    },
```
JeremyWuuuuu, line 44: This is good, because I have seen many components still tries to implement compatible API for both `value & v-model`
Can we make this to a general compositions? Like taking a `function` as parameter to handling user defined stuffs

packages/checkbox/src/useCheckbox.ts
```diff
@@ -1,23 +1,183 @@
...
+
+    set(val: unknown) {
```
JeremyWuuuuu, line 46: When this composition became general, you may mark this as a generic type so that you won't need to worry about the type

packages/checkbox/src/useCheckbox.ts
```diff
@@ -1,23 +1,183 @@
...
+}
+
+const useCheckboxStatus = (props: ICheckboxProps, { model }: PartialReturnType<typeof useModel>) => {
+  const { isGroup, checkboxGroup, elFormItemSize, ELEMENT } = useCheckboxGroup()
+  const focus = ref(false)
+  const size = computed<string|undefined>(() => checkboxGroup.checkboxGroupSize?.value || elFormItemSize.value || (ELEMENT || {}).size)
+  const isChecked = computed(() => {
+    const value = model.value
+    if (toTypeString(value) === '[object Boolean]') {
+      return Boolean(value)
```
JeremyWuuuuu, line 78: Don't you think these two lines of code does not make any sense?
When you tries to know if `value` is Boolean or not, when it's already boolean, why would you use `Boolean(val)` on a `boolean` value?
Since whether `value` is `Object<Boolean>` or `Boolean Literal` applying `toString` on `value` always gets you `[object Boolean]`

packages/checkbox/src/useCheckbox.ts
```diff
@@ -1,23 +1,183 @@
...
+  const isLimitDisabled = computed(() => {
+    const max = checkboxGroup.max?.value
+    const min = checkboxGroup.min?.value
+    return !!(max || min) && (model.value.length >= max && !isChecked.value) ||
+      (model.value.length <= min && isChecked.value)
+  })
+  const isDisabled = computed(() => {
+    return isGroup.value
+      ? checkboxGroup.disabled?.value || props.disabled ||  (elForm as any || {} as any).disabled?.value || isLimitDisabled.value
+      : props.disabled || (elForm as any || {} as any).disabled?.value
```
JeremyWuuuuu, line 114: This is some sort of disaster for both the chained short cut operator and type casting.
Also, when `isGroup.value` goes to the falsy branch and props.disabled is `false` also elForm has not been given by injection, the final result would be `undefined` instead of a `boolean` type that I assume you need here.

packages/checkbox/src/useCheckbox.ts
```diff
@@ -1,23 +1,183 @@
...
+    elFormItem.changeEvent?.(val)
+  })
+
+  return {
+    handleChange,
+  }
+}
+
+export const useSetAria = (props: ICheckboxProps) => {
+  const instance = getCurrentInstance()
```
JeremyWuuuuu, line 160: why not put the line of code into `onMounted` callback body instead of holding the reference closure when this hook gets invoked?

jw-foss (changes requested):

packages/checkbox/src/checkbox.d.ts
```diff
@@ -0,0 +1,30 @@
...
+  name?: string
+  trueLabel?: string| number
+  falseLabel?: string| number
+  id?: string
+  controls?: string
+  border?: boolean
+  size?: string
+}
+
+export type PartialReturnType<T extends (...args: unknown[]) =>  unknown> = Partial<ReturnType<T>>
```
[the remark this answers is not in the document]
zouhangwithsweet: I hope others need this type

packages/checkbox/src/useCheckbox.ts
zouhangwithsweet: Yep, maybe we need more discuss

packages/checkbox/src/useCheckbox.ts
zouhangwithsweet: I agree this point.

packages/checkbox/src/checkbox-button.vue
```diff
@@ -84,105 +82,20 @@ export default defineComponent({
...
     const activeStyle = computed(() => {
       return {
-        backgroundColor: _checkboxGroup.fill || '',
-        borderColor: _checkboxGroup.fill || '',
-        color: _checkboxGroup.textColor || '',
-        'box-shadow': '-1px 0 0 0 ' + _checkboxGroup.fill,
+        backgroundColor: checkboxGroup.fill?.value || '',
+        borderColor: checkboxGroup.fill?.value || '',
+        color: checkboxGroup.textColor?.value || '',
+        boxShadow: '-1px 0 0 0 ' + checkboxGroup.fill?.value,
```
JeremyWuuuuu, line 95: Do you mind to change these to
```suggestion
backgroundColor: checkboxGroup.fill?.value ?? '',
borderColor: checkboxGroup.fill?.value ?? '',
color: checkboxGroup.textColor?.value ?? '',
boxShadow: '-1px 0 0 0 ' + checkboxGroup.fill?.value ?? '',
```
for consistency? I saw you using `??` syntax in your changes. also the last line is a potential bug if the evaluation returns `undefined` the value of `boxShadow` will be `-1px 0 0 0 undefined'

jw-foss (changes requested):

packages/checkbox/src/checkbox-button.vue
JeremyWuuuuu: Everything else LGTM, address here then it'll get merged

jw-foss (approved):

PR #234, opened 2020-08-31
Fix/image: fix the bug cause by $attrs

Ryan2128:
- [ ] Component Migration

Fix the bug cause by `v-bind="$attrs"`

In vue3, `$attrs` includes `style` and `class`, but not in vue2. In vue3, one of its uses is pass attrs to `Fragment`, so it should include `style` and `class`. 

So should filte `style` and `class` or use `inheritAttrs: false` to pass attrs to child node directly when use `v-bind="$attrs"` to pass attrs to child node.

Because current progress is migrate and the CSS is not ready, so I fix this bug by separating `$attrs`, let only `alt` attach to ``, and other attrs apply to root node.

The error behavior in below: 
image
image

### DIFF
--- a/packages/image/src/index.vue
+++ b/packages/image/src/index.vue
@@ -1,5 +1,10 @@
 <template>
-  <div ref="container" class="el-image">
+  <div
+    ref="container"
+    class="el-image"
+    :class="rootClass"
+    :style="rootStyle"
+  >
     <slot v-if="loading" name="placeholder">
       <div class="el-image__placeholder"></div>
     </slot>
@@ -35,7 +40,7 @@ import throttle from 'lodash/throttle'
 import isServer from '@element-plus/utils/isServer'
 import { on, off, getScrollContainer, isInContainer } from '@element-plus/utils/dom'
 import { t } from '@element-plus/locale'
-import ImageViewer from './image-viewer'
+import ImageViewer from './image-viewer.vue'
 
 const isSupportObjectFit = () => document.documentElement.style.objectFit !== undefined
 const isHtmlEle = e => e && e.nodeType === 1
@@ -55,7 +60,10 @@ export default defineComponent({
   components: {
     ImageViewer,
   },
+  inheritAttrs: false,
   props: {
+    class: String,
+    style: Object,
     src: {
       type: String,
       default: '',
@@ -86,8 +94,8 @@ export default defineComponent({
     // init here
     const hasLoadError = ref(false)
     const loading = ref(true)
-    const imgWidth = ref(false)
-    const imgHeight = ref(false)
+    const imgWidth = ref(0)
+    const imgHeight = ref(0)
     const showViewer = ref(false)
     const container = ref<HTMLElement | null>(null)
     const show = ref(props.lazy)
@@ -115,7 +123,7 @@ export default defineComponent({
       return Array.isArray(previewSrcList) && previewSrcList.length > 0
     })
     const imageIndex = computed(() => {
-      const { src , previewSrcList } = props
+      const { src, previewSrcList } = props
       let previewIndex = 0
       const srcIndex = previewSrcList.indexOf(src)
       if (srcIndex >= 0) {
@@ -170,13 +178,13 @@ export default defineComponent({
       // so it can behave consistently
       Object.keys(attrs)
         .forEach(key => {
-          const value = attrs[key]
+          const value = attrs[key] as string
           img.setAttribute(key, value)
         })
       img.src = props.src
     }
 
-    function handleLoad(e: Event, img: Any) {
+    function handleLoad(e: Event, img: HTMLImageElement) {
       imgWidth.value = img.width
       imgHeight.value = img.height
       loading.value = false
@@ -254,6 +262,10 @@ export default defineComponent({
       props.lazy && removeLazyLoadListener()
     })
 
+    // class is keyword of JS, so cannot pass it on root, like :class="class".
+    const rootClass = computed(() => props.class)
+    const rootStyle = computed(() => props.style)
+
     return {
       loading,
       hasLoadError,
@@ -268,6 +280,8 @@ export default defineComponent({
       closeViewer,
       container,
       t,
+      rootClass,
+      rootStyle,
     }
   },
 })


bastarder:
```
props: {
  style: {  ...  },
  class: {  ... },
}
```

i think we just add `style`, 'class' in props, then `attr` will not include them

Ryan2128:
> ```
> props: {
>   style: {  ...  },
>   class: {  ... },
> }
> ```
> 
> i think we just add `style`, `class` in props, then `attr` will not include them

Yes, this way is also possible, but it is just a little strange, because it is seem to two props which don't appear in docs.

But code like this is more clearer.

I'll change it in a few latter.

Ryan2128:
> ```
> props: {
>   style: {  ...  },
>   class: {  ... },
> }
> ```
> 
> i think we just add `style`, `class` in props, then `attr` will not include them


And should add `inheritAttrs: false`? 

If add it, some attr will not attatch to root `div`, like `titile`? If not, `title` will be attach to both of them.

If do this, all attrs passed to root node should in props?

bastarder:
> > ```
> > props: {
> >   style: {  ...  },
> >   class: {  ... },
> > }
> > ```
> > 
> > 
> > i think we just add `style`, `class` in props, then `attr` will not include them
> 
> And should add `inheritAttrs: false`?
> 
> If add it, some attr will not attatch to root `div`, like `id`, `name`, `titile`? If not, `title` will be attach to both of them.
> 
> If do this, all attrs passed to root node should in props?
> 
> How about the idea separating attrs that child node need and the rest are attached to the root node?
> 
> @bastarder

`element-ui` also add `inheritAttrs`, so the `el-image` only bind `style`, `class` by default, this dom will not have `id`, 'title'

Ryan2128:
According to the current situation, if add `inheritAttrs: false`, there will be no style and class on the root node.

Maybe this sentence means it would prevent attrs from being bound to the root node, including class and style, unlike vue2, class and style would still be attached to the root node?

@bastarder

Ryan2128:
> `element-ui` also add `inheritAttrs`, so the `el-image` only bind `style`, `class` by default, this dom will not have `id`, `title`
> 
> i saw the `Vue 3.0 doc`, also have `Note: this option does not affect class and style bindings.` , but actually we can not find `class, style` on `el-image`, i dont know is this a bug

In vue3, `inheritAttrs: false` will effect `style` and `class`, but not in vue2

image

image

packages/image/src/index.vue
```diff
@@ -29,13 +34,13 @@
...
+import { t } from '@element-plus/locale'
+import { getScrollContainer, isInContainer, off, on } from '@element-plus/utils/dom'
+import isServer from '@element-plus/utils/isServer'
 import { isString } from '@vue/shared'
 import throttle from 'lodash/throttle'
-import isServer from '@element-plus/utils/isServer'
-import { on, off, getScrollContainer, isInContainer } from '@element-plus/utils/dom'
-import { t } from '@element-plus/locale'
-import ImageViewer from './image-viewer'
+import { computed, defineComponent, onBeforeUnmount, onMounted, ref, watch } from 'vue'
```
zouhangwithsweet, line 42: Why not put modules of `vue` to the top

packages/image/src/index.vue
```diff
@@ -83,6 +91,7 @@ export default defineComponent({
   },
   emits: ['error'],
   setup(props, { emit, attrs }) {
+    console.log(attrs)
```
zouhangwithsweet, line 94: remove `console.log`

packages/image/src/index.vue
```diff
@@ -1,5 +1,10 @@
 <template>
-  <div ref="container" class="el-image">
+  <div
+    ref="container"
+    class="el-image"
+    :class="$props['class']"
```
zouhangwithsweet, line 5: Is there another way to bind it?

zouhangwithsweet (changes requested):

packages/image/src/index.vue
Ryan2128: Webstorm sort the import, I will revert it

packages/image/src/index.vue
Ryan2128: Two way:
1. Make `style` and `class` in props and bind `$props['class']` or `$props.class` to `div`(`class` is js keywords, so cannot bind `class` directly).

2. Make `style` and `class` not in props, bind `$attrs.class` and `$attrs.style` to `div` and filte `style` and `class` in `setup` and bind rest attrs to `img`.

Which is better?

JeremyWuuuuu:
> > ```
> > props: {
> >   style: {  ...  },
> >   class: {  ... },
> > }
> > ```
> > 
> > 
> > i think we just add `style`, `class` in props, then `attr` will not include them
> 
> Yes, this way is also possible, but it is just a little strange, because it is seem to two props which don't appear in docs.
> 
> Coding like this is more clearer.

IMHO, first above all, the component is a customized `Vue` component, instead of raw `img` tag, it is quite ok to have APIs like `style, class`, customized React components does this all the time (passing the `className` `style` as component props), if you have these APIs, simply reflect them in the document, that is alright.
Also, if the property passed down from the user is what we don't want, we could also add some property checking services like a filter to give a warning about it. Suppressing `inheritAttrs` by default it not what we wanted.

packages/image/src/index.vue
JeremyWuuuuu: This is not making any sense, when you declare `style` and `class` in `props`, simply use `style` and `class` would've been enough, you do not need syntax like `$props['class']` also , accessing a member when you already know it's name you do not need the `[]`, simply using dot operator to reach the member

Ryan2128: `class` is keyword of Javascript, cannot use directly.

bastarder (approved):

packages/image/src/index.vue
```diff
@@ -1,5 +1,10 @@
 <template>
-  <div ref="container" class="el-image">
+  <div
+    ref="container"
+    class="el-image"
+    :class="$props.class"
```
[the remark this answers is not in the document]
Ryan2128: The newest code:
```
:class="rootClass"
:style="rootClass"
```
Because `class` is keyword of js, cannot use `:class="class"` directly, so I wrap it and `style` in setup.

HerringtonDarkholme:
IMHO this fix is good enough and makes sensible trade-off between Vue2 and Vue3.

HerringtonDarkholme (approved):

PR #236, opened 2020-08-31
feat(table): Table component

justwiner:
- [ ] Component Migration


packages/utils/util.ts
```diff
@@ -156,3 +156,18 @@ export function getRandomInt(max: number) {
 }
 
 export { isVNode } from 'vue'
+
+// TODO: use native Array.find, Array.findIndex when IE support is dropped
```
zazzaz, line 160: Just use Array.find, please

zazzaz:
Cool~

@justwiner Just a reminder that it's better to rebase the origin/dev branch daily to reduce conflicts.

packages/utils/util.ts
justwiner: Ok!
Optimization will be carried out later, including type checking

packages/table/src/store/helper.ts
```diff
@@ -1,5 +1,5 @@
 import Store from './index'
-import debounce from 'throttle-debounce/debounce'
```
zazzaz: please use `lodash` instead, package throttle-debounce will be removed later

PR #248, opened 2020-09-02
Feature/drawer

### LINKED ISSUE #150: ### Drawer
Assign to @luxueyan
Drawer component

luxueyan:
- [x] drawer component done

Fix #150


JeremyWuuuuu:
You need to remove `.pretterrc` and revert changes in `eslintrc`

luxueyan:
I rebase from dev branch , but then run yarn test.  The try catch block statement has gone. How do I resolve it @JeremyWuuuuu 
```
TypeError: Illegal invocation

      94 |   Utils.IgnoreUtilFocusChanges = true
      95 |   // Remove the old try catch block since there will be no error to be thrown
    > 96 |   element.focus?.()
         | ^
      97 |   Utils.IgnoreUtilFocusChanges = false
      98 |   return document.activeElement === element
      99 | 
 at Function.focus (node_modules/jsdom/lib/jsdom/living/generated/HTMLElement.js:114:15)
      at attemptFocus (packages/utils/aria.ts:96:1)
      at Object.focusFirstDescendant (packages/utils/aria.ts:155:1)
      at Object.focusFirstDescendant (packages/utils/aria.ts:155:1)
      at packages/drawer/src/index.vue:146:1
```

JeremyWuuuuu:
> I have revert the unecessary commit.
> I rebase from dev branch , but then run yarn test. The try catch block statement has gone. How do I resolve it . @JeremyWuuuuu
> 
> ```
> TypeError: Illegal invocation
> 
>       94 |   Utils.IgnoreUtilFocusChanges = true
>       95 |   // Remove the old try catch block since there will be no error to be thrown
>     > 96 |   element.focus?.()
>          | ^
>       97 |   Utils.IgnoreUtilFocusChanges = false
>       98 |   return document.activeElement === element
>       99 | 
>  at Function.focus (node_modules/jsdom/lib/jsdom/living/generated/HTMLElement.js:114:15)
>       at attemptFocus (packages/utils/aria.ts:96:1)
>       at Object.focusFirstDescendant (packages/utils/aria.ts:155:1)
>       at Object.focusFirstDescendant (packages/utils/aria.ts:155:1)
>       at packages/drawer/src/index.vue:146:1
> ```
> 
> For test case passed , I add try catch back.

This is not an error caused by us, you need to handle it yourself. All code on the dev branch from the origin repo is legit, you need to figure it out on your own

JeremyWuuuuu:
BTW, I just looked through your code, the current implementation is not what we wanted since `Vue` itself has already provided `Teleport` feature for you to implementing teleporting APIs like `appendToBody` if you do not have any idea on how to implementing this API with `Teleport`, you could easily find an example in `Popper`

packages/drawer/src/index.vue
```diff
@@ -0,0 +1,239 @@
...
+              }}</span>
+            </slot>
+            <button
+              v-if="showClose"
+              :aria-label="`close ${title || 'drawer'}`"
+              class="el-drawer__close-btn"
+              type="button"
+              @click="closeDrawer"
+            >
+              <i class="el-dialog__close el-icon el-icon-close"></i>
```
caozhong1996, line 48: `el-dialog__close` maybe should be `el-draw__close`, I think it's legacy.

luxueyan: OK！I will fix it。

luxueyan:
> BTW, I just looked through your code, the current implementation is not what we wanted since `Vue` itself has already provided `Teleport` feature for you to implementing teleporting APIs like `appendToBody` if you do not have any idea on how to implementing this API with `Teleport`, you could easily find an example in `Popper`

Ok I get it!

zhixiaotong:
I am paying attention to element-plus, but I have less practice on typescript. Could I make a suggestion? In the element 2.x version of the draw component, if i set mask to false, i still cannot operate other elements on the page, but I personally think this is necessary. Do you agree or any plans?

luxueyan:
> I am paying attention to element-plus, but I have less practice on typescript. Could I make a suggestion? In the element 2.x version of the draw component, if i set mask to false, i still cannot operate other elements on the page, but I personally think this is necessary. Do you agree or any plans?

If we refer to the non-modal window of the PC software, it should be allowed to operate the content below the window. But I found that similar dialog components also have this problem. Maybe later versions will unify standards.
If you want to  achieve the same effect, You can hack to set style of 'el-drawer__wrapper' with 'pointer-events:none' and meanwhile set ’pointer-events: initial‘ to 'el-drawer'

zhixiaotong:
> > I am paying attention to element-plus, but I have less practice on typescript. Could I make a suggestion? In the element 2.x version of the draw component, if i set mask to false, i still cannot operate other elements on the page, but I personally think this is necessary. Do you agree or any plans?
> 
> If we refer to the non-modal window of the PC software, it should be allowed to operate the content below the window. But I found that similar dialog components also have this problem. Maybe later versions will unify standards.
> If you want to achieve the same effect, You can hack to set style of 'el-drawer__wrapper' with 'pointer-events:none' and meanwhile set ’pointer-events: initial‘ to 'el-drawer'

Thank you very much for your answer. I look forward to it. In my work practice, Draw is compared to Dialog. It is usually used for similar retractable sidebars, but there is almost no such need for Dialog.

luxueyan:
> > I have revert the unecessary commit.
> > I rebase from dev branch , but then run yarn test. The try catch block statement has gone. How do I resolve it . @JeremyWuuuuu
> > ```
> > TypeError: Illegal invocation
> > 
> >       94 |   Utils.IgnoreUtilFocusChanges = true
> >       95 |   // Remove the old try catch block since there will be no error to be thrown
> >     > 96 |   element.focus?.()
> >          | ^
> >       97 |   Utils.IgnoreUtilFocusChanges = false
> >       98 |   return document.activeElement === element
> >       99 | 
> >  at Function.focus (node_modules/jsdom/lib/jsdom/living/generated/HTMLElement.js:114:15)
> >       at attemptFocus (packages/utils/aria.ts:96:1)
> >       at Object.focusFirstDescendant (packages/utils/aria.ts:155:1)
> >       at Object.focusFirstDescendant (packages/utils/aria.ts:155:1)
> >       at packages/drawer/src/index.vue:146:1
> > ```
> > 
> > 
> > For test case passed , I add try catch back.
> 
> This is not an error caused by us, you need to handle it yourself. All code on the dev branch from the origin repo is legit, you need to figure it out on your own

This seems to be a bug of @swc-node. I have submitted a issue https://github.com/Brooooooklyn/swc-node/issues/62. The  transformed code of  optional method call seems missing right context.

luxueyan:
@JeremyWuuuuu  I have update the code with Teleport。But in the 'aira.ts', the optional method call was transformed wrongly by  @swc-node. Brooooooklyn/swc-node#62.

JeremyWuuuuu:
@luxueyan  thanks Lu! I'll go ahead take a look at your code tomorrow, I saw some changes in the `aria.ts` file, has the bug fix gets merged? If it has, would you mind upgrade the dependency with it? That'd be good, but the change you made is legit to me though. Great job

luxueyan:
> @luxueyan thanks Lu! I'll go ahead take a look at your code tomorrow, I saw some changes in the `aria.ts` file, has the bug fix gets merged? If it has, would you mind upgrade the dependency with it? That'd be good, but the change you made is legit to me though. Great job

Haha! You are welcome! It is my pleasure to study with every collaborators.As to this bug.Happily, the author has fixed this problem in time. I have upgraded the dependency to latest. And I rebase my commit for pruning redundant commit. Thanks for your reviewing and sorry for my poor english.
-----
BTW, what formatter do you use in vsc. I use prettier, but some formatted code don't conform to the eslint rule.Such as：
```vue
<!--Disallow self-closing on HTML void elements (<input/>).eslintvue/html-self-closing -->
<input type="radio" :checked="direction === 'ltr'" @click="change('ltr')" />

<!-- :direction' should be on a new line.eslintvue/max-attributes-per-line -->
<el-drawer v-model:visible="visible" size="50%" title="我是标题" :direction="direction">
```

JeremyWuuuuu:
> > @luxueyan thanks Lu! I'll go ahead take a look at your code tomorrow, I saw some changes in the `aria.ts` file, has the bug fix gets merged? If it has, would you mind upgrade the dependency with it? That'd be good, but the change you made is legit to me though. Great job
> 
> Haha! You are welcome! It is my pleasure to study with every collaborators.As to this bug.Happily, the author has fixed this problem in time. I have upgraded the dependency to latest. And I rebase my commit for pruning redundant commit. Thanks for your reviewing and sorry for my poor english.
> 
> BTW, what formatter do you use in vsc. I use prettier, but some formatted code don't conform to the eslint rule.Such as：
> 
> ```
> 
> <input type="radio" :checked="direction === 'ltr'" @click="change('ltr')" />
> 
> 
> <el-drawer v-model:visible="visible" size="50%" title="我是标题" :direction="direction">
> ```

Thing is, we had this intension not adding `prettier` in the project to make sure developers checks on their commits themselves and not being too dependent on tools like `prettier`. The way that I'd recommend is you run `yarn lint` every time you finish your coding, even though this is relatively waste of time.

packages/drawer/__tests__/drawer.spec.ts
```diff
@@ -0,0 +1,343 @@
...
+    template,
+    data,
+    ...otherObj,
+  })
+
+const sleep = (time = 250) => new Promise(resolve => setTimeout(resolve, time))
+export const timeout = async (fn, time = 250) => {
+  await sleep(time)
+  fn()
+}
```
JeremyWuuuuu, line 20: I will recommend you to use `jest.useFakeTimer` instead of `sleep`  check this out: [jest.useFakeTimers](https://jestjs.io/docs/en/jest-object#jestusefaketimersimplementation-modern--legacy)

packages/drawer/src/popup.ts
```diff
@@ -0,0 +1,228 @@
...
+
+// XXX this may be replaced with other usePopup as Dialog?
+const usePopup = ({ props, $el, vVm }) => {
+  let _popupId: string
+  let _opening = false
+  let _closing = false
+  // let _closeTimer = null // XXX These code will be removed ？
+  // let _openTimer = null
+
+  const data = reactive({
```
JeremyWuuuuu, line 32: I think the name `states` or `drawerState` is better than `data`

packages/drawer/src/popup.ts
```diff
@@ -0,0 +1,228 @@
...
+    opened: false,
+    bodyPaddingRight: null,
+    computedBodyPaddingRight: 0,
+    withoutHiddenClass: true,
+    rendered: false,
+  })
+
+  function open(options = {}) {
```
JeremyWuuuuu, line 40: You need to define the `options`'s type

packages/drawer/src/drawer.vue
```diff
@@ -0,0 +1,238 @@
...
+        }
+      },
+    )
+
+    // onUnmounted(() => {
+    //   // if appendToBody is true, remove DOM node after destroy
+    //   if (props.appendToBody && root && root.value.parentNode) {
+    //     root.value.parentNode.removeChild(root.value)
+    //   }
+    // })
```
JeremyWuuuuu, line 222: You may remove these comments since you have already implemented it with `Teleport` API

packages/drawer/src/drawer.vue
```diff
@@ -0,0 +1,238 @@
...
+    watch(
+      () => props.visible,
+      val => {
+        if (val) {
+          closed.value = false
+          ctx.emit('open')
+          // if (props.appendToBody) {
+          //   document.body.appendChild(root.value)
+          // }
+          prevActiveElement.value = document.activeElement as HTMLElement
```
JeremyWuuuuu, line 202: I actually had a composition hook in #197, but lets stick with this for now, I'll do the optimizations after this gets merged

packages/drawer/src/drawer.vue
```diff
@@ -0,0 +1,238 @@
...
+
+    function handleClose() {
+      // This method here will be called by PopupManger, when the `closeOnPressEscape` was set to true
+      // pressing `ESC` will call this method, and also close the drawer.
+      // This method also calls `beforeClose` if there was one.
+      closeDrawer()
+    }
+
+    watch(
+      () => props.visible,
```
JeremyWuuuuu, line 194: I think we need to use the new `v-model` API for now, since `.sync` modifier has already been deprecated. Refer to [.sync](https://v3.vuejs.org/guide/migration/v-model.html#using-v-bind-sync)

packages/drawer/src/index.vue
```diff
@@ -0,0 +1,29 @@
...
+    const drawer = ref<typeof Drawer>(null)
+    const appendToBody = computed(() => ctx.attrs['append-to-body'])
+    return {
+      drawer,
+      appendToBody,
+      closeDrawer: () => drawer.value.closeDrawer(),
+    }
+  },
+})
+</script>
```
JeremyWuuuuu, line 29: By wrapping `Drawer` like this we ended up with extra overheads which is rendering an additional components (including all lifecycles), I think we should keep `Drawer` minimal when you actually do not need to wrap it up like this, also `functional` components is recommended if you insist doing so.

packages/drawer/src/popup.ts
```diff
@@ -0,0 +1,228 @@
...
+  hasClass,
+} from '@element-plus/utils/dom'
+import { merge } from 'lodash'
+import getScrollBarWidth from '@element-plus/utils/scrollbar-width'
+
+let idSeed = 1
+let scrollBarWidth: number
+
+// XXX this may be replaced with other usePopup as Dialog?
+const usePopup = ({ props, $el, vVm }) => {
```
JeremyWuuuuu, line 25: You need to specify the type here.

packages/drawer/src/popup.ts
```diff
@@ -0,0 +1,228 @@
...
+          open()
+        }
+      } else {
+        close()
+      }
+    },
+  )
+  return { popupData: data, open }
+}
+
```
JeremyWuuuuu, line 227: We need to replace this with `modalValue` A.K.A. `v-model`

jw-foss (changes requested):
Even though there are some code that we do not need for v3 that I have implemented in #197, let's just fix the places that I pointed out, and I'll do the next job of extracting the common code

packages/drawer/__tests__/drawer.spec.ts
luxueyan: ok!

packages/drawer/src/popup.ts
luxueyan: ok

packages/drawer/src/index.vue
luxueyan: Ok! But the project don't seem to support jsx. Using functional component looks like a painful work. Maybe I 
 have to use a string template for the conditional switch template. Is that a ok resolution?

packages/drawer/src/drawer.vue
luxueyan: My understanding is to simplely replace 'visible' with 'modelValue' and 'update:visible' with 'update:modelValue'?

packages/drawer/src/popup.ts
luxueyan: Got it.

packages/drawer/src/popup.ts
luxueyan: Like the above .Just simply replace it with 'modelValue‘?

luxueyan:
Maybe I have made a mistake.Sorry for the force push. I didn’t figure out why it diverged every time I rebase from dev.

JeremyWuuuuu:
> Maybe I have made a mistake.Sorry for the force push. I didn’t figure out why it diverged every time I rebase from dev.

Don't worry, diverging is correct result, because you both had different commits

jw-foss (approved):

JeremyWuuuuu:
@luxueyan , you need to rebase `dev` then force push again, then we can merge this

luxueyan:
> @luxueyan , you need to rebase `dev` then force push again, then we can merge this

Ok. I have rebased.

PR #250, opened 2020-09-02
test(utils): remove utils/jestMock.ts

Ryan2128:
- [ ] Component Migration

When I wrote `jestMock`, I missed this folder.

Now move `jestMock` to `test-utils/mock.ts`.


packages/test-utils/jest-spyOn.ts
```diff
@@ -1,4 +1,4 @@
-export const jestMock = (object: any, method: string, implementation: () => any, accessType?: 'get' | 'set') => {
+export const jestSpyOn = (object: any, method: string, implementation: () => any, accessType?: 'get' | 'set') => {
```
SimonaliaChen, line 1: I don't think this wrapper is necessary since it's  just a shorthand for a chain calls.

packages/test-utils/jest-spyOn.ts
```diff
@@ -1,4 +1,4 @@
...
   let mockSpy = null
   if(!accessType){
```
SimonaliaChen, line 3: This condition judgement seems redundant. `jest.spyOn` already has take `accessType: undefined` into condiseration

Ryan2128: The declear of this function is below:
image

I think it is more clear though the implement of `spyOn` already check the `accessType`

packages/test-utils/jest-spyOn.ts
Ryan2128: You can use chain calls directly, I think it is too long, so extract it.

packages/test-utils/jest-spyOn.ts
SimonaliaChen: > The declear of this function is below:
> image
> 
> I think it is more clear though the implement of `spyOn` already check the `accessType`

You can do it with ts declaration too but not in function implemention. BTW, I still do NOT think it's necessary to extract `jestSpyOn`

Ryan2128: Actually Jerry Wuuuuu asked me to extract it, I'll take a look again and delete it if it is not necessary.

bastarder: emmm , i dont think it's necessary too.

Ryan2128: OK, I will delete this file in this PR later.

jw-foss (approved):

Issue #162, opened 2020-08-19
☝️ Breaking Change List


zazzaz:
Here is the list of breaking changes from Element 2.x to Element Plus

zazzaz:
## Time Picker

### Attributes
1. Delete picker-options (Move picker-options to component tag)
2. Delete picker-options.format (Use the only picker format)
3. Delete picker-options.selectableRange (New selectableRange API )
4. Add disabledHours disabledMinutes disabledSeconds APIs

bastarder:
## Timeline

### Attributes
1. Delete reverse (You should update data to reverse time-line，not bind props reverse)

JeremyWuuuuu:
## Tooltip

### Attributes
1. popperOptions: since we have replaced the previous version to `@popperjs/core` v2, the popperOptions are completely different from v1.
~~~diff
--- popper.js parameters
+++ once the user has provided this option, the popper is under the user's control, `Popper` will not initialize `popper` from the default options anymore, all attributes such as `offset` will not be effective
~~~

xiaoeyuZZZ:
## Popconfirm

### Attributes
1. onConfirm is replaced by confirm
2. onCancel is replaced by cancel

zazzaz:
## Calendar

### Attributes
1. value / v-model Date Object only
2. range Date Object only, do not have to be start day of week / end day of week
3. remove first-day-of-week, use Day.js locale instead
4. slot data add date (JS Date)

### Doc
How to set Day.js locale

zazzaz:
## Date Picker

### Attributes
1. remove first-day-of-week, use Day.js locale instead

### Doc
Update editable description

HerringtonDarkholme:
# Form

1. validate now returns a promise that reject invalidFields instead of a single boolean if no callback is provided.
2. the library now assumes Promise is always available and users should provide a sensible polyfill if they need ever to support deprecate browsers at all.

zazzaz:
## Locale

1. Default language English
2. Locale file name in lower case

zazzaz:
## TimePicker

#### Attributes

- Delete picker-options option (Move to top-level options separately)

edsolater:
> 
> ## Date Picker
> ### Attributes
> 
>     1. remove first-day-of-week, use Day.js locale instead
> 
>     2. Delete picker-options option (Move to top-level options separately)
> 
>     3. Add shortcut option, change accepted values [{text, value}]
> 
>     4. Add disabledDate option
> 
[...]

PR #252, opened 2020-09-03
feat(popconfirm:100%)

xiaoeyuZZZ:
- [ ] Component Migration

Fix #


packages/popconfirm/src/index.vue
```diff
@@ -0,0 +1,103 @@
...
+        type: String
+      },
+      confirmButtonText: {
+        type: String,
+        default: '确定'
+      },
+      cancelButtonText: {
+        type: String,
+        default: '取消'
+      },
```
Ryan2128, line 56: Default value should use locale.

packages/popconfirm/src/index.vue
```diff
@@ -0,0 +1,103 @@
...
+        default: false
+      }
+    },
+    components: {
+      ElPopper,
+    },
+    setup(props,ctx){
+      const visible = ref(false)
+
+      const confirm = (): void => {
```
Ryan2128, line 84: `void` is not need.

Same for the following.

packages/popconfirm/src/index.vue
```diff
@@ -0,0 +1,103 @@
...
+        visible.value = false;
+          ctx.emit('onConfirm');
```
Ryan2128, line 86: I think event name should be uniform and not start with `on`, thougn it maybe a breaking change.

Same for the following.

Ryan2128 (changes requested):
Please lint your code firstly. Many point I have not comment, like spacing after comma, extra semicolon.

packages/popconfirm/src/index.vue
Ryan2128: @JeremyWuuuuu are you agree this point?

JeremyWuuuuu: Yep, ur right, BTW, the correct grammar here is "Do you agree", when it comes with a verb you use "Do/Did/Does" instead of "Am/Are/Is"

packages/popconfirm/src/index.vue
JeremyWuuuuu: Correct, typescript automatically infers the return type as `void`, it is redundant.
BTW "is not need(ed)"  a passive statement here.

JeremyWuuuuu:
@xiaoeyuZZZ You need to fix your format before requesting for review. @Ryan2128 let him finish what needs to be done, then you can do your review here.

packages/popconfirm/src/index.vue
Ryan2128: An inattentive error.😂

Ryan2128:
> @xiaoeyuZZZ You need to fix your format before requesting for review. @Ryan2128 let him finish what needs to be done, then you can do your review here.

✔

packages/popconfirm/src/index.vue
xiaoeyuZZZ: Can we be multilingual too

packages/popconfirm/src/index.vue
xiaoeyuZZZ: Ok, I will modify this

packages/popconfirm/src/index.vue
Ryan2128: Yes, `locale` should be used here to support internationalization. `t` is in `packages/locale/index.ts`.

bastarder:
is there any unit test?

xiaoeyuZZZ:
> is there any unit test?

No unit tests, manual sample tests, weekend unit tests

Ryan2128:
Please add breaking change to #162 about `popconfirm`'s event.

xiaoeyuZZZ:
> Please add breaking change to #162 about `popconfirm`'s event.

No major events have changed

Ryan2128:
> > Please add breaking change to #162 about `popconfirm`'s event.
> 
> No major events have changed

`onConfirm` -> `confirm`
`onCancel` -> `cancel`

packages/popconfirm/src/index.vue
xiaoeyuZZZ: ok

Issue #254, opened 2020-09-04
Checkbox: lack default value of inject


Ryan2128:
The newest code lacks default value of inject of `elForm`, `elFormItem`, `CheckboxGroup`.

```
const elForm = inject<any>('elForm')
const elFormItem = inject<any>('elFormItem')
const checkboxGroup = inject<ICheckboxGroupInstance>('CheckboxGroup')
```

![image](https://user-images.githubusercontent.com/33176053/92230325-82ed6300-eedd-11ea-9e22-2d6cc77b3866.png)

@zouhangwithsweet

PR #255, opened 2020-09-04
Feat/transfer

### LINKED ISSUE #161: ### Transfer
The Transfer component denpends on Input component, I want to begin to refactor this component after Input refactored.

Ryan2128:
- [x] Component Migration

Fix #161 

Refactoring progress:
- refactor component ✔
- add test file ✔
- add docs file ✔

Two bugs about other components:
-  `Input`: `$attrs`'s behavior
- ~~`Checkbox`: lack default value of inject #254~~  ✔


Ryan2128:
Ready for review, PTAL.

Ryan2128:
@HerringtonDarkholme Do you have time to review my PR now? I cannot request a review.

packages/transfer/src/option-content.vue
```diff
@@ -0,0 +1,26 @@
+<script lang="ts">
```
HerringtonDarkholme, line 1: I would suggest preferring to use plain ts over vue if there is no tempalte.

packages/transfer/src/option-content.vue
```diff
@@ -0,0 +1,26 @@
...
+  },
+  setup() {
+    const defaultScopedSlots: ComputedRef<undefined | Slot> = inject('defaultScopedSlots')
+    return {
+      defaultScopedSlots,
+    }
+  },
+  render() {
+    return this.renderContent
+      ? this.renderContent(h, this.option)
```
HerringtonDarkholme, line 20: `h` is context-free in Vue 3 so we don't need to pass it to `renderContent`

packages/transfer/src/transfer-panel.vue
```diff
@@ -0,0 +1,177 @@
...
+        >
+          <option-content
+            :option="item"
+            :render-content="renderContent"
+            :label-prop="labelProp"
+            :key-prop="keyProp"
+          />
+        </el-checkbox>
+      </el-checkbox-group>
+      <p v-show="hasNoMatch" class="el-transfer-panel__empty">
```
HerringtonDarkholme, line 54:
```suggestion
      <p v-show="hasNoMatch || data.length === 0" class="el-transfer-panel__empty">
        {{hasNoMatch ?  t('el.transfer.noMatch') :  t('el.transfer.noData') }}
      </p>
```

packages/transfer/src/transfer-panel.vue
```diff
@@ -0,0 +1,177 @@
...
+    format: Object,
+    filterMethod: Function,
+    defaultChecked: Array,
+    props: Object,
+  },
+
+  emits: [CHECKED_CHANGE_EVENT],
+
+  setup(props, { emit, slots }) {
+    const initData = reactive({
```
HerringtonDarkholme, line 111: `initData` looks like a misleading variable name, since we will assign `initData.query` to it. `panelState` might be a better name.

packages/transfer/src/index.vue
```diff
@@ -0,0 +1,235 @@
...
+      },
+    },
+    rightDefaultChecked: {
+      type: Array,
+      default() {
+        return []
+      },
+    },
+    renderContent: {
+      type: Function as PropType<(option) => VNode>,
```
HerringtonDarkholme, line 123: My bad, renderContent is a public API and its first argument is `h` render function. We cannot remove it here for backward compatibility....

packages/transfer/src/index.vue
```diff
@@ -0,0 +1,235 @@
...
+
+    const leftPanelTitle = computed(() => props.titles[0] || t('el.transfer.titles.0'))
+
+    const rightPanelTitle = computed(() => props.titles[1] || t('el.transfer.titles.1'))
+
+    const panelFilterPlaceholder = computed(() => props.filterPlaceholder || t('el.transfer.filterPlaceholder'))
+
+    watch(() => props.modelValue, (val: Key[]) => emit(UPDATE_MODEL_EVENT, val))
+
+    provide('defaultScopedSlots', computed(() => slots.default))
```
HerringtonDarkholme, line 209: I would suggest to use an `optionRender` for `renderContent` props.

```suggestion
  const optionRender = computed(() => option => {
     if (props.renderContent) return props.renderContent(h, option)
     if (slots.default) return slots.default(option)
     return // default implementation
  })
```

HerringtonDarkholme: To render vnodes in template, we can use something like

```ts
components: {
  VNodes: ({vnodes}) => vnodes
}
```

And in template

```
<VNodes :vnodes="rendered">
```

packages/transfer/src/index.vue
```diff
@@ -0,0 +1,235 @@
...
+      :placeholder="panelFilterPlaceholder"
+      :title="leftPanelTitle"
+      :filterable="filterable"
+      :format="format"
+      :filter-method="filterMethod"
+      :default-checked="leftDefaultChecked"
+      :props="props"
+      @checked-change="onSourceCheckedChange"
+    >
+      <template #default>
```
HerringtonDarkholme, line 16: No need to use template default here.

packages/transfer/src/index.vue
```diff
@@ -0,0 +1,235 @@
...
+    return {
+      sourceData,
+      targetData,
+      onSourceCheckedChange,
+      onTargetCheckedChange,
+      addToLeft,
+      addToRight,
+
+      leftChecked,
+      rightChecked,
```
HerringtonDarkholme, line 225:
```suggestion
   ...toRefs(initData)
```

packages/transfer/src/index.vue
```diff
@@ -0,0 +1,235 @@
...
+
+  emits: [
+    UPDATE_MODEL_EVENT,
+    CHANGE_EVENT,
+    LEFT_CHECK_CHANGE_EVENT,
+    RIGHT_CHECK_CHANGE_EVENT,
+  ],
+
+  setup(props, { emit, slots }) {
+    const initData = reactive({
```
HerringtonDarkholme, line 168: Similarly `checkedState` is better

packages/transfer/src/index.vue
Ryan2128: I just have tried it, but falied. I can print `option` in `optionRender`, but vnodes was not apeared. Did I write something wrong?

image

image

image

image

HerringtonDarkholme: try this?

```
  const optionRender = computed(() => {
     if (props.renderContent) return (option) => props.renderContent(h, option)
     if (slots.default) return option => slots.default(option)
     return option => // default implementation
  })
```

HerringtonDarkholme: You can also add some logging in OptionContent's body.

Ryan2128: I found the reason why the label is not displayed, because I used `:v-nodes=""`, which may be regarded as a drective.

After I changed it, normal render was right, but `slot.default` and `renderContent` both failed. 

The Console's log is in below and you can also use this [test branch](https://github.com/Ryan2128/element-plus/tree/test/transfer) to reproduce it :

image

Ryan2128: OK, I has fixed it, the reason is the param of slot.

Ryan2128:
PTAL, all requested changes has been done. @HerringtonDarkholme

HerringtonDarkholme (approved):

HerringtonDarkholme:
@Ryan2128 I have updated several test cases. Please check out. Generally we don't allow sleep or setTimeout in test code.

Ryan2128:
> @Ryan2128 I have updated several test cases. Please check out. Generally we don't allow sleep or setTimeout in test code.

OK, I know.

Ryan2128:
@JeremyWuuuuu PTAL. 

If there is no problem, I think it can be merged

PR #262, opened 2020-09-06
Fix/steps: fix error in deploy

Ryan2128:
- [x] Bug Fix

- add `change` event

  `change` event has been removed when refactoring. I think there is no necessary to do a breaking change.

- fix error in deploy
  
  image

### DIFF
--- a/packages/steps/src/index.vue
+++ b/packages/steps/src/index.vue
@@ -7,6 +7,8 @@
 <script lang="ts">
 import { defineComponent, watch, ref, provide } from 'vue'
 
+const CHANGE_EVENT = 'change'
+
 export default defineComponent({
   name: 'ElSteps',
   props: {
@@ -42,7 +44,8 @@ export default defineComponent({
       validator: (val: string): boolean => ['wait', 'process', 'finish', 'error', 'success'].indexOf(val) > -1,
     },
   },
-  setup(props) {
+  emits: [CHANGE_EVENT],
+  setup(props, { emit }) {
     const steps = ref([])
 
     watch(steps, () => {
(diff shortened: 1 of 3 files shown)


packages/steps/src/index.vue
```diff
@@ -7,6 +7,8 @@
 <script lang="ts">
 import { defineComponent, watch, ref, provide } from 'vue'
 
+const CHANGE_EVENT = 'change'
```
JeremyWuuuuu, line 10: I need you to move this to `utils/constants`, later we will replace all the event constants with this

packages/steps/src/item.vue
```diff
@@ -58,7 +58,7 @@ interface IStepsProps {
 
 interface IStepInstance extends ComponentInternalInstance {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  ctx: any
+  setupState: any
```
JeremyWuuuuu, line 61: Do you mind enhancing the type definitions here? Since you already know what has been returned by the setups
BTW, this is not a mandatory request, if you do not have the time or you do not feel to do this, you can simply resolve my request here

jw-foss (changes requested):

packages/steps/src/index.vue
Ryan2128: OK, I will add it later.

packages/steps/src/item.vue
Ryan2128: OK, I will add it later.

JeremyWuuuuu: Thanks @Ryan2128 !

packages/steps/src/index.vue
```diff
@@ -42,17 +43,22 @@ export default defineComponent({
       validator: (val: string): boolean => ['wait', 'process', 'finish', 'error', 'success'].indexOf(val) > -1,
```
JeremyWuuuuu, line 43: Can we use `includes` here? `indexOf` is a bit of old fashioned way for `includes`

jw-foss (changes requested):

jw-foss (approved):

PR #267, opened 2020-09-07
feat(infinite-scroll): add inifinite scroll

### LINKED ISSUE #111: ### InfiniteScroll

SimonaliaChen:
- [x] Component Migration

Fix #111


packages/test-utils/make-scroll.ts
```diff
@@ -1,8 +1,9 @@
 import { sleep } from '@element-plus/test-utils'
 
 const makeScroll = async (dom: Element, name: 'scrollTop' | 'scrollLeft', offset: number) => {
+  const eventTarget = dom === document.documentElement ? window : dom
   dom[name] = offset
-  dom.dispatchEvent(new CustomEvent('scroll'))
+  eventTarget.dispatchEvent(new CustomEvent('scroll'))
```
Ryan2128, line 6: Because `new CustomEvent()` can not be used in IE11 and vue 3 resease version will be compatible with IE11, so I think should be judged whether it is IE11 for compatibility.

image

Ryan2128 (changes requested):

packages/test-utils/make-scroll.ts
SimonaliaChen: > Because `new CustomEvent()` can not be used in IE11 and vue 3 resease version will be compatible with IE11, so I think should be judged whether browser is IE11 for compatibility.
> 
> image

You're right, I will raise another pr to fix it later

SimonaliaChen: Sorry, I was out of my mind earlier. It's no need to worry about the compatibility for that it just a mock in test. @Ryan2128

Ryan2128: YES, I also didn't notice it.

packages/infinite-scroll/src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+const InfiniteScroll: ObjectDirective<InfiniteScrollEl, InfiniteScrollCallback> = {
+  async mounted(el, binding) {
+    const { instance, value: cb } = binding
+
+    if (!isFunction(cb)) {
+      throwError(SCOPE, '\'v-infinite-scroll\' binding value must be a function')
+    }
+
+    // ensure parentNode mounted
+    await nextTick()
```
JeremyWuuuuu, line 113: I think at the time which `directive` gets attached, the DOM should already be mounted, I don't think you need to do the extra `nextTick` right here. Refer to https://v3.vuejs.org/guide/migration/custom-directives.html#_3-x-syntax, you could see that according the documentation, by the time `beforeMounted` gets invoked, `el` has already been mounted on DOM tree. have you encountered any problems here?

packages/infinite-scroll/src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+    if (!container) return
+
+    el[SCOPE] = { instance, container, containerEl, cb, onScroll }
+
+    if (immediate) {
+      const observer = new MutationObserver(throttle(checkFull.bind(null, el, cb), CHECK_INTERVAL))
+      el[SCOPE].observer = observer
+      observer.observe(el, { childList: true, subtree: true })
+      checkFull(el, cb)
+    }
```
JeremyWuuuuu, line 129: What happens if `immediate` is `false`?

jw-foss (changes requested):
So for your testing cases, I strongly suggest you to use `jest.useFakeTimer` instead of `sleep`, because simply **sleep** you r program is not how we wanted to test our code's functionality. You can also mock `_.throttle` like I did. Basically, using `sleep` is a bad habit that could potentially cover up the edge cases we did not know.

packages/infinite-scroll/src/index.ts
SimonaliaChen: The purpose of `infinite-scroll-immediate` is whether to execute the load method immediate in order to avoid that the scroll container is not overflow. If `immediate` is `false`, no such check will be done. The overflow state of scroll container all depends on outsiders.

SimonaliaChen:
> So for your testing cases, I strongly suggest you to use `jest.useFakeTimer` instead of `sleep`, because simply **sleep** you r program is not how we wanted to test our code's functionality. You can also mock `_.throttle` like I did. Basically, using `sleep` is a bad habit that could potentially cover up the edge cases we did not know.

I think it's approviate to use `sleep` in my case because I calculated the time accurately to verify that `infinite-scroll-delay` is working. But It's kind of redundant and a waste of time. I have tried test `throttle` with `jest.useFakeTimer`, but it doesn't work. I will try to mock `_.throttle` and verify `infinite-scroll-delay`  in another way  instead.

packages/infinite-scroll/src/index.ts
SimonaliaChen: Truely, when `mounted` gets invoked, the current DOM has been mounted. But, there is no guarantee that all the parent nodes have been mounted. So I have to wait a moment until all the parent nodes are mounted in order to get the scroll container which might not included in el. I haven't read the source code and not sure its reliability, but it does work in my case.

JeremyWuuuuu: Well, this is my mistake, the way how you handling this is correct.

packages/infinite-scroll/src/index.ts
JeremyWuuuuu: I have checked the rest of your code, the only place you use `observer.observe` is here, if this is just a single call, I think I'm ok with this

JeremyWuuuuu:
> > So for your testing cases, I strongly suggest you to use `jest.useFakeTimer` instead of `sleep`, because simply **sleep** you r program is not how we wanted to test our code's functionality. You can also mock `_.throttle` like I did. Basically, using `sleep` is a bad habit that could potentially cover up the edge cases we did not know.
> 
> I think it's approviate to use `sleep` in my case because I calculated the time accurately to verify that `infinite-scroll-delay` is working. But It's kind of redundant and a waste of time. I have tried test `throttle` with `jest.useFakeTimer`, but it doesn't work. I will try to mock `_.throttle` and verify `infinite-scroll-delay` in another way instead.

Well, what I was talking about is that you can utilize `fakeTimers` with advanced APIs, because I think your code should not be depended on how `setTimeout` or `throttle` implemented, because the functionality is guaranteed by the provider, the only thing you need to care about is "whether you invoked the function correctly or not", that to be said `de-coupled`, no matter how the function was implemented, you should only care about how your code gets processed. You can check the documentation here [Jest.useFakeTimers](https://jestjs.io/docs/en/jest-object#jestadvancetimersbytimemstorun) this section covers what I believe what you tried to convince me about.

SimonaliaChen:
> > > So for your testing cases, I strongly suggest you to use `jest.useFakeTimer` instead of `sleep`, because simply **sleep** you r program is not how we wanted to test our code's functionality. You can also mock `_.throttle` like I did. Basically, using `sleep` is a bad habit that could potentially cover up the edge cases we did not know.
> > 
> > 
> > I think it's approviate to use `sleep` in my case because I calculated the time accurately to verify that `infinite-scroll-delay` is working. But It's kind of redundant and a waste of time. I have tried test `throttle` with `jest.useFakeTimer`, but it doesn't work. I will try to mock `_.throttle` and verify `infinite-scroll-delay` in another way instead.
> 
> Well, what I was talking about is that you can utilize `fakeTimers` with advanced APIs, because I think your code should not be depended on how `setTimeout` or `throttle` implemented, because the functionality is guaranteed by the provider, the only thing you need to care about is "whether you invoked the function correctly or not", that to be said `de-coupled`, no matter how the function was implemented, you should only care about how your code gets processed. You can check the documentation here [Jest.useFakeTimers](https://jestjs.io/docs/en/jest-object#jestadvancetimersbytimemstorun) this section covers what I believe what you tried to convince me about.

Already fixed, PTAL @JeremyWuuuuu

jw-foss (approved):

PR #268, opened 2020-09-07
Feat/color picker

Hades-li:
- [x] Component migration

Fix # 132


JeremyWuuuuu:
Sorry, if you are not familiar with `Git`, you should probably check this out https://git-scm.com/ it is free, you need to exclude commits which were not committed by you

Hades-li:
this pr is discard

PR #308, opened 2020-09-15
fix(core): Remove website/play/index.vue out of commit scope

jw-foss:
- [ ] Component Migration

Fix #


g3r4n (commented):
Why do we want to remove this file ? can you fix the conflict ?

JeremyWuuuuu:
@g3r4n So this PR is meant to removing it from git tracking (1st step), then create a same file via Github GUI, thus developers can obtain a copy of `play.vue` and also making git not tracking this file, IMHO, this file would be modified constantly because building the full size website costs more time and consumes more memory, it is a light weighted way to develop feature/fix bug.

zazzaz (approved):

Issue #235, opened 2020-08-31
Discussion about v-bind="$attrs"


Ryan2128:
In vue3, `$attrs` includes `style` and `class`, but not in vue2. In vue3, one of its uses is pass attrs to `Fragment`, so it should include `style` and `class`. 

There is at least a styling error if use `v-bind="$attrs"` apply to a child node which has single root node.

Just like:
[First demo](https://github.com/element-plus/element-plus/pull/216#pullrequestreview-478143604)
[Second demo](https://github.com/element-plus/element-plus/pull/234#issue-476275046)

So I think it is necessary to discuss the question about `v-bind="$attrs"`.

I think this is actually a logical problem of component style.

But due to the CSS is not ready and many migrate task still need to be done, I think we should separate the `$attrs` at this stage instead of refactoring CSS.

Maybe we need open a new issue and write some temporary agreement at this stage?

cc @JeremyWuuuuu @zouhangwithsweet @zazzaz @hanx316

JeremyWuuuuu:
@Ryan2128 @HerringtonDarkholme @zazzaz @zouhangwithsweet @bastarder @SimonaliaChen 

Hey, guys, I do think this issue is valuable due to the current issue we have been encountering with.

With the  change of `Vue` itself,  the element now inherit attributes like `style` `class` etc, to the component's root which was not the same behavior in **Vue 2.x**.

Due to this issue, once the child templates uses the syntax `v-bind="$attrs"` would inherit properties such as `style` `class` unexpectedly.

So we must have to come up with a solution for this issue like in #234, the ideal way is not preventing user to pass down properties like `class` to child component by disabling the `inheritAttrs` option, we need to prevent the children inheriting unwanted properties. We need a hook that essentially excludes attributes like `class` `style` etc out of the setup context.

Let's keep tracking on this issue.

cc @SimonaliaChen can you help with this?

SimonaliaChen:
> @Ryan2128 @HerringtonDarkholme @zazzaz @zouhangwithsweet @bastarder @SimonaliaChen
> 
> Hey, guys, I do think this issue is valuable due to the current issue we have been encountering with.
> 
> With the change of `Vue` itself, the element now inherit attributes like `style` `class` etc, to the component's root which was not the same behavior in **Vue 2.x**.
> 
> Due to this issue, once the child templates uses the syntax `v-bind="$attrs"` would inherit properties such as `style` `class` unexpectedly.
> 
[...]

PR #317, opened 2020-09-16
feat(hooks): add useAttrs hook

SimonaliaChen:
Smooth out the differences between vue2 and vue3 when using `$attrs`,  `class` `style` and other specific attributes will not bind to child node but just root node.

Fix  #234 , #235

### DIFF
--- a/packages/hooks/use-attrs/index.ts
+++ b/packages/hooks/use-attrs/index.ts
@@ -1,69 +1,16 @@
 import {
   getCurrentInstance,
-  onMounted,
   reactive,
   shallowRef,
-  watch,
   watchEffect,
 } from 'vue'
-import { entries, isUndefined, isEqual } from '@element-plus/utils/util'
-import { addClass, removeClass, setStyle, removeStyle } from '@element-plus/utils/dom'
-import type { ComponentInternalInstance } from 'vue'
-
-interface Attrs extends Record<string, unknown> {
-  class?: string
-  style?: Hash<string>
-}
+import { entries } from '@element-plus/utils/util'
 
 const DEFAULT_EXCLUDE_KEYS = ['class', 'style']
 const LISTENER_PREFIX = /^on[A-Z]/
 
-function syncAttribute<T extends string, P = Attrs[T]>(
-  instance: ComponentInternalInstance,
-  attributeName: T,
-  handler: (instance: ComponentInternalInstance, newVal: P, oldVal?: P) => void,
-) {
-  const attrs = instance.attrs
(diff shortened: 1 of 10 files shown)


packages/hooks/use-attrs/index.ts
```diff
@@ -0,0 +1,95 @@
...
+          !(excludeListeners && LISTENER_PREFIX.test(key))
+        ) {
+          acm[key] = val
+        }
+
+        return acm
+      }, {})
+
+    attrs.value = res
+  })
```
JeremyWuuuuu, line 87: To me this seems a little overkill for this feature, because you have you idea when the `attrs` gets changed, which means you run this whenever the component gets updated.

JeremyWuuuuu: IMHO holding a cache of `attrs` to avoid unnecessary updates could solve the problem. 
BTW, since you have made `attrs` reactive, can you use `watch(() => attrs)` that only updates when `attrs` gets changed?
Also, I think at this particular point, you could set `instance.type.inheritAttrs` to `false` to avoid automatically inheritance to `root element` 
These were all based on my understanding, correct me when you feel like you should.

SimonaliaChen: You may have some misunderstanding of `watchEffect` , it only calls when dependencies change but not whenever component updates, so `watchEffect` could make same effect as `watch(() => attrs)`. And I'm afraid that I can't argree with you about modifying `instance.type.inheritAttrs`, cause it's kind of impure and unnecessary.

JeremyWuuuuu:
LGTM.

packages/input/__tests__/input.spec.ts
```diff
@@ -2,6 +2,7 @@ import { ref } from 'vue'
 import { mount } from '@vue/test-utils'
 import { sleep, defineGetter } from '@element-plus/test-utils'
 import Input from '../src/index.vue'
+import input from '..'
```
JeremyWuuuuu, line 5: What is this for?

jw-foss (changes requested):

packages/input/__tests__/input.spec.ts
JeremyWuuuuu: Remove this unused code.

SimonaliaChen: Done. It should have been imported automatically by accident.

HerringtonDarkholme (approved):

jw-foss (approved):

PR #319, opened 2020-09-17
feat(popper): allow the use of element to dynamically set the trigger

justwiner:
- [ ] Popper

Fix #

### DIFF
--- a/packages/popper/__tests__/popper.spec.ts
+++ b/packages/popper/__tests__/popper.spec.ts
@@ -335,5 +335,31 @@ describe('Popper.vue', () => {
       await trigger.trigger(CLICK_EVENT)
       expect(popper.vm.visibility).toBe(false)
     })
+
+    test('change trigger', async () => {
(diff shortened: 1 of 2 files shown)


JeremyWuuuuu:
@justwiner 
Manually attach reference trigger the user need to attach events and manipulate the popper's state, I do not think this is a good way, you could see the code in `renderer/trigger.ts`, normally we clone `VNode` given by the user and attach events via `cloneVNode`, in your way of implementing this there will be no chance that we could attach events to the reference node. Is this what you wanted?

justwiner:
> @justwiner
> Manually attach reference trigger the user need to attach events and manipulate the popper's state, I do not think this is a good way, you could see the code in `renderer/trigger.ts`, normally we clone `VNode` given by the user and attach events via `cloneVNode`, in your way of implementing this there will be no chance that we could attach events to the reference node. Is this what you wanted?

yes, you're right
I will improve it later

PR #331, opened 2020-09-19
fix(steps): fix bug cause by setupState

Ryan2128:
Fix bug cause by setupState.

After changes:
- Passed the test case
- Passed the demo test of element2
- Passed the demo test of `basic.vue` of `Steps`


SimonaliaChen:
Already fixed in #330 which is more elegant IMHO

Ryan2128:
> Already fixed in #330 which is more elegant IMHO

OK, I didn't see it when Caaalabash asked me to fix it.

Issue #169, opened 2020-08-20
Popper bug list


zazzaz:

zazzaz:
It's better to provide a pure popper with no default styles like padding, border color
<img width="396" alt="image" src="https://user-images.githubusercontent.com/10731096/90731903-e6339e80-e2fc-11ea-977a-7103fc9669e2.png">

zazzaz:
wrong arrow direction in `effect="light"`
<img width="404" alt="image" src="https://user-images.githubusercontent.com/10731096/90732159-4591ae80-e2fd-11ea-95c9-3ca188c68187.png">

JeremyWuuuuu:
> It's better to provide a pure popper with no default styles like padding, border color (no transitions maybe?)
> <img alt="image" width="396" src="https://user-images.githubusercontent.com/10731096/90731903-e6339e80-e2fc-11ea-977a-7103fc9669e2.png">

I am considering a new API called `shouldDecorate` as `boolean` to  fix this one.

JeremyWuuuuu:
loop @ Ryan2128 in to get updates

Ryan2128:
> loop @Ryan2128 in to get updates

Should I wait for your update first? I don't understand what you mean too much.

JeremyWuuuuu:
> > loop @Ryan2128 in to get updates
> 
> Should I wait for your update first? I don't understand what you mean too much.

Nope, you can go ahead implement features related to tooltip with popper, the current documentation might be fragmentary, 
Here are some APIs you might need:
1. manualMode (if you want to control the tooltip manually like `manual` in tooltip in v2
2. value (if manualMode is set to true) you are going to need it to control the visibility
3. placement as same as tooltip's in v2
4. effect

Ryan2128:
Ok, thank you for your explanation.😁

JeremyWuuuuu:
ElPopper needs to add a minimum width ~20px to `el-popper` in order to keep the arrow placement correct

Hades-li:
![image](https://user-images.githubusercontent.com/12124478/92063242-54806280-edcd-11ea-96ae-dbaca7a53785.png)
I need v-model one value。but，is not work。
![image](https://user-images.githubusercontent.com/12124478/92063403-a2956600-edcd-11ea-8e9d-d2dbd5991815.png)
maybe you need change this code

JeremyWuuuuu:
[...]

PR #334, opened 2020-09-20
fix(popper): fix popper styling bug/adding transition events

jw-foss:
- Fix styling bug https://github.com/element-plus/element-plus/issues/169#issuecomment-686891549
- Fix trigger bug https://github.com/element-plus/element-plus/issues/169#issuecomment-694720899
- Add events for transition.

### DIFF
--- a/packages/popper/src/usePopper.ts
+++ b/packages/popper/src/usePopper.ts
@@ -10,11 +10,23 @@ import {
 } from 'vue'
 import { createPopper } from '@popperjs/core'
 
-import { generateId, clearTimer, isBool, isHTMLElement, isArray } from '@element-plus/utils/util'
+import {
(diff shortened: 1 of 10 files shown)


packages/popper/src/usePopper.ts
```diff
@@ -125,7 +123,7 @@ export default (props: IPopperOptions, { emit }: SetupContext) => {
   }
 
   function onPopperMouseLeave() {
-    if (props.trigger.length === 1 && props.trigger[0] === 'click') return
+    if ((props.trigger.length === 1 && props.trigger[0] === 'click') || props.trigger === 'click') return
```
bastarder, line 126: what if `focus` ?

bastarder (approved):

PR #336, opened 2020-09-22
Badge组件添加点击事件&button组件click阻止冒泡

xiaofeng-bm:
- [ ] Component Migration

Fix #

### DIFF
--- a/packages/badge/src/index.vue
+++ b/packages/badge/src/index.vue
@@ -12,6 +12,7 @@
             'is-dot': isDot
           }
         ]"
+        @click.stop="handleClick"
         v-text="content"
       >
       </sup>
@@ -54,7 +55,7 @@ export default {
       },
     },
   },
-  setup(props: IBadgeProps): IBadgeSetups {
+  setup(props: IBadgeProps, ctx): IBadgeSetups {
     const content = computed(() => {
       if (props.isDot) {
         return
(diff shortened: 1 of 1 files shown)


JeremyWuuuuu:
Reason for this?

xiaofeng-bm:
> Reason for this?

When I encounter a business that clicks on the Badge component to show some details, I personally think I can add a click event to make it more flexible. Of course, I can turn off the isseue if you don't think it's necessary.

JeremyWuuuuu:
Well, I do not think this is necessary anyway.
If you want to catch the click event, you should wrap `<el-tag>` under some other tag like `div`, then you attach `@click.stop` upon it.

packages/badge/src/index.vue
```diff
@@ -12,6 +12,7 @@
             'is-dot': isDot
           }
         ]"
+        @click.stop="handleClick"
```
JeremyWuuuuu, line 15: BTW, you can just change the code here as
```suggestion
@click.stop="$emit('click')"
```

xiaofeng-bm:
> Well, I do not think this is necessary anyway.
> If you want to catch the click event, you should wrap `<el-tag>` under some other tag like `div`, then you attach `@click.stop` upon it.
> @HerringtonDarkholme @zazzaz any suggestions?
Thanks, you're right, I didn't think of it that way.

PR #369, opened 2020-09-29
fix(tree): replace keyCode

Caaalabash:
+ replace `keyCode`  in `el-tree` component
+ remove unused type `ComponentInternalInstance` in `el-step`
+ some code style.

### DIFF
--- a/packages/utils/aria.ts
+++ b/packages/utils/aria.ts
@@ -120,9 +120,9 @@ const Utils = {
    */
   focusFirstDescendant: function(element: HTMLElement): boolean {
     for (let i = 0; i < element.childNodes.length; i++) {
-      const child = element.childNodes[i]
+      const child = element.childNodes[i] as HTMLElement
       if (
-        attemptFocus(child as HTMLElement) ||
+        attemptFocus(child) ||
         this.focusFirstDescendant(child)
       ) {
         return true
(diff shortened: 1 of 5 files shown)


packages/utils/vnode.ts
```diff
@@ -55,11 +55,7 @@ export const getFirstValidNode = (
   maxDepth = 3,
 ): ReturnType<typeof getChildren> => {
   if (Array.isArray(nodes)) {
-    for (let i = 0; i < nodes.length; i++) {
-      const node = nodes[i] as VNode
-      const child = getChildren(node as VNode, maxDepth)
-      return child
-    }
```
JeremyWuuuuu, line 62: What error have you encountered?

packages/utils/aria.ts
```diff
@@ -120,9 +120,9 @@ const Utils = {
    */
   focusFirstDescendant: function(element: HTMLElement): boolean {
     for (let i = 0; i < element.childNodes.length; i++) {
-      const child = element.childNodes[i]
+      const child = element.childNodes[i] as HTMLElement
       if (
-        attemptFocus(child as HTMLElement) ||
```
JeremyWuuuuu, line 125: what reason did you have on changing this?

jw-foss (commented):
The rest LGTM, I need answers for the two comments

packages/utils/vnode.ts
Caaalabash:
````
    for (let i = 0; i < nodes.length; i++) {
      const node = nodes[i] as VNode
      const child = getChildren(node as VNode, maxDepth)
      return child
    }
````

only take the first node in nodes

packages/utils/aria.ts
Caaalabash:
````
if (attemptFocus(child as HTMLElement) ||  this.focusLastDescendant(child) ) {
   return true
}
````

`focusLastDescendant` got warning: `Argument type ChildNode is not assignable to parameter type HTMLElement `

so change to 

````
const child = element.childNodes[i] as HTMLElement
````

jw-foss (approved):

PR #370, opened 2020-09-29
fix(build): externalize vue dependency as Vue

cheqianxiao:

### DIFF
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -80,7 +80,7 @@ module.exports = {
     extensions: ['.ts', '.tsx', '.js', '.vue', '.json'],
   },
   externals: {
-    vue: 'vue',
+    vue: 'Vue',
   },
   plugins: [
     new VueLoaderPlugin(),


JeremyWuuuuu:
I'm not sure if this change is making any sense, since the current version I didn't see any external `Vue` lib in templates
cc @zazzaz 
image

zazzaz:
We should add a CDN link to index.tpl to make this optimization really work.

cheqianxiao:
The current build version in not working in browser due to this bug. 
[the reproduction link](https://codepen.io/oobakoyo/pen/XWdyGqj?editors=1010). @JeremyWuuuuu 

image

JeremyWuuuuu:
Oh, sorry, you were right. I thought this was the config file under `website/` my bad.

jw-foss (approved):

cheqianxiao:
uh... I was also wrong,  found it not working with webpack after this commit. fixed it again in https://github.com/element-plus/element-plus/pull/380 .

PR #402, opened 2020-10-10
feat(switch): add loading attribute

MBearo:
- [ ] Component Migration

Fix #


packages/switch/src/index.vue
```diff
@@ -195,6 +203,8 @@ export default defineComponent({
       const coreEl = core.value
       coreEl.style.borderColor = newColor
       coreEl.style.backgroundColor = newColor
+      console.log(coreEl.children[0])
```
JeremyWuuuuu, line 206: You need to remove this line.

jw-foss (commented):
I mean, what is the point of adding an attribute like this? We have provided `v-loading` directive that you can easily apply upon any components that you have to give the component `loading` effect. Why adding this anyway?

JeremyWuuuuu:
What kind of problem does this solve? https://ant.design/components/switch/ according to `AntDesign` they had a specific design on this one, but for now we do not have resources for UI Design, I personally cannot give the best suggestion here in order to match the original design language, I do not know if this change breaks the consistency of `ElementUI`'s design language.

MBearo:
> What kind of problem does this solve? https://ant.design/components/switch/ according to `AntDesign` they had a specific design on this one, but for now we do not have resources for UI Design, I personally cannot give the best suggestion here in order to match the original design language, I do not know if this change breaks the consistency of `ElementUI`'s design language.

Ok, maybe I don't fully understand `Element UI`'s design language. My idea is to treat the `switch` component as a special `button` component, switch between true and false, and also have a loading attribute. Just to give users a more delicate experience.

zouhangwithsweet:
I think it's a good feature. Can you show us a gif of the `loading switch`?

zouhangwithsweet:
I think it's a good feature. Can you show us a gif of the `loading switch`?

MBearo:
Screen Recording 2020-10-19 at 09

jw-foss (approved):

JeremyWuuuuu:
Alright, merged. Nice feature!

PR #406, opened 2020-10-12
feat(calendar): add value type String

Tvinsh:
change for v-model and range to support String type, change docs for new feature, and remove no use
test case

- [ calendar] Calendar add some feature


packages/calendar/__tests__/calendar.spec.ts
```diff
@@ -61,29 +72,4 @@ describe('Calendar.vue', () => {
     expect(/2019.*May/.test((titleEl.element as HTMLElement).innerHTML)).toBeTruthy()
     expect(cell.classList.contains('is-selected')).toBeTruthy()
   })
-
```
zazzaz, line 64: Any reason we should delete this test?

zazzaz:
Thanks for the update of the docs.

IMHO, components that are related to date and time, should accept and emit native Date type only. In this way, we can make the best compatibility among all browsers.

packages/calendar/__tests__/calendar.spec.ts
Tvinsh: firstDayOfWeek prop is no use in calendar code

Tvinsh (commented):
> Thanks for updating the docs.
> 
> IMHO, components that are related to date and time, should accept and emit native Date type only. In this way, we can make the best compatibility among all browsers.

so what problems may string value ​​cause

zazzaz:
> so what problems may string value ​​cause

The problem is not the string value itself.  The date object is a better way to dementrating date and time, it contains more information that a simple date string like uct and offset.

packages/calendar/__tests__/calendar.spec.ts
zazzaz: Although it's not used in the code, firstDayOfWeek is still a valid setting configed by dayjs.locale() in the global scope. 

The test is still needed.

Tvinsh:
> > so what problems may string value ​​cause
> 
> The problem is not the string value itself. The date object is a better way to dementrating date and time, it contains more information that a simple date string like uct and offset.

Thanks for replying, I think this does not affect the string value, it can reduce the breaking change, if you do not think this,i will close this pr

zazzaz:
> > > so what problems may string value ​​cause
> > 
> > 
> > The problem is not the string value itself. The date object is a better way to dementrating date and time, it contains more information that a simple date string like uct and offset.
> 
> Thanks for replying, I think this does not affect the string value, it can reduce the breaking change, if you do not think this,i will close this pr

We can keep the part related to the doc update.

JeremyWuuuuu:
> > so what problems may string value ​​cause
> 
> The problem is not the string value itself. The date object is a better way to dementrating date and time, it contains more information that a simple date string like uct and offset.

Agreed, Date Objects are better for containing extra informations but string value is not.

JeremyWuuuuu:
@Tvinsh Could you please revert the part of code which made changes towards the type? We can still use the documenation part.

Tvinsh:
@zazzaz @JeremyWuuuuu Thanks for replying, change is revert

zazzaz:
Cool, thanks.

PR #409, opened 2020-10-13
fix(form): fix form label render

Tvinsh:
- Fix calendar docs <Date> tag 
- Fix form label render, inline style 
- Change the select default size value to make it same as the input default height '40px'
- Fix getRules -> getPropByPath, and now the dynamic doc demo can run without error


Tvinsh:
@zazzaz

Tvinsh:
cc @HerringtonDarkholme

packages/form/src/label-wrap.ts
```diff
@@ -81,7 +81,7 @@ export default defineComponent({
           slots.default?.(),
         )
       } else {
-        return h('div', { ref: el }, slots.default?.())
+        return slots.default?.()
```
JeremyWuuuuu, line 84: Why did you modify this line anyway? After this modification, how do you think that we can using ref to access the element?

packages/form/src/form-item.vue
```diff
@@ -265,8 +265,7 @@ export default defineComponent({
       const selfRules = props.rules
       const requiredRule =
         props.required !== undefined ? { required: !!props.required } : []
-
-      const prop = getPropByPath(formRules, props.prop || '', true)
+      const prop = getPropByPath(formRules, props.prop || '')
```
JeremyWuuuuu, line 268: I simply do not understand why do you have to change this, your modification based on `getPropByPath` is 
~~~diff
-export function getPropByPath(obj: any, path: string, strict: boolean): {
+export function getPropByPath(obj: any, path: string, strict?: boolean):
~~~
Tell me why you need to remove the parameter flag?

packages/select/src/useSelect.ts
```diff
@@ -20,7 +20,7 @@ import {
 } from '@element-plus/utils/util'
 import isEqual from 'lodash/isEqual'
 
-const ELEMENT = { size: 'medium' }
+const ELEMENT = { size: '' }
```
JeremyWuuuuu, line 23: Reason?

packages/utils/util.ts
```diff
@@ -51,7 +51,7 @@ export function getPropByPath(obj: any, path: string, strict: boolean): {
 
   const keyArr = path.split('.')
   let i = 0
-  for (i; i < keyArr.length - 1; i++) {
+  for (const len = keyArr.length; i < len - 1; i++) {
```
JeremyWuuuuu, line 54: Do you think this modification means anything? Was it improving any readability?

jw-foss (changes requested):

JeremyWuuuuu:
I don't see any point of this PR, please detail explain questions that I asked above.

packages/form/src/label-wrap.ts
Tvinsh: you can see [docs](https://element-plus.org/#/en-US/component/form) Inline form
first, inline mode is no need div to wrap label
then, i think the label width is computed in form-item,inline mode is no need this ref

packages/form/src/form-item.vue
Tvinsh: also the doc, please run Delete or add form items dynamically demo and check the console

packages/select/src/useSelect.ts
Tvinsh: the select default size is '', height '40px', if default 'medium', height is '36px', so the default height is different in form

packages/form/src/label-wrap.ts
JeremyWuuuuu: The definition of inline mode is not who the DOM tree looks like, is how we represent it in the UI frame.

packages/utils/util.ts
Tvinsh: Personal habits, and same with v2 version， i can revert this

packages/form/src/label-wrap.ts
Tvinsh: so you think inline mode also needs div wrap? at least you need to change the style

JeremyWuuuuu: I double checked the V2 version, it indeed does not have any 'div' wrapped the label. As far as I could tell, enabling  `inline` mode, the method `getLabelWidth` still get called because it only depends on `isAutoWidth`

packages/select/src/useSelect.ts
JeremyWuuuuu: This one is a general bug, needed to be fixed for all form items.

Tvinsh:
> I don't see any point of this PR, please detail explain questions that I asked above.

okay, already detailed, and you can see [doc](https://element-plus.org/#/en-US/component/form) to find what i mentioned

packages/form/src/label-wrap.ts
JeremyWuuuuu: > so you think inline mode also needs div wrap? at least you need to change the style

What about I pass down a template with structure like
```vue
<div>
  <label>
  </label>
</div>
```

packages/select/src/useSelect.ts
Tvinsh: Validation demo in doc, input near select, so i found this

JeremyWuuuuu: You can refer to #417 , ElementPlus now needs a global configuration to make all attributes required from a single source

packages/form/src/label-wrap.ts
JeremyWuuuuu: Maybe wrap the `$slots.default` with a `div` with `inlined` display property is more flexible on my understanding.

JeremyWuuuuu: Besides this, I've got no questions about this PR

Tvinsh: > Maybe wrap the `$slots.default` with a `div` with `inlined` display property is more flexible on my understanding.

I don’t know if my understanding is correct，if do this, the inner label with `class="el-form-item__label" :style="labelStyle"` is lose it's effect

Issue #420, opened 2020-10-15
The button component has an incorrect type assertion


wozien:
the button prop 'size' should use type assertion with IButtonSize  but now use IButtonType
![image](https://user-images.githubusercontent.com/17634819/96149151-84587500-0ef8-11eb-8881-3882038d308a.png)

zazzaz:
fixed

PR #421, opened 2020-10-15
fix(button): update prop type assertions

wozien:
fix(button): update prop 'size' type assertions to IButtonSize #420


zazzaz:
Thanks

PR #431, opened 2020-10-19
improvement(project): remove mitt #419

### LINKED ISSUE #419: Proposal: giving up a "mitt"
I think "mitt" could be given up. At the moment it's used exclusively to change parent stage through events. This task can be accomplished by reactive provide/inject functionality which is well-described in documentation: https://v3.vuejs.org/guide/composition-api-provide-inject.html#mutating-reactive-properties

In our case take a form example. Right now Form provides mitt instance to make FormItem to be able to call 2 events on it: addField and removeField.

Instead of this we could provide 2 functions from Form (addField and removeField) which are to be called in FormItem to mutate parent state.

In that case we would give up needless dependency and would use vue functionality.

It would be looks like this.
In the form:

image
In the form-item
image
image

If you agree with the idea, I could make a PR a little bit later.

fat4lix:
Giving up mitt from all components. Instead of this, we can use:

1) We can use provide/inject, if we want to trigger parent function
2) Just passing 'ref' with provide/inject end 'watch' it inside child. Because in Vue 3, provided data is reactive
3) Or combine 'shallowRef' and 'triggerRef' if you need to trigger effects manually
 For example, if we want notify child component to do some stuff, something like signaling event.

### DIFF
--- a/packages/form/src/form-item.vue
+++ b/packages/form/src/form-item.vue
@@ -61,7 +61,6 @@ import AsyncValidator from 'async-validator'
 import { RuleItem } from 'async-validator'
 import LabelWrap from './label-wrap'
 import { getPropByPath, useGlobalConfig } from '@element-plus/utils/util'
-import mitt from 'mitt'
 import { elFormKey, elFormItemKey, ValidateFieldCallback } from './token'
 
 export default defineComponent({
@@ -93,7 +92,6 @@ export default defineComponent({
     size: String,
   },
   setup(props) {
-    const formItemMitt = mitt()
     const $ELEMENT = useGlobalConfig()
 
     const elForm = inject(elFormKey)
@@ -136,8 +134,8 @@ export default defineComponent({
 
     const labelFor = computed(() => props.for || props.prop)
     const labelStyle = computed(() => {
-      if (elForm.labelPosition === 'top') return {}
-      const labelWidth = props.labelWidth || elForm.labelWidth
+      if (elForm?.labelPosition === 'top') return {}
+      const labelWidth = props.labelWidth || elForm?.labelWidth
       if (labelWidth) {
         return {
           width: labelWidth,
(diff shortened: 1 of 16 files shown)


JeremyWuuuuu:
Nice suggestion, but have you considered another scenario of `broadcasting` which child components need to listen to whatever the parents emitted.

fat4lix:
For now I can't image a situation when it would be useful.

But still, if we are talking about the situation when a child should listen to the something in the parent, we just could pass parent reactive data and it will listen to them by itself.

If we just need to do something in child component when event happens in parent, we could use a method with shallowRef and triggerRef. Which I've described to show how to manually call an effect.

At the end it's a matter of design.
I think such things like events bus could be useful in the result application in order to build communication for the whole interface.
But when we are talking about communication between components within a single branch, events bus seems to be excessive.

Especially when vue 3 provides a plenty of ways to handle described problems.

JeremyWuuuuu:
> For now I can't image a situation when it would be useful.
> 
> But still, if we are talking about the situation when a child should listen to the something in the parent, we just could pass parent reactive data and it will listen to them by itself.
> 
> If we just need to do something in child component when event happens in parent, we could use a method with shallowRef and triggerRef. Which I've described to show how to manually call an effect.
> 
> At the end it's a matter of design.
> I think such things like events bus could be useful in the result application in order to build communication for the whole interface.
> But when we are talking about communication between components within a single branch, events bus seems to be excessive.
> 
> Especially when vue 3 provides a plenty of ways to handle described problems.

Right, we simply pass down a ref to the children to watch, that should give us a better flexibility. 
Thanks for making this PR, later I'll get it merged.

packages/form/src/form-item.vue
```diff
@@ -193,16 +191,20 @@ export default defineComponent({
...
     const validate = (trigger: string, callback?: ValidateFieldCallback) => {
+      if (trigger === 'change' && validateDisabled.value) {
+        validateDisabled.value = false
+        return
+      }
       validateDisabled.value = false
       const rules = getFilteredRule(trigger)
       if ((!rules || rules.length === 0) && props.required === undefined) {
-        callback()
+        if (callback) callback()
```
JeremyWuuuuu, line 207: Can we change here like `callback?.()`, I think this is the perfect place for us using this syntax.

packages/form/src/form-item.vue
```diff
@@ -288,76 +290,49 @@ export default defineComponent({
...
     const formItemClass = computed(() => [
       {
-        'el-form-item--feedback': elForm && elForm.statusIcon,
+        'el-form-item--feedback': elForm?.statusIcon,
         'is-error': validateState.value === 'error',
         'is-validating': validateState.value === 'validating',
         'is-success': validateState.value === 'success',
         'is-required': isRequired.value || props.required,
-        'is-no-asterisk': elForm && elForm.hideRequiredAsterisk,
+        'is-no-asterisk': elForm?.hideRequiredAsterisk,
```
JeremyWuuuuu, line 329: Nice

packages/select/src/option-group.vue
```diff
@@ -37,10 +37,11 @@ export default defineComponent({
 
     const select = inject(selectKey)
 
-    const queryChange = () => {
+    const groupQueryChange = toRef(select, 'groupQueryChange')
+
+    watch(groupQueryChange, () => {
       visible.value = select?.options?.some(option => option.visible === true)
```
JeremyWuuuuu, line 43: Do we need a default value for this? Though the chaining fallback should be `undefined`

JeremyWuuuuu: But the type of `visible` should be `boolean` instead of `undefined | boolean` right?

jw-foss (approved):

zazzaz:
Hi @fat4lix, would you please fix the conflict before this PR get merged?

YunYouJun:
Thanks for your contribution.

Because https://github.com/element-plus/element-plus/pull/3356 has been merged, and this pr is stale, we will close it.

PR #469, opened 2020-10-26
fix(switch): fix switch has outline

Ther-su:
fix switch has outline, the class el-switch need to add outline:none

- [ switch] Component Migration

Fix #468

PR #474, opened 2020-10-27
Revert "fix(switch): fix switch has outline"

zazzaz:
Reverts element-plus/element-plus#469

a global fix here #472


jw-foss (approved):

JeremyWuuuuu:
Reverting this PR due to #472, this fix this issue from the root

PR #483, opened 2020-10-27
fix: modify createElement to h

BeADre:
- [ ] docs
 - changes: change $createElement to h


jw-foss (approved):

JeremyWuuuuu:
Seems good, thanks for your help

Issue #475, opened 2020-10-27
[docs] Some component documents report property errors!


Sphinm:
### Components behavior
- [Cascader](https://element-plus.org/#/zh-CN/component/cascader) 
![image](https://user-images.githubusercontent.com/16483962/97267648-71816100-1865-11eb-840c-32c811621527.png)

- [Upload](https://element-plus.org/#/zh-CN/component/upload)
![image](https://user-images.githubusercontent.com/16483962/97267735-9bd31e80-1865-11eb-8644-5721b974d983.png)

- [Transfer](https://element-plus.org/#/zh-CN/component/transfer)
![image](https://user-images.githubusercontent.com/16483962/97267809-c329eb80-1865-11eb-8f32-bb430645681e.png)

### Expectation
> The component document is previewed normally and no error is reported

JeremyWuuuuu:
Hi, @Sphinm 
We are still in progress and some components haven't been implemented yet, or the documentation haven't been updated yet, so the error was expected but certainly not correct, we also need help for updating the documentation.
For now we have 3 components on the way, right after the end of migration, we will begin working on documentations immediately.
Sorry for the inconvenience that we made.

Sphinm:
> Hi, @Sphinm
> We are still in progress and some components haven't been implemented yet, or the documentation haven't been updated yet, so the error was expected but certainly not correct, we also need help for updating the documentation.
> For now we have 3 components on the way, right after the end of migration, we will begin working on documentations immediately.
> Sorry for the inconvenience that we made.

Got it~

Ryan2128:
I will update `Transfer`'s docs today, the reason of errors is syntax of slots.

zazzaz:
fixed, thanks

PR #494, opened 2020-10-29
docs(transfer): fix error

Ryan2128:
update `transfer` demo to vue 3

### DIFF
--- a/website/docs/en-US/transfer.md
+++ b/website/docs/en-US/transfer.md
@@ -10,8 +10,8 @@ You can customize list titles, button texts, render function for data items, che
   <p style="text-align: center; margin: 0 0 20px">Customize data items using render-content</p>
   <div style="text-align: center">
     <el-transfer
+      v-model="leftValue"
       style="text-align: left; display: inline-block"
-      v-model="value"
       filterable
       :left-default-checked="[2, 3]"
       :right-default-checked="[1]"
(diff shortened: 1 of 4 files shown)


Ryan2128:
#475

zazzaz:
@Ryan2128 Hi, can you please keep the change of new slot syntax, but revert to the previous option API instead of  composition api

Ryan2128:
> @Ryan2128 Hi, can you please keep the change of new slot syntax, but revert to the previous option API instead of composition api

OK

Ryan2128:
Ready for review. @zazzaz

packages/transfer/src/useCheck.ts
```diff
@@ -17,7 +17,7 @@ export const useCheck = (props: TransferPanelProps, panelState: TransferPanelSta
       if (typeof props.filterMethod === 'function') {
         return props.filterMethod(panelState.query, item)
       } else {
-        const label = item[labelProp.value] || item[keyProp.value].toString()
+        const label = item[labelProp.value] || item[keyProp.value].toString && item[keyProp.value].toString() || item[keyProp.value]
```
JeremyWuuuuu, line 20: Why not `item[keyProp.value].toString?.()`?

jw-foss (approved):

JeremyWuuuuu:
It looks good to me after you fix the comment I mentioned.

packages/transfer/src/useCheck.ts
Ryan2128: Because the `key` of `item` may be a `number`, the type of `number` does't have the method of `toString`.

Ryan2128: I did something wrong. It was a demo that was wrong,  the reason is `item[keyProp.value]` might be `undefined`. I will correct it later.

JeremyWuuuuu:
is this ready to get merged？

Ryan2128:
> is this ready to get merged？

Yes.

PR #526, opened 2020-11-01
fix(input-number): fix bugs, finish test and correct docs

FengXingyuan:
- [ ] Component Migration

Fix #


packages/directives/__tests__/repeat-click.spec.ts
```diff
@@ -26,7 +26,7 @@ describe('Directives.vue', () => {
     const wrapper = _mount()
     const block = wrapper.find('#block')
     block.trigger('mousedown')
-    const testTime = 330
+    const testTime = 399
```
JeremyWuuuuu, line 29: What is the reason for this magic number?

JeremyWuuuuu: I personally do not agree with testing with `sleep`

packages/input-number/__tests__/input-number.spec.ts
```diff
@@ -1,32 +1,173 @@
...
+    const wrapper = _mount({
+      template: '<el-input-number :precision="2" v-model="num" />',
+      setup() {
+        const num = ref(0)
+        return {
+          num,
+        }
+      },
+    })
+    await wrapper.find('input').setValue(1.1111111111)
```
JeremyWuuuuu, line 83: Maybe we need to test whether manually setting value equals to `1.11` or something? This should be tested, correct?

packages/input-number/__tests__/input-number.spec.ts
```diff
@@ -1,32 +1,173 @@
...
+        const num = ref(0)
+        return {
+          num,
+        }
+      },
+    })
+    await wrapper.find('.el-input-number__increase').trigger('keydown.enter')
+    expect(wrapper.find('input').element.value).toEqual('0')
+    await wrapper.find('.el-input-number__decrease').trigger('keydown.enter')
+    expect(wrapper.find('input').element.value).toEqual('0')
```
JeremyWuuuuu, line 100: I think the most important testing against this `disabled` feature is to test the event handler gets triggered or not. Testing its value is the "side effect" caused by "Clicking the increase/decrease button"

packages/input-number/__tests__/input-number.spec.ts
```diff
@@ -1,32 +1,173 @@
...
+        const num = ref(0)
+        return {
+          num,
+        }
+      },
+    })
+    const increase = wrapper.find('.el-input-number__increase')
+    expect(increase.exists()).toBe(false)
+    const decrease = wrapper.find('.el-input-number__decrease')
+    expect(decrease.exists()).toBe(false)
```
JeremyWuuuuu, line 115: What about
```suggestion

expect(wrapper.find('.el-input-number__increase').exists()).toBe(false)
expect(wrapper.find('.el-input-number__decrease').exists()).toBe(false)
```

packages/input-number/__tests__/input-number.spec.ts
```diff
@@ -1,32 +1,173 @@
...
+        const num = ref(0)
+        return {
+          num,
+        }
+      },
+    })
+    const increase = wrapper.find('.el-icon-arrow-down')
+    expect(increase.exists()).toBe(true)
+    const decrease = wrapper.find('.el-icon-arrow-up')
+    expect(decrease.exists()).toBe(true)
```
JeremyWuuuuu, line 130: Same here.

packages/input-number/src/index.vue
```diff
@@ -41,8 +41,8 @@
       :label="label"
       @keydown.up.prevent="increase"
       @keydown.down.prevent="decrease"
-      @blur="event => $emit('blur', event)"
-      @focus="event => $emit('focus', event)"
```
JeremyWuuuuu, line 45: Reason behind this change?

packages/input-number/src/index.vue
```diff
@@ -103,14 +103,17 @@ export default defineComponent({
...
+    },
     disabled: {
       type: Boolean,
       default: false,
     },
     size: {
-      type: String as PropType<'large' | 'small'>,
-      validator: (val: string) => ['large', 'small'].includes(val),
+      type: String as PropType<'medium' | 'small' | 'mini'>,
+      validator: (val: string) => ['medium', 'small', 'mini'].includes(val),
```
JeremyWuuuuu, line 116: We had a global union type `ComponentSize` you can use it directly

packages/input-number/src/index.vue
```diff
@@ -248,6 +251,20 @@ export default defineComponent({
...
+      emit('blur', event)
+    }
+    const handleFocus = event => {
+      emit('focus', event)
+    }
+    const focus = () => {
+      nextTick(() => {
+        input.value.focus()
+      })
+    }
```
JeremyWuuuuu, line 264: Seems this function has not been used

packages/input-number/__tests__/input-number.spec.ts
FengXingyuan: got it, already correct it

packages/input-number/src/index.vue
FengXingyuan: got it

packages/input-number/__tests__/input-number.spec.ts
FengXingyuan: got it

PR #535, opened 2020-11-03
Docs/avatar

rottenpen:
- [ ] update docs

update avatar docs


JeremyWuuuuu:
Hey, I am closing this PR due to there was no reason to change all the documentation to VCA, also, I wanted to point out that when you have some static data like the `url` which never gets changed, you DO NOT NEED TO make it a reactive.

JeremyWuuuuu:
We do not want to put all the documentation to VCA, because that for some users they found it difficult or less useful than options API.

rottenpen:
Any there any Todolist about it？


 原始邮件 
发件人: jeremywu<notifications@github.com>
收件人: element-plus/element-plus<element-plus@noreply.github.com>
抄送: 滑威<rottenpen@outlook.com>; Author<author@noreply.github.com>
发送时间: 2020年11月4日(周三) 10:26
主题: Re: [element-plus/element-plus] Docs/avatar (#535)


We do not want to put all the documentation to VCA, because that for some users they found it difficult or less useful than options API.

PR #558, opened 2020-11-05
Fix(transition): install ElCollapseTransition 

rottenpen:
- [ ] fix ElCollapseTransition bug

Fix ElCollapseTransition isn't installed . and can't run like this
image


JeremyWuuuuu:
We should export this member from the main package, consider registering it locally.

rottenpen:
@JeremyWuuuuu what is the meaning of this？

JeremyWuuuuu:
> We should export this member from the main package, consider registering it locally.

Typo **Should not** instead of should...

JeremyWuuuuu:
cc @bastarder. But seems the original version had this one installed globally, maybe I was wrong about this.

jw-foss (approved):

PR #562, opened 2020-11-09
fix(popper): fix popper related style

jw-foss:
- Fix popper related style issue
- Add css minifier into website


packages/dropdown/src/dropdown.vue
```diff
@@ -6,7 +6,7 @@
     :effect="effect"
     :manual-mode="true"
     :trigger="[trigger]"
-    popper-class="el-dropdown-popper"
+    popper-class="el-dropdown--popper"
```
SimonaliaChen, line 9: use `el-dropdown__popper` instead since `popper` is an element

packages/theme-chalk/src/popper.scss
```diff
@@ -1,101 +1,86 @@
...
+  line-height: 1.2;
+  min-width: 10px;
+  word-wrap: break-word;
+  visibility: visible;
 
-  .popper__arrow {
-    border-width: $--popover-arrow-size;
-    filter: drop-shadow(0 2px 12px rgba(0, 0, 0, 0.03))
-  }
+  $arrow-selector: #{& + '__arrow'};
```
SimonaliaChen, line 15: cool

SimonaliaChen (approved):

PR #584, opened 2020-11-12
fix: fix color-picker popover popconfirm style

zazzaz:

PR #595, opened 2020-11-14
fix: remove outline and padding 

BeADre:
**outline**: like this

image

**padding**: like this

image


zazzaz:
fixed in #584

BeADre:
padding的问题似乎已经修了  那文档的outline呢？我修改当前的pr还是你那边来修复？

zazzaz:
seems to fix the outline in popper is better?

BeADre:
but the outline I modified seems to have nothing to do with popper

PR #610, opened 2020-11-17
docs(timeline): remove the description about the reverse

BeADre:
according to the description in the source code, the timeline will not support prop 'reverse' ?

image

### DIFF
--- a/website/docs/en-US/timeline.md
+++ b/website/docs/en-US/timeline.md
@@ -4,20 +4,12 @@ Visually display timeline.
 
 ### Basic usage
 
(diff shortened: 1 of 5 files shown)


zazzaz:
Thanks

zazzaz (approved):

BeADre:
对了 日文文档的描述部分我没有更新 因为我实在看不懂  所以只改了代码的部分

BeADre:
然后还有[#483](https://github.com/element-plus/element-plus/pull/483)相关部分的日文文档似乎也没有更新

zazzaz:
Hi, @BeADre you can simply ignore the description that you did not familiar with. 

It will be a great help if you could update the code part in #483

BeADre:
ok, I'll update it later

Issue #496, opened 2020-10-29
Bug with select component


kooriookami:
1. Default size of `el-select`, it's height should be 40px, actually it's 36px.
![image](https://user-images.githubusercontent.com/38392315/97518668-ac54d780-19d2-11eb-9a73-f6cf88f30d8e.png)
2. The popover is missing a corner.
![image](https://user-images.githubusercontent.com/38392315/97518628-9515ea00-19d2-11eb-9074-a4aef470501d.png)
3. When click `el-select` twice, it should open and then close, actually it won't close.

PR #619, opened 2020-11-18
fix: fix the color picker can't colse on click twice and add z-index for the notification

### LINKED ISSUE #602: A Few Bug List
I'm not sure whether you have known these bugs, I just list them for your information.
1. The slider button above on the popover.
image
2. The color picker can't colse on click twice.
3. Notification under the doc header.
image
4. The message appeared in no animation.
image
5. The message hasn't bind functions like `this.$message[type]`(type of success/warning/info/error)

Ther-su:
Fix #602 the color picker can't colse on click twice.
In my opinion,the bug and solution is similar to #496.

Fix #602 Notification under the doc header.


Ther-su:
color-picker
This is showed.

Ther-su:
In my solution,I add a father div that is inline-block for the component so that it can fix the bug. Does it can be received? Or do you have a better idea?

zazzaz:
Thanks. I did not check this bug yet. However, I do think it is not a good idea to change the dom structure.

Ther-su:
Thanks,and let me see if there's a better idea.haha.

jw-foss (approved):

JeremyWuuuuu:
@Ther-su  Do you have further moving on this bug? ..I just merged it 😂, if you do please raise another PR for the fix.

PR #634, opened 2020-11-20
fix: fix the weird tooltip

### LINKED ISSUE #633: [Bug] [Website] Demo box `try it` section tooltip seems weird
The demo box's tooltip seems odd but the rest tooltips seems to be just fine. I believe this is a bug which needs to be fixed
image

BeADre:
fix #633

### DIFF
--- a/packages/theme-chalk/src/popper.scss
+++ b/packages/theme-chalk/src/popper.scss
@@ -19,6 +19,7 @@
     background: $--color-text-primary;
     #{$arrow-selector}::before {
       background: $--color-text-primary;
+      right: 0;
     }
   }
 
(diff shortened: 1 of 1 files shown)


jw-foss (approved):

JeremyWuuuuu:
So does this change covers all cases? like when the space is sufficient tooltip appears on the right side, and when it's not on the left side? I think this change is legit, maybe with a screenshot could be better.

BeADre:
我查看了文档中所有使用了tooltip组件的情况，目前似乎一切表现正常，下面是部分截图
___
image

JeremyWuuuuu:
Alright. Let's get it merged.

Issue #638, opened 2020-11-20
Some small things


kooriookami:
Hey, it's me again. I find some small things this time.
1. `Select` seems has two panel when animating. And the shadow is heavier than other popver. The small triangle is bigger than others.
![image](https://user-images.githubusercontent.com/38392315/99771728-8db2ae00-2b44-11eb-8928-e3db07dae222.png)![image](https://user-images.githubusercontent.com/38392315/99771787-a622c880-2b44-11eb-99b5-d70aafc8f3f1.png)![image](https://user-images.githubusercontent.com/38392315/99771813-b33fb780-2b44-11eb-9e19-546346359c78.png)
2. `Popver` trigger="focus", when blur, panel still open.
![image](https://user-images.githubusercontent.com/38392315/99772310-7de79980-2b45-11eb-913f-8cce71c816e0.png)
3. `Notification` when open, has no animation.
4. `MessageBox prompt`, default value is null.
![image](https://user-images.githubusercontent.com/38392315/99772616-11b96580-2b46-11eb-9d8c-ceb5767027e0.png)
5. `Cascader`
(1). Doc has wrong style.
(2). When click a menu that has children(for example: "设计原则"), panel closed.
(3). When mouse move on the panel, sometimes the panel jumped left and right cause of no more space.
![image](https://user-images.githubusercontent.com/38392315/99772821-71b00c00-2b46-11eb-99e9-c02d66a1d4b1.png)

JeremyWuuuuu:
Thanks for raising this issue!
One thing that I wanted to ask is which version of `EP` and `Vue` you are current using?
Could you provide me some information?
Or this is what you found in the documentation website?

JeremyWuuuuu:
This is definitely helpful for making this product better!

JeremyWuuuuu:
For the first issue I think @SimonaliaChen had a fix last month to fix this by adding `transform-origin` in order to the the animation ordering correct.

JeremyWuuuuu:
Recording for the `Popover` issue, seems that it can be correctly responded by blurring out of focus.
![Screen Recording 2020-11-20 at 4 13 57 PM](https://user-images.githubusercontent.com/15975785/99776515-ea659700-2b4b-11eb-97d8-7381195aa06b.gif)

kooriookami:
> Thanks for raising this issue!
> One thing that I wanted to ask is which version of `EP` and `Vue` you are current using?
> Could you provide me some information?
[...]

PR #642, opened 2020-11-20
Fix/message-box: messageBox prompt default value is null

Ryan2128:

* [x] Make sure you follow Element's contributing guide ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).

> messageBox prompt, default value is null.

#### Fix the bug in #638

- fix validate failed style error
- fix input's value is a 'null' string when modelValue is null

### DIFF
--- a/packages/input/src/index.vue
+++ b/packages/input/src/index.vue
@@ -243,7 +243,7 @@ export default defineComponent({
       resize: props.resize,
     }))
     const inputDisabled = computed(() => props.disabled || elForm.disabled)
-    const nativeInputValue = computed(() => String(props.modelValue))
+    const nativeInputValue = computed(() => (props.modelValue === null || props.modelValue === undefined) ? '' : String(props.modelValue))
     const upperLimit = computed(() => ctx.attrs.maxlength)
     const showClear = computed(() => {
       return props.clearable &&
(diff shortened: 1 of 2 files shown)


packages/input/src/index.vue
```diff
@@ -243,7 +243,7 @@ export default defineComponent({
       resize: props.resize,
     }))
     const inputDisabled = computed(() => props.disabled || elForm.disabled)
-    const nativeInputValue = computed(() => String(props.modelValue))
+    const nativeInputValue = computed(() => props.modelValue === null || props.modelValue === undefined ? '' : String(props.modelValue))
```
JeremyWuuuuu, line 246: Is this intentional or it was a mistake? 
If you were trying to express with "if props.modelValue is nullish, then return empty string, otherwise return Stringified modelValue" then this line is wrong, you need a  parenthesis to wrap the expression

jw-foss (approved):

PR #648, opened 2020-11-21
Fixed a problem where the Cascader selector panel cannot be displayed when selected

xiaoxian521:
Use version：
vue：3.0.0
element-plus：1.0.1-alpha.16

problem：
After the cascade is selected, clicking the panel and the outside of the panel area will be hidden directly

Solution results：
After the cascade is selected, the click panel will not be hidden, and the outside of the click panel area will be hidden。The test passed

### DIFF
--- a/packages/cascader/src/index.vue
+++ b/packages/cascader/src/index.vue
@@ -445,7 +445,6 @@ export default defineComponent({
     }
 
     const handleExpandChange = (value: CascaderValue) => {
-      if (popperVisible.value) return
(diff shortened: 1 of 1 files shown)


JeremyWuuuuu:
Well, I am closing this issue due to the bug which causing `Cascader` acting weird is a bug that might cause the other component whom depends on `Popper` acting weird as well. So the fix for this should happen on `Popper` instead of `Cascader` itself.

xiaoxian521:
ok

Issue #661, opened 2020-11-23
`select`  component  `size` property not working when set `multiple`


weiyuc:
```
<el-select size='mini' multiple>
  <el-option :value='1' label='label 1'></el-option>
</el-select>
```
expect: el-select height = 28px
in fact: el-select height = 40px

Ryan2128:
Can you provide you version? In my test, it works well.

[DEMO](https://codesandbox.io/s/prod-architecture-q90jz?file=/src/App.vue)

weiyuc:
it not working on latest version,
[demo](https://codesandbox.io/s/wizardly-framework-hx5y4?file=/src/App.vue)

zazzaz:
The problem finally becomes, a select component inside a form does not size correctly.

JeremyWuuuuu:
Then this is some bug that we need to fix. Assigning it to me.

PR #663, opened 2020-11-23
fix bug: 修复下拉框设置多选时size参数失效BUG

weiyuc:

#661

### DIFF
--- a/packages/select/src/select.vue
+++ b/packages/select/src/select.vue
@@ -346,15 +346,13 @@ export default defineComponent({
         currentPlaceholder.value = ''
       }
       addResizeListener(selectWrapper.value as any, handleResize)
-      if (reference.value && reference.value.$el) {
-        const sizeMap = {
-          medium: 36,
-          small: 32,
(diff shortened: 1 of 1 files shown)


Ryan2128:
Please update your version and test it again.

packages/select/src/select.vue
```diff
@@ -346,15 +346,13 @@ export default defineComponent({
...
-        }
-        const input = reference.value.$el
-        states.initialInputHeight = input.getBoundingClientRect().height || sizeMap[selectSize.value]
+      const sizeMap = {
+        medium: 36,
+        small: 32,
+        mini: 28,
       }
+      const input = reference.value.$el
+      states.initialInputHeight = input.getBoundingClientRect().height || sizeMap[selectSize.value]
```
JeremyWuuuuu, line 355: So this change did not cover the case when `input` is `undefined`, then `input.getBoundingClientRect()` would throw an error which says cannot get member from undefined, in this case the code would panic.

PR #736, opened 2020-12-01
feat(table): complete show-overflow-tooltip

justwiner:
* [ table ] complete show-overflow-tooltip


packages/table/src/util.ts
```diff
@@ -293,3 +295,54 @@ export function walkTreeNode(
...
+  trigger.onmouseenter = () => {
+    popperInstance.update()
+  }
+  trigger.onmouseleave = () => {
+    try {
+      popperInstance.destroy()
+      document.body.removeChild(content)
+    } catch {}
+  }
+}
```
JeremyWuuuuu, line 348: Using `on` `off` in `utils/dom`

zazzaz (approved):

PR #772, opened 2020-12-02
feat(form): add method that can get what attribute in model has changed.

Ther-su:
re #676
If it's not necessary to make this feature PR, plz close this.


JeremyWuuuuu:
I think we might need to put this feature on hold, because we are not sure if this is a good idea for exposing an API like this, with this implementation we might encounter performance issues, because of all those watchers. Also the concept of two way binding, the upstream always know what changed via a single watcher, every time the `modelValue` gets updated, it should be notified, then cast their own diff method to do that, instead of adding all those watchers on us
Let's keep this open

sxzz:
closed due to oudated

PR #785, opened 2020-12-02
fix: picker transition with 'el-zoom-in-top'

SimonaliaChen:
- Add prop `gpuAcceleration` to `Popper`
- Fix all picker's transition and styles, including `Autocomplete`, `Dropdown`, `Select`, `TimerPicker`, `DatePicker`
and so on, which transition should be 'el-zoom-in-top' and without duplication
- Set prop `popperAppendToBody`'s default value to `false` of `Select` and `Autocomplete` to keep consistency with `Popper`


packages/popper/src/use-popper/build-modifiers.ts
```diff
@@ -41,7 +43,8 @@ export default function buildModifier(props: ModifierProps, externalModifiers: S
     {
       name: 'computeStyles',
       options: {
-        adaptive: true,
+        gpuAcceleration,
+        adaptive: gpuAcceleration,
```
JeremyWuuuuu, line 47: These two were not conflicted?

jw-foss (approved):

packages/time-picker/src/time-picker-com/panel-time-range.vue
```diff
@@ -1,76 +1,69 @@
...
-              :spinner-date="maxDate"
-              :disabled-hours="disabledHours_"
-              :disabled-minutes="disabledMinutes_"
-              :disabled-seconds="disabledSeconds_"
-              @change="handleMaxChange"
-              @set-option="onSetOption"
-              @select-range="setMaxSelectionRange"
-            />
-          </div>
+  <div v-if="actualVisible" class="el-time-range-picker el-picker-panel">
```
SimonaliaChen, line 2: Can you take a look? @zazzaz I am not sure whether it is a good solution. I try to remove  `v-if` for better performance, but some errors will be thrown, such as `can not read xx from undefined`. I suppose there are something wrong with the default value, but the component are toooo complicated. Could you help me out？

packages/popper/src/use-popper/build-modifiers.ts
SimonaliaChen: When set `gpuAcceleration` to false, adaptive should also be false, otherwise will get a warning.

JeremyWuuuuu: Got it

PR #825, opened 2020-12-04
fix(select): fix error style of select-dropdown

### LINKED ISSUE #812: Error style of el-select-dropdown
image
```html
<el-form label-width="auto" style="width: 400px">
    <el-form-item label="123">
        <el-select v-model="form.value" style="width: 100%">
            <el-option v-for="item in 2" :key="item" :label="item" :value="item.value"></el-option>
        </el-select>
    </el-form-item>
</el-form>
```

kooriookami:
fix #812

### DIFF
--- a/packages/select/src/select-dropdown.vue
+++ b/packages/select/src/select-dropdown.vue
@@ -8,11 +8,12 @@
   </div>
 </template>
 
-<script  lang="ts">
+<script lang="ts">
 import {
   computed,
   onMounted,
   inject,
+  ref,
 } from 'vue'
 import {
   selectKey,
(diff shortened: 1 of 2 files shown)


packages/select/src/select-dropdown.vue
```diff
@@ -29,11 +30,15 @@ export default {
...
     const popperClass = computed(() => select.props.popperClass)
     const isMultiple = computed(() => select.props.multiple)
-    const minWidth = computed(() => select.selectWrapper?.getBoundingClientRect().width + 'px')
+    const minWidth = ref('')
 
     onMounted(() => {
       // TODO: updatePopper
       // popper.value.update()
+
+      setTimeout(() => {
```
JeremyWuuuuu, line 39: Instead of using magical `setTimeout` we need to make sure that whenever `getBoudingClientRect()` is called, the ref is actually there. So the way you making it `ref` is correct, but then you need to watch `select.selectWrapper` for it's existence like `watch(() => select.selectWrapper, (val) => minWidht.value = val.getBoundingClientRect() + 'px');

JeremyWuuuuu: setTimeout could not save us from fail every single time.

kooriookami: Ok, I use resize-observer instead.

kooriookami:
@JeremyWuuuuu Sorry, I commit a error commit. I just close it and fix it later.

JeremyWuuuuu:
- - Sure.

bastarder:
sorry about that, I didn't notice that you had fixed the bug, thx for your pr!

PR #828, opened 2020-12-04
fix(table): avoiding the error of el-table-column rendering

### LINKED ISSUE #793: Bug with slot of el-table-column
When use `scope.row.xxx.yyy` in el-table-column, There is an error: TypeError: Cannot read property 'yyy' of undefined. I can use `scope.row.xxx?.yyy`. It seems a bug with lifecycle.

justwiner:
fix #793

* [table] fix:#793

### DIFF
--- a/packages/table/src/table-column/index.ts
+++ b/packages/table/src/table-column/index.ts
@@ -88,9 +88,6 @@ export default defineComponent({
     const instance = getCurrentInstance() as TableColumn
     const columnConfig = ref({})
     const props = (prop as unknown) as TableColumnCtx
-    const row = ref({})
-    const r = ref({})
-    const $index = ref(0)
     const owner = computed(() => {
       let parent = instance.parent as any
       while (parent && !parent.tableId) {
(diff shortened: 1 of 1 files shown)


packages/table/src/table-column/index.ts
```diff
@@ -202,24 +199,22 @@ export default defineComponent({
...
-        row: {},
-        $index: undefined,
-      }),
-    )
+    try {
+      return h(
+        'div',
+        this.$slots.default?.({
+          store: {},
+          _self: {},
```
JeremyWuuuuu, line 210: Without passing any data, how could users get data from?

justwiner: Rendering within `el-table-column` helps to generate column structures, cell rendering is called actively in `table-body` by `slots.default`
https://github.com/element-plus/element-plus/blob/fe42f4404e522f8bd6730fb51e0245d014a03512/packages/table/src/table-body/render-helper.ts#L81-L87
https://github.com/element-plus/element-plus/blob/fe42f4404e522f8bd6730fb51e0245d014a03512/packages/table/src/table-body/render-helper.ts#L104-L116
https://github.com/element-plus/element-plus/blob/fe42f4404e522f8bd6730fb51e0245d014a03512/packages/table/src/table-column/render-helper.ts#L111-L131

jw-foss (approved):

PR #830, opened 2020-12-04
fix(form): fix form-item auto width bug

### LINKED ISSUE #770: Form label-width=“auto” will not change automatically
When the form set `label-width=“auto”`, and the form-item is dynamic, the label width will not be correct.

bastarder:
Fix when el-form display none, auto label width not work bug
#770

### DIFF
--- a/packages/form/src/form.vue
+++ b/packages/form/src/form.vue
@@ -32,9 +32,8 @@ function useFormLabelWidth() {
 
   function getLabelWidthIndex(width: number) {
     const index = potentialLabelWidthArr.value.indexOf(width)
-    // it's impossible
     if (index === -1) {
-      throw new Error('[ElementForm]unpected width ' + width)
+      console.warn('[Element Warn][ElementForm]unexpected width ' + width)
     }
     return index
   }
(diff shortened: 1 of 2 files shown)


kooriookami:
> Fix when el-form display none, auto label width not work bug
> #770

`label-wrap.ts`
image
Is it possible to add a judgment to prevent reporting errors?

packages/form/src/form.vue
```diff
@@ -32,9 +32,8 @@ function useFormLabelWidth() {
 
   function getLabelWidthIndex(width: number) {
     const index = potentialLabelWidthArr.value.indexOf(width)
-    // it's impossible
     if (index === -1) {
-      throw new Error('[ElementForm]unpected width ' + width)
+      console.warn('[Element Warn][ElementForm]unpected width ' + width)
```
JeremyWuuuuu, line 36: Typo **[unexpected]**

jw-foss (approved):

JeremyWuuuuu:
Resolve that typo issue, also add one space after each word like `[ElmentPlus] [ElForm]`

packages/form/src/form.vue
```diff
@@ -50,7 +49,7 @@ function useFormLabelWidth() {
 
   function deregisterLabelWidth(val: number) {
     const index = getLabelWidthIndex(val)
-    potentialLabelWidthArr.value.splice(index, 1)
+    index > -1 && potentialLabelWidthArr.value.splice(index, 1)
```
JeremyWuuuuu, line 52: @kooriookami Will, with this line, the concern is certainly unnecessary, because if this value was not found, the function won't be invoked anyway.

jw-foss (approved):

Issue #930, opened 2020-12-09
[Bug Report] El-switch starts the Change method on the first render


XingXiaoWu:
# Element-Plus version
"element-plus": "^1.0.1-beta.5"

# OS/Browsers version
Mozilla/5.0 (Macintosh; Intel Mac OS X 11_0_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/87.0.4280.88 Safari/537.36

# Steps to reproduce
set switch @change

# What is Expected?
It will not fire the first time it is rendered

# What is actually happening?
It will fire the first time you render

This is example:
https://codepen.io/xingxiaowu/pen/KKggqpZ

I compared the element  2.14.1, which is normal：
https://codepen.io/xingxiaowu/pen/QWKKgjq

Ryan2128:
Thanks. It will be fixed in next version.

Ryan2128:
I check the source code and test the demo you provide just now, I find there is an error in your usage.

```
if (!~[props.activeValue, props.inactiveValue].indexOf(actualValue.value)) {
  ctx.emit('update:modelValue', props.inactiveValue)
  ctx.emit('change', props.inactiveValue)
  ctx.emit('input', props.inactiveValue)
}
```

You passed the `active-value` and `inactive-value`, but the `v-model` is `undefind`. So `switch` will emit a `change` event firstly.

Ryan2128:
If there are no doubts, I will close the issue later.

XingXiaoWu:
> I check the source code and test the demo you provide just now, I find there is an error in your usage.
> 
> ```
> if (!~[props.activeValue, props.inactiveValue].indexOf(actualValue.value)) {
>   ctx.emit('update:modelValue', props.inactiveValue)
>   ctx.emit('change', props.inactiveValue)
>   ctx.emit('input', props.inactiveValue)
> }
> ```
> 
> You passed the `active-value` and `inactive-value`, but the `v-model` is `undefind`. So `switch` will emit a `change` event firstly.
> 
> ```
> <el-switch
>   @change="switchChange"
>   v-model="value"
>   active-color="#13ce66"
>   inactive-color="#ff4949"
>   active-value="100"
>   inactive-value="0">
> </el-switch>
> 
> var Main = {
>   data() {
>     return {
>       value1: true,
>       value2: true
>     };
>   },
>   methods: {
>     switchChange(val) {
>       console.log("123");
>     }
>   }
> };
> ```

[...]

PR #933, opened 2020-12-09
fix(docs): change switch docs #930

XingXiaoWu:

### DIFF
--- a/website/docs/en-US/switch.md
+++ b/website/docs/en-US/switch.md
@@ -143,7 +143,7 @@ Switch is used for switching between two opposing states.
 
 | Attribute      | Description          | Type      | Accepted Values       | Default |
 |-----| ----| ----| ----|---- |
-| value / v-model | binding value | boolean / string / number | — | — |
+| value / v-model | binding value,that must be equal to active-value or inactive-value and default to a Boolean type | boolean / string / number | — | — |
 | disabled | whether Switch is disabled | boolean | — | false |
 | loading  | whether Switch is in loading state | boolean | — | false |
 | width | width of Switch | number | — | 40 |
--- a/website/docs/en-US/switch.md
+++ b/website/docs/en-US/switch.md
@@ -143,7 +143,7 @@ Switch is used for switching between two opposing states.
 
 | Attribute      | Description          | Type      | Accepted Values       | Default |
 |-----| ----| ----| ----|---- |
-| value / v-model | binding value,that must be equal to active-value or inactive-value and default to a Boolean type | boolean / string / number | — | — |
+| value / v-model | binding value, it should be equivalent to either `active-value` or `inactive-value`, by default it's `boolean` type | boolean / string / number | — | — |
 | disabled | whether Switch is disabled | boolean | — | false |
 | loading  | whether Switch is in loading state | boolean | — | false |
 | width | width of Switch | number | — | 40 |
--- a/website/docs/fr-FR/switch.md
+++ b/website/docs/fr-FR/switch.md
@@ -145,7 +145,7 @@ Switch est utilisé pour choisir entre deux états opposés.
 
 | Attribut      | Description          | Type      | Valeurs acceptées       | Défaut |
 | ----| ----| ----| ----|---- |
-| value / v-model | La valeur liée.Il doit être active-value ou inactive-value, par défaut type booléan | boolean / string / number | — | — |
+| value / v-model | La valeur liée. Elle doit être équivalente à une `active-value` ou une `inactive-value`, par défaut c'est un type `boolean`. | boolean / string / number | — | — |
 | disabled | Si le switch est désactivé. | boolean | — | false |
 | loading  | whether Switch is in loading state | boolean | — | false |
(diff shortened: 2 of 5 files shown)


XingXiaoWu:
@Ryan2128 https://github.com/element-plus/element-plus/issues/930 
Thank you for your answer. I didn't think the documentation was very detailed, so I changed it

website/docs/zh-CN/switch.md
```diff
@@ -143,7 +143,7 @@
 
 | 参数      | 说明    | 类型      | 可选值       | 默认值   |
 |---------- |-------- |---------- |-------------  |-------- |
-| value / v-model | 绑定值 | boolean / string / number | — | — |
+| value / v-model | 绑定值,必须等于active-value或inactive-value,默认为boolean类型 | boolean / string / number | — | — |
```
kooriookami, line 146: Latin words and Chinese characters should have spaces directly, and Chinese punctuation should use full angles.Chinese punctuation should be full-angle.

JeremyWuuuuu: Agreed, for Japanese, use 。instead of .

JeremyWuuuuu: Also I do not think using translation service to do this is a good idea, translator sometimes translate to odd statement for native speakers

JeremyWuuuuu:
Do you mind marking strings in `es` `jp` `fr` to English?

XingXiaoWu:
Ok, thanks to your guidance, I will revise it later and submit PR again according to the Contributing Guide

website/docs/en-US/switch.md
```diff
@@ -143,7 +143,7 @@ Switch is used for switching between two opposing states.
 
 | Attribute      | Description          | Type      | Accepted Values       | Default |
 |-----| ----| ----| ----|---- |
-| value / v-model | binding value | boolean / string / number | — | — |
+| value / v-model | binding value,that must be equal to active-value or inactive-value and default to a Boolean type | boolean / string / number | — | — |
```
JeremyWuuuuu, line 146: Use `should` to make it more soft.

it should be equivalent to either `active-value` or `inactive-value`, by default it's `boolean` type

XingXiaoWu:
According to your guidance, I revised the document.
But I didn't know much about Spanish, French, or Japanese, so I used translation software, but I put the sentences into different translators and got relatively consistent semantics.
Please let me know if you have any more questions. Thank you very much. 🙏🙏🙏

website/docs/jp/switch.md
```diff
@@ -143,7 +143,7 @@
 
 | Attribute      | Description          | Type      | Accepted Values       | Default |
 |-----| ----| ----| ----|---- |
-| value / v-model | バインディング値 | boolean / string / number | — | — |
+| value / v-model | バインド値は、"active-value "または "inactive-value "と等しくなければなりません。デフォルトの型は "boolean "です。 | boolean / string / number | — | — |
```
JeremyWuuuuu, line 146: According to the context `バインディング` is the katakana expression of `binding` which can be written as `bindingu` in roman. So this value should stay the same as before. Also I think, not translating them is better than translating them without knowing the correctness of the translation, I'd like to ask you to change them into English instead of using translator which might cause misunderstanding to native speakers, unless you speaks these language. I as a reviewer only speaks Chinese/English/limited Japanese. So I can only review them.

website/docs/fr-FR/switch.md
```diff
@@ -145,7 +145,7 @@ Switch est utilisé pour choisir entre deux états opposés.
 
 | Attribut      | Description          | Type      | Valeurs acceptées       | Défaut |
 | ----| ----| ----| ----|---- |
-| value / v-model | La valeur liée. | boolean / string / number | — | — |
+| value / v-model | La valeur liée. Elle doit être équivalente à une `active-value` ou une `inactive-value`, par défaut c'est un type `boolean`. | boolean / string / number | — | — |
```
JeremyWuuuuu, line 148: @g3r4n are these strings correct in both grammar and words?

website/docs/jp/switch.md
XingXiaoWu: I understand. Because I'm not good for other languages, so can I only change Chinese documents and pull request?
I didn't submit it before, so I subconsciously thought I needed to make sure all the documents were consistent.

website/docs/fr-FR/switch.md
g3r4n, line 148:
```suggestion
| value / v-model | Valeur liée. Elle doit être équivalente à `active-value` ou `inactive-value`, par défaut elle est de type `boolean`. | boolean / string / number | — | — |
```
Here is some improvement to the translation. Happy to help for French translation

g3r4n (changes requested):

website/docs/fr-FR/switch.md
JeremyWuuuuu: Thank you! cc @XingXiaoWu

jw-foss (approved):

PR #957, opened 2020-12-10
fix(locale): setting locale does not work for DayJS

wukang0718:
Setting locale does not work for DayJS

Vue：3.0.4
element-plus： 1.0.1-beta.7

I set the locale, but the week display in the calendar component is not in Chinese


porcelainHeart:
Add tests here for making sure it won't break again, thanks.

zazzaz:
It is better to load Day.js locale by your self. 

We will update the doc shortly。

PR #967, opened 2020-12-11
fix(select): fix content overflow in multiple select

### LINKED ISSUE #919: el-select 多选 显示有问题，内容溢出
2020-12-09 

版本号： @1.0.1-beta.6

代码复现

[code pen](https://codepen.io/luoriwusheng/pen/KKgggaj)

期望：
做成像单选框那样， 即使是多选，也不要溢出

单选框是没有溢出的

Ther-su:
fix content overflow in multiple select

fix #919


Ther-su:
color-picker
The result is showed like this.

packages/select/src/select.vue
```diff
@@ -37,7 +37,7 @@
                 disable-transitions
                 @close="deleteTag($event, selected[0])"
               >
-                <span class="el-select__tags-text">{{ selected[0].currentLabel }}</span>
+                <span class="el-select__tags-text" :style="{ 'text-overflow':'ellipsis',verticalAlign: 'bottom','max-width': inputWidth - 120 + 'px', 'display':'inline-block','overflowX':'hidden' }">{{ selected[0].currentLabel }}</span>
```
JeremyWuuuuu, line 40: Writing this into a variable and bind it, would've been more clean. Also, add tests against it.

Ther-su:
Can you give me much detail? I'm not sure what features to test. @JeremyWuuuuu

JeremyWuuuuu:
> Can you give me much detail? I'm not sure what features to test. @JeremyWuuuuu

Just test whether the styles gets attached when the condition gets satisfied. 
For example, in your case, you test the style with `inputWidth` value.
Also, I think that static style rules should be added into `select.scss` file instead of alone with the `:style` binding.

Ther-su:
I just use `inputWidth` to control the `maxWidth` of `el-select__tags-text` so that the select will not overflow, And I think it is strange to make a test case.

Ther-su:
This is my code about test case.But it seems have some bugs that I can't find.
image

JeremyWuuuuu:
Try debugging them step by step, in this process you'll learn the entire lifecycle of a vue component.
You can use `console.log` or `debugger` connecting to Chrome or VSCode.

JeremyWuuuuu:
The first thing you need to know is what is to be tested, then you'll figure out a way to test.

Ther-su:
I  try many way, like `document.querySelector` or `wrapper.find().attributes` and so on.But the maxWidth I get don't contain `inputWidth`, just `-90px`.It makes me doubt.
image
image
image

Ther-su:
What is more, when I try to get the `width` of `.el-input__inner` as `inputWidth`, like `document.querySelector().getBoundingClientRect().width`. It is 0 however.

Ther-su:
In the case, I would like to compare the maxWidth and inputWidth. It is my original intention to finish the test case.

JeremyWuuuuu:
> What is more, when I try to get the `width` of `.el-input__inner` as `inputWidth`, like `document.querySelector().getBoundingClientRect().width`. It is 0 however.

This is because, `JSDOM` does not have a layout engine, so `getBoundingRect` won't be able to retrieve the layout width and height that you were trying to get.

JeremyWuuuuu:
Here is a work around that might be helpful:
image
You can also search `getBoundingClientRect` in the codebase and find solutions.

JeremyWuuuuu:
The reason that we choose `JSDOM` over `Puppeteer` is that even though `Puppeteer` is `Headless Chrome` just like `Chrome` but the package size is way too big also consumes lots of memory while activating, `JSDOM` is rather small but also have the capability to test almost every case.

Ther-su:
After I mock ` getBoundingClientRect `, in test, I find `handleResize` don't trigger, so `restInputWidth` can't trigger. And the `inputWidth` can get the value that `getBoundingClientRect ` provide.

JeremyWuuuuu:
One thing that you have to know is that **THERE WERE NO LAYOUT ENGINE AT ALL**. 
Here is how the `ResizeObserver` implemented:
image

Ther-su:
Well, since `ResizeObserver ` can't work in non-browser environment, the `inputWidth` can't pass in the component when test, right?

Ther-su:
I got it. So how can I make sure that `inputWidth` can pass in the component since `ResizeObserver` can't work without `layout engine`, thanks?

Ther-su:
> I got it. So how can I make sure that `inputWidth` can pass in the component since `ResizeObserver` can't work without `layout engine`, thanks?
Hey! @JeremyWuuuuu for what I say above, can you give me some advice? `inputWidth` maybe important in test case, but I can't find a good way to let it pass to `JSDOM`.

JeremyWuuuuu:
What I meant was that you can trigger `ResizeObserver` via event `resize`, which you said you couldn't trigger the event passed to `ResizeObserver`, after this, you need to mock the `getBoundingClientRect` method then get the desired result that you wanted. For instance:

I mock this method to return a mock `rect` model, then I will trigger this method via `triggering ResizeObserver` via triggering `resize` event on the document.

Ther-su:
Done,BTW, for the multiple select with `collapseTags`, I just make sure that it can work well when the number < 100.

JeremyWuuuuu:
@Ther-su Nice work!

jw-foss (approved):

PR #971, opened 2020-12-11
fix(scrollbar): fix style of scrollbar

### LINKED ISSUE #964: Bug & improve with scrollbar
1. In many components like `select`, `time-select`, once click the scrollbar, when mouse up, it still scroll.
2. image
image
image
Scrollbar should has default `height: 100%`, even though, the horizontal scrollbar has a native bar.

kooriookami:
hide native scrollbar horizontal, give default height: 100%

fix #964

### DIFF
--- a/packages/theme-chalk/src/scrollbar.scss
+++ b/packages/theme-chalk/src/scrollbar.scss
@@ -19,6 +19,10 @@
     overflow: auto;
     height: 100%;
 
+    &::-webkit-scrollbar {
+      display: none;
+    }
+
     @include m(hidden-default) {
       scrollbar-width: none;
       &::-webkit-scrollbar {
--- a/packages/theme-chalk/src/scrollbar.scss
+++ b/packages/theme-chalk/src/scrollbar.scss
@@ -16,10 +16,14 @@
   }
 
   @include e(wrap) {
-    overflow: scroll;
+    overflow: auto;
     height: 100%;
-    &::-webkit-scrollbar {
-      display: none;
+
+    @include m(hidden-default) {
+      scrollbar-width: none;
+      &::-webkit-scrollbar {
+        opacity: 0;
+      }
     }
   }
 
--- a/packages/theme-chalk/src/scrollbar.scss
+++ b/packages/theme-chalk/src/scrollbar.scss
@@ -4,6 +4,7 @@
 @include b(scrollbar) {
   overflow: hidden;
   position: relative;
+  height: 100%;
 
   &:hover,
   &:active,
@@ -17,13 +18,8 @@
   @include e(wrap) {
     overflow: scroll;
     height: 100%;
-
-    @include m(hidden-default) {
-      scrollbar-width: none;
-      &::-webkit-scrollbar {
-        width: 0;
-        height: 0;
-      }
+    &::-webkit-scrollbar {
+      display: none;
     }
   }
 
--- a/packages/theme-chalk/src/scrollbar.scss
+++ b/packages/theme-chalk/src/scrollbar.scss
@@ -16,12 +16,8 @@
   }
 
   @include e(wrap) {
-    overflow: auto;
+    overflow: scroll;
     height: 100%;
-
-    &::-webkit-scrollbar {
-      display: none;
-    }
 
     @include m(hidden-default) {
       scrollbar-width: none;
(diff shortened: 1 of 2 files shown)


kooriookami:
https://test.kooriookami.top/deploy-preview-971--element-plus/#/zh-CN

jw-foss (approved):

JeremyWuuuuu:
cc @SimonaliaChen

packages/theme-chalk/src/scrollbar.scss
```diff
@@ -17,13 +18,8 @@
   @include e(wrap) {
     overflow: scroll;
     height: 100%;
-
-    @include m(hidden-default) {
```
SimonaliaChen, line 21: This is only useful when `native = false`, do not remove this rule.

-

SimonaliaChen (changes requested):

SimonaliaChen:
After investigation, I found out that the root cause you mentioned in #964 is that the `margin-bottom` style not working. So I prefer only change https://github.com/element-plus/element-plus/blob/dev/packages/scrollbar/src/index.vue#L8 to `native ? '' : 'el-scrollbar__wrap--hidden-default'` to fix it. Would you like to recorrect it ? or I will submit another pr.

kooriookami:
> After investigation, I found out that the root cause you mentioned in #964 is that the `margin-bottom` style not working. So I prefer only change https://github.com/element-plus/element-plus/blob/dev/packages/scrollbar/src/index.vue#L8 to `native ? '' : 'el-scrollbar__wrap--hidden-default'` to fix it. Would you @kooriookami like to recorrect it ? or I will submit another pr.

Only change this will cause another problems. The bottem was cut off.
image

JeremyWuuuuu:
I have merged the dependency request, you may rebase that then deploy again, so that we can see the result, also, is this only reproducible on a Windows machine? Because the scrollbar on Windows doesn't act as same as a Mac machine.

SimonaliaChen:
> > After investigation, I found out that the root cause you mentioned in #964 is that the `margin-bottom` style not working. So I prefer only change https://github.com/element-plus/element-plus/blob/dev/packages/scrollbar/src/index.vue#L8 to `native ? '' : 'el-scrollbar__wrap--hidden-default'` to fix it. Would you @kooriookami like to recorrect it ? or I will submit another pr.
> 
> Only change this will cause another problems. The bottem was cut off.
> image

I see. `margin-bottom` only works in certain cases. And try this.

```diff
    @include m(hidden-default) {
      scrollbar-width: none;
      &::-webkit-scrollbar {
-       width: 0;
-       height: 0;
+       opacity: 0;
      }
    }
```

kooriookami:
> > > After investigation, I found out that the root cause you mentioned in #964 is that the `margin-bottom` style not working. So I prefer only change https://github.com/element-plus/element-plus/blob/dev/packages/scrollbar/src/index.vue#L8 to `native ? '' : 'el-scrollbar__wrap--hidden-default'` to fix it. Would you @kooriookami like to recorrect it ? or I will submit another pr.
> > 
> > 
> > Only change this will cause another problems. The bottem was cut off.
> > image
> 
> I see. `margin-bottom` only works in certain cases. And try this.
> 
> ```diff
>     @include m(hidden-default) {
>       scrollbar-width: none;
>       &::-webkit-scrollbar {
> -       width: 0;
> -       height: 0;
> +       opacity: 0;
>       }
>     }
> ```

Well, it still weird. The visiable height of the div is 17 px less.
image

SimonaliaChen:
I'm afraid it's the best solution as I can imagine, both elegant and simple. Shall we take that for the temporary fix, pr is still welcome later

kooriookami:
> I'm afraid it's the best solution as far as I can imagine, both elegant and simple. Shall we take that for the temporary fix, pr is still welcome later

I made changes according to what you said.

packages/theme-chalk/src/scrollbar.scss
```diff
@@ -15,14 +16,17 @@
   }
 
   @include e(wrap) {
-    overflow: scroll;
+    overflow: auto;
     height: 100%;
 
+    &::-webkit-scrollbar {
+      display: none;
```
SimonaliaChen, line 23: Like I said before, native scrollbar will display when `native=true`. Do not add this

SimonaliaChen (changes requested):

packages/theme-chalk/src/scrollbar.scss
kooriookami: OK

SimonaliaChen (approved):

PR #993, opened 2020-12-14
fix dialog can't scroll if the content too large

### LINKED ISSUE #991: dialog can't scroll if the content to large
### Version
element-plus: 1.0.1-beta.8
vue: 3.0.4

### Environment
macOS 10.15.7

### Reproduction link
[view](https://codesandbox.io/s/distracted-ritchie-zpih5-zpih2?file=/src/App.vue)

### What is expected?
dialog can't scroll when the content to large

dialog scroll when the content to large

Ther-su:
fix #991


Ther-su:
image

packages/theme-chalk/src/dialog.scss
```diff
@@ -12,6 +12,8 @@
   box-shadow: $--dialog-box-shadow;
   box-sizing: border-box;
   width: 50%;
+  max-height: 70%;
```
JeremyWuuuuu, line 15: Why magical 70%?

Ther-su: I just causally to set the value.

Ther-su:
What is more, at first, I had two schemes to control the scrolling, one was to let the whole dialog scroll, the other was to let its body scroll, but I found the latter difficult to achieve with `max-width`.

packages/theme-chalk/src/dialog.scss
JeremyWuuuuu: I think we need to use `flex` to layout the UI, set the content to `flex:1`, also set `overflow` to `scroll`, since `70%` isn't a legit number to me.

JeremyWuuuuu:
> What is more, at first, I had two schemes to control the scrolling, one was to let the whole dialog scroll, the other was to let its body scroll, but I found the latter difficult to achieve with `max-width`.

Only the content of the dialog can be scrolled, although I think horizontal scrolling is not a good experience for the users.
One thing to notice is that, dialog **should be a light-weighted component** throughout the entire application.

packages/theme-chalk/src/dialog.scss
Ther-su: In your way, the height of dialog will still overflow its father div. It seems that the value of margin can't let the height of dialog compress.
image

JeremyWuuuuu: Maybe we can try something like making the `el-overlay` to be able to be scrollable.
Like when the content gets very long, the whole background can be scrolled.
Maybe this is a better way to do this, I suppose. BTW, the v2 version acts like this.

JeremyWuuuuu: Although the idea is that dialog shouldn't be putting too much stuff inside, but we cannot control how the developers/PMs behavior, so we can try this way. 
The idea is to find a better, elegant way to solve the issue, I want to hear your voice here.

Ther-su: Well, I use the way of v2. I think it may be a good way because we can't control the height of dialog.

JeremyWuuuuu:
Seems legit, approved.

jw-foss (approved):

PR #999, opened 2020-12-15
fix(table): get global config

### LINKED ISSUE #992: Global config of size not working with table
### Version
element-plus: 1.0.1-beta.8
vue: 3.0.4

### Environment
macOS 10.15.7

### Step to reproduce

1. app.use(ElementPlus, { size: 'small', zIndex: 3000 });
2. Use table in vue file
3. The size of the table is always as medium

### What is expected?
Global config of size working with table and other components

Global config of size just working with button

justwiner:
* [table] fix #992

### DIFF
--- a/packages/table/src/table/style-helper.ts
+++ b/packages/table/src/table/style-helper.ts
@@ -14,6 +14,7 @@ import {
   TableColumnCtx,
   fn,
 } from '../table.type'
+import { useGlobalConfig } from '@element-plus/utils/util'
 
 function useStyle(
   props: TableProps,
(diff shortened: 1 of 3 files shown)


packages/table/src/table/style-helper.ts
```diff
@@ -174,7 +187,7 @@ function useStyle(
     }
   }
   const tableSize = computed(() => {
-    return props.size
+    return props.size || getConfig('size')
```
JeremyWuuuuu, line 190: use `useGlobalConfig` in `@element-plus/utils/util` instead

jw-foss (approved):

PR #1016, opened 2020-12-16
fix(input): add resizeTextarea method

Laxse:
i think we need export resizeTextarea to support dynamic size like element2
```javascript
this.$refs.textarea.resizeTextarea();
```

### DIFF
--- a/packages/input/__tests__/input.spec.ts
+++ b/packages/input/__tests__/input.spec.ts
@@ -276,14 +276,24 @@ describe('Input.vue', () => {
     test('method:resizeTextarea', async () => {
       const testContent = 'TEXT:resizeTextarea'
       const wrapper = _mount({
-        template: `<el-input  ref="textarea"  type="textarea" v-model="text" />`,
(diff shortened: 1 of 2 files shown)


JeremyWuuuuu:
Add a test case to support the changes plz

packages/input/__tests__/input.spec.ts
```diff
@@ -273,6 +273,18 @@ describe('Input.vue', () => {
...
+      const testContent = 'TEXT:resizeTextarea'
+      const wrapper = _mount({
+        template: `<el-input  ref="textarea"  type="textarea" v-model="text" />`,
+        data() {
+          return {
+            text: testContent,
+          }
+        },
+      })
+      wrapper.vm.$refs.textarea.resizeTextarea()
```
JeremyWuuuuu, line 286: Could you make it measurable? Because I can't see what you were trying to test. For instance, test some value that changes inside `resizeTextarea`

Laxse: i got it

PR #1032, opened 2020-12-17
feat(textarea):export method resizeTextarea

Laxse:
i think we need export resizeTextarea to support dynamic size like element2

### DIFF
--- a/packages/input/__tests__/input.spec.ts
+++ b/packages/input/__tests__/input.spec.ts
@@ -289,7 +289,7 @@ describe('Input.vue', () => {
 
       _ref.autosize.minRows = 5
       _ref.autosize.maxRows = 5
-      
+
       _ref.resizeTextarea()
       // Textarea height will change
       setTimeout(()=>{
--- a/packages/input/__tests__/input.spec.ts
+++ b/packages/input/__tests__/input.spec.ts
@@ -283,18 +283,14 @@ describe('Input.vue', () => {
           }
         },
       })
-      const _ref = wrapper.vm.$refs.textarea
-      const originHeight = _ref.$el.clientHeight
+      const ref = wrapper.vm.$refs.textarea
+      const originMinHeight  = ref.textareaStyle.minHeight
 
-      _ref.autosize.minRows = 5
-      _ref.autosize.maxRows = 5
-
-      _ref.resizeTextarea()
-      // Textarea height will change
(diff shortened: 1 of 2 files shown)


packages/input/__tests__/input.spec.ts
```diff
@@ -273,6 +273,31 @@ describe('Input.vue', () => {
...
+
+      _ref.autosize.minRows = 5
+      _ref.autosize.maxRows = 5
+
+      _ref.resizeTextarea()
+      // Textarea height will change
+      setTimeout(()=>{
+        const nowHeight = _ref.$el.clientHeight
+        expect(originHeight===nowHeight).toBe(false)
+      },1000)
```
JeremyWuuuuu, line 298: We've had a `sleep` function in `@element-plus/test-utils/sleep`, use that instead.
```suggestion
const nowHeight = _ref.$el.clientHeight
expect(originHeight).not.toEqual(nowHeight)
```

Laxse: i get it

JeremyWuuuuu: Thank you for being patient with my comments.

packages/input/__tests__/input.spec.ts
```diff
@@ -284,19 +284,17 @@ describe('Input.vue', () => {
...
       _ref.resizeTextarea()
       // Textarea height will change
-      setTimeout(()=>{
+      sleep(200).then(() => {
         const nowHeight = _ref.$el.clientHeight
-        expect(originHeight===nowHeight).toBe(false)
-      },1000)
-
+        expect(originHeight).not.toEqual(nowHeight)
+      })
```
JeremyWuuuuu: Better use await syntax to keep consistency here.

```ts
await sleep(200)
const nowHeight = _ref.$el.clientHeight
expect(originHeight).not.toEqual(nowHeight)
```

Change this will be the last request.

Laxse: In jest dom doesn't render,so clientHeight always zero, i  consider textareaStyle can reflect the change

JeremyWuuuuu: That's totally ok to me. The only thing you need to do is to test it see if it matches the expected results.

jw-foss (approved):

PR #1067, opened 2020-12-21
fix: app.use(comp.vue) type is not compatible

XiongAmao:
Adding `install` function to the default import `.vue` sfc does not generate the type of `install`.
By reassigning a variable and declaring its type, it's possible to export d.ts that have both the
component type and the `install` type, but only work for `.vue` sfc.

```ts
// component index.ts
import { App } from 'vue'
import SomeComp from './SomeComp.vue'

const _SomeComp: typeof SomeComp & { install(app: App): void } = SomeComp

// vue-shim.d.ts
declare module '*.vue' {
  import { App, defineComponent } from 'vue'
  const component: ReturnType<typeof defineComponent> & {
    install(app: App): void
  }
  export default component
}
```

### DIFF
--- a/packages/aside/index.ts
+++ b/packages/aside/index.ts
@@ -1,5 +1,5 @@
 import { App } from 'vue'
-import { SFCWithInstall } from '../utils/types'
+import type { SFCWithInstall } from '@element-plus/utils/types'
 import Aside from '../container/src/aside.vue'
 
 Aside.install = (app: App): void => {
--- a/packages/aside/index.ts
+++ b/packages/aside/index.ts
@@ -1,9 +1,12 @@
 import { App } from 'vue'
+import { SFCWithInstall } from '../utils/types'
 import Aside from '../container/src/aside.vue'
 
 Aside.install = (app: App): void => {
   app.component(Aside.name, Aside)
 }
 
-export default Aside
+const _Aside: SFCWithInstall<typeof Aside> = Aside
+
+export default _Aside
 
--- a/packages/autocomplete/index.ts
+++ b/packages/autocomplete/index.ts
@@ -1,5 +1,5 @@
 import { App } from 'vue'
-import { SFCWithInstall } from '../utils/types'
+import type { SFCWithInstall } from '@element-plus/utils/types'
 import Autocomplete from './src/index.vue'
 
 Autocomplete.install = (app: App): void => {
--- a/packages/autocomplete/index.ts
+++ b/packages/autocomplete/index.ts
@@ -1,8 +1,11 @@
 import { App } from 'vue'
+import { SFCWithInstall } from '../utils/types'
 import Autocomplete from './src/index.vue'
 
 Autocomplete.install = (app: App): void => {
   app.component(Autocomplete.name, Autocomplete)
 }
 
-export default Autocomplete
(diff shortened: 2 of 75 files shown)


JeremyWuuuuu:
Seems legit. BTW, are you using `Volar`? Now seems Evan recommends us to use `Volar` as an alternative to `Vetur`, I need to know if this works with `Volar`

packages/autocomplete/index.ts
```diff
@@ -1,8 +1,11 @@
...
 import Autocomplete from './src/index.vue'
 
 Autocomplete.install = (app: App): void => {
   app.component(Autocomplete.name, Autocomplete)
 }
 
-export default Autocomplete
+const _Autocomplete: SFCWithInstall<typeof Autocomplete> = Autocomplete
+
+export default _Autocomplete
```
JeremyWuuuuu, line 11: Can we do something like this? Rather than reassigning it, which is totally meaningless statement.
image

JeremyWuuuuu: Using `cast` is better than the way you did IMHO.

packages/aside/index.ts
```diff
@@ -1,9 +1,12 @@
 import { App } from 'vue'
+import { SFCWithInstall } from '../utils/types'
```
JeremyWuuuuu, line 2: Also, use `import type { SFCWithInstall } from '@element-plus/utils/types'` instead.

packages/autocomplete/index.ts
XiongAmao: Using `cast` seems to generate the d.ts file incorrectly. It's missing `import Comp from 'path'` and the inclusion of all the types.

image

XiongAmao: I don't think it's a good idea to reassign values either, but I haven't found any other way.
And i don't really know how tsc generates d.ts file.

packages/aside/index.ts
XiongAmao: Done.
The packaged `lib` does not have the `el-utils` directory, the correct directory is `utils`., the correct directory is `utils`. It may be a problem with `gen-type.js`
image

XiongAmao:
> Seems legit. BTW, are you using `Volar`? Now seems Evan recommends us to use `Volar` as an alternative to `Vetur`, I need to know if this works with `Volar`

I've just heard about `Volar` in these days.  I will go test to see if it works correctly under `Volar`.

packages/autocomplete/index.ts
JeremyWuuuuu: Seems reasonable to me.

packages/aside/index.ts
JeremyWuuuuu: Aye, util was packaged independently. We need to fix that.

XiongAmao: I push a commit to fix `@element-plus/uitls` replacement. Should i push a sinlge PR for this? 
d2bc49644307a9eec3cabf572c1eb744a34ecaca

zazzaz (approved):

BeADre:
This pr seems to cause some ts errors

image

XiongAmao:
Maybe you should check if the local `typing/vue-shim.d.ts` is in effect. Because the exported `*.vue` module type actually comes from here.
image

PR #1109, opened 2020-12-24
docs(popover): change value/v-model to visible

Vgbire:
Fix that the popover component value/v-model doesn't work, because the doc document doesn't synchronize the code changes, and the source code is actually visible to control


Vgbire:
#1110 fix

Vgbire:
> Fix that the popover component value/v-model doesn't work, because the doc document doesn't synchronize the code changes, and the source code is actually visible to control

fix #1109

Issue #1119, opened 2020-12-25
[bug report] after inputnumbe is cleared, the focus is lost and the value of V-model is not changed when the value of Min is changed


wzohxia:
### Element Plus version
1.0.1-beta.14

### OS/Browsers version
Google Chrome  87.0.4280.88

### Vue version
3.0.4

### Reproduction Link
https://codepen.io/wzohxia/pen/MWjOKxp

### Steps to reproduce
1、清空input的值
2、input的值变成1
3、num的值是undefined

### What is Expected?
num的值也是1

### What is actually happening?
num的值是undefined

zazzaz:
working on this via #1252

PR #1237, opened 2021-01-08
fix(input-number): update value

### LINKED ISSUE #1229: [Bug Report] delete value of number-input, increase and decrease button will not work
### Element Plus version
1.0.1-beta.23

### OS/Browsers version
chrome87

### Vue version
3.0.5

### Reproduction Link
https://element-plus.org/#/zh-CN/component/input-number

### Steps to reproduce
delete value of number-input, increase and decrease button will not work

### What is Expected?
increase and decrease button will work

increase and decrease button will not work

BeADre:
fix #1229


JeremyWuuuuu:
Seems this is conflict with #1252. cc @Ryan2128

Ryan2128:
Sorry, I haven't see #1229 before, there are two issues #1119 and #1229, both of them have been fixed in #1252.

BeADre:
ok, I'll close the pr

PR #1270, opened 2021-01-13
fix(message): fix message component trigger many times bug #1269

iceshou:

### DIFF
--- a/packages/message/__tests__/message.spec.ts
+++ b/packages/message/__tests__/message.spec.ts
@@ -118,7 +118,8 @@ describe('Message.vue', () => {
 
       const closeBtn = wrapper.find('.el-message__closeBtn')
       expect(closeBtn.exists()).toBe(true)
-      wrapper.vm.destroyElement()
+      wrapper.vm.visible = false
+      wrapper.vm.onClose()
       expect(onClose).toHaveBeenCalled()
     })
 
@@ -184,7 +185,7 @@ describe('Message.vue', () => {
       wrapper.vm.close()
       expect(wrapper.vm.closed).toBe(true)
       await nextTick()
-      await wrapper.trigger('transitionend')
+      await wrapper.vm.onClose()
       expect(onClose).toHaveBeenCalledTimes(1)
     })
   })
--- a/packages/message/__tests__/message.spec.ts
+++ b/packages/message/__tests__/message.spec.ts
@@ -85,20 +85,6 @@ describe('Message.vue', () => {
       expect(domExports.off).toHaveBeenCalled()
     })
 
-    test('should listen to transitionend on close', async () => {
-      jest.spyOn(domExports, 'on')
-      jest.spyOn(domExports, 'off')
-      const wrapper = _mount({
-        slots: { default: AXIOM },
-      })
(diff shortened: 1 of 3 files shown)


packages/message/src/index.vue
```diff
@@ -112,9 +99,11 @@ export default defineComponent({
     off(document, 'keydown', this.keydown)
   },
   methods: {
+    handleAfterLeave() {
+      this.destroyElement()
```
JeremyWuuuuu, line 103: Why not use `destroyElement` directly?

packages/message/src/index.vue
```diff
@@ -112,9 +99,11 @@ export default defineComponent({
...
+    },
     destroyElement() {
       this.visible = false
```
JeremyWuuuuu, line 106: Since `this.visible` is set to `false` in `destroyElement` which would be called in `close` method. Setting `this.visible` here is redundant. Since this `destroyElement` and `handleAfterLeave` only calls `this.onClose`. You only need to pass `onClose` to `after-leave`. That'd be all.

packages/message/src/index.vue
```diff
@@ -135,6 +124,8 @@ export default defineComponent({
     // Event handlers
     close() {
       this.closed = true
+      this.visible = false
+      this.onClose()
```
JeremyWuuuuu, line 128: Delete this line.

packages/message/src/index.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <transition name="el-message-fade">
+  <transition name="el-message-fade" @after-leave="handleAfterLeave">
```
JeremyWuuuuu, line 2:
```suggestion
<transition name="el-message-fade" @after-leave="onClose">
```

packages/message/__tests__/message.spec.ts
```diff
@@ -85,20 +85,6 @@ describe('Message.vue', () => {
...
-        slots: { default: AXIOM },
-      })
-      wrapper.vm.close()
-      await nextTick()
-      expect(domExports.on).toHaveBeenLastCalledWith(
-        wrapper.vm.$el,
-        'transitionend',
-        wrapper.vm.destroyElement,
-      )
-    })
```
JeremyWuuuuu, line 101: Also, test `after-leave` hook please. You don't just delete test case.

packages/message/src/index.vue
fabin5: 'this.onClose()' can be deleted, but 'this.visible = false' cannot be deleted.

packages/message/src/index.vue
fabin5: OK

packages/message/src/index.vue
fabin5: use `onClose`

packages/message/src/index.vue
JeremyWuuuuu: What I said is `DELETE THIS LINE` which means the marked line 128. Thanks. I didn't use  plural here.

jw-foss (approved):

PR #1276, opened 2021-01-14
feat(utils): getPropByPath function adjust

jectychen:
getPropByPath 方法目前默认会对 path 进行解析，现实中会有不需要解析的情况。
因此这份修改判断 obj 是否存在 key 为 path 的情况，如果是的话，则直接返回对应的数据。


zazzaz:
Can you give us an example or demo code about the adjustment, please?

jectychen:
@zazzaz 
For example, in some cases, we may have some data source like this:
```javascript
const model = {
  'a.b.c': 'abc'
}
```

And we have the html structure link this:
```html
  <a-form-model :model="model">
    <a-form-model-item prop="a.b.c"></a-form-model-item>
  </a-form-model>
```

Our purpose is to use the a.b.c key and it's value directly, not the model[a][b][c].
Such case is not supported well right now.

sxzz (approved):

sxzz:
Thanks. Sorry for the looooooong delay.

PR #1284, opened 2021-01-15
improvement(directives): add v-same-dom

Vgbire:


JeremyWuuuuu:
Also modify `el-message-box`

Vgbire:
> Also modify `el-message-box`

OK,I think maybe I should modify the overlay to optimize it.you do the same thing in it.I will make a new pull requrest after.

Vgbire:
@JeremyWuuuuu OK.I didn't know at first that overlay should be optimized instead of changing other business components.

packages/dialog/src/index.vue
```diff
@@ -11,11 +11,13 @@
         :mask="modal"
         :overlay-class="modalClass"
         :z-index="zIndex"
-        @click="onModalClick"
+        :close="onModalClick"
```
JeremyWuuuuu, line 14: we use it as a event emitter, why would you change this as a value?

JeremyWuuuuu:
Seemed to me that this is way more complicated than I expected, I think.
I think the more appropriated version should be a hook like this:

### useClickSameDom.ts
```typescript
export default (emitter: () => void) {
    let mousedownTarget = false
    let mouseupTarget = false
    // refer to this https://javascript.info/mouse-events-basics
    // events fired in the order: mousedown -> mouseup -> click
    // we need to set the mousedown handle to false after click
    // fired.
    const onClick = (e: MouseEvent) => {
      // due to these two value were set only when props.mask is true
      // so there is no need to do any extra judgment here.
      // if and only if
      if (mousedownTarget && mouseupTarget) {
        emitter()
      }
      mousedownTarget = mouseupTarget = false
    }
    return onClick
} 

```

Vgbire:
I'm not sure if this change meets your expectations, if not, please let me know.

PR #1288, opened 2021-01-15
style(notification): fix style when its type set and title not set

Vgbire:
When you set type but not title in your notification, I don't think its style is centered. You can find this problem in the example on the official website.

In this modification, I checked 4 cases of setting and not setting type and title respectively, and the style is normal.

### DIFF
--- a/packages/theme-chalk/src/notification.scss
+++ b/packages/theme-chalk/src/notification.scss
@@ -37,8 +37,7 @@
 
   @include e(content) {
     font-size: $--notification-content-font-size;
-    line-height: 21px;
-    margin: 6px 0 0 0;
+    line-height: 24px;
     color: $--notification-content-color;
     text-align: justify;
 
--- a/packages/theme-chalk/src/notification.scss
+++ b/packages/theme-chalk/src/notification.scss
@@ -38,6 +38,7 @@
   @include e(content) {
     font-size: $--notification-content-font-size;
     line-height: 24px;
+    margin: 6px 0 0 0;
     color: $--notification-content-color;
     text-align: justify;
 
--- a/packages/notification/src/index.vue
+++ b/packages/notification/src/index.vue
@@ -20,7 +20,7 @@
         :class="{ 'is-with-icon': typeClass || iconClass }"
       >
         <h2 class="el-notification__title" v-text="title"></h2>
-        <div v-show="message" class="el-notification__content">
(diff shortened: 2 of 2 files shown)


packages/theme-chalk/src/notification.scss
```diff
@@ -37,8 +37,7 @@
 
   @include e(content) {
     font-size: $--notification-content-font-size;
-    line-height: 21px;
-    margin: 6px 0 0 0;
+    line-height: 24px;
```
Ryan2128, line 40:
```suggestion
    line-height: 24px;
    margin: 6px 0 0 0;
```

I think the margin style should be retained. When title is none, it should be overwritten with `margin:0;`.

So could you modify the `index.vue` of notification to implement it?

Ryan2128 (changes requested):

packages/theme-chalk/src/notification.scss
JeremyWuuuuu: Agreed

Vgbire: you mean just change the line-height to 24px?but i try if it has 6px top margin, it will be this.
I don't know why margin should be retained.Can you explain it in detail?
image

JeremyWuuuuu: What @Ryan2128 meant was, using javascript to determine whether notification had content or not. When no content is presented, you need to set the `margin` to `0` via javascript, so that you do not need to remove that line of `margin: 6px 0 0`

JeremyWuuuuu: Something like 

```vue
<!-- pseudo code -->
<template>
  <title :class="{has-body: !!$slots.body}">
    <icon />
  </title>
</template>
<style>
  .has-body {
     margin: 0;
  }
</style>
```

Vgbire: Oh, I know what that means? You want to keep this margin to make it more spaced when it has a title, right. I've taken this into consideration before, but I don't think it's necessary to make such a fuss.like this.
margin: 0
image
margin: 6px 0 0 0
image

Vgbire: so I won't change it, you can fix it by your own, it is easy.I'm a minimalist.

Ryan2128 (approved):

PR #1342, opened 2021-01-24
perf(scrollbar): when scrollbar isn't shown, hide the bar

kooriookami:


Ryan2128:
Not familiar with this component.💀

kooriookami:
> Not familiar with this component.💀

Haha, you can just look the effect of srcollbar.

packages/theme-chalk/src/scrollbar.scss
```diff
@@ -70,3 +59,16 @@
...
   }
 }
+
+.el-scrollbar-fade-enter-active {
+  transition: opacity 340ms ease-out;
+}
+
+.el-scrollbar-fade-leave-active {
+  transition: opacity 120ms ease-out;
+}
```
JeremyWuuuuu, line 69: Where did you get the transition duration time?

kooriookami: image
The original sass

packages/scrollbar/src/bar.vue
```diff
@@ -0,0 +1,123 @@
...
+
+      wrap.value[bar.value.scroll] = (thumbPositionPercentage * wrap.value[bar.value.scrollSize] / 100)
+    }
+
+    const startDrag = e => {
+      e.stopImmediatePropagation()
+      cursorDown.value = true
+      on(document, 'mousemove', mouseMoveDocumentHandler)
+      on(document, 'mouseup', mouseUpDocumentHandler)
+      document.onselectstart = () => false
```
JeremyWuuuuu, line 64: Shouldn't we store the original `onselectstart` handle and restore it after `mouseup`? We can hijack the event, but we can't just set it to null, what do you think?

kooriookami: sounds good

JeremyWuuuuu: You don't need a ref to store this. use a local variable instead.

jw-foss (approved):

PR #1364, opened 2021-01-26
fix(popover): fix z-index won't update when popover is shown

kooriookami:

### DIFF
--- a/packages/popover/src/usePopover.ts
+++ b/packages/popover/src/usePopover.ts
@@ -1,5 +1,5 @@
 import type { SetupContext } from 'vue'
-import { onMounted, ref, watch, watchEffect } from 'vue'
+import { ref, watch, watchEffect } from 'vue'
 import { isString } from '@element-plus/utils/util'
 import type { IPopperOptions } from '@element-plus/popper'
 import { usePopper } from '@element-plus/popper'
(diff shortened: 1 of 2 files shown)


packages/popover/src/usePopover.ts
```diff
@@ -1,38 +1,45 @@
-import { computed, watch } from 'vue'
+import type { SetupContext } from 'vue'
+import { onMounted, ref, watch, watchEffect } from 'vue'
```
JeremyWuuuuu, line 2: Remove `onMounted`

packages/popover/src/usePopover.ts
```diff
@@ -1,38 +1,45 @@
...
     }
-  })
+  }
 
-  const popperProps = usePopper(props, ctx)
+  const popperProps = usePopper(props, ctx as SetupContext<EmitType[]>)
+
+  watchEffect(() => {
+    updatePopperStyle()
+  })
```
JeremyWuuuuu, line 37: I think we only need to update the `zIndex` when `visibility` is set to true.

JeremyWuuuuu: Like the other components do, such as `Dialog` `Drawer`

kooriookami: when width changes, the unit test won't pass.

JeremyWuuuuu: I mean `width` and `zIndex` should be separated, because they are not updated via the same condition, so I think it's better to split it up.

jw-foss (approved):

PR #1381, opened 2021-01-28
fix(table): fix that the column width is calculated incorrectly

### LINKED ISSUE #1355: [Bug Report] Drag table columns，the width calculation not in effect
### Element Plus version
1.0.2-beta.29

### OS/Browsers version
chrome87

### Vue version
3.0.5

### Reproduction Link
https://codesandbox.io/s/olvof

### Steps to reproduce
Drag table columns

### What is Expected?
Recalculate width

Abnormal width

justwiner:
Fix the bug that the column width is calculated incorrectly when dragging the header

fix #1355

### DIFF
--- a/packages/table/src/table-body/styles-helper.ts
+++ b/packages/table/src/table-body/styles-helper.ts
@@ -2,7 +2,7 @@ import { getCurrentInstance } from 'vue'
 import { TableBodyProps } from './table-body'
 import { Table, AnyObject, TableColumnCtx } from '../table.type'
 
-function useStyles(props: TableBodyProps) {
+function useStyles (props: TableBodyProps) {
   const instance = getCurrentInstance()
   const parent = instance.parent as Table
   const isColumnHidden = index => {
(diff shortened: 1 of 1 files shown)


Ryan2128:
I think this bug is cause by the `show-overflow-tooltip` of props, but your fix seems to have nothing to do with this prop. Is this fix correct?

This is just my question after the test reproduction, I don’t really understand the logic of it.

justwiner:
image

justwiner:
This is a style bug.  

image
I don't think the width here needs to exist, because the width of the cell has been set in `colgroup`, the content inside the cell only needs to be 100%  

image

justwiner:
haha, there seems to be some problems. I'll have a look later

justwiner:
image
At present, I think this is the best solution. The reason for this issue is that the width of the column is calculated incorrectly. If the width of the column is not set, the `realwidth` of the column will be `null` after dragging the header. In fact, the real width is the width attribute.  

But on the other hand, you don't need to set width, it just needs to inherit the width of the parent container.

image

Ryan2128:
OK

Ryan2128 (approved):

PR #1384, opened 2021-01-28
refactor(docs): refactor docs site layout

Ryan2128:

refactor(docs): refactor docs site layout

### DIFF
--- a/website/app.vue
+++ b/website/app.vue
@@ -1,14 +1,10 @@
-<template>
-  <div id="app" :class="{ 'is-component': isComponent }">
-    <main-header v-if="lang !== 'play'" />
-    <div class="main-cnt">
-      <router-view />
-    </div>
-    <main-footer v-if="lang !== 'play' && !isComponent" />
-  </div>
-</template>
-
 <script>
+import { defineComponent, h, computed, watch, getCurrentInstance, onMounted } from 'vue'
+import { RouterView, useRoute } from 'vue-router'
+import { ElScrollbar } from 'element-plus'
+import { ElMessageBox } from 'element-plus'
+import MainHeader from './components/header'
+import MainFooter from './components/footer'
 import { use } from '@element-plus/locale'
 import zhLocale from '@element-plus/locale/lang/zh-cn'
 import enLocale from '@element-plus/locale/lang/en'
@@ -38,36 +34,17 @@ const localize = lang => {
 }
 localize(lang)
 
-export default {
+export default defineComponent({
   name: 'App',
 
-  computed: {
-    lang() {
-      return this.$route.path.split('/')[1] || Language.CN
-    },
-    isComponent() {
-      return /^component-/.test(this.$route.name || '')
-    },
-  },
+  setup(){
+    const route = useRoute()
 
-  watch: {
-    lang(val) {
-      if (val === Language.CN) {
-        this.suggestJump()
-      }
-      localize(val)
-    },
-  },
+    const lang = computed(() => route.path.split('/')[1] || Language.CN)
 
-  mounted() {
-    localize(this.lang)
-    if (this.lang === Language.CN) {
-      this.suggestJump()
-    }
-  },
+    const isComponent = computed(() => /^component-/.test(route.name || ''))
 
-  methods: {
-    suggestJump() {
+    const suggestJump = () => {
       if (process.env.NODE_ENV !== 'production') return
 
       const href = location.href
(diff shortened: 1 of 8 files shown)


website/app.vue
```diff
@@ -38,53 +33,73 @@ const localize = lang => {
...
 
       const href = location.href
       const preferGithub = localStorage.getItem('PREFER_GITHUB')
       const cnHref = href.indexOf('element-plus.gitee.io') > -1
       if (cnHref || preferGithub) return
       setTimeout(() => {
-        if (this.lang !== Language.CN) return
-        this.$confirm('建议大陆用户访问部署在国内的站点，是否跳转？', '提示')
+        if (lang.value !== Language.CN) return
+        _this.appContext.config.globalProperties.$confirm('建议大陆用户访问部署在国内的站点，是否跳转？', '提示')
```
JeremyWuuuuu, line 57: Maybe use `import { ElNotification } from `element-plus` could reduce this complexity.

website/app.vue
```diff
@@ -38,53 +33,73 @@ const localize = lang => {
...
+    }
   },
-}
+
+  render() {
+    const mainHeader = this.lang !== 'play' ? h(MainHeader, {
+      style: 'position: fixed;top: 0;width: 100%;z-index: 2000',
+    }) : null
+
+    const mainFooter = this.lang !== 'play' && !this.isComponent ? h(MainFooter) : null
```
JeremyWuuuuu, line 89: const isPlay = this.lang === 'play'

website/app.vue
```diff
@@ -38,53 +33,73 @@ const localize = lang => {
...
+
+    const content = [h('div', {
+      class: 'main-cnt',
+    }, [h(RouterView)]), mainFooter]
+
+    const contentWrapper = !this.isComponent ? h(ElScrollbar, null, { default: () => content }): content
```
JeremyWuuuuu, line 95: `const notComponent = !isComponent` since you have used twice.
also either break branches into lines or add a space before the `colon`

JeremyWuuuuu:
```suggestion
const contentWrapper = !this.isComponent 
  ? h(ElScrollbar, null, { default: () => content })
  : content
```
or
```suggestion
    const contentWrapper = !this.isComponent ? h(ElScrollbar, null, { default: () => content }) : content

```

website/components/right-nav.vue
```diff
@@ -38,6 +45,7 @@ export default defineComponent({
       await nextTick()
       scrollContainer = document.querySelector('.el-scrollbar.page-component__scroll>.el-scrollbar__wrap.el-scrollbar__wrap--hidden-default')
       const content = document.querySelector('.content.element-doc.content')
+      if (!content) return
```
JeremyWuuuuu, line 48: Can we use `DOM ref` do to the exact same thing? Since you can add ref to `scrollbar` I think thats achievable

website/pages/resource.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div class="page-container page-resource">
     <h2>{{ langConfig[1] }}</h2>
-    <p>{{ langConfig[placeholder2] }}</p>
+    <p>{{ langConfig['placeholder2'] }}</p>
```
JeremyWuuuuu, line 4: use `langConfig.placeholder2` instead.

JeremyWuuuuu: Shouldn't here be like `2`? instead of `placeholder2`? You can see that here are incremental numbers like `1, 3, 4, 5` I think this should be `2`.

Ryan2128: No, `resources`'s data structure is in blow:

```
"resource": {
  "1": "Resource",
  "2": "Under construction.",
  "3": "Axure Components",
  "4": "By importing Element UI in Axure, you can easily apply all the components we provide during interaction design.",
  "5": "Download",
  "6": "Sketch Template",
  "7": "Apply components from Element template, so you can improve design efficiency while keeping a unified visual style.",
  "8": "Interaction Design Documentation",
  "9": "Take a closer look at the interaction design documentation. Learn more details of each component from a microcosmic perspective.",
  "paraHeight": "1.6",
  "placeholder1": "Under construction",
  "placeholder2": "More resources are being developed"
}
```

I will change it to `langConfig.placeholder2`.

website/components/right-nav.vue
Ryan2128: Maybe it is a little difficult.

For the `scrollContainer`, because `scrollContainer` is innermost `div` of `scrollbar` component which in `component.vue`, though I  can provide the ref of the `scrollbar`, I also need query the innermost `div`.

For the `content`, this dom is in the component which generate by `md-loader`, the code in below:

```
const result = `
<template>
  <section class="content element-doc">
    ${output.join('')}
    <right-nav />
  </section>
</template>
${pageScript}
`
```

If use `ref` instead of querying dom, I need to modify `pageScript` which is a string value generate by code, I think it is unnecessary.

jw-foss (approved):

PR #1397, opened 2021-01-30
fix(select): fix not scroll to default option

### LINKED ISSUE #1356: [bug report] uses El select to select a data by default. After opening the drop-down list, it does not jump to the currently selected data anchor, but only displays the top data
### Element Plus version
1.0.2-beta.30

### OS/Browsers version
 86.0.4240.198（正式版本）谷歌

### Vue version
3.0.5

### Steps to reproduce
no

### What is Expected?
应该跳转到当前选中的数据

没有跳转当前选中数据的锚地

Ryan2128:

Fixed #1356

### DIFF
--- a/packages/select/src/useSelect.ts
+++ b/packages/select/src/useSelect.ts
@@ -520,16 +520,22 @@ export const useSelect = (props, states: States, ctx) => {
     }
   }
 
+  const getEl = option => {
+    const options = states.options.filter(item => item.value === option.value)
+    if (options.length > 0) {
+      return options[0].$el
+    }
+  }
+
   const scrollToOption = option => {
-    const target = Array.isArray(option) ? option[0]?.$el : option.$el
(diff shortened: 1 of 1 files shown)


packages/select/src/useSelect.ts
```diff
@@ -521,15 +521,21 @@ export const useSelect = (props, states: States, ctx) => {
   }
 
   const scrollToOption = option => {
-    const target = Array.isArray(option) ? option[0]?.$el : option.$el
+    const getEl = option => {
```
JeremyWuuuuu, line 524: Ssems to me that this function will only be called once. The only thing different is that you need to pick options out of multiple options, I don't see the benefit of encapsulating a function for that

Ryan2128: Yes, I will extract it to the outside.

zazzaz (approved):

PR #1433, opened 2021-02-05
fix(table): fix column default rendering errors

justwiner:
When rendering columns by default, default parameters should not be provided. Default parameters cannot be passed in as props of other components, which will cause other components to report errors

### DIFF
--- a/packages/table/src/table-column/index.ts
+++ b/packages/table/src/table-column/index.ts
@@ -7,6 +7,7 @@ import {
   getCurrentInstance,
   h,
   onBeforeUnmount,
+  Fragment,
 } from 'vue'
 import { cellStarts } from '../config'
 import { mergeOptions, compose } from '../util'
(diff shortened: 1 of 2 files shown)


packages/table/src/table-column/index.ts
```diff
@@ -209,20 +210,15 @@ export default defineComponent({
...
-          _self: {},
-          column: {},
-          row: {},
-          $index: -1,
-        }),
-      )
+      children = this.$slots
+        .default?.()
+        .filter(childNode => childNode.type?.name === 'ElTableColumn')
     } catch {
```
JeremyWuuuuu, line 219: You need to handle `v-for` when the root tag is `Fragment`, you have to filter at least 2 twice.

JeremyWuuuuu: For example, when use give you template like:
```vue
<el-table>
  <el-column v-for="" />
</el-table>
```

JeremyWuuuuu: The expression `this.$slots.default?.()` will resolve a value: `VNode<Fragment>` which only have length of 1 and its type will be `Fragment === childNode.type.name`

justwiner: yes, now it's fixed

Ryan2128 (approved):

PR #1448, opened 2021-02-08
feat(loading): Add request data for empty logic display and reuse

1063428973:
#

https://user-images.githubusercontent.com/32738523/107175359-914d2700-6a07-11eb-9ce2-bc68512af3ef.mp4


packages/loading/src/createLoadingComponent.ts
```diff
@@ -1,5 +1,6 @@
-import { createVNode, h, reactive, ref, render, toRefs, Transition, VNode, vShow, withCtx, withDirectives } from 'vue'
+import { createVNode, h, reactive, ref, render, toRefs, Transition, VNode, vShow, withCtx, withDirectives, defineComponent } from 'vue'
 import { removeClass } from '@element-plus/utils/dom'
+import Empty from '../../empty/src/index.vue'
```
nurlansu, line 3: Can't we use global import here?

```suggestion
import Empty from '@element-plus/empty'
```

packages/loading/src/createLoadingComponent.ts
```diff
@@ -87,31 +92,49 @@ export function createLoadingComponent({
 
       const spinnerText = h('p', { class: 'el-loading-text' }, [this.text])
 
+      const loadingContent = h('div', {
+        class: 'el-loading-spinner',
+      }, [
+        !this.spinner ? spinner : noSpinner,
```
nurlansu, line 98: Can you change the order for better reading?

```suggestion
        this.spinner ? noSpinner : spinner,
```

suyundukov (changes requested):

Ryan2128:
Sorry, I think this pr should not be merged, beacuse we should only provide atomic component.

Your requirements can be obtained by combining loading and emptying, instead of adding new functions to atomic components.

What's your advice? @JeremyWuuuuu

kooriookami:
Thanks for your PR, but this might not a good change. So I close it.

PR #1505, opened 2021-02-24
fix(time-select): computed time-select value

siam-ese:
time-select组件modevalue更改时,里面的value没有同步更改

[codepen bug demo](https://codepen.io/siamzz/pen/MWbOMgg)

### DIFF
--- a/packages/time-select/src/time-select.vue
+++ b/packages/time-select/src/time-select.vue
@@ -31,7 +31,7 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, ref, computed } from 'vue'
+import { defineComponent, computed } from 'vue'
 import ElSelect from '@element-plus/select'
 import ElOption from '@element-plus/option'
 interface Time {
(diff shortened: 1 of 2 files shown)


conan2008:
changed computed ， but it still doesn't work

packages/time-select/src/time-select.vue
```diff
@@ -144,7 +144,7 @@ export default defineComponent({
   emits: ['change', 'blur', 'focus', 'update:modelValue'],
   setup(props) {
     // data
-    const value = ref(props.modelValue)
+    const value = computed(() => props.modelValue)
```
BeADre, line 147: I guess the `value` is unnecessary, we can do `:model-value="modelValue"` in el-select

JeremyWuuuuu: Yep, that's the essential idea of not introducing more variables. The less the better.

JeremyWuuuuu:
Checkout #1725, the value is till being wrapped into `computed`, rebase the `origin/dev` branch then get do something with the `computed` statement. You can actually remove it, the computed is just some sort duplication.

PR #1582, opened 2021-03-05
fix(cascader): prevent log error for before filter

soc221b:

reproduce link: https://codesandbox.io/s/element-plus-spa-forked-ourpv?file=/src/App.vue

reproduce steps: 
1. open console
2. filter and type 'a'

Expected:
There should not log error.

### DIFF
--- a/packages/cascader/src/index.vue
+++ b/packages/cascader/src/index.vue
@@ -504,6 +504,7 @@ export default defineComponent({
 
       if (isPromise(passed)) {
         passed.then(calculateSuggestions)
+          .catch(() => { /* prevent log error */ })
       } else if (passed !== false) {
         calculateSuggestions()
       } else {


Ryan2128:
Why need to do this?

iendeavor:
Ummm, Why not?

packages/cascader/src/index.vue
```diff
@@ -504,6 +504,7 @@ export default defineComponent({
 
       if (isPromise(passed)) {
         passed.then(calculateSuggestions)
+          .catch(() => { /* prevent log error */ })
```
JeremyWuuuuu, line 507: Why should we hide error from bubbling up to the users?

JeremyWuuuuu: I personally think that the error should always be handled from the upper layer instead of us.

iendeavor: @JeremyWuuuuu 

Users can catch that error in the corresponding callback too:

https://codesandbox.io/s/element-plus-spa-forked-jfhlz?file=/src/App.vue:290-321

In addition, IMHO, user have to write more code (in `onErrorCaptured` hook) to catch that the error bubbled from component, and that would be annoying.

JeremyWuuuuu: Well, point made. If we bubbles the error to the user layer the whole application crashes, which is not something we would like to see.

jw-foss (approved):

PR #1585, opened 2021-03-06
fix(cascader-panel): click checkbox should works after lazy node loaded

soc221b:

Reproduce link: https://codesandbox.io/s/element-plus-spa-forked-vnzbv?file=/src/App.vue

Reproduce step:

1. Click the `Option - 1's *checkbox*` to load children.
2. Wait one second.
2. Click the checkbox again.

Expected:
It should be checked/indeterminate.

Actual:
Nothing happened. But if you click the checkbox again, it works.

This is due to the Cascader Panel's Checkbox use [one-way binding](https://github.com/element-plus/element-plus/blob/dev/packages/cascader-panel/src/node.vue#L24-L28), and the first click event is for expand/load children rather than change the checked state.

### DIFF
--- a/packages/cascader-panel/src/node.vue
+++ b/packages/cascader-panel/src/node.vue
@@ -121,7 +121,7 @@ export default defineComponent({
 
     const doCheck = () => {
       const { node } = props
-      panel.handleCheckChange(node, !node.checked)
+      panel.handleCheckChange(node, multiple.value ? !node.checked : true)
     }
 
     const doLoad = () => {
--- a/packages/cascader-panel/src/node.vue
+++ b/packages/cascader-panel/src/node.vue
@@ -119,10 +119,9 @@ export default defineComponent({
       panel.expandNode(props.node)
     }
 
-    const doCheck = (checked: boolean) => {
+    const doCheck = () => {
       const { node } = props
-      if (checked === node.checked) return
-      panel.handleCheckChange(node, checked)
+      panel.handleCheckChange(node, !node.checked)
     }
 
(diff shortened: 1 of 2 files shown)


packages/cascader-panel/src/node.vue
```diff
@@ -148,17 +147,17 @@ export default defineComponent({
       if (isHoverMenu.value && !isLeaf.value) return
 
       if (isLeaf.value && !isDisabled.value && !checkStrictly.value && !multiple.value) {
-        handleCheck(true)
```
SimonaliaChen, line 151: Clicking the option should always check it in single-select mode, remove this line will cause another interaction bug.

SimonaliaChen (changes requested):

packages/cascader-panel/src/node.vue
iendeavor: It is done, but I found another issue, It will emit close even it already checked.
Should I fix it?

SimonaliaChen (changes requested):

SimonaliaChen:
I just found out this pr and #1583 will cause other effects , such as unable to uncheck nodes whose child nodes are not fully loaded. since the new checked state depends on previous state, but with [loaded check](https://github.com/element-plus/element-plus/pull/1583/files#diff-2a3bcf35c2a7d4b660bf210cac58678dec17cbcf833dee2a9254c7e9a501c794L151), the `newCheckedState` will always be `true`. I will close this pr and submit another to fix this issue. Thanks for your contribution !

PR #1611, opened 2021-03-11
feat(space): Some minor adjustments

### LINKED ISSUE #1474: [Feature Request] Provide an option to make the margin-right of the last el-space-item 0, or set it to 0 by default
### Existing Component
Yes

### Component Name
el-space

### Description
I think the spacing should be set only between these elements, and the last element is not required spacing, and some scenarios need to align the layout of the two ends of multiple lines, but the last element has set margin-right value, resulting in some visual distance contrast

yinfoyuan:
> 1.Accomplish #1474 

> 2.Fixed Chinese text line wrapping issues

### DIFF
--- a/packages/space/src/index.ts
+++ b/packages/space/src/index.ts
@@ -4,6 +4,7 @@ import {
   createVNode,
   createTextVNode,
   isVNode,
+  unref,
 } from 'vue'
 import {
   PatchFlags,
@@ -34,6 +35,7 @@ export default defineComponent({
       itemStyle,
       spacer,
       prefixCls,
+      direction,
     } = ctx
 
     const children = renderSlot($slots, 'default', { key: 0 }, () => [])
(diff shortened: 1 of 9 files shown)


yinfoyuan:
#1474

website/docs/en-US/space.md
```diff
@@ -279,6 +279,7 @@ Setting this attribute can adjust the alignment of child nodes, the desirable va
 | prefixCls | Prefix for space-items | string | el-space | - |
 | style     | Extra style rules  | string / Array<Object \| String> / Object | - | - |
 | spacer    | Spacer    | string / number / VNode | - | - |
+| spacer-align      | Spacer align | string | center / left / right  | 'center' |
```
JeremyWuuuuu, line 282: I didn't see you adding the attribute to in file `useSpace.ts` `defaultProps`.

packages/space/src/index.ts
```diff
@@ -49,12 +51,33 @@ export default defineComponent({
...
+                if (direction === 'horizontal') {
+                  nodeStype.paddingBottom = unrefItemStyle.paddingBottom
+                } else {
+                  nodeStype.paddingBottom = '0'
+                }
+              } else {
+                nodeStype.marginRight = unrefItemStyle.marginRight
+                nodeStype.paddingBottom = unrefItemStyle.paddingBottom
+              }
+
```
JeremyWuuuuu, line 75: Actually here the type annotation is wrong, we should use `Unref<ReturnType<typeof useSpace>>` instead of `ReturnType<typeof useSpace>` here. The actual type of `itemStyle` is `Object` (despite the fact that this `itemStyle` is a computedRef, but it is a plain object here). We only need to mark the type. as I mentioned above.

JeremyWuuuuu: Refer to the image here:
image
image

JeremyWuuuuu: That was my bad.

website/docs/en-US/space.md
yinfoyuan: Error to submit Error to submit 😥

sxzz:
Hi, sorry to delay so long.
Could you please resolve conflicts? We changed file structure a lot recently.

PR #1633, opened 2021-03-14
fix(time-picker): fix the bug of clicking cancel

Ther-su:
fix the bug of clicking cancel

fix #1561

PR #1635, opened 2021-03-14
fix(time-picker): update oldValue when visible change

### LINKED ISSUE #1561: [bug report] TimePicker Click to cancel the bug
### Element Plus version
1.0.2-beta.33

### OS/Browsers version
all

### Vue version
3.0.7

### Reproduction Link
https://jsfiddle.net/iamkun/ods20kt/

### Steps to reproduce
重现链接直接看官网例子即可：https://element-plus.org/#/zh-CN/component/time-picker
1、更改时间并确定
2、再次打开更改面板，点击取消

### What is Expected?
点击取消后，时间不变化

点击取消后，时间变成了初始值，element2.0没有这个bug

inottn:
fix #1561

I found #1633 was trying to solve the same problem. I prefer to use watch to update oldValue. Because in the current version, there is more than one place that needs to do this operation. For example, click on the outside of pick panel.

In addition, range mode also had this problem. So I fixed it together.

### DIFF
--- a/packages/time-picker/src/time-picker-com/panel-time-range.vue
+++ b/packages/time-picker/src/time-picker-com/panel-time-range.vue
@@ -71,6 +71,7 @@ import {
   defineComponent,
   ref,
   computed,
+  watch,
   PropType,
   inject,
 } from 'vue'
@@ -124,6 +125,13 @@ export default defineComponent({
 
     const minSelectableRange = ref([])
     const maxSelectableRange = ref([])
+
+    // watch
+    watch(() => props.visible, val => {
+      if (!val) {
+        oldValue.value = props.parsedValue
+      }
+    })
 
     const handleConfirm = (visible = false) => {
       ctx.emit('pick', [minDate.value, maxDate.value], visible)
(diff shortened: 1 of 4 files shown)


packages/time-picker/src/time-picker-com/panel-time-range.vue
```diff
@@ -125,6 +126,13 @@ export default defineComponent({
     const minSelectableRange = ref([])
     const maxSelectableRange = ref([])
 
+    // watch
+    watch(() => props.visible, val => {
+      if (!val) {
+        oldValue.value = props.parsedValue
+      }
+    })
+
```
JeremyWuuuuu, line 135: Since the code is sort of the same here, can we encapsulate them together with some `useXxxx` hook to avoid duplication?

inottn: Yes, it is better.

inottn: done~

jw-foss (approved):

PR #1689, opened 2021-03-25
fix(image): avoid onload to be overwritten

### LINKED ISSUE #1688: e-image doesn't emit load event
version: 1.0.2-beta.35

el-image doesn't emit load event.
And when I listen on load event, img would fail to load.

Here is  the code in el-image/index.js

image

[codepen demo](https://codepen.io/vonlyinno/pen/XWpmVzR)

inottn:
fix #1688

### DIFF
--- a/packages/image/src/index.vue
+++ b/packages/image/src/index.vue
@@ -185,6 +185,8 @@ export default defineComponent({
       // so it can behave consistently
       Object.keys(attributes)
         .forEach(key => {
+          // avoid onload to be overwritten
+          if (key.toLowerCase() === 'onload') return
           const value = attributes[key]
           img.setAttribute(key, value)
         })
--- a/packages/image/__tests__/image.spec.ts
+++ b/packages/image/__tests__/image.spec.ts
@@ -99,5 +99,18 @@ describe('Image.vue', () => {
     expect(result).toBeTruthy()
   })
 
+  test('emit load event', async () => {
+    const handleLoad = jest.fn()
+    const wrapper = mount(Image, {
+      props: {
+        src: IMAGE_SUCCESS,
+        onLoad: handleLoad,
+      },
+    })
+    await doubleWait()
+    expect(wrapper.find('.el-image__inner').exists()).toBe(true)
+    expect(handleLoad).toBeCalled()
+  })
+
   //@todo lazy image test
 })


Ryan2128:
Thanks for your contribution.

What do you think about extracting the load into the vue event? In this case, only need to emit it in `handleLoad` function. This should be a little clearer.

packages/image/src/index.vue
```diff
@@ -185,6 +185,8 @@ export default defineComponent({
       // so it can behave consistently
       Object.keys(attributes)
         .forEach(key => {
+          // avoid onload to be overwritten
+          if (key.toLowerCase() === 'onload') return
```
JeremyWuuuuu, line 189: What about we do something like this:
```ts
Object.keys(attributes)
        .forEach(key => {
          // avoid onload to be overwritten
          const value = attributes[key]
          img.setAttribute(key, value)
        })
const prevOnload = image.onload;
image.onload = (e) => {
   handleLoaded(e, img)
   prevOnload?.()
}
```

JeremyWuuuuu: The main idea behind this is that we need to hijack the event instead of prevent it from binding, sometimes user need to know the exact moment that image is loaded.

inottn:
Thanks for reviews. @Ryan2128 @JeremyWuuuuu 

First of all, we didn’t prevent the event to be passed. In vue3, event listeners are part of $attrs. So we have passed the onload event  listener to `el-image__inner` in [this line of code](https://github.com/element-plus/element-plus/blob/a776dbc45f1a741e6c7d1209486755b49ccf52c4/packages/image/src/index.vue#L16).

In this case, what we need to do is make sure that the  [handleLoad method](https://github.com/element-plus/element-plus/blob/a776dbc45f1a741e6c7d1209486755b49ccf52c4/packages/image/src/index.vue#L196) is called and the loading ref is changed from true to false when the [img](https://github.com/element-plus/element-plus/blob/a776dbc45f1a741e6c7d1209486755b49ccf52c4/packages/image/src/index.vue#L180) is loaded. 

Keep behavior as same as element, the load event should be triggered by `el-image__inner`, not img.

In addition, the second parameter of setAttribute should be passed string. It is invalid to pass function.

JeremyWuuuuu:
> Thanks for reviews. @Ryan2128 @JeremyWuuuuu
> 
> First of all, we didn’t prevent the event to be passed. In vue3, event listeners are part of $attrs. So we have passed the onload event listener to `el-image__inner` in [this line of code](https://github.com/element-plus/element-plus/blob/a776dbc45f1a741e6c7d1209486755b49ccf52c4/packages/image/src/index.vue#L16).
> 
> In this case, what we need to do is make sure that the [handleLoad method](https://github.com/element-plus/element-plus/blob/a776dbc45f1a741e6c7d1209486755b49ccf52c4/packages/image/src/index.vue#L196) is called and the loading ref is changed from true to false when the [img](https://github.com/element-plus/element-plus/blob/a776dbc45f1a741e6c7d1209486755b49ccf52c4/packages/image/src/index.vue#L180) is loaded.
> 
> Keep behavior as same as element, the load event should be triggered by `el-image__inner`, not img.
> 
> In addition, the second parameter of setAttribute should be passed string. It is invalid to pass function.

Right, I forgot that part. Then I have no problem with this issue.

jw-foss (approved):

PR #1725, opened 2021-03-30
fix(time-select): input value not changed with v-model bindings

### LINKED ISSUE #1724: [Bug Report] el-time-picker value not change
### Element Plus version
1.0.2-beta.36

### OS/Browsers version
Windows 10 / Chrome 84 & Firefox 87.0

### Vue version
3.0.7

### Reproduction Link
https://codepen.io/hakurouken/pen/oNBzrLP

### Steps to reproduce
1. open https://codepen.io/hakurouken/pen/oNBzrLP
2. Click "Update Value" button.

### What is Expected?
The value of TimePicker should be changed to `18:15`.

Nothing happened.

HakurouKen:

fix #1724

### DIFF
--- a/packages/time-select/src/time-select.vue
+++ b/packages/time-select/src/time-select.vue
@@ -1,6 +1,6 @@
 <template>
   <el-select
-    v-model="value"
+    :model-value="value"
     :disabled="!editable"
     :clearable="clearable"
     :clear-icon="clearIcon"
@@ -8,10 +8,14 @@
     :placeholder="placeholder"
     default-first-option
     filterable
+    @update:model-value="
+      (value) => {
+        $emit('update:modelValue', value)
+      }
+    "
     @change="
       (value) => {
         $emit('change', value)
-        $emit('update:modelValue', value)
       }
     "
     @blur="(event) => $emit('blur', event)"
@@ -143,9 +147,8 @@ export default defineComponent({
   },
   emits: ['change', 'blur', 'focus', 'update:modelValue'],
   setup(props) {
-    // data
-    const value = ref(props.modelValue)
     // computed
+    const value = computed(() => props.modelValue)
     const items = computed(() => {
       const result = []
       if (props.start && props.end && props.step) {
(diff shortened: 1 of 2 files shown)


packages/time-select/src/time-select.vue
```diff
@@ -1,17 +1,21 @@
...
     :size="size"
     :placeholder="placeholder"
     default-first-option
     filterable
+    @update:model-value="
+      (value) => {
+        $emit('update:modelValue', value)
+      }
+    "
     @change="
```
JeremyWuuuuu, line 16: I don't quite understand why do you have to do this
since `v-model` is a syntax sugar for two way bindings which will be unwraps to `modelValue` and `update:modelValue`.
What is the point of rewriting this?

JeremyWuuuuu: I think I might know the root cause on this issue, the simplest way to fix this utilizing the props.modelValue instead of wrapping it with `computed`.

JeremyWuuuuu: cc @zazzaz

JeremyWuuuuu: The easiest way to fix this is that we bind it with `props.modelValue` instead of creating a new value, could you try that locally see if it works?

HakurouKen: Nope, `v-model` is shorten for `:model-value="v"` and `@update:model-value="v = $event"` (not `update:model-value="$emit('update:model-value', $event)"`), which will cause a error when clicking an option item in dropdown menu. I'll add a test case for that later.
image

HakurouKen: Also, maybe related to #1505 ?

JeremyWuuuuu: Right, make sense, I was wrong about the `@update` thing.  LGTM.

jw-foss (approved):

JeremyWuuuuu:
I think the code still had some flaws in it. But I merged it anyway. The `computed` statement is not really needed, using `props.modelValue` should be ok.

PR #1720, opened 2021-03-30
chore: 🤖 gen-type el-locale to locale

wukang0718:
gen-type el-locale to locale

After successful packaging, the file type that used utils and locale was imported to an incorrect address

PR #1736, opened 2021-04-01
Build: fix build bug cause import path error in xxx.d.ts

kaimeng:
Fix build bug.

Original build result, in lib/index.d.ts file and some other xxx.d.ts files,
import errors caused by 'import el-local' and' import el-utils' because local and utils modules don't have el- prefix.


Ryan2128 (approved):

Ryan2128:
Sorry, duplicate #1720

PR #1745, opened 2021-04-02
fix(checkbox): fix checkbox when modelvalue length less than min limit

### LINKED ISSUE #1742: [Bug Report] el-checkbox-group min
### Element Plus version
1.0.2-beta.36

### OS/Browsers version
mac os 11.2.3/ chrome  89.0.4389.90

### Vue version
3.0.11

### Reproduction Link
https://codepen.io/xiaolannuoyi/pen/ZELerQq

### Steps to reproduce
min>1 and `v-model` is `[]`

The checkbox cannot be selected

### What is Expected?
The checkbox can be selected

The checkbox cannot be selected

cokemine:
fix #1742

isLimitExceeded shouldn't be able to limit by min.
This may cause bug when model length < min - 1.

### DIFF
--- a/packages/components/checkbox/__tests__/checkbox.spec.ts
+++ b/packages/components/checkbox/__tests__/checkbox.spec.ts
@@ -294,34 +294,43 @@ describe('check-button', () => {
       `
       <el-checkbox-group
         v-model="checkList"
-        :min="1"
-        :max="2"
+        :min="2"
+        :max="3"
       >
         <el-checkbox-button label="a" ref="a"></el-checkbox-button>
         <el-checkbox-button label="b" ref="b"></el-checkbox-button>
         <el-checkbox-button label="c" ref="c"></el-checkbox-button>
         <el-checkbox-button label="d" ref="d"></el-checkbox-button>
+        <el-checkbox-button label="e" ref="e"></el-checkbox-button>
       </el-checkbox-group>
       `,
       () => ({
-        checkList: ['a'],
+        checkList: ['a', 'b'],
         lastEvent: null,
       })
     )
     const vm = wrapper.vm
-    expect(vm.checkList.length).toBe(1)
+    expect(vm.checkList.length).toBe(2)
     await wrapper.findComponent({ ref: 'a' }).trigger('click')
     vm.$nextTick(async () => {
-      expect(vm.checkList.length).toBe(1)
-      await wrapper.findComponent({ ref: 'b' }).trigger('click')
       expect(vm.checkList.length).toBe(2)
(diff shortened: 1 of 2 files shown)


packages/checkbox/src/useCheckbox.ts
```diff
@@ -50,9 +50,6 @@ const useModel = (props: ICheckboxProps) => {
       if (isGroup.value && Array.isArray(val)) {
         isLimitExceeded.value = false
 
-        if (checkboxGroup.min !== undefined && val.length < checkboxGroup.min.value) {
-          isLimitExceeded.value = true
-        }
```
JeremyWuuuuu, line 55: Changes here makes sense, but we need to find a way to utilize `min` property which should report error to the upper layer like `el-form` or somewhere else, we need to let the users know that the limit has not been reached.

JeremyWuuuuu: I think the value here `isLimitExceeded` was wrong in the first place, because setting the `min` property will never get the limitation exceeded, it should be something like `hasNotReached` flag, anyways, we need to find a way to report the error. 
I need to to work on that.

CokeMine: Yes. min should not get limitation exceeded.  El-form has it's own validation. I think it can be validated by the validater from elform

JeremyWuuuuu: Yeah, the `min` validation should be checked when its wrapped by `el-form`.

CokeMine:
@JeremyWuuuuu Hi, Will this PR be merged? This PR only solved one problem that `isLimitExceeded` was limited by `min`.If so, I will add some tests and mark it as `Ready for review`.
I think it will get a big change if validate `min` automatically when wrapped by `el-form` and wait for your work.

Now it can be validated manually such as,
```
data() {
    return {
      Form: {
        cities: [],
      },
      cityOptions,
      rules: {
        city: {
          message: '不合法',
          validator: () => this.Form.cities.length > 2 && this.Form.cities.length < 5,
          trigger: 'change',
        },
      },
    }
  }

sxzz:
Hi, sorry for delay so long.
Could you please resolve conflicts? We changed file structure a lot recently.

cokemine:
> Hi, sorry for delay so long.
> Could you please resolve conflicts? We changed file structure a lot recently.

Sorry I almost forgot about it. This PR does not completely fix the issue, if  need to fix it, may need to use `provide` and `inject`.
However I still think the code I deleted should be removed.

jw-foss (approved):

PR #1770, opened 2021-04-06
fix(avatar): fix src attributes default value

yuzhang9804:
A watch source can only be a getter/effect function, a ref, a reactive object, or an array of these types.

### DIFF
--- a/packages/avatar/src/index.vue
+++ b/packages/avatar/src/index.vue
@@ -38,7 +38,10 @@ export default defineComponent({
       },
     },
     icon: String,
-    src: String,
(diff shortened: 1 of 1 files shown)


JeremyWuuuuu:
What issue does this PR solve anyway? I don't see the point, if you set the props' type to primitive types, Vue will grand default value to it. There was no reason to do so.

yuzhang9804:
If the src attribute is undefined when not passed, Vue will display a warning
image
image

jw-foss (approved):

Issue #1763, opened 2021-04-06
[Feature Request] The TimePicker should hide the disabled hours, minutes, seconds


angela-1:
### Existing Component
Yes

### Component Name
TimePicker 

### Description
When I set some disabled minutes, the disabled minutes is still in the drop down list.
For example, I want to choose the minutes every 5 minutes([0,5,10,15, 20,25,30,35,40,45,50,55), and I set disabled the minutes not end with 0/5,  but the disabled minutes(2,3,4,6,7,8...) is still there.

It makes me inconvenient to fast choose time but the purpose i disable minutes is to fast choose time.

JeremyWuuuuu:
The request is quite legit, let's see if we can do anything with this. cc @zazzaz

bigcatLiu:
新版的time-picker没有options属性，不能设置step，如果我想设置分钟的时间间隔，改怎么设置呢？

xuxing409:
This feature is indeed necessary

LucaCuello:
I agree this is a needed feature

urani-dev:
Yep, implement this feature plaz

PR #1774, opened 2021-04-07
feat: hide disabled hours/minutes/seconds (#1763)

angela-1:

Hi, I have made a pr for my feature request #1763 . I tried on my pc and it looks like work fine.
I am a newbie and not good at testing, so maybe the modification of testing is not so good. 
Thanks for the library!

### DIFF
--- a/packages/date-picker/__tests__/date-time-picker.spec.ts
+++ b/packages/date-picker/__tests__/date-time-picker.spec.ts
@@ -257,12 +257,12 @@ describe('Datetime Picker', () => {
     const disabledHours = [].slice
       .call(hoursEl.querySelectorAll('.disabled'))
       .map(node => Number(node.textContent))
-    expect(disabledHours).toStrictEqual(disabledHoursArr)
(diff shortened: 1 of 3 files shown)


JeremyWuuuuu:
LGTM, cc @zazzaz

0song:
There's a question here, should we hide the disabled options or just make them disable?

JeremyWuuuuu:
We need to consult our designers to get approval on the change, because changing the using is not simply change the appearance, it also changes the way our user experience, so we cannot make the decision on our own. 
Thank you for contributing though. I will let you know once the result gets back.

angela-1:
> should

Like the #1763 mentioned, I want to choose minutes every 5 minutes(0,5,10,15,20,...), the disabled
minutes(1,2,3,4,11,12,...) are still shown, it makes me hard to choose the right minute, 
because between 0 to 5 has 4 disabled value, the scroll has problem to move from 0 to 5.

In my opinion, the disabled value are no need to shown as they have been "disabled" by user.

Issue #1692, opened 2021-03-25
[Feature Request] keep Drawer open


Fanna1119:
### Existing Component
Yes

### Component Name
Drawer
https://element-plus.org/#/en-US/component/drawer

### Description
the problem is when clicking outside the drawer it will want to close.

it would be good to have a 

`close-outside-click="false"` prop so if i still want to navigate my page while the drawer is open

JeremyWuuuuu:
If you do need the drawer to not having the backdrop set `modal` attribute to `false` also if you need to make the content scrollable, set `lock-scroll` to false.

Fanna1119:
> If you do need the drawer to not having the backdrop set `modal` attribute to `false` also if you need to make the content scrollable, set `lock-scroll` to false.

thank you this fixed it.

the documentation does not mention the `lock-scroll` prop

blackhu0804:
> > If you do need the drawer to not having the backdrop set `modal` attribute to `false` also if you need to make the content scrollable, set `lock-scroll` to false.
> 
> thank you this fixed it partially.
> it removed backdrop and keeps open, but i still am not able to click things, what css class to override to allow this?
> 
> the documentation does not mention the `lock-scroll` prop



> If you do need the drawer to not having the backdrop set `modal` attribute to `false` also if you need to make the content scrollable, set `lock-scroll` to false.



> If you do need the drawer to not having the backdrop set `modal` attribute to `false` also if you need to make the content scrollable, set `lock-scroll` to false.

I have a same question. How can I solve this problem？

JeremyWuuuuu:
I am sorry for the late response.
Right, since `el-overlay` will still be rendered into the DOM and holds the popups,  seems we need to have some 
discussion about this. @blackhu0804 had a fix for this, after the fix is released, you can override the style by using !important modifier(even though this is not suggested to use in you program).

PR #1932, opened 2021-04-30
fix(drawer): add modal false div class

blackhu0804:


blackhu0804:
To solve this problem. [#1692](https://github.com/element-plus/element-plus/issues/1692)

jw-foss (approved):

PR #2038, opened 2021-05-21
feat(click-outside):  return mousedown event

BelinChung:
Return mousedown event for binding function.

Sometimes it is necessary to prevent the click event for which components append to body or outside container.
```js
handleClickOutside(e) {
   const target = e.target
    if (target.closest('.ow-flow__handle')) return false

    this.$emit('update:visible', false)
}
```

### DIFF
--- a/packages/directives/click-outside/index.ts
+++ b/packages/directives/click-outside/index.ts
@@ -71,7 +71,7 @@ function createDocumentHandler(
     ) {
       return
     }
-    binding.value(mousedown)
+    binding.value(mouseup, mousedown)
   }
 }
 
--- a/packages/directives/click-outside/index.ts
+++ b/packages/directives/click-outside/index.ts
@@ -71,7 +71,7 @@ function createDocumentHandler(
     ) {
       return
     }
-    binding.value()
+    binding.value(mousedown)
   }
 }
 
--- a/packages/directives/__tests__/click-outside.spec.ts
+++ b/packages/directives/__tests__/click-outside.spec.ts
@@ -4,7 +4,15 @@ import ClickOutside from '../click-outside'
 const AXIOM = 'Rem is the best girl'
 const TRIGGER = 'trigger'
 const OTHER_CLASS = 'other-class'
-const handler = jest.fn()
+
+// init some local variables
+let mousedownObject
+let mouseupObject
+// mock handler with implementations which makes assignment to the local variable that we registered above.
+const handler = jest.fn((eMouseup, eMousedown) => {
+  mouseupObject = eMouseup
+  mousedownObject = eMousedown
+})
 const Component = {
   template: `
   <div>
@@ -52,6 +60,9 @@ describe('Directives.vue', () => {
     const mouseup = document.createEvent('MouseEvents')
     mouseup.initMouseEvent('mouseup', true, true, window, 0, 0, 0, 0, 0, false, false, false, false, 0, null)
     document.dispatchEvent(mouseup)
-    expect(handler).toHaveBeenCalledTimes(1)
+    // here is the different part
+    // we test the existence of the local variable.
+    expect(mousedownObject).toBeDefined()
+    expect(mouseupObject).toBeDefined()
   })
 })
--- a/packages/directives/__tests__/click-outside.spec.ts
+++ b/packages/directives/__tests__/click-outside.spec.ts
@@ -37,6 +37,10 @@ const _mount = () => mount(Component, {
 
 describe('Directives.vue', () => {
   beforeEach(() => {
+    // clear the previously assigned event object
+    mousedownObject = null
+    mouseupObject = null
+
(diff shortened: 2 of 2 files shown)


packages/directives/click-outside/index.ts
```diff
@@ -71,7 +71,7 @@ function createDocumentHandler(
     ) {
       return
     }
-    binding.value()
+    binding.value(mousedown)
```
JeremyWuuuuu, line 74: maybe we should expose both event to the handler so that they can do comparison to the `mousedown` target and `mouseup` target.

Also make sure your code is tested.

JeremyWuuuuu:
Please add test case against your change.

BelinChung:
Maybe this change requires adding a test case in `packages/directives/__tests__/click-outside.spec.ts`
But it may be a little difficult for me
This code is working for me
```js
handleClickOutside(mouseup, mousedown) {
  const target = mousedown.target
  if (target.closest('.ow-flow__handle')) return false

  this.$emit('update:visible', false)
}
```

JeremyWuuuuu:
> Maybe this change requires adding a test case in `packages/directives/__tests__/click-outside.spec.ts`
> But it may be a little difficult for me
> This code is working for me
> 
> ```js
> handleClickOutside(mouseup, mousedown) {
>   const target = mousedown.target
>   if (target.closest('.ow-flow__handle')) return false
> 
>   this.$emit('update:visible', false)
> }
> ```

I can walk you through on how to add test against your changes

I will take this chunk of code as an example:
```ts
test('should trigger handler', async () => {
  // init some local variables
  let mousedownObject
  let mouseupObject
  // mock handler with implementations which makes assignment to the local variable that we registered above.
  const handler = jest.fn((eMousedown, eMouseup) => {
    mousedownObject = eMousedown
    mouseupObject = eMouseup
  })

  const Component = {
    template: `
    <div>
      <div class="${OTHER_CLASS}">Hello</div>
      <div class="${TRIGGER}" v-click-outside="handler">${AXIOM}</div>
    </div>
  `,
    setup() {
      return {
        handler,
      }
    },
  }
    const wrapper = _mount()
    await wrapper.find(`.${TRIGGER}`).trigger('mousedown')
    await wrapper.find(`.${TRIGGER}`).trigger('mouseup')
    expect(handler).toHaveBeenCalledTimes(0)

    const mousedown = document.createEvent('MouseEvents')
    mousedown.initMouseEvent('mousedown', true, true, window, 0, 0, 0, 0, 0, false, false, false, false, 0, null)

    document.dispatchEvent(mousedown)
    const mouseup = document.createEvent('MouseEvents')
    mouseup.initMouseEvent('mouseup', true, true, window, 0, 0, 0, 0, 0, false, false, false, false, 0, null)
    document.dispatchEvent(mouseup)
    // here is the different part
    // we test the existence of the local variable.
    expect(mousedownObject).toBeDefined()
    expect(mouseupObject).toBeDefined()  
  })
```

In fact that I have just composed the entire test case that you needed to add. For this case I hope that I can give you the basic concept on how to add tests for your code.

JeremyWuuuuu:
Lol, by adding tests you need to add it locally which means you add it inside your own test case, do not change the previous setups for this special needs.

```ts

test('', () => {
  
  const handler = jest.fn(() => {
      // the code here.
   })
   const wrapper = mount({})

})

```

BelinChung:
em... I think the entire test case only needs a unique defined handler. It will expose both mouseup and mousedown event for where is it needed.
It has passed the local test. (but the project has one failure `FAIL  packages/element-plus/__tests__/ssr.spec.ts `)

JeremyWuuuuu:
Nah, by doing what you did, we ended up assigning to the same variable multiple times. Which is not wrong in correctness way, but it is wrong for context way, we do not want to do all the assignment stuff with each case. 

I'll explain in Chinese.

正确性上确实没问题，但是语义上是有问题的，因为每次都会给那个变量赋值，但是我们其他的测试并不关心这个变量的值，只有这一个测试用例关心而已，如果有好几个测试用例都关心这个值，那么应该将这个测试包装到另一个大的测试下，来测试这个值相关的代码。

BelinChung:
Yes. I agree with you.
But, 从这个组件的测试用例来说，`trigger handler` 应该作为核心测试用例进行保证，它的可读性应该优先保证。而且重复的 `handler`，`Component` 对于整个测试用例会严重干扰阅读。
我觉得从后续可预见性来说，2个鼠标事件仅对于 `render` 用例产生干扰，后续的用例可能均需要保证鼠标事件的准确性。利是大于弊的

JeremyWuuuuu:
If we go with your way, you will need to clear the previously assigned event object to null inside the `beforeEach` or `afterEach` block.

jw-foss (approved):

JeremyWuuuuu:
Great! Let's get mered.

PR #2100, opened 2021-05-28
feat(cascader): can be selected by clicking label (leaf node only)

Suzichen:
Cascader: Can be selected by clicking label (leaf node only) and auto matically closed

Cascader -> props.multiple = true
Cascader -> props.checkStrictly = true


JeremyWuuuuu:
I think this is by design.

PR #2120, opened 2021-06-01
feat: DatePicker support value-format

msidolphin:

reference #1923

### DIFF
--- a/packages/time-picker/src/common/picker.vue
+++ b/packages/time-picker/src/common/picker.vue
@@ -130,12 +130,12 @@ import {
   watch,
   provide,
 } from 'vue'
-import dayjs from 'dayjs'
+import dayjs, { Dayjs } from 'dayjs'
 import { ClickOutside } from '@element-plus/directives'
 import ElInput from '@element-plus/input'
 import ElPopper from '@element-plus/popper'
 import { EVENT_CODE } from '@element-plus/utils/aria'
-import { useGlobalConfig } from '@element-plus/utils/util'
+import { useGlobalConfig, isEmpty } from '@element-plus/utils/util'
 import { elFormKey, elFormItemKey } from '@element-plus/form'
 import { defaultProps } from './props'
 import type { ElFormContext, ElFormItemContext } from '@element-plus/form'
@@ -178,6 +178,15 @@ const valueEquals = function(a, b) {
     return dateEquals(a, b)
   }
   return false
+}
+
+const parser = function(date: Date | string, format: string): Dayjs {
+  const day = isEmpty(format) ? dayjs(date) : dayjs(date, format)
+  return day.isValid() ? day : undefined
+}
+
+const formatter = function(date: Date,  format: string) {
+  return isEmpty(format) ? date : dayjs(date).format(format)
 }
 
 export default defineComponent({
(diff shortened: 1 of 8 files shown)


JeremyWuuuuu:
Thank you for contributing, seems like you have a member which was not used and the linter complains about it. Please fix it ;)

iamkun:
Nice catch, seems better to add this feature back

packages/time-picker/src/common/picker.vue
```diff
@@ -274,15 +289,18 @@ export default defineComponent({
...
+          result = parser(props.modelValue as Date, props.valueFormat)
         }
       }
 
       if (pickerOptions.value.getRangeAvaliableTime) {
         result = pickerOptions.value.getRangeAvaliableTime(result)
       }
+      if (Array.isArray(result) && result.some(_ => !_)) {
+        result = []
+      }
```
0song, line 303: What's this line used for, please?

packages/time-picker/src/common/picker.vue
```diff
@@ -223,7 +232,13 @@ export default defineComponent({
...
     const emitInput = val => {
       if (!valueEquals(props.modelValue, val)) {
-        ctx.emit('update:modelValue', val)
+        let formatValue
+        if (Array.isArray(val)) {
+          formatValue = val.map(_ => formatter(_, props.valueFormat))
+        } else if(val) {
+          formatValue = formatter(val, props.valueFormat)
+        }
+        ctx.emit('update:modelValue', val ? formatValue : val)
```
0song, line 241: ctx.emit('update:modelValue', val || formatValue)

packages/time-picker/src/common/picker.vue
msidolphin: When the daterange selector used this property and the value is set to an illegal format, the value of parsedValue is [undefined, undefined], and an error will be reported when listening to parsedValue in panel-date-range.vue

当日期范围选择器使用这个属性并且将绑定的value设置为非法格式的时候，parsedValue的值是[undefined, undefined]，在panel-date-range.vue中监听parsedValue的时候会报错

packages/time-picker/src/common/picker.vue
msidolphin: ctx.emit('update:modelValue', formatValue || val)

0song: what I mean is to simplify it from `val ? formatValue : val` to `val || formatValue`

msidolphin: when empty case, I did't change the origin way

msidolphin: for example：click clear button

msidolphin: actually  `ctx.emit('update:modelValue', val ? formatValue : val)` is the right way

msidolphin:
> Thank you for contributing, seems like you have a member which was not used and the linter complains about it. Please fix it ;)

what issue?

0song:
> > Thank you for contributing, seems like you have a member which was not used and the linter complains about it. Please fix it ;)
> 
> what issue?

image

iamkun (approved):
LGTM

PR #2393, opened 2021-06-30
fix: table selectable callback index incorrect issue

msidolphin:

### DIFF
--- a/packages/table/__tests__/table-column.spec.ts
+++ b/packages/table/__tests__/table-column.spec.ts
@@ -944,4 +944,141 @@ describe('table column', () => {
       wrapper.unmount()
     })
   })
+
+  describe('tree table', () => {
+
+    const getTableData = () => {
+      return [{
+        id: 1,
+        date: '2016-05-02',
+        name: 'Wangxiaohu',
+        address: '1518 Jinshajiang Road, Putuo District, Shanghai',
+        index: 1,
+      }, {
+        id: 2,
+        date: '2016-05-04',
+        name: 'Wangxiaohu',
+        address: '1518 Jinshajiang Road, Putuo District, Shanghai',
+        index: 2,
+      }, {
+        id: 3,
+        date: '2016-05-01',
+        name: 'Wangxiaohu',
+        address: '1518 Jinshajiang Road, Putuo District, Shanghai',
+        index: 3,
+        children: [{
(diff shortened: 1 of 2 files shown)


JeremyWuuuuu:
Should we add test cases for verifying?

msidolphin:
> Should we add test cases for verifying?

I already added test case

packages/table/__tests__/table-column.spec.ts
```diff
@@ -944,4 +944,139 @@ describe('table column', () => {
...
+      const wrapper = createTable({
+        selectable(row, index) {
+          expect(row.index - 1).toEqual(index)
+          return !row.disabled
+        },
+      })
+      await nextTick()
+      wrapper.vm.$refs.table.toggleAllSelection()
+      wrapper.unmount()
+    })
```
JeremyWuuuuu, line 1079: use assertion like this:

```ts
const expectToBeCalled = jest.fn()
const wrapper = createTable({
        selectable(row, index) {
          expectToBeCalled((row.index - 1) === index)
          return !row.disabled
        },
      })
await nextTick()
wrapper.vm.$refs.table.toggleAllSelection()

expect(expectToBeCalled).toHaveBeenCalledWith(true)

```

msidolphin: I need to make sure that the result is correct every time, but I refactored the code anyway

JeremyWuuuuu: Humm, I think you can do that by asserting your result to `expect(toBeCalled).toHaveBeenNthCalled(order, params)`. Use a for loop to assert your result

packages/table/__tests__/table-column.spec.ts
```diff
@@ -944,4 +944,141 @@ describe('table column', () => {
...
+      const result = []
+      const wrapper = createTable({
+        selectable(row, index) {
+          result.push((row.index - 1) === index)
+          return !row.disabled
+        },
+      })
+      await nextTick()
+      wrapper.vm.$refs.table.toggleAllSelection()
+      expect(result.every(item => item)).toBeTruthy()
```
JeremyWuuuuu, line 1079: This is a way to test this case. Nice work.

jw-foss (approved):

PR #2396, opened 2021-06-30
feat(form): label-width supports number type

kooriookami:


packages/form/src/form-item.vue
```diff
@@ -47,7 +47,7 @@ import { computed, defineComponent, getCurrentInstance, inject, nextTick, onBefo
 import { NOOP } from '@vue/shared'
 import AsyncValidator, { RuleItem } from 'async-validator'
 import LabelWrap from './label-wrap'
-import { getPropByPath, useGlobalConfig } from '@element-plus/utils/util'
+import { addUnit, getPropByPath, useGlobalConfig } from '@element-plus/utils/util'
 import { isValidComponentSize } from '@element-plus/utils/validators'
```
JeremyWuuuuu, line 51: We should move these lines above `./label-wrap`

packages/form/src/form.vue
```diff
@@ -11,15 +11,9 @@
...
-  computed, reactive, toRefs, PropType,
-} from 'vue'
+import { computed, defineComponent, PropType, provide, reactive, ref, toRefs, watch } from 'vue'
 import mitt from 'mitt'
-import {
-  elFormKey, ElFormItemContext as FormItemCtx,
-  elFormEvents, ValidateFieldCallback,
-} from './token'
+import { elFormEvents, ElFormItemContext as FormItemCtx, elFormKey, ValidateFieldCallback } from './token'
 import { FieldErrorList } from 'async-validator'
```
JeremyWuuuuu, line 17: Minor here, this was not your case to fix but I'd like here before the imported member to have a `type` keyword.

JeremyWuuuuu:
Left 2 minor comments, please address it, thank you!

packages/form/src/form.vue
```diff
@@ -11,17 +11,13 @@
 </template>
 
 <script lang="ts">
-import {
-  defineComponent, provide, watch, ref,
-  computed, reactive, toRefs, PropType,
-} from 'vue'
+import { computed, defineComponent, PropType, provide, reactive, ref, toRefs, watch } from 'vue'
```
JeremyWuuuuu, line 14: `PropType` should be a type so we need to import it as a type.

jw-foss (approved):

JeremyWuuuuu:
You need to address the comment I left and also resolve the conflicts due to the recently merged commits.

jw-foss (approved):

packages/form/src/form-item.vue
```diff
@@ -145,7 +146,7 @@ export default defineComponent({
       if (!props.label && !props.labelWidth && isNested.value) {
         return {}
       }
-      const labelWidth = props.labelWidth || elForm.labelWidth
+      const labelWidth = addUnit(props.labelWidth) || addUnit(elForm.labelWidth)
```
equt, line 149: `addUnit` here will introduce a `warning`, if the `labelWidth` is by default using the value assigned from the `ElForm` component. It's better to first check its existence, then perform the conversion.

LBJDJW: thank you!

xxyGoTop: If the label-width is not set for the ElFormItem component, a warning will still be triggered，Now the label-width must be set?

plehnen: Is there already a bug report for this?
Working with beta.55 is very difficult with all those error messages showing up...

equt: @plehnen #2490

plehnen: @equt: that's the thing which breaks. because `props.labelWidth` doesn't exists (when only `elForm.labelWidth` is used), addUnit will produce a warning: "**ElementPlusError: [Util] binding value must be a string or number**"

plehnen: ah, nevermind. my e-mail client was showing me the comment from above. got me confused.

PR #2466, opened 2021-07-07
fix(table-column): parse width and minWidth

### LINKED ISSUE #1999: [bug report] dynamic modification of table column width results in style change
### Element Plus version
1.0.2-beta.44

### OS/Browsers version
chrome 90

### Vue version
3.1.0-beta.3

### Reproduction Link
https://codepen.io/zhoujiang/pen/yLMeWbx

### Steps to reproduce
初始化列宽度，点击修改宽度按钮， 表格列宽度变无效

### What is Expected?
可以动态修改列宽度

无法动态修改列宽度

BeADre:
fix #1999

### DIFF
--- a/packages/table/src/table-column/watcher-helper.ts
+++ b/packages/table/src/table-column/watcher-helper.ts
@@ -1,6 +1,6 @@
 import { watch, getCurrentInstance, ComputedRef } from 'vue'
 import { hasOwn } from '@vue/shared'
-import { TableColumnCtx, TableColumn } from './defaults'
+import type { TableColumnCtx, TableColumn, ValueOf } from './defaults'
 import { parseMinWidth, parseWidth } from '../util'
 
 function useWatcher<T>(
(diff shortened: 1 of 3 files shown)


packages/table/src/table-column/watcher-helper.ts
```diff
@@ -24,8 +25,15 @@ function useWatcher<T>(
         watch(
           () => props_[columnKey],
           newVal => {
-            instance.columnConfig.value[columnKey] = newVal
-            instance.columnConfig.value[key] = newVal
+            let value: any = newVal
```
JeremyWuuuuu, line 28: Since both `parseWidth` and `parseMinWidth` returns `number`, we should mark the type here as number I suppose?

BeADre: 只有进入下面的`if`条件语句之后`value`才一定会是`number`，其他的时候可能为其他类型

JeremyWuuuuu: Can we specify the type here? 
Say:

```ts
let value: numer | OtherType = newVal
```

JeremyWuuuuu: If we can't, then I think this is acceptable.

BeADre: done

jw-foss (approved):

JeremyWuuuuu:
Nice!

PR #2468, opened 2021-07-08
fix(menu): fix el-menu activeIndex bug

iwusong:
fix el-menu activeIndex bug when the vue-router navigation is failed

### DIFF
--- a/packages/menu/src/menu.vue
+++ b/packages/menu/src/menu.vue
@@ -205,24 +205,21 @@ export default defineComponent({
       }
 
       if (props.router && router && hasIndex) {
-        routeToItem(item, error => {
-          activeIndex.value = oldActiveIndex
-          if (error) {
-            // vue-router 3.1.0+ push/replace cause NavigationDuplicated error
-            // https://github.com/ElemeFE/element/issues/17044
-            if (error.name === 'NavigationDuplicated') return
-            console.error(error)
-          }
-        })
-      }
-    }
+        let route = item.route || item.index
+        router?.push(route)
+          .then(navigationResult => {
+            //vue-router NavigationFailureType duplicated
+            const DUPLICATE_ROUTE = 16
+            if (navigationResult && navigationResult.type !== DUPLICATE_ROUTE) {
+              activeIndex.value = oldActiveIndex
+              throw navigationResult
+            }
+          })
+          .catch(error => {
+            activeIndex.value = oldActiveIndex
+            throw error
+          })
 
-    const routeToItem = (item, onError) => {
-      let route = item.route || item.index
-      try {
-        router?.push(route, () => null, onError)
-      } catch (e) {
-        console.error(e)
       }
     }
 


itreetree:
router.push and all the other navigation methods return a Promise that allows us to wait til the navigation is finished and to know if it succeeded or failed. We will talk more about that in Navigation Handling.

packages/menu/src/menu.vue
```diff
@@ -205,24 +206,19 @@ export default defineComponent({
...
-          }
-        })
-      }
-    }
+        let route = item.route || item.index
+        router?.push(route)
+          .then(navigationResult => {
+            if (navigationResult && !isNavigationFailure(navigationResult, NavigationFailureType.duplicated)) {
+              activeIndex.value = oldActiveIndex
+              console.log(navigationResult)
```
JeremyWuuuuu, line 214: Remove these `console.log`

itreetree: now is ok

jw-foss (approved):

JeremyWuuuuu:
LGTM, need @Liiked to double check it.

packages/menu/src/menu.vue
```diff
@@ -50,6 +50,7 @@ import {
 import Menubar from '@element-plus/utils/menu/menu-bar'
 import ElMenuCollapseTransition from './menu-collapse-transition.vue'
 import useMenuColor from './useMenuColor'
+import { isNavigationFailure, NavigationFailureType } from 'vue-router'
```
Liiked, line 53: Combination of other library could be a burden on Element-Plus. Please remove this dependency and change the implementation.

Liiked (changes requested):
Do not introduce unnecessary library into Element-Plus.

packages/menu/src/menu.vue
JeremyWuuuuu: Agreed, I had this doubt at the beginning of reviewing this code, but I am not quite sure since the `vue-router` is not actually coupled with this component, that was why we fetched the object from `appContext` instead of importing it directly.

Even with tree-shaking I don't think that importing members from `vue-router` is a good idea. We should definitely put a line of comment here to fill in the context.

packages/menu/src/menu.vue
```diff
@@ -205,24 +205,19 @@ export default defineComponent({
...
-            if (error.name === 'NavigationDuplicated') return
-            console.error(error)
-          }
-        })
-      }
-    }
+        let route = item.route || item.index
+        router?.push(route)
+          .then(navigationResult => {
+            if (navigationResult && navigationResult.type !== 16) {
```
JeremyWuuuuu, line 211: I'm sorry that we have to avoid using magic numbers so that can you make 16 as a local variable like:

```ts
const DUPLICATE_ROUTE = 16

if (navigationResult && navigationResult.type !== DUPLICATE_ROUTE) {
  // ... other code
}
```
🙏

JeremyWuuuuu: This is the one last change request, after you change this we should be able to merge this PR. Thank you so much for collaborating. 🙇

itreetree: ok了:smile:

JeremyWuuuuu: 太棒了 谢谢配合！

Liiked (approved):
Good job!

PR #2507, opened 2021-07-13
fix(select): filtered empty groups

### LINKED ISSUE #2506: [Bug Report] Select filterable: labels of empty option groups should be filtered out
### Element Plus version
1.0.2-beta.55

### OS/Browsers version
Windows 10 / Edge Version 91.0.864.67

### Vue version
3.1.4

### Reproduction Link
https://codepen.io/maelga/pen/dyWOZRV

### Steps to reproduce
See codepen, using a filterable select with options grouping.
Try this: type "sh" in the filterable select will return 'Shanghai' and 'Shenzhen'.
However, option group labels with no options (i.e. "Australia", "Indonesia", "India") should not show.

### What is Expected?
When using filterable, the label of empty option groups should not show if no options are available. There is no benefits for users to see labels (e.g. country names in this codepen example) when no options (e.g. cities) can be selected. The select should show only option group labels (e.g. country names) if options are available (e.g. cities).

Labels of empty option groups are currently displayed even if there are no options displayed for the user to select.

msidolphin:
fix #2506

### DIFF
--- a/packages/select/src/option-group.vue
+++ b/packages/select/src/option-group.vue
@@ -10,7 +10,7 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, provide, inject, ref, reactive, toRefs } from 'vue'
+import { defineComponent, provide, inject, ref, reactive, toRefs, getCurrentInstance, onMounted } from 'vue'
 import { selectGroupKey, selectKey, selectEvents } from './token'
 
 export default defineComponent({
@@ -24,9 +24,10 @@ export default defineComponent({
       default: false,
     },
   },
-
   setup(props) {
     const visible = ref(true)
+    const instance = getCurrentInstance()
+    const children = ref([])
 
     provide(
       selectGroupKey,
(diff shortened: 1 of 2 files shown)


packages/select/src/option-group.vue
```diff
@@ -37,8 +38,32 @@ export default defineComponent({
 
     const select = inject(selectKey)
 
+    onMounted(() => {
+      children.value = flattedChildren(instance.subTree)
+    })
```
XingXiaoWu, line 43: I have been looking for a long time, but I have no idea how to get the options under the current group, I learned it

XingXiaoWu (approved):

packages/select/src/option-group.vue
JeremyWuuuuu: Although I think this is not the best way or the best approach to solve this issue you can see we took different approach in `ElSelectV2` which used data with explicit structure that hints us to render group or item. 

AntDesign took both ways for the users which I think the structured data would be more elegant.

jw-foss (approved):

PR #2522, opened 2021-07-14
feat(formItem): Change slot label formItem to scopedSlots

jdinartejesus:
Changed the slot from label on FormItem to be also a scoped slot to follow the same standard has error slot.

### DIFF
--- a/packages/form/src/form-item.vue
+++ b/packages/form/src/form-item.vue
@@ -13,7 +13,9 @@
         class="el-form-item__label"
         :style="labelStyle"
       >
-        <slot name="label">{{ label + elForm.labelSuffix }}</slot>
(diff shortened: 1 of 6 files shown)


jw-foss (approved):

JeremyWuuuuu:
@jdinartejesus 

The code LGTM, but could you update all documentations? If you don't speak these language that's ok, update them with the same English content so that we can have the others to work on the translations.

jdinartejesus:
@JeremyWuuuuu sure, maybe should I also move the label to scoped table on the documentation like the error?

image

JeremyWuuuuu:
Nice! Thank you for making ElementPlus better ;)

PR #2531, opened 2021-07-14
refactor(typography): add css var for typography & simplify example code

YunYouJun:
- use `map.get($--font-size, 'small')` instead of `$--font-size-small`
- set `--el-font-size-small: map.get($--font-size, 'small')`
- replace the variables used in each component
- simplify the typography sample document code

---

### DIFF
--- a/packages/theme-chalk/src/mixins/_var.scss
+++ b/packages/theme-chalk/src/mixins/_var.scss
@@ -1,3 +1,5 @@
+@use "sass:map";
+
 @mixin set-css-color-primary($i) {
   --el-color-primary-light-#{$i}: #{map.get($--colors, 'primary', 'light-#{$i}')};
 }
--- a/packages/theme-chalk/src/mixins/_var.scss
+++ b/packages/theme-chalk/src/mixins/_var.scss
@@ -1,0 +1,13 @@
+@mixin set-css-color-primary($i) {
+  --el-color-primary-light-#{$i}: #{map.get($--colors, 'primary', 'light-#{$i}')};
+}
+
+@mixin set-css-color-type($type) {
+  --el-color-#{$type}: #{map.get($--colors, $type, 'base')};
+  --el-color-#{$type}-light: #{map.get($--colors, $type, 'light')};
+  --el-color-#{$type}-lighter: #{map.get($--colors, $type, 'lighter')};
+}
+
+@mixin set-css-var-type($name, $type, $--variables) {
+  --el-#{$name}-#{$type}: #{map.get($--variables, $type)};
+}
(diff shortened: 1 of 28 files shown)


packages/theme-chalk/src/mixins/_var.scss
```diff
@@ -0,0 +1,13 @@
...
+}
+
+@mixin set-css-color-type($type) {
+  --el-color-#{$type}: #{map.get($--colors, $type, 'base')};
+  --el-color-#{$type}-light: #{map.get($--colors, $type, 'light')};
+  --el-color-#{$type}-lighter: #{map.get($--colors, $type, 'lighter')};
+}
+
+@mixin set-css-var-type($name, $type, $--variables) {
+  --el-#{$name}-#{$type}: #{map.get($--variables, $type)};
```
JeremyWuuuuu, line 12: I think we should name this variable more explicitly for instance: `$--variable-map` or something, or maybe that's just redundant given we have `map.get` here. WDTY?

packages/theme-chalk/src/mixins/_var.scss
```diff
@@ -0,0 +1,13 @@
+@mixin set-css-color-primary($i) {
```
JeremyWuuuuu, line 1: I didn't see the `@use "sass:map"` declaration here, is that ok for using this without explicitly importing?

jw-foss (approved):
@YunYouJun  I left 2 minor comments, let me know WDYT.

packages/theme-chalk/src/mixins/_var.scss
YunYouJun: Yeah, it can work. Because it is only called in `common/var.scss` which has `sass:map`.

Of course, just in case, I can add it later.

packages/theme-chalk/src/mixins/_var.scss
YunYouJun: Which is better, `$--variable-map` or `$--variables-map`?

> Because it is an internal variable, it has no effect on the user.

YunYouJun: checked the definitions of some other ui libraries and the examples given by sass. They usually don't use map as a suffix.
So I think `$--variables` is better.

YunYouJun: `$--variables` `s` means that it is a combination of many variables.

packages/theme-chalk/src/mixins/_var.scss
JeremyWuuuuu: So for better consistency here I think we need to explicitly add it so that would not cause any misleading.

YunYouJun (commented):
Done.

PR #2547, opened 2021-07-16
feat: new hook: `useCssVar` for inject custom css vars

zouhangwithsweet:
## Motivation

Easy for users to create custom themes

## How to use

In main.ts:
 > In fact, this step is not necessary, you can set some basic vars here
 ```ts
  const themeVars = {
   '--el-color-primary': '#f44336',
   '--el-color-white': '#2196f3',
  }
  app.provide(themeVarsKey, themeVars)
  ```
> Usually you need to use this function in the root component.
> If you pass a Ref vars param, it will generate a watch handler.  
 
In components' steup:
  ```ts
  setup() {
   const themeVars = ref({
     '--el-button-default-background-color': '#f44336',
     '--el-button-default-font-color': '#2196f3',
   })
   useCssVar(themeVars)
 }
 ```


JeremyWuuuuu:
Would you add a simple use case here?

packages/hooks/use-css-var/index.ts
```diff
@@ -0,0 +1,43 @@
+import { watch, isRef } from 'vue'
+import type { Ref, WatchStopHandle } from 'vue'
+
+let stopWatchCssVar: WatchStopHandle | null = null
+
+export function useCssVar(
+  vars:
+    | Ref<Record<string, string>>
+    | Record<string, string>,
+  target?: HTMLElement,
```
JeremyWuuuuu, line 10: Since the hook i invoked during `setup`, the target could have the possibilities to be a `RefElement`, by the time it's passed into this hook, the value would always be undefined.
I suggest we dispatch the action(attaching vars) at the `onMouned` hook stage so that the target could just be valid here.


```suggestion
target?: Ref<HTMLElement> | HTMLElement.
```

packages/hooks/use-css-var/index.ts
```diff
@@ -0,0 +1,43 @@
...
+) {
+  const elRef = target || window?.document?.documentElement
```
JeremyWuuuuu, line 12: This is not even a ref though.

packages/hooks/use-css-var/index.ts
```diff
@@ -0,0 +1,43 @@
...
+      vars,
+      val => {
+        setVars(val)
+      },
+      {
+        immediate: true,
+      },
+    ))
+    : setVars(vars)
+
```
JeremyWuuuuu, line 41: Right here we say:

```ts
import { onMounted } from 'vue'

setup() {
  onMounted(() => {
      isRef(target) ? setVars(target.value) : setVars(target)
  })

}

```

packages/hooks/use-css-var/index.ts
```diff
@@ -0,0 +1,43 @@
...
+export function useCssVar(
+  vars:
+    | Ref<Record<string, string>>
+    | Record<string, string>,
+  target?: HTMLElement,
+) {
+  const elRef = target || window?.document?.documentElement
+  const VAR_PREFIX = '--el-'
+
+  const setVars = (val: Record<string, string>) => {
```
JeremyWuuuuu, line 15:
```ts
const setVars = (el: HTMLElement,val: Record<string, string>) => {}
```

jw-foss (commented):
Left few comments, let me know WDYT.

msidolphin:
Can we not require the user to write the -el prefix?

JeremyWuuuuu:
> Can we not require the user to write the -el prefix?

I don't think we need to, since the user needs to know what they are doing, which means they are doing whatever at their own risk.

packages/hooks/use-css-var/index.ts
```diff
@@ -0,0 +1,105 @@
...
+ *  })
+ *  useCssVar(themeVars)
+ * }
+ * ```
+ */
+export function useCssVar(
+  vars: VarsType,
+  target?: Ref<HTMLElement> | HTMLElement,
+) {
+  let stopWatchCssVar: WatchStopHandle | null = null
```
JeremyWuuuuu, line 67: I more thing, we should do something like this:

```ts
import isServer from '@element-plus/util/isServer`
setup() {
    if (isServer) return
}
```

JeremyWuuuuu: Since this whole css thing would work without actual DOM elements, so we can say that this judgement is legit.

zouhangwithsweet: `optional chaining` prevents this error from occurring

JeremyWuuuuu: I understand that but isn't early terminating better than optional chaining? Besides, even though we use optional chaining throughout the project I think this is a better way to do that instead of relying on optional chaining (Plus, deep optional chaining gets compiled into ugly code even though we don't have to read that code but it definitely increases the bundle size with `window ? window.document ? window.document.documentElement : null : null : null`

jw-foss (approved):

Issue #2535, opened 2021-07-15
[bug report] 'carousel' click 'indicator' to switch the middle item to the first or last item, and the white screen will appear


RadiumAg:
### Element Plus version
1.0.2-beta.55

### OS/Browsers version
Chrome   91.0.4472.124

### Vue version
3.1.4

### Reproduction Link
https://user-images.githubusercontent.com/48897151/125716567-99f35767-8521-4280-ad39-de3b767fa31f.gif

### Steps to reproduce
Carousel组件点击`indicator`切换中间项到第一项或者最后一项

### What is Expected?
点击`indicator`切换中间项到第一项或者最后一项时，应该有平滑的过渡动画，其实更应该像这样 https://user-images.githubusercontent.com/48897151/125718313-539dd59f-0b53-423c-b35b-35a12180485b.gif     更加统一一点。

### What is actually happening?
点击`indicator`切换中间项到第一项或者最后一项时，出现白屏

binvb:
i can help in this

RadiumAg:
> i can help in this

thank you very much

PR #2548, opened 2021-07-16
fix(carousel): blank happened when using indicator

binvb:
reference: https://github.com/element-plus/element-plus/issues/2535  

在走马灯切换时候先等新的active模块切换好了之后，再把原位置的banner切换走

### DIFF
--- a/packages/carousel/src/item.vue
+++ b/packages/carousel/src/item.vue
@@ -9,7 +9,7 @@
       'is-hover': data.hover,
       'is-animating': data.animating,
     }"
-    :style="itemStyle"
+    :style="{transition: data.animating ? (data.active ? 'transform .4s ease-in-out' : 'transform .4s ease-in-out .4s') : 'transform 0s ease-in-out',...itemStyle}"
     @click="handleItemClick"
   >
     <div
@@ -123,6 +123,7 @@ export default defineComponent({
     ) => {
       const parentType = injectCarouselScope.type
       const length = injectCarouselScope.items.value.length
+
       if (parentType !== 'card' && oldIndex !== undefined) {
         data.animating = index === activeIndex || index === oldIndex
       }


JeremyWuuuuu:
I think the issue here is that the whole functionality is weird. 

Say if we have something like going from item 2 to item 4, we have 2 ways of doing so:
1. Moving forward, via 2 -> 3 -> 4.
2. Moving backward, via 2 -> 1 -> 4. This is the current moving strategies that we have which I think is weird. Logically we think that from 2 to 4 we normally use the first approach which is 2 -> 3 -> 4, but here instead of the normal approach we used 2 -> 1 -> 4 which naturally caused the the issue.

Also your change broke the animation which directly skipped the middle state 3, I think you might need to refine your change though.

binvb:
> I think the issue here is that the whole functionality is weird.
> 
> Say if we have something like going from item 2 to item 4, we have 2 ways of doing so:
> 
> 1. Moving forward, via 2 -> 3 -> 4.
> 2. Moving backward, via 2 -> 1 -> 4. This is the current moving strategies that we have which I think is weird. Logically we think that from 2 to 4 we normally use the first approach which is 2 -> 3 -> 4, but here instead of the normal approach we used 2 -> 1 -> 4 which naturally caused the the issue.
> 
> Also your change broke the animation which directly skipped the middle state 3, I think you might need to refine your change though.

i think it's only active item has transition: transform .4s ease(other item is default) caused the blank problem. but your are right, it should be in a reasonable order. i will refine this commit later,thank you

JeremyWuuuuu:
> > I think the issue here is that the whole functionality is weird.
> > Say if we have something like going from item 2 to item 4, we have 2 ways of doing so:
> > 
> > 1. Moving forward, via 2 -> 3 -> 4.
> > 2. Moving backward, via 2 -> 1 -> 4. This is the current moving strategies that we have which I think is weird. Logically we think that from 2 to 4 we normally use the first approach which is 2 -> 3 -> 4, but here instead of the normal approach we used 2 -> 1 -> 4 which naturally caused the the issue.
> > 
> > Also your change broke the animation which directly skipped the middle state 3, I think you might need to refine your change though.
> 
> i think it's only active item has transition: transform .4s ease(other item are defaults) caused the blank problem. but your are right, it should be in a reasonable order. i will refine this commit later,thank you

Sorry for the late response, and thank you for being collaborative, we look forward to see the updated version of this 👍

xiaoxian521:

PR #2550, opened 2021-07-16
fix(style): ensure that the loading icon is inline

anguiao:

When using some frameworks that include CSS reset, the svg element will be set to block, which will cause the style problem of the loading icon.
screenshot

### DIFF
--- a/packages/theme-chalk/src/loading.scss
+++ b/packages/theme-chalk/src/loading.scss
@@ -52,6 +52,7 @@
   }
 
   .circular {
+    display: inline;
     height: $--loading-spinner-size;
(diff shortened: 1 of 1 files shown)


JeremyWuuuuu:
I think we'd better control that via inline styling instead of globally change it since some users are using it as it is but this change could break that.

anguiao:
@JeremyWuuuuu Thanks for the reply.
This will only affect cases that use the default loading animation, and svg is inline by default, so it shouldn't cause break.

JeremyWuuuuu:
@anguiao I am happy to approve as it is

jw-foss (approved):

PR #2556, opened 2021-07-16
refactor(hooks): refactor transition

jw-foss:
- Introduce new hook named useTransition
- Update transition.scss with css vars

### DIFF
--- a/packages/theme-chalk/src/common/var.scss
+++ b/packages/theme-chalk/src/common/var.scss
@@ -14,23 +14,25 @@
 //  https://cubic-bezier.com/#p1,p2,p3,p4 (change px as your function parameter)
 
 :root {
-  --el-transition-duration: 300ms;
-  --el-ez-in-out-bezier-function: cubic-bezier(0.645, 0.045, 0.355, 1);
-  --el-fast-bezier-transition: cubic-bezier(0.23, 1, 0.32, 1);
-}
 
-$--all-transition: all var(--el-transition-duration)
-  var(--el-ez-in-out-bezier-function) !default;
-$--fade-transition: opacity var(--el-transition-duration)
-  var(--el-fast-bezier-transition) !default;
-$--fade-linear-transition: opacity var(--el-transition-duration) linear !default;
-$--md-fade-transition: transform var(--el-transition-duration)
-    var(--el-fast-bezier-transition),
-  opacity var(--el-transition-duration) var(--el-fast-bezier-transition) !default;
-$--border-transition-base: border-color var(--el-transition-duration)
-  var(--el-ez-in-out-bezier-function) !default;
-$--color-transition-base: color var(--el-transition-duration)
-  var(--el-ez-in-out-bezier-function) !default;
+  --el-transition-duration: 0.3s;
+  --el-fast-transition-duration: 0.2s;
+  --el-ease-in-out-bezier-function: cubic-bezier(0.645, 0.045, 0.355, 1);
+  --el-fast-bezier-transition: cubic-bezier(0.23, 1, 0.32, 1);
+
+  --el-all-transition: all var(--el-transition-duration)
+    var(--el-ease-in-out-bezier-function);
+  --el-fade-transition: opacity var(--el-transition-duration)
+    var(--el-fast-bezier-transition);
+  --el-md-fade-transition: transform var(--el-transition-duration)
+      var(--el-fast-bezier-transition),
+    opacity var(--el-transition-duration) var(--el-fast-bezier-transition);
+  --el-fade-linear-transition: opacity var(--el-fast-transition-duration) linear;
+  --el-border-transition-base: border-color var(--el-fast-transition-duration)
+    var(--el-ease-in-out-bezier-function);
+  --el-color-transition-base: color var(--el-fast-transition-duration)
+    var(--el-ease-in-out-bezier-function);
+}
 
 /* Color
 -------------------------- */
--- a/packages/theme-chalk/src/common/var.scss
+++ b/packages/theme-chalk/src/common/var.scss
@@ -10,14 +10,27 @@
 
 /* Transition
 -------------------------- */
-$--all-transition: all 0.3s cubic-bezier(0.645, 0.045, 0.355, 1) !default;
-$--fade-transition: opacity 300ms cubic-bezier(0.23, 1, 0.32, 1) !default;
-$--fade-linear-transition: opacity 200ms linear !default;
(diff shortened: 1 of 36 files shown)


packages/theme-chalk/src/common/var.scss
```diff
@@ -10,14 +10,27 @@
...
 /* Transition
 -------------------------- */
-$--all-transition: all 0.3s cubic-bezier(0.645, 0.045, 0.355, 1) !default;
-$--fade-transition: opacity 300ms cubic-bezier(0.23, 1, 0.32, 1) !default;
-$--fade-linear-transition: opacity 200ms linear !default;
-$--md-fade-transition: transform 300ms cubic-bezier(0.23, 1, 0.32, 1),
-  opacity 300ms cubic-bezier(0.23, 1, 0.32, 1) !default;
-$--border-transition-base: border-color 0.2s
-  cubic-bezier(0.645, 0.045, 0.355, 1) !default;
-$--color-transition-base: color 0.2s cubic-bezier(0.645, 0.045, 0.355, 1) !default;
```
YunYouJun, line 20: Are these `0.2s` -> `0.3s`?

Besides, `300ms` or `0.3s` which is better? 

> I see that bootstrap uses `0.2s`.

packages/theme-chalk/src/common/var.scss
```diff
@@ -10,14 +10,27 @@
...
+// refer to this website to get the bezier motion function detail
+//  https://cubic-bezier.com/#p1,p2,p3,p4 (change px as your function parameter)
+
+:root {
+  --el-transition-duration: 300ms;
+  --el-ez-in-out-bezier-function: cubic-bezier(0.645, 0.045, 0.355, 1);
+  --el-fast-bezier-transition: cubic-bezier(0.23, 1, 0.32, 1);
+}
+
+$--all-transition: all var(--el-transition-duration)
```
YunYouJun, line 22: Now that we have removed the variable, then we can use `--el-all-transition` instead of `$--all-transition`.

packages/theme-chalk/src/common/var.scss
```diff
@@ -10,14 +10,27 @@
...
-  opacity 300ms cubic-bezier(0.23, 1, 0.32, 1) !default;
-$--border-transition-base: border-color 0.2s
-  cubic-bezier(0.645, 0.045, 0.355, 1) !default;
-$--color-transition-base: color 0.2s cubic-bezier(0.645, 0.045, 0.355, 1) !default;
+// refer to this website to get the bezier motion function detail
+//  https://cubic-bezier.com/#p1,p2,p3,p4 (change px as your function parameter)
+
+:root {
+  --el-transition-duration: 300ms;
+  --el-ez-in-out-bezier-function: cubic-bezier(0.645, 0.045, 0.355, 1);
```
YunYouJun, line 18: Does `ez` stand for `ease`? 
Maybe it's better to use the full name.

packages/theme-chalk/src/common/var.scss
JeremyWuuuuu: Yup, will address this later.

packages/theme-chalk/src/common/var.scss
JeremyWuuuuu: I'm not sure since I have referenced from another project and I think integer here should be more precise than float, even though JavaScript doesn't have a very clear boundary on this.
Also the `0.2` was intentionally changed to `0.3` but I thought this could be a bad thing, do you have any suggestion of naming `0.2s` duration?

packages/theme-chalk/src/common/var.scss
JeremyWuuuuu: yeah, ease should be better for this, ez is a short term of easy which is not appropriate here.

packages/theme-chalk/src/common/var.scss
YunYouJun: `0.2s` or `200ms`, I have no idea, I checked several UI libraries, such as `bootstrap`, `vuetify`, `ant-design`.
They all use the form of `0.2s` instead of `200ms`. I think we can keep the same with other UI libraries.

In the jQuery era, I remember `slow` `normal` `fast` to describe animation.

- fast: 200ms
- normal: 400ms
- slow: 600ms

So I think we can reserve three variables `--el-transition-duration-fast` `--el-transition-duration-normal` `--el-transition-duration-slow`.

Of course we can do this first, and I will deal with this part later.

packages/theme-chalk/src/common/var.scss
```diff
@@ -10,14 +10,29 @@
...
-$--border-transition-base: border-color 0.2s
-  cubic-bezier(0.645, 0.045, 0.355, 1) !default;
-$--color-transition-base: color 0.2s cubic-bezier(0.645, 0.045, 0.355, 1) !default;
+// refer to this website to get the bezier motion function detail
+//  https://cubic-bezier.com/#p1,p2,p3,p4 (change px as your function parameter)
+
+:root {
+
+  --el-transition-duration: 0.3s;
+  --el-fast-transition-duration: 0.2s;
```
YunYouJun, line 19: emm, I see that the naming rules for some variables usually put the `type: success/warning` behind the `kind: color/button`.

Should we use `--el-transition-duration-fast` instead of `--el-fast-transition-duration`?

packages/theme-chalk/src/select.scss
```diff
@@ -76,7 +76,7 @@
         transform: rotateZ(180deg);
         border-radius: $--border-radius-circle;
         color: $--select-input-color;
-        transition: $--color-transition-base;
+        transition: var(--el-color-transition-base)e;
```
YunYouJun, line 79: Hey, i find a typo.

packages/theme-chalk/src/select-v2.scss
```diff
@@ -217,7 +217,7 @@ $--input-inline-start: 7px !default;
       transform: rotateZ(180deg);
       border-radius: $--border-radius-circle;
       color: $--select-input-color;
-      transition: $--color-transition-base;
+      transition: var(--el-color-transition-base)e;
```
YunYouJun, line 220: `e`!

packages/theme-chalk/src/common/var.scss
JeremyWuuuuu: Nice catch, I'll update the variable name

packages/theme-chalk/src/select.scss
JeremyWuuuuu: Nice catch, my bad.

JeremyWuuuuu:
@YunYouJun  Done

YunYouJun (approved):

PR #2565, opened 2021-07-19
refactor(color): add error var

YunYouJun:
- add `$--color-error` variables
- generate `--el-color-error` `--el-message-error-font-color` ...
- `success, info, warning, error` instead of `success, info, warning, danger`

---

### DIFF
--- a/packages/theme-chalk/src/alert.scss
+++ b/packages/theme-chalk/src/alert.scss
@@ -14,7 +14,7 @@
   --el-alert-icon-size: 16px;
   --el-alert-icon-large-size: 28px;
 
-  @each $type in (success, info, warning, danger) {
+  @each $type in (success, info, warning, error) {
     --el-alert-#{$type}-color: #{map.get($--colors, $type, 'lighter')};
   }
 }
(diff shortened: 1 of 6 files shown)


jw-foss (approved):

packages/theme-chalk/src/alert.scss
```diff
@@ -14,7 +14,7 @@
   --el-alert-icon-size: 16px;
   --el-alert-icon-large-size: 28px;
 
-  @each $type in (success, info, warning, danger) {
+  @each $type in (success, info, warning, error) {
```
JeremyWuuuuu, line 17: I think we can apply some redundant here to allow `danger` and `error` to exist at the same time since for uses in mainland China we tend to use `error` but for users outside mainland China, `danger` would be more syntaxially appropriate.

YunYouJun: Is alert special, or all components?

YunYouJun: alert `type` is `success/warning/info/error`.

If we want to add `danger`, that means we should add `danger` here too.

It may not be necessary.

Issue #2563, opened 2021-07-19
[Bug Report] use down-direction to move hoverIndex not right


CoderMonkie:
### Element Plus version
1.0.2-beta.55

### OS/Browsers version
Windows10 / Chrome 91.0.4472.124

### Vue version
3.2.0-beta.1

### Reproduction Link
https://codepen.io/codermonkie/pen/YzVQQpz

### Steps to reproduce
attributes:
    `multiple`
    `filterable`
    `allow-create`

1. type "foo" and down-direction and Enter (will create a "foo" tag that not in dropdown)
2. type "bar" and down-direction and Enter (will create a "bar" tag that not in dropdown)
3. click remove icon button to remove both "foo" and "bar" tag
4. keep input box focused and type direction key (up or down)
5. [BUG] to find hover style changing not correctly, it will stop at options not vilible

### What is Expected?
there are 3 options in dropdown, so take 3 times pressing down-direction for a round is better ?

### What is actually happening?
it will take 5 times pressing down-direction for a round

PR #2581, opened 2021-07-19
fix: select option does not delete in onBeforeMount

zhaozhongyu:

fix bug https://github.com/element-plus/element-plus/issues/2563, select option does not delete in onBeforeMount


packages/select/src/option.vue
```diff
@@ -67,20 +67,21 @@ export default defineComponent({
     } = toRefs(states)
 
     const vm = getCurrentInstance().proxy
+    const key = (vm as unknown as SelectOptionProxy).value
```
JeremyWuuuuu, line 70: The proxy is essentially the instance of `Option` which owned this `value` but `Option` gets the `value` via its parent which is `props.value` rewriting this doesn't make any change on this I assume?

JeremyWuuuuu: Would you elaborate the reason why were you doing this?

zhaozhongyu: 当我们手动输入一个值的时候, 创建option实例时使用了`vm.value`值作为`state.options`和`state.cachedOptions`的key, 此时可能只是输入了一个字符
image

当继续输入更多字符后再去触发销毁事件时, 以前的代码里面尝试使用当前的`props.value`进行删除, 此时`state.options`和`state.cachedOptions`找不到对应的key, 导致内存泄漏和按方向键选中时会有隐藏options的问题
image

PR #2664, opened 2021-07-25
refactor(locale): i18n migration

jw-foss:
- Update i18n implementation for files with initial C.

### DIFF
--- a/packages/calendar/src/index.vue
+++ b/packages/calendar/src/index.vue
@@ -8,18 +8,18 @@
             size="mini"
             @click="selectDate('prev-month')"
           >
-            {{ t('el.datepicker.prevMonth') }}
+            {{ locale.t('el.datepicker.prevMonth') }}
           </el-button>
           <el-button size="mini" @click="selectDate('today')">
             {{
-              t('el.datepicker.today')
+              locale.t('el.datepicker.today')
             }}
           </el-button>
           <el-button
             size="mini"
             @click="selectDate('next-month')"
           >
-            {{ t('el.datepicker.nextMonth') }}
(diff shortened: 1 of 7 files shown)


packages/calendar/src/index.vue
```diff
@@ -98,6 +100,7 @@ export default defineComponent({
 
   emits: ['input', 'update:modelValue'],
   setup(props, ctx) {
+    const locale = useLocaleInject()
```
YunYouJun, line 103: Just a small problem, why don‘t we use `const { t } = useLocaleInject()` instead of `const locale = useLocaleInject()`?

So we don't even have to modify the code we already have.

```
- return `${date.value.year()} ${t('el.datepicker.year')} ${t(pickedMonth)}`
+ return `${date.value.year()} ${locale.t('el.datepicker.year')} ${locale.t(pickedMonth)}`
```

YunYouJun (approved):

packages/calendar/src/index.vue
JeremyWuuuuu: Since useLocaleInject returns a `reactive` instead of a `ref`

YunYouJun: May be we can make it be same as this form of `vue-i18n` API.

> <https://github.com/intlify/vue-i18n-next/blob/1d7843afad0aa1fae51728dec1505e19b5d1058f/examples/composition/started.html#L47>

JeremyWuuuuu: ur right, I'm going to update the code to this style

YunYouJun (approved):

PR #2764, opened 2021-07-29
refactor(badge/card/dropdown/rate): migrate css var & use sass:map

YunYouJun:
- migrate `badge` `dropdown` `rate` `card` css var
- I'm trying to provide a new way for users to override variables using `sass` or `css var` at the same time.
 
WTDY

---

### DIFF
--- a/packages/theme-chalk/src/dropdown.scss
+++ b/packages/theme-chalk/src/dropdown.scss
@@ -1,7 +1,12 @@
 @import 'mixins/mixins';
+@import 'mixins/var';
 @import 'common/var';
 @import 'button';
 @import 'popper';
+
+:root {
+  @include set-component-css-var('dropdown', $--dropdown);
+}
 
 @include b(dropdown) {
   display: inline-block;
@@ -16,7 +21,7 @@
     @include picker-popper(
       $--color-white,
       1px solid var(--el-border-color-light),
-      $--dropdown-menu-box-shadow
+      var(--el-dropdown-menu-box-shadow)
     );
 
     .#{$namespace}-dropdown-menu {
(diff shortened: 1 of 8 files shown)


JeremyWuuuuu:
LGTM,  how do you intend to expose these APIs for users then?

jw-foss (approved):

YunYouJun:
Uses can use this way to custom sass variables.

```scss
$--card: (
  'padding': 10px
);
```

At the same time, they can use `--el-card-padding`.

Finally, I think it is necessary to have a special article about it.
Finally, ideally, write the variable name for each component last.

packages/theme-chalk/src/dropdown.scss
```diff
@@ -28,7 +33,7 @@
     }
 
     @include b(scrollbar__bar) {
-      z-index: #{$--dropdown-menu-index + 1};
+      z-index: calc(var(--el-dropdown-menu-index + 1));
```
tolking, line 36: There is a styles error same like #2760 on here, it many be `calc(var(--el-dropdown-menu-index) + 1)`

tolking (commented):
Sorry for post here, I found a style writing error that also appeared before #2760

packages/theme-chalk/src/dropdown.scss
YunYouJun: Thanks. I will fix it quickly.

YunYouJun: Do you want to commit for it as a finder?

tolking: Of course, I can create a PR

PR #2957, opened 2021-08-17
update the docs of table component to match the vue3 syntax

yingshaoxo:
It's just a sample for making sure that I'm doing the right modification.

Any suggestion is welcomed.

### DIFF
--- a/website/docs/en-US/table.md
+++ b/website/docs/en-US/table.md
@@ -31,8 +31,9 @@ Basic table is just for data display.
   </template>
 
   <script>
-    export default {
-      data() {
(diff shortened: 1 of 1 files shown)


iamkun:
Thanks.

Would you please refer to this pull request, https://github.com/element-plus/element-plus/pull/2363/files

to add a setup tag, to support both options API and composition api like this.
```
<!--
<setup>

</setup>
-->
```

You can test it here https://element-plus.org/#/en-US/component/radio
image

sxzz:
We have migrated to setup syntax, thank you for your contribution.

PR #3005, opened 2021-08-21
fix(descriptions): avoid conflict with el-table

### LINKED ISSUE #2910: [Bug Report] el-table inside el-descriptions width changing
### Element Plus version
1.0.2-beta.70

### OS/Browsers version
MacOS Big Sur 11.4 / Chrome 91.0.4472.114

### Vue version
3.1.2

### Reproduction Link
https://codesandbox.io/s/element-plus-spa-forked-rtf3s?file=/src/App.vue

### Steps to reproduce
Use table inside descriptions component. The table width is automatically increasing (see horizontal scrollbar in repro)

### What is Expected?
Table width should not change.

Table width infinitely increasing.

adaex:
fix #2910

### DIFF
--- a/packages/components/descriptions/src/descriptions-cell.ts
+++ b/packages/components/descriptions/src/descriptions-cell.ts
@@ -45,13 +45,13 @@ export default defineComponent({
       case 'label':
         return h(this.tag, {
           style: style,
-          class: ['el-descriptions__label', { 'is-bordered-label': this.descriptions.border }, labelAlign, labelClassName],
+          class: ['el-descriptions__cell', 'el-descriptions__label', { 'is-bordered-label': this.descriptions.border }, labelAlign, labelClassName],
           colSpan: this.descriptions.direction === 'vertical' ? span : 1,
         }, label)
       case 'content':
         return h(this.tag, {
           style: style,
-          class: ['el-descriptions__content', align, className],
+          class: ['el-descriptions__cell', 'el-descriptions__content', align, className],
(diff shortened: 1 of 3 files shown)


adaex:
This problem is difficult to troubleshoot. Special thanks to @imswk for positioning this problem, please add him as a co-contributor.

kooriookami:
image
Well, the problem is on `table` component. So, I think table needs a fix.

adaex:
To be honest, I thought so for the first time. But in fact, the reason for this problem is that the border style of `el-descriptions` affects the border style of `el-table`
As shown in the screenshot below
image

When `el-descriptions` set the `border` property, the `el-table` also displays border at the influence of the `td` style. However, the `el-table` component does not set the `border` property, so the actual display width of the `el-table` component will be different, so falling into the recalculation logic of the infinite loop.

You can try it when the `el-descriptions` does not set the border property, there will be no such problem.



So I use a simple way to change the `td` tag style of `el-descriptions` to use a class to avoid the `td` tags affecting the child elements.

JeremyWuuuuu:
@kooriookami Please help @adaex on this one. Thanks.

kooriookami:
Yes, it's seems good. And I think `table` component also need to improve `td tr` like this pr.

jw-foss (approved):

adaex:
@kooriookami  Ok, if you agree to use this way to solve the problem, I will finish it later.

kooriookami:
> @kooriookami Ok, if you agree to use this way to solve the problem, I will finish it later.

It's nice of you to solve this problem.

PR #3046, opened 2021-08-25
fix: fix the error of build

BeADre:
I am not sure if this modification is correct, please let me know if there is any problem @JeremyWuuuuu


JeremyWuuuuu:
@BeADre Thank you for contributing, but currently these scripts were not used in the build stage, you can run 

```shell
$ yarn build:comps # for components
$ yarn build:theme # for theme
$ yarn build:directives # for directives, as same as hooks, utils, token 
$ yarn build:full-bundle # for generating full bundle 
```

JeremyWuuuuu:
The scripts you were trying to fix is out dated(deprecated) see `scripts/monorepo.sh` for more information. Thank you!

BeADre:
OK, thank you

PR #3117, opened 2021-08-28
fix(components): [el-table] width init error (close #3087)

### LINKED ISSUE #3087: [Bug Report] el-space incrementally growing horizontally by itself without end
### Element Plus version
1.1.0-beta.6

### OS/Browsers version
Mac OS Big Sur - Firefox/Safari

### Vue version
3.2.6

### Reproduction Link
https://jsfiddle.net/1bd4sLto/

### Steps to reproduce
Simply placing an el-table inside an el-space with :fill=true will cause the div to expand forever

            <el-space :fill="true">
                <el-table>
                    <el-table-column>This will be expanding horizontally</el-table-column>
                </el-table>
            </el-space>

### What is Expected?
Expected for the el-space to take the max available width and stop moving

From t=0 of render time, the width of <main class="el-main"> element keeps increasing forever

imswk:
fix(#3087)

### DIFF
--- a/packages/components/table/src/table/style-helper.ts
+++ b/packages/components/table/src/table/style-helper.ts
@@ -6,6 +6,7 @@ import {
 import throttle from 'lodash/throttle'
 import { parseHeight } from '../util'
 import { useGlobalConfig } from '@element-plus/utils/util'
+import { on, off } from '@element-plus/utils/dom'
 
 import type { ResizableElement } from '@element-plus/utils/resize-event'
 import type { Table, TableProps } from './defaults'
@@ -159,7 +160,7 @@ function useStyle<T>(
     if (props.fit) {
       addResizeListener(table.vnode.el as ResizableElement, resizeListener)
     } else {
-      window.addEventListener('resize', doLayout)
+      on(window, 'resize', doLayout)
     }
   }
   onUnmounted(() => {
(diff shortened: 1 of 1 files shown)


JeremyWuuuuu:
May I ask why do we remove `resize` event listener here?

imswk:
> May I ask why do we remove `resize` event listener here?

#3087 

`el-space` style` min-width: 100%;` It will affect the value of `resize`. I don't think it's good to delete it directly. I've modified it.

packages/components/table/src/table/style-helper.ts
```diff
@@ -153,22 +153,24 @@ function useStyle<T>(
...
 
   const bindEvents = () => {
-    window.addEventListener('resize', doLayout)
     table.refs.bodyWrapper.addEventListener('scroll', syncPostion, {
       passive: true,
     })
     if (props.fit) {
       addResizeListener(table.vnode.el as ResizableElement, resizeListener)
+    } else {
+      window.addEventListener('resize', doLayout)
```
JeremyWuuuuu, line 162: Use `on` and `off` exported from `@element-plus/utils/dom`

imswk: OK，I'm done 🙇

sxzz (approved):

PR #3124, opened 2021-08-30
feat(components): space add gap compatible style

SeahorZhang:

### DIFF
--- a/packages/components/space/src/useSpace.ts
+++ b/packages/components/space/src/useSpace.ts
@@ -73,6 +73,27 @@ export const defaultProps = {
   },
 }
 
+function flexGapSupported (){
+  if (!window.document?.documentElement) return false
+  let flexGapSupported: boolean | null = null
+
+  // create flex container with row-gap set
+  const flex = document.createElement('div')
+  flex.style.display = 'flex'
+  flex.style.flexDirection = 'column'
+  flex.style.rowGap = '1px'
+
+  // create two, elements inside it
+  flex.appendChild(document.createElement('div'))
+  flex.appendChild(document.createElement('div'))
+
+  // append to the DOM (needed to obtain scrollHeight)
+  document.body.appendChild(flex)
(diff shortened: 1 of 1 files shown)


packages/components/space/src/useSpace.ts
```diff
@@ -116,21 +137,26 @@ export function useSpace(props: ExtractPropTypes<typeof defaultProps>) {
...
-        ? { flexWrap: 'wrap', marginBottom: `-${verticalSize.value}px` }
-        : null
+    const flexGapStyles: CSSProperties = flexGapSupported()
+      ? { gap:`${verticalSize.value}px ${horizontalSize.value}px` }
+      : { marginBottom: `-${verticalSize.value}px` }
+
+    const wrapKls: CSSProperties = props.wrap || props.fill
+      ? { ...flexGapStyles , flexWrap: 'wrap' }
+      : flexGapSupported() ? { gap:`${verticalSize.value || horizontalSize.value}px` } : null
+
```
JeremyWuuuuu, line 147: Could we cache the result? So that we can avoid the overhead of constantly creating and inserting and remove DOM element?

Even though computed does cache the result. But I think we could cache the result since the environment detecting is a one time thing.

Plus, you should avoid using raw DOM APIs without detecting the current environment. E.g. SSR context.

JeremyWuuuuu: BTW, the code seems legit to me.

Miss-Sixty: 很抱歉提出了一个考虑不那么成熟的代码，我会在时间允许的情况下去修复它的这些问题。

JeremyWuuuuu: It is ok. This is why code review is so important. If you need any help on contributing, please feel free to ping me in any thread. 

BTW, if you need to detect whether you are in `browser` you should try

```
import isServer from '@element-plus/utils/isServer`
```

PR #3189, opened 2021-09-02
fix(components): lazy image is not work when it changes src

zycoJamie:

el-image has a problem.
image opens lazy props , if i do not trigger lazy-load of image ,  i change src of image , picture will load immediately. lazy loading is not work.

### DIFF
--- a/packages/components/image/src/index.vue
+++ b/packages/components/image/src/index.vue
@@ -301,12 +301,17 @@ export default defineComponent({
       emit('switch', val)
     }
 
-    watch(
-      () => props.src,
-      () => {
+    watch(() => props.src, () => {
+      if (props.lazy) {
+        // reset status
+        loading.value = true
+        hasLoadError.value = false
+        removeLazyLoadListener()
+        nextTick(addLazyLoadListener)
+      } else {
         loadImage()
       }
-    )
+    })
 
     onMounted(() => {
       if (props.lazy) {


msidolphin:
Can you provide a demo?

zycoJamie:
https://codesandbox.io/s/friendly-pond-4ht68?file=/src/ImageDemo.vue

zycoJamie:
WeChatce7eb6aafd42f1c1dc2d552b60a7c7f4
WeChatc5334d1db6e0836e2c98fd7636d45920

change src, all picture will be loading.

JeremyWuuuuu:
@zycoJamie Can you record a video of after this modification how it behaves? 
Also I would like you to open a issue for tracking this bug

zycoJamie:
#3210

zycoJamie:
fix
after the fixes... when i change src of image, only loading the visible image, lazy loading do work.

msidolphin:
I think it is caused by you using the wrong key, try to use a unique id instead of idx

zycoJamie:
> I think it is caused by you using the wrong key, try to use a unique id instead of idx

when props.src is changed, the "loadImage" function must be called in the source code. So the problem has nothing to do with key..

JeremyWuuuuu:
LGTM, I would like to ask you for writing a unit test, for testing whether the `loadImage` is called right away. Note that we can also create a follow up issue for the unit test. This **is not** mandatory request.

jw-foss (approved):

msidolphin:
> > I think it is caused by you using the wrong key, try to use a unique id instead of idx
> 
> when props.src is changed, the "loadImage" function must be called in the source code. So the problem has nothing to do with key..
> 
> /packages/components/image/src/index.vue
> 
> ```js
> 286  watch(() => props.src, () => {
> 287   loadImage()
> 288  })
> ```

Since it's the same component instance, maybe using a unique key is a better solution, using the idx as the key is a bad habit

zycoJamie:
> LGTM, I would like to ask you for writing a unit test, for testing whether the `loadImage` is called right away. Note that we can also create a follow up issue for the unit test. This **is not** mandatory request.

umm...i can not think out how to check calling of the `loadImage ` in jest, i am not good at unit testing. sry..

JeremyWuuuuu:
@zycoJamie Please rebase the upstream so we can get this merged. About the test w can add a follow up issue for it. :bow:

sxzz (approved):

JeremyWuuuuu:
Let's get this merged! 🥂

PR #3274, opened 2021-09-07
fix(components): [el-date-picker] change shortcuts value

### LINKED ISSUE #3261: [Issue] el-date-picker change shortcuts language
Vue file
```
<el-date-picker
          :clearable="false"
          v-model="currentScheduleDate"
          ref="datePicker"
          type="date"
          format="DD/MM/YYYY"
          :shortcuts="shortcuts"
          :disabledDate="disabledDate"
          @change="onScheduleDatepickerChanged"
        >
        </el-date-picker>
```

ts
```
inside watch:
 this.shortcuts = this.setShortcuts();
 const datePicker = this.$refs.datePicker;
 datePicker.shortcuts = this.setShortcuts();
```
// I have tried to do it with a computed property as well and it doesn't work
datepicker doesn't change the language.
How can I change shortcuts language when someone changes the language?

imswk:
fix(#3261)


packages/components/date-picker/src/date-picker-com/panel-date-range.vue
```diff
@@ -681,6 +680,16 @@ export default defineComponent({
       clearable,
     } = pickerBase.props
 
+    const shortcuts = ref([])
+    shortcuts.value = pickerBase.props
```
JeremyWuuuuu, line 684: This is a little confusing, when you have `shortcut` props injected into `pickerBase.props` why you assign `pickerBase.props` to `shortcuts.value`?

Here we have
```ts
    const shortcuts = ref([])
    shortcuts.value = pickerBase.props
```
is essentially equivalent to
```ts
    const shortcuts = ref(pickerBase.props)
```
Also the statement here **IS WRONG**

it should be either

```ts
    const shortcuts = ref([])
    shortcuts.value = pickerBase.props.shortcuts
```

or

```ts
    const shortcuts = ref(pickerBase.props.shortcuts || [])
```

with the second way you are free to remove `immediate` from there.

JeremyWuuuuu: If you want to keep the `immediate` part, you will only need https://github.com/element-plus/element-plus/pull/3274/files#diff-c6cfc3c5d64d776c7962e059460dc01995f9f692faab74c6d2e3175eed0d80d6R683

packages/components/date-picker/src/date-picker-com/panel-date-pick.vue
```diff
@@ -617,14 +617,23 @@ export default defineComponent({
...
-      shortcuts,
       disabledDate,
       cellClassName,
       defaultTime,
       defaultValue,
       arrowControl,
     } = pickerBase.props
 
+    const shortcuts = ref([])
+    shortcuts.value = pickerBase.props
```
JeremyWuuuuu, line 628: Same thing goes here.

packages/components/date-picker/src/date-picker-com/panel-month-range.vue
```diff
@@ -118,7 +118,7 @@ export default defineComponent({
     const leftDate = ref(dayjs().locale(lang.value))
     const rightDate = ref(dayjs().locale(lang.value).add(1, 'year'))
 
-    const hasShortcuts = computed(() => !!shortcuts.length)
+    const hasShortcuts = computed(() => !!shortcuts.value.length)
```
JeremyWuuuuu, line 121: I think we can do better like:

```ts
const hasShortcuts = computed(() => shortcuts.value.length > 0)
```

the negation here is making no sense to me.

packages/components/date-picker/src/date-picker-com/panel-month-range.vue
```diff
@@ -255,7 +255,17 @@ export default defineComponent({
...
+
+    const shortcuts = ref([])
+    shortcuts.value = pickerBase.props
+    watch(
+      () => pickerBase.props.shortcuts,
+      (val) => {
+        shortcuts.value = val
+      },
+      { immediate: true }
+    )
```
JeremyWuuuuu, line 268: Same happens here
I see these code in 3 different places, I think we should consider encapsulate the code to avoid code duplication. Please use a `useShortcut` to encapsulate it.

packages/components/time-picker/src/common/picker.vue
```diff
@@ -557,8 +557,17 @@ export default defineComponent({
...
 
+    const baseProps: any = ref({ ...props })
+    watch(
+      () => props.shortcuts,
+      (val) => {
+        baseProps.value.shortcuts = val
+      },
+      { immediate: true }
+    )
+
```
JeremyWuuuuu, line 568: First:

Do not destruct `props` which will lose its reactivity,

convert this like

```
import { toRefs } from 'vue'
const baseProps = toRefs(props) // plus do not use any annotation unless you cannot work around.
```

2. With `toRef` method, `baseProps` is naturally reactive, refer to: https://v3.vuejs.org/api/refs-api.html#torefs
You do not have to watch `props.shortcuts` because THEY WERE CONNECTED.

packages/components/time-picker/src/common/picker.vue
```diff
@@ -557,8 +557,17 @@ export default defineComponent({
...
     provide('EP_PICKER_BASE', {
-      props,
+      props: baseProps.value,
```
JeremyWuuuuu, line 570: Under here we provide the ref-ed object directly:

```ts
provide({
  pickerProps: baseProps,
})
```

Then we use it like:

```ts

const pickerProps = inject(key, {})

pickerProps.shortcuts.value // note that `shortcuts` is a  `Ref`

```

imswk:
@JeremyWuuuuu  I am sorry for my behavior. Write code without thinking carefully. And the most embarrassing thing is that I knew I hadn't finished writing, but I forgot it later.

sxzz:
Hi, don't forget merge/rebase to the latest dev branch, the unit test workflow has been modified.

JeremyWuuuuu:
> @JeremyWuuuuu I am sorry for my behavior. Write code without thinking carefully. And the most embarrassing thing is that I knew I hadn't finished writing, but I forgot it later.

There is nothing to apologize for, take as much steps as you need to learn, no shame on that. :pray:

sxzz (approved):

imswk:
> > @JeremyWuuuuu I am sorry for my behavior. Write code without thinking carefully. And the most embarrassing thing is that I knew I hadn't finished writing, but I forgot it later.
> 
> There is nothing to apologize for, take as much steps as you need to learn, no shame on that. 🙏

thanks, I will 🙇

sxzz:
LGTM

tolking:

Sorry, it seems that this issue has already been fixed by other PR.

PR #3290, opened 2021-09-08
docs: migrate documents

adaex:
migrate:

- [x] migrate datetime-picker
- [x] migrate descriptions
- [x] migrate dialog
- [x] migrate divider
- [x] migrate drawer
- [x] migrate drapdown
- [x] migrate empty
- [x] migrate form

### DIFF
--- a/docs/en-US/component/descriptions.md
+++ b/docs/en-US/component/descriptions.md
@@ -1,201 +1,56 @@
-## Descriptions
+# Descriptions
 
 Display multiple fields in list form.
 
-### Basic usage
-
-:::demo
-
-```html
-<el-descriptions title="User Info">
-  <el-descriptions-item label="Username">kooriookami</el-descriptions-item>
-  <el-descriptions-item label="Telephone">18100000000</el-descriptions-item>
-  <el-descriptions-item label="Place">Suzhou</el-descriptions-item>
-  <el-descriptions-item label="Remarks">
-    <el-tag size="small">School</el-tag>
-  </el-descriptions-item>
-  <el-descriptions-item label="Address"
-    >No.1188, Wuzhong Avenue, Wuzhong District, Suzhou, Jiangsu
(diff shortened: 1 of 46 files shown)


docs/en-US/component/descriptions.md
```diff
@@ -1,201 +1,56 @@
-## Descriptions
+# Descriptions
 
 Display multiple fields in list form.
 
-### Basic usage
+<style lang="scss">
```
JeremyWuuuuu, line 5: Should we add `scoped` here? Since if we visit this page then visit other page we might pollute other page's style

docs/examples/datetime-picker/default-time.vue
```diff
@@ -0,0 +1,48 @@
...
+      defaultTime1: [new Date(2000, 1, 1, 12, 0, 0)], // '12:00:00'
+      defaultTime2: [
+        new Date(2000, 1, 1, 12, 0, 0),
+        new Date(2000, 2, 1, 8, 0, 0),
+      ], // '12:00:00', '08:00:00'
+    })
+
+    return {
+      ...toRefs(state),
+    }
```
JeremyWuuuuu, line 45: I think simply return

```ts
return toRefs
```
would be better, since this is useless code.

docs/examples/datetime-picker/date-and-time.vue
```diff
@@ -0,0 +1,72 @@
...
+      ],
+      value1: '',
+      value2: '',
+      value3: '',
+      defaultTime: new Date(2000, 1, 1, 12, 0, 0), // '12:00:00'
+    })
+
+    return {
+      ...toRefs(state),
+    }
```
JeremyWuuuuu, line 69: Same goes here.

adaex:
@JeremyWuuuuu  Updated.

PR #3356, opened 2021-09-11
Feat/remove mitt

### LINKED ISSUE #419: Proposal: giving up a "mitt"

purepear:
# Remove `mitt` as build dependency by using provide/inject instead.

Discussion https://github.com/element-plus/element-plus/discussions/3158

This PR is slightly based on https://github.com/element-plus/element-plus/pull/431

_It'll probably be easier for reviewers do the initial review commit by commit_

## Affected components: 
* collapse / collapse-item
* menu / submenu / menu-item 
* tree (drag n drop)
* select / option-group / option: we are using `shallowRef` and `triggerRef` to simulate the existing mitt behavior but i imagine the **select filtering** functionality could be refactored in the future(and probably not rely on shallowRef/triggerRef as they seem slightly difficult from maintainability point of view)
* form / form-item: refactor the `validate()` functionality requires small changes in most form controls(cascader, checkbox, color-picker, input, radio, select, slider, switch, time-picker, transfer)

I removed `mitt` from 
* `packages/components/package.json`
* `packages/element-plus/package.json`
* `packages/tokens/package.json`

I left `mitt` in the root `package.json` as it is used by the `website` events [bus](https://github.com/element-plus/element-plus/blob/dev/website/bus.js) in order to fade in/out the website navigation. 

I believe removing mitt from the packages should be enough to remove it from the library build.

I also saw mitt in the [webpack build config](https://github.com/element-plus/element-plus/blob/dev/build/webpack.config.js#L30) 🤔 ~Shall i remove  it?~ **I removed it.**


JeremyWuuuuu:
@sxzz Can you follow through this PR review process? I will check this PR once you have done the reviewing.

sxzz:
Of course.

packages/components/collapse/src/collapse-item.vue
```diff
@@ -49,7 +49,7 @@
 </template>
 <script lang="ts">
 import { defineComponent, PropType, inject, computed, ref } from 'vue'
-import { CollapseProvider } from './collapse'
+import type { CollapseProvider } from './collapse.type'
```
sxzz, line 52: Could you please move this line 52 to 55?

sxzz (approved):

packages/components/collapse/src/collapse-item.vue
purepear: Yep, makes sense to move `import type` after the imports 👍 . Done :)

sxzz:
@JeremyWuuuuu Review please.

packages/components/form/src/form-item.vue
```diff
@@ -296,43 +289,20 @@ export default defineComponent({
...
-
     const elFormItem = reactive({
       ...toRefs(props),
       size: sizeClass,
       validateState,
       $el: formItemRef,
-      formItemMitt,
-      removeValidateEvents,
-      addValidateEvents,
+      evaluateValidationEnabled,
```
JeremyWuuuuu, line 305: Praise: This is good!

packages/components/form/src/form-item.vue
```diff
@@ -258,16 +256,11 @@ export default defineComponent({
...
       } else {
         prop.o[prop.k] = initialValue
       }
-      // reset validateDisabled after onFieldChange triggered
-      nextTick(() => {
-        validateDisabled.value = false
-      })
     }
 
     const getRules = () => {
```
JeremyWuuuuu, line 266: I was just wondering can we make this one as `computed` since we could be invoking this for a lot of times.

NB: we can and we should put this into separate PRs

packages/components/select/src/useOption.ts
```diff
@@ -3,10 +3,12 @@ import {
   computed,
   getCurrentInstance,
   watch,
-  onBeforeUnmount,
+  Ref,
```
JeremyWuuuuu, line 6: **Suggestion**:  we can move this to 

```ts
import type { Ref } from 'vue'
```

packages/components/select/src/useOption.ts
```diff
@@ -3,10 +3,12 @@ import {
...
+  toRaw,
+  unref,
 } from 'vue'
 import { getValueByPath, escapeRegexpString } from '@element-plus/utils/util'
-import { selectKey, selectGroupKey, selectEvents } from './token'
+import { selectKey, selectGroupKey, QueryChangeCtx } from './token'
```
JeremyWuuuuu, line 11: Same as above

packages/components/select/src/useSelect.ts
```diff
@@ -253,8 +262,10 @@ export const useSelect = (props, states: States, ctx) => {
           }
           handleQueryChange(states.query)
           if (!props.multiple && !props.remote) {
-            states.selectEmitter.emit('elOptionQueryChange', '')
-            states.selectEmitter.emit('elOptionGroupQueryChange')
+            queryChange.value.query = ''
+
+            triggerRef(queryChange)
+            triggerRef(groupQueryChange)
```
JeremyWuuuuu, line 268: **Praise:** I love this idea! 👍

jw-foss (approved):

JeremyWuuuuu:
The code looks legit to me. I did leave a few comments, you can decide on your own if you want to change it. I think the change in `webpack.config.js` can be reverted since we do not use that for bundling. :bow:

This is really meaningful to us, thank you for your contribution. 👍

packages/components/form/src/form-item.vue
purepear: Yes, i agree, sth we can do in another PR :)

JeremyWuuuuu:
Let's make the roll!

PR #3383, opened 2021-09-13
feat(hooks): extracting size injection for form items

jw-foss:
- Extract common code for form items
- Apply extracted code for el-button


packages/components/button/src/button.vue
```diff
@@ -25,31 +25,23 @@
 
 <script lang="ts">
 import { computed, inject, defineComponent } from 'vue'
-import { elFormKey, elFormItemKey } from '@element-plus/tokens'
-import { useGlobalConfig } from '@element-plus/utils/util'
+import { useFormItem } from '@element-plus/hooks'
 import { elButtonGroupKey } from '@element-plus/tokens'
 
+import type { ComponentSize } from '@element-plus/utils/types'
```
sxzz, line 31: import order

packages/components/button/src/button.vue
```diff
@@ -25,31 +25,23 @@
...
-    const elBtnGroup = inject(elButtonGroupKey)
-
-    const buttonSize = computed(
-      () => props.size || elBtnGroup?.size || elFormItem?.size || $ELEMENT.size
-    )
-    const buttonDisabled = computed(() => props.disabled || elForm?.disabled)
+  setup(_, { emit }) {
+    const elBtnGroup = inject(elButtonGroupKey, undefined)
+    const { size: buttonSize, disabled: buttonDisabled } = useFormItem({
+      size: computed(() => elBtnGroup?.size as ComponentSize),
```
sxzz, line 43:
```suggestion
      size: computed(() => elBtnGroup?.size),
```
I think `as ComponentSize` can be removed.

packages/hooks/use-form-item/index.ts
```diff
@@ -0,0 +1,53 @@
...
+export const useFormItemProps = {
+  size: buildProp({
+    type: String,
+    values: sizes,
+    default: '',
+  } as const),
+  disabled: Boolean,
+} as const
+
+export type IUseFormItemProps = ExtractPropTypes<typeof useFormItemProps>
```
sxzz, line 19:
```suggestion
export type UseFormItemProps = ExtractPropTypes<typeof useFormItemProps>
```
I think `I` prefix is redundant.

packages/components/button/src/button.vue
JeremyWuuuuu: No.  since the size here is type `undefined | ComponentSize` if you remove the cast the assertion would fail because undefined is not type of `ComponentSize`

packages/hooks/use-form-item/index.ts
JeremyWuuuuu: This is more like a convention to prefix the type with `I` which stands for `interface` even though there it is a type.

packages/components/button/src/button.vue
sxzz: Emmm, you're right. But I think you should change `LocalFallbacks['size']` as `Ref<IUseFormItemProps['size'] | undefined>`.

packages/hooks/use-form-item/index.ts
sxzz: But it seems that the type naming convention of this project starts with a capital letter, and no  `I`.

packages/components/button/src/button.vue
JeremyWuuuuu: I think the question mark is already marking the member as optional so that we don't need to add `undefined` annotation which seems to be redundant.

sxzz: In fact, `elBtnGroup?.size` could be `undefined`.

packages/hooks/use-form-item/index.ts
JeremyWuuuuu: Alright, let's update it.

packages/components/button/src/button.vue
sxzz: The use of forced type conversion here is not in compliance with the specification.

packages/hooks/use-form-item/index.ts
sxzz: Don't be hurry. You only changed one line, and there are other references to `UseFormItemProps`.

packages/components/button/src/button.vue
JeremyWuuuuu: Yes, and that is acceptable since we have handled the situation in `use-form-item`. But it seems like the compiler is acting weird which I think the syntax here is correct by marking the member optional should provide the type `undefined` for it.

sxzz: Never mind. I will handle this tomorrow...

sxzz (approved):

PR #3401, opened 2021-09-14
feat(components): ElSelect remote search auto load data

### LINKED ISSUE #2974: [Feature Request] ElSelect remote search auto load data
### Existing Component
Yes

### Component Name
ElSelect

### Description
I use ElSelect remote search,  I can see nothing when i focus select， except  i press some key,  Hoping add a propery: remote-auto-load[bool]
that control get remote data when init  

我在使用ElSelect时， 只有当我输入了字符时才能看到选项， 默认进去为空，希望可能增加remote-auto-load: bool, 支持是否默认获取远程数据的功能

blackhu0804:


blackhu0804:
To solve [#2974](https://github.com/element-plus/element-plus/issues/2974)

sxzz (approved):

JeremyWuuuuu:
@blackhu0804 can we add this into documentations? Since this is a new API, so that we need to mention it in the docs. Thanks!

blackhu0804:
> @blackhu0804 can we add this into documentations? Since this is a new API, so that we need to mention it in the docs. Thanks!

Of course, I have added the documentation.

msidolphin:
I think this attribute is redundant, because the OPTIONS are provided by the user, don't you think?

blackhu0804:
> I think this attribute is redundant, because the OPTIONS are provided by the user, what do you think?

In this case, I currently don't know how to solve this problem.

msidolphin:
> > I think this attribute is redundant, because the OPTIONS are provided by the user, what do you think?
> 
> In this case, I currently don't know how to solve this problem.

https://github.com/element-plus/element-plus/issues/2974#issuecomment-927525654

PR #3402, opened 2021-09-14
fix(components): [el-dialog] unmounted the popover in dialog

Alanscut:

fix #2908

### DIFF
--- a/packages/components/dialog/src/use-dialog.ts
+++ b/packages/components/dialog/src/use-dialog.ts
@@ -149,6 +149,15 @@ export const useDialog = (
     }
   )
 
+  watch(
+    () => visible.value,
+    () => {
+      if (!visible.value) {
+        rendered.value = false
+      }
+    }
+  )
+
   onMounted(() => {
     if (props.modelValue) {
       visible.value = true


sxzz:
This will automatically enable `destroy-on-close`.

packages/components/dialog/src/use-dialog.ts
```diff
@@ -149,6 +149,15 @@ export const useDialog = (
...
   )
 
+  watch(
+    () => visible.value,
+    () => {
+      if (!visible.value) {
+        rendered.value = false
+      }
+    }
+  )
```
JeremyWuuuuu, line 159: 1. In the code level, we simply do not use the watched value in the callback. The suggested usage is:

```ts
watch(
    () => visible.value,
    (val) => {
      if (!val) {
        rendered.value = false
      }
    }
  )
```
2. In the logic level

By doing so, you are unmounting the component that's inside the dialog, which in some cases is not a really good idea, because some user wants to preserve the state of their component(Although it should be done by using value binding which I think is the correct way).

Let me know WDYT. 🙏

Alanscut: Agree! I re-thinked it last night, and thought `destroy-on-close` is more appropriate to control the component in `dialog`.

The reason I create this PR is because  @xiaoxian521  mentioned this issue in https://github.com/element-plus/element-plus/issues/2908#issuecomment-918940142  last night, and I found that problem is still in latest version. As I said in previous https://github.com/element-plus/element-plus/issues/2908#issuecomment-896642947, `destroy-on-close` could be a workaroud.

Alanscut: Hi @JeremyWuuuuu , maybe change the default value of `destroyOnClose` to `true` is a good solution. If users really hope to keep the component inside dialog, then just set `destroyOnClose` to `false`.  Should I update or close this PR?

JeremyWuuuuu: I think we can keep the old way.

JeremyWuuuuu:
Closing since we have `destroy-on-close`

PR #3404, opened 2021-09-14
fix(build): fix exclude dependency

### LINKED ISSUE #3391: 1.1.0-beta.10 version problem
theme-chalk/display.css 文件没了是有替代方案还是什么

sxzz:

fix https://github.com/element-plus/element-plus/issues/3391#issuecomment-918955010


build/components.ts
```diff
@@ -73,6 +53,23 @@ const pathsRewriter = (id) => {
...
+}
+
+function pathsRewriter(id) {
+  if (id.startsWith(`${EP_PREFIX}/components`))
+    return id.replace(`${EP_PREFIX}/components`, '..')
+  if (id.startsWith(EP_PREFIX) && excludes.every((e) => !id.endsWith(e)))
+    return id.replace(EP_PREFIX, '../..')
+  return id
+}
+
```
sxzz, line 72: Not changed, adjust order only.

build/build-locale.ts
```diff
@@ -3,12 +3,19 @@ import fs from 'fs'
...
 const fileList = fs.readdirSync(localePath)
 
-const transform = function (filename, name, cb) {
-  require('@babel/core').transformFile(
+const transform = function (
+  filename: string,
+  name: string,
+  cb: FileResultCallback
+) {
+  babel.transformFile(
```
JeremyWuuuuu, line 18: In fact this file is already deprecated right now since we applied `gulp` for building locale code. (But this change does seem legit.

build/components.ts
```diff
@@ -81,15 +69,15 @@ async function buildComponents() {
...
-        return (
-          id.startsWith(VUE_REGEX) ||
-          id.startsWith(VUE_MONO) ||
-          id.startsWith(EP_PREFIX) ||
-          externals.some((i) => id.startsWith(i))
-        )
+      const rollupConfig = {
+        input: entry,
+        plugins,
+        external: getExternals({ full: false }),
```
JeremyWuuuuu, line 75: Why not `getExternals(false)`? Seems like there won't be any extension on this.

build/build-locale.ts
sxzz: So could I remove `build-locale.ts` and `yarn build:locale-umd`?

build/components.ts
sxzz: Maybe we can customize the specified package later.
```
getExternals({ pkg: ['aa', 'bb', 'cc'] })
```

build/components.ts
```diff
@@ -81,15 +69,15 @@ async function buildComponents() {
...
       }
-      const esm = {
+      const bundle = await rollup.rollup(rollupConfig)
+
+      // ESM
+      await bundle.write({
```
JeremyWuuuuu, line 80: Actually I was wondering if we can put `Promise.all(writers)` here to enable a parallel writing(hypothetically, I am not very familiar with NodeJS stuff. cc @adaex is that possible?)

sxzz: I think it works.

sxzz: Maybe I will try to improve it in next PR. Also, we can execute multiple commands in parallel in build.sh

build/size-reporter.ts
```diff
@@ -1,68 +1,17 @@
 import chalk from 'chalk'
+import { FileSizeReporter } from 'rollup-plugin-filesize'
```
JeremyWuuuuu, line 2: Although for consistency I would request you to add `type` keyword in front of the importing member but since this is not public using file I will consider this minor.

build/utils.ts
```diff
@@ -0,0 +1,49 @@
+import path from 'path'
+import os from 'os'
+import { getPackagesSync } from '@lerna/project'
+
+import { compRoot } from './paths'
+import chalk from 'chalk'
```
JeremyWuuuuu, line 6: I think this like should be above line#3

build/size-reporter.ts
sxzz: Yeah, next PR~🤪

build/utils.ts
sxzz: Done.

scripts/build.sh
```diff
@@ -1,18 +1,19 @@
-#! /usr/bin/bash
+#!/bin/sh
 
 set -e
 
 yarn bootstrap
 yarn clean:lib
-yarn update:version
 yarn gen:version
```
JeremyWuuuuu, line 7: This line must followed by line#7 since this generates a file that contains the current releasing version information, so that I would say the you could break this into 3 pieces of code.
`pre-build` `build` `post-build` which co-responds to:

1. `prepare` stage 
2. build` stage
3. `publish` stage.

in stage `prepare` we need to collect informations from the environment variables to generate correct metadata.
in stage `build` we do what it means to do, as well as  `publish`

build/components.ts
adaex: > Actually I was wondering if we can put `Promise.all(writers)` here to enable a parallel writing

Yeah, it's perfectly fine to write to both `es` and `cjs` formats in parallel here.

JeremyWuuuuu: Let's move this into separate PRs for faster build speed.

jw-foss (approved):

PR #3406, opened 2021-09-14
fix(utils): add type remove the ts error

btea:

### DIFF
--- a/packages/utils/scrollbar-width.ts
+++ b/packages/utils/scrollbar-width.ts
@@ -22,7 +22,7 @@ export default function (): number {
   outer.appendChild(inner)
 
   const widthWithScroll = inner.offsetWidth
-  outer.parentNode.removeChild(outer)
+  outer.parentNode!.removeChild(outer)
   scrollBarWidth = widthNoScroll - widthWithScroll
 
   return scrollBarWidth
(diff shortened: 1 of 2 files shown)


packages/utils/scrollbar-width.ts
```diff
@@ -22,7 +22,7 @@ export default function (): number {
   outer.appendChild(inner)
 
   const widthWithScroll = inner.offsetWidth
-  outer.parentNode.removeChild(outer)
+  outer.parentNode!.removeChild(outer)
```
JeremyWuuuuu, line 25: I think the use of non-nullish operator should be very careful which could cause trouble. I suggest that we use `optional` chaining operator here instead of `non-nullish assertion` to prevent error.

JeremyWuuuuu:
Left 1 request for you. Let me know if you need any help. ;)

packages/utils/scrollbar-width.ts
btea: Yes, you are right.

btea:
The test seems unstable~

jw-foss (approved):

JeremyWuuuuu:
> The test seems unstable~

We are going to fix that for stability.

PR #3416, opened 2021-09-14
build: parallel build

sxzz:


build/full-bundle.ts
```diff
@@ -33,65 +34,56 @@ import { getExternals } from './utils'
...
   })
 
-  const rewriter = (id) => {
+  const rewriter = (id: string) => {
     if (id.startsWith(`${EP_PREFIX}/components`))
       return id.replace(`${EP_PREFIX}/components`, './components')
-    if (id.startsWith(EP_PREFIX) && excludes.every((e) => !id.endsWith(e)))
+    else if (id.startsWith(EP_PREFIX) && excludes.every((e) => !id.endsWith(e)))
       return id.replace(EP_PREFIX, '.')
+    else return id
```
JeremyWuuuuu, line 83: With `return` here there is actually no need to use `else`. Even though it is semantically correct.

JeremyWuuuuu:
TBH, when there is no test cases against the bundled product, I think we better not get our hands on the bundling script, since the previous version is tested by releasing multiple versions. If you do need to add some trivial change like abstracting the `chalk` methods, I am fine with it, but if you want to change those script for bundling, I suggest you not to. I am not saying that we are afraid of making mistakes, I am saying that at the stage of a project used by thousands we should be careful handling stuff like this, if we can guarantee the quality of the product I would be fine with it(since we can't).

build/full-bundle.ts
sxzz: If don’t return, typescript will add `undefined` to the return value.

build/build-helper.ts
```diff
@@ -1,8 +1,8 @@
-/* eslint-disable @typescript-eslint/no-var-requires */
 import helper from 'components-helper'
 import path from 'path'
 import { epRoot } from './paths'
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
 const { name, version } = require(path.resolve(epRoot, './package.json'))
 
 import icon from '../website/icon.json'
```
JeremyWuuuuu, line 8: I think you need to rebase on the `dev` for getting the newest code.

sxzz:
@JeremyWuuuuu PTAL.

build/helper.ts
```diff
@@ -1,49 +1,24 @@
...
+const reComponentName: InstallOptions['reComponentName'] = (title: string) =>
+  `el-${title
     .replace(/\B([A-Z])/g, '-$1')
     .replace(/[ ]+/g, '-')
     .toLowerCase()}`
-}
 
-function reDocUrl(fileName, header) {
+const reDocUrl: InstallOptions['reDocUrl'] = (fileName, header) => {
   const docs = 'https://element-plus.org/#/en-US/component/'
```
JeremyWuuuuu, line 14: Actually here we need to update to
```ts
const docs = 'https://element-plus.org/en-US/component/'
```

But we can do that in separate PRs.

build/utils.ts
```diff
@@ -1,54 +1,16 @@
...
+//   const workspacePkgs = (await getWorkspaceNames()).filter((pkg) =>
+//     pkg.startsWith(EP_PREFIX)
+//   )
+//   return (module: Module) => (id: string) => {
+//     const config = buildConfig[module]
+//     if (workspacePkgs.some((pkg) => id.startsWith(pkg)))
+//       return id.replace(EP_PREFIX, config.bundle.path)
+//     else return ''
+//   }
+// }
```
JeremyWuuuuu, line 16: Why are we keeping this file anyway?

jw-foss (approved):
Approving as-is

JeremyWuuuuu:
@sxzz got some conflicts, you can merge this PR after resolving them.

PR #3531, opened 2021-09-21
fix(components): [el-popper] visible default value to false

btea:

### DIFF
--- a/packages/components/popper/src/use-popper/defaults.ts
+++ b/packages/components/popper/src/use-popper/defaults.ts
@@ -143,7 +143,7 @@ export default {
   },
   visible: {
     type: Boolean,
-    default: undefined,
+    default: false,
   },
   stopPopperMouseEvent: {
     type: Boolean,
--- a/packages/components/popper/src/use-popper/index.ts
+++ b/packages/components/popper/src/use-popper/index.ts
@@ -74,17 +74,23 @@ export default function (
     get() {
       if (props.disabled) {
         return false
-      } else {
-        return isBool(props.visible) ? props.visible : state.visible
       }
+      return state.visible
     },
     set(val) {
       if (isManualMode()) return
-      isBool(props.visible)
-        ? emit(UPDATE_VISIBLE_EVENT, val)
-        : (state.visible = val)
+      if (isBool(props.visible)) {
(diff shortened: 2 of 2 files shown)


sxzz (approved):

sxzz:
It seems that unit test failed. Could you please fix it?

btea:
@sxzz ok, I will fix it.

btea:
@sxzz I tried to fix it, and I found the reason of cause the unit test failure seems here, the emit seems invalid.
https://github.com/element-plus/element-plus/blob/d71c87344a36f1837027683fd847a85aa5e4ba17/packages/components/popper/src/use-popper/index.ts#L83-L85  
As so far, I can't resolve it, Could you give me some help? Thank you!

btea:
@sxzz I changed the visibility's `get` and `set` and I'm not sure if it will cause other problems.  Please check it! Thanks!

sxzz:
OK. I will this later.

JeremyWuuuuu:
Well, actually it was set to default `undefined` for reasons, for users that didn't want to use `v-model` binding we use that as an indicator for determining whether this is the case. May I ask based on what purpose we need to change this?

btea:
> Well, actually it was set to default `undefined` for reasons, for users that didn't want to use `v-model` binding we use that as an indicator for determining whether this is the case. May I ask based on what purpose we need to change this?

I want to change it because here has a ts error
image

JeremyWuuuuu:
> > Well, actually it was set to default `undefined` for reasons, for users that didn't want to use `v-model` binding we use that as an indicator for determining whether this is the case. May I ask based on what purpose we need to change this?
> 
> I want to change it because here has a ts error, and I don't think the default value of boolean type should be undefined.
> image

But we need to figure out a way to indicate the value was not passed into `popper`, actually I think I had found a way before you can checkout `hooks/use-popper` for more detail, my question is that, changing the code this way **might** break some functionalities, we need to be careful on this one though(not saying that we wouldn't want your PR, I think it is legit).

btea:
@JeremyWuuuuu Yes, I agree with you. At first, I thought it was simple, just changing a default value. But,  When I commit  the change, the unit test failed. To past the test, I changed the visibility and I'm not sure the change will affect other functions.~~

PR #3536, opened 2021-09-21
refactor(components): el-button icon

adaex:


imswk:
image
@adaex document example needs to be modified

JeremyWuuuuu:
@adaex @imswk For now this segment 
```ts
import { Search, Edit, Check, Message, Star, Delete } from '@element-plus/icons'
```
might not work for demo since the icon was not added into the demo as dependencies. So that the online demo will probably be unavailable for this case.

docs/en-US/component/button.md
```diff
@@ -72,18 +72,18 @@ button/size
```
JeremyWuuuuu, line 72: **Suggestion:**

Since I cannot add comment for unchanged lines so I am adding it here, after the refactor our user is able to pass `string` and `Component` to `icon` attribute to use icon, we need to specify that. So that in this file at line37 we should update the description from 

> Use the `icon` attribute to add icon. You can find the icon list in Element Plus icon component. Adding icons to the right side of the text is achievable with an `<i>` tag. Custom icons can be used as well.

to 

Use the `icon` attribute to add icon. You can pass either string for the component name (registered in advance) or the component itself which is a SVG Vue component. Element Plus has provided a set of icon that you can find at [icon](/en-US/component/icon)

NB: **Don't forget adding these descriptions for our user to make a smooth transition for each changed component**, we can always use a `tip` or `warning` to draw the attention help them to notice that we have made changes on the icons.

PR #3677, opened 2021-09-27
feat(build): support esm and minified full bundle

sxzz:

### DIFF
--- a/build/full-bundle.ts
+++ b/build/full-bundle.ts
@@ -6,7 +6,7 @@ import commonjs from '@rollup/plugin-commonjs'
 import vue from 'rollup-plugin-vue'
 import esbuild from 'rollup-plugin-esbuild'
 import replace from '@rollup/plugin-replace'
-import { parallel, series } from 'gulp'
+import { parallel } from 'gulp'
 import { genEntryTypes } from './entry-types'
 import { RollupResolveEntryPlugin } from './rollup-plugin-entry'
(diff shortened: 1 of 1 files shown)


build/full-bundle.ts
```diff
@@ -6,64 +6,78 @@ import commonjs from '@rollup/plugin-commonjs'
...
+      },
+      sourcemap: minify,
     },
-  })
-  green('index.full.js generated')
+    {
+      format: 'esm',
+      file: path.resolve(
+        epOutput,
+        `dist/index.full${minify ? '.min' : ''}.cjs`
```
JeremyWuuuuu, line 73: I think we can remove `full` here.
Also would it be weird since you have `esm` format, yet you use the extension name `.cjs`?

sxzz: This is almost the norm. Such as `vueuse`, tsup.

JeremyWuuuuu: I don't think this is legit naming for bundle with `esm` format with `cjs` extension.

I'm going to ask you to export co-responding file with correct name, if you want to output `cjs` you should write the file with `cjs` format, also applies to `esm`.

You can always output both `cjs` and `esm`, I am ok with this way.

sxzz: 😅 看错了

JeremyWuuuuu: We could aways follow this way:
image

jw-foss (approved):

PR #3736, opened 2021-09-29
chore: fix the source value error in web-types

tolking:

fix the source value error in web-types, ensure that in theory you can jump to the type definition files by the components name in the webstorm. 

```diff
- "source": {"symbol": "Affix"},
+ "source": {"symbol": "ElAffix"},
```

### DIFF
--- a/build/helper.ts
+++ b/build/helper.ts
@@ -20,9 +20,11 @@ const reDocUrl: InstallOptions['reDocUrl'] = (fileName, header) => {
 }
 
 const reWebTypesSource: InstallOptions['reWebTypesSource'] = (title) => {
-  const symbol = `El${title.replace(/-/g, ' ').replace(/^\w|\s+\w/g, (item) => {
-    return item.trim().toUpperCase()
-  })}`
+  const symbol = `El${title
+    .replaceAll(/-/g, ' ')
+    .replaceAll(/^\w|\s+\w/g, (item) => {
+      return item.trim().toUpperCase()
+    })}`
 
   return { symbol }
 }
--- a/build/helper.ts
+++ b/build/helper.ts
@@ -20,7 +20,7 @@ const reDocUrl: InstallOptions['reDocUrl'] = (fileName, header) => {
 }
 
 const reWebTypesSource: InstallOptions['reWebTypesSource'] = (title) => {
-  const symbol = `El${title.replace(/-/, ' ').replace(/^\w|\s+\w/g, (item) => {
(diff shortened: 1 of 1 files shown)


build/helper.ts
```diff
@@ -19,6 +19,14 @@ const reDocUrl: InstallOptions['reDocUrl'] = (fileName, header) => {
   return docs + fileName + (_header ? `#${_header}` : '')
 }
 
+const reWebTypesSource: InstallOptions['reWebTypesSource'] = (title) => {
+  const symbol = `EL${title.replace(/-/, ' ').replace(/^\w|\s+\w/g, (item) => {
```
JeremyWuuuuu, line 23: Should the symbol be started with `El`? I think `EL` seems wrong to me.

tolking: Yes this is a typo

build/helper.ts
```diff
@@ -19,6 +19,14 @@ const reDocUrl: InstallOptions['reDocUrl'] = (fileName, header) => {
   return docs + fileName + (_header ? `#${_header}` : '')
 }
 
+const reWebTypesSource: InstallOptions['reWebTypesSource'] = (title) => {
+  const symbol = `El${title.replace(/-/, ' ').replace(/^\w|\s+\w/g, (item) => {
```
sxzz, line 23:
```suggestion
  const symbol = `El${title.replace('-', ' ').replaceAll(/^\w|\s+\w/, (item) => {
```

tolking: When using regular, the replace is the same as the replaceAll. For easy reading?

sxzz: Yeah. I think `replaceAll` is better.

build/helper.ts
```diff
@@ -19,6 +19,14 @@ const reDocUrl: InstallOptions['reDocUrl'] = (fileName, header) => {
   return docs + fileName + (_header ? `#${_header}` : '')
 }
 
+const reWebTypesSource: InstallOptions['reWebTypesSource'] = (title) => {
+  const symbol = `El${title.replace(/-/g, ' ').replace(/^\w|\s+\w/g, (item) => {
```
sxzz, line 23:
```suggestion
  const symbol = `El${title.replaceAll('-', ' ').replaceAll(/^\w|\s+\w/, (item) => {
```

tolking:
```js
.replaceAll(/^\w|\s+\w/,
```

Is the wrong way to use, it should be

```js
.replaceAll(/^\w|\s+\w/g,
```

It's the same as replace

sxzz: My mistake. You're right.

sxzz (approved):

PR #3865, opened 2021-10-13
refactor: [basic-date-table] add rowType refactor getCellClasses

btea:

### DIFF
--- a/packages/components/date-picker/src/date-picker-com/basic-date-table.vue
+++ b/packages/components/date-picker/src/date-picker-com/basic-date-table.vue
@@ -41,6 +41,7 @@ import { defineComponent, computed, ref } from 'vue'
 import dayjs from 'dayjs'
 import { useLocaleInject } from '@element-plus/hooks'
 import { coerceTruthyValueToArray } from '@element-plus/utils/util'
+import type { rowType } from '../date-picker.type'
 import type { PropType } from 'vue'
 
 import type { Dayjs } from 'dayjs'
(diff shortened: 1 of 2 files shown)


packages/components/date-picker/src/date-picker-com/basic-date-table.vue
```diff
@@ -389,7 +373,7 @@ export default defineComponent({
 
     const isWeekActive = (cell) => {
       if (props.selectionMode !== 'week') return false
-      let newDate = props.date.startOf('day')
+      let newDate = props.date?.startOf('day')!
```
JeremyWuuuuu, line 376: I think this is somehow conflict since `date` is marked as  `optional` then you get the `!` mark for `startOf` method, question is why don't we do
```ts
      let newDate = props.date!.startOf('day')
```
instead?

packages/components/date-picker/src/date-picker-com/basic-date-table.vue
```diff
@@ -245,56 +247,38 @@ export default defineComponent({
...
-      }
-
-      if (cell.customClass) {
-        classes.push(cell.customClass)
+        start && classes.push('start-date')
+        end && classes.push('end-date')
       }
+      disabled && classes.push('disabled')
+      selected && classes.push('selected')
+      customClass && classes.push(customClass)
```
JeremyWuuuuu, line 281: I'd prefer something like this:

```
const classes = {
  disabled,
  selected,
  customClass,
}

return Object.entries(classes).filter(([kls, val]) => val).map(([kls]) => kls).join(' ')
```

packages/components/date-picker/src/date-picker-com/basic-date-table.vue
btea: OK, I'll replace it.

btea: OK, I'll replace it.

packages/components/date-picker/src/date-picker-com/basic-date-table.vue
btea: This is well written, and other classes can also be handled in this way.

btea:
It takes a long time and there are many code conflicts. I'll close the PR and create a new one.

PR #3979, opened 2021-10-20
fix(components): ensure table size affects inner selection

### LINKED ISSUE #3629: [bug report] the size attribute of El table does not take effect under specific circumstances
### Element Plus version
1.1.0-beta.16

### OS/Browsers version
macOS Big Sur-11.4    /  Google Chrome-94.0.4606.61

### Vue version
3.2.18

### Reproduction Link
https://codesandbox.io/embed/element-plus-spa-forked-thzs5?fontsize=14&hidenavigation=1&theme=dark

### Steps to reproduce
将 el-table 的size属性设置为 mini,正常情况下是生效的

将 el-table 的size属性设置为 mini,并将某一列的类型设置为 selection 复选框,el-table的 mini 就失效

### What is Expected?
将 el-table 的size属性设置为 mini,不论列的类型设置什,el-table的的size属性不会失效

将 el-table 的size属性设置为 mini,并将某一列的类型设置为 selection 复选框,el-table的大小，就会变成small样式

cokemine:
fix #3629

### DIFF
--- a/packages/components/table/src/store/watcher.ts
+++ b/packages/components/table/src/store/watcher.ts
@@ -46,7 +46,7 @@ const doFlattenColumns = (columns) => {
 
 function useWatcher<T>() {
   const instance = getCurrentInstance() as Table<T>
-  const { size: tableSize } = toRefs(instance.props)
+  const { size: tableSize } = toRefs(instance.proxy?.$props as any)
   const rowKey: Ref<string> = ref(null)
   const data: Ref<T[]> = ref([])
   const _data: Ref<T[]> = ref([])
--- a/packages/components/table/src/store/watcher.ts
+++ b/packages/components/table/src/store/watcher.ts
@@ -1,4 +1,4 @@
-import { ref, getCurrentInstance, unref, watch } from 'vue'
+import { ref, getCurrentInstance, unref, watch, toRefs } from 'vue'
 import { hasOwn } from '@vue/shared'
 import {
   getKeysMap,
@@ -46,6 +46,7 @@ const doFlattenColumns = (columns) => {
 
 function useWatcher<T>() {
   const instance = getCurrentInstance() as Table<T>
+  const { size: tableSize } = toRefs(instance.props)
   const rowKey: Ref<string> = ref(null)
   const data: Ref<T[]> = ref([])
   const _data: Ref<T[]> = ref([])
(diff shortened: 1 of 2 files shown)


packages/components/table/src/store/watcher.ts
```diff
@@ -46,6 +46,7 @@ const doFlattenColumns = (columns) => {
 
 function useWatcher<T>() {
   const instance = getCurrentInstance() as Table<T>
+  const { size: tableSize } = toRefs(instance.props)
```
JeremyWuuuuu, line 49: I think we should use `instance.$props` instead of `instance.$proxy.$props`(this one is actually reactive)

JeremyWuuuuu: cc @cokemine (

cokemine: It seems that the instance object obtained by getCurrentInstance Hook does not have $props or $proxy properties, we can only get the props through instance.props or instance.ctx(proxy).$props (still reactive)

JeremyWuuuuu: It was marked as optional because you can define a component without props so that would be accurate. For us we know that there must be this $props object you can use `!.` to bypass the type checking.

cokemine: But actually it is always `undefined` when I try to output it

JeremyWuuuuu: that's weird cc @sxzz can you have a look on this? Until we have found the root cause let's stick with your way.

sxzz: It's `instance.proxy.$props`. /cc @JeremyWuuuuu @cokemine 
image

sxzz: @cokemine Emmm, we just use instance.proxy.$props
```suggestion
  const { size: tableSize } = toRefs(instance.proxy.$props)
```

cokemine: What's the different between using `instance.props` and `instance.proxy.$props`.
I only know that `instance.ctx` can be only used in dev mode.
BTW, it seems like `instance.proxy.$props` will may lose types

jw-foss (approved):

packages/components/table/src/store/watcher.ts
JeremyWuuuuu: Doesn't matter, since `$props` is defined by Vue so we can simply cast it.

JeremyWuuuuu:
Thank you so much for making Element Plus better! :pray:

PR #4069, opened 2021-10-28
fix(components): [el-select] Automatically trigger verification

Alanscut:

fix #4052


Alanscut:
When select is multiple selection, the initial value should be bound to an array. It's just that when the type of the initial value is a string, it will be converted to an array. Everything can run normally, except that the calibration will be triggered automatically. Maybe it is better for us to remove the automatic trigger verification?

PR #4171, opened 2021-11-02
fix(build): windows executes build correctly

Yueyanc:

### DIFF
--- a/build/utils/process.ts
+++ b/build/utils/process.ts
@@ -4,7 +4,37 @@ import { projRoot } from './paths'
 
 export const run = async (command: string, cwd: string = projRoot) =>
   new Promise<void>((resolve, reject) => {
-    const [cmd, ...args] = command.split(' ')
+    let [cmd, ...args] = command.split(' ')
+    args.forEach((item, index) => {
+      console.log(item)
+      args[index] = item.replace(/\//g, '\\')
+    })
+    if (process.platform === 'win32') {
+      switch (cmd) {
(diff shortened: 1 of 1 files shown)


sxzz (commented):
I think we should refactor this part. And using `fs-extra` to copy and remove file

ChenAdien:
> 我认为我们应该重构这部分。并`fs-extra`用于复制和删除文件
Ok, I'll try to reconstruct it myself

ChenAdien:
> I think we should refactor this part. And using `fs-extra` to copy and remove file

Ok, I'll try to reconstruct it myself

JeremyWuuuuu:
So umm, I myself is not very familiar with Windows Platform APIs for `rsync` etc, but for me personally I think for Windows users, WSL2 would be a better solution without all those API mappings.

Just in case we should add some documentation here for recording the mapped APIs such as `rsync`.

build/utils/process.ts
```diff
@@ -4,7 +4,37 @@ import { projRoot } from './paths'
 
 export const run = async (command: string, cwd: string = projRoot) =>
   new Promise<void>((resolve, reject) => {
-    const [cmd, ...args] = command.split(' ')
+    let [cmd, ...args] = command.split(' ')
+    args.forEach((item, index) => {
+      console.log(item)
```
JeremyWuuuuu, line 9: Also please don't forget to remove the console part.

PR #4223, opened 2021-11-05
chore(components): change the wait in the el-carousel tests to sleep

emojiiii:

### DIFF
--- a/packages/components/carousel/__tests__/carousel.spec.ts
+++ b/packages/components/carousel/__tests__/carousel.spec.ts
@@ -1,10 +1,8 @@
 import { nextTick } from 'vue'
 import { mount } from '@vue/test-utils'
+import { sleep } from '@element-plus/test-utils'
 import Carousel from '../src/main.vue'
 import CarouselItem from '../src/item.vue'
-
-const wait = (ms = 100) =>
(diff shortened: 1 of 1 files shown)


undefined-degisn:
看单测的错误，似乎与我此次提交无关，本地跑单测也通过了，很奇怪

msidolphin:
Consider using `jest.runAllTimers()` instead of sleep

undefined-degisn:
`carousel`组件因为存在一个`interval`的参数，并不是渲染的问题，所以使用`jest.runAllTimers()`似乎不太合适

JeremyWuuuuu:
> `carousel`组件因为存在一个`interval`的参数，并不是渲染的问题，所以使用`jest.runAllTimers()`似乎不太合适

The `sleep` API is not recommended to use because we cannot know the what was going on inside the component itself, normally I would suggest using `jest.fakeTimer` API for testing components like this, with `fakeTimer` rich APIs we developers can test components exactly the way it should be functioning. 

The interval is considered as constant when doing unit testing, via `fakeTimer` it enables us to test the functions without waiting for the timer to resolve, because we need to be assuming that the timer API provided by the browser (which our apps run in) is 100% correct. Timer is just nothing but a constant. 

Since the component should behave as we designed it to, there will be no trouble testing.

PR #3113, opened 2021-08-27
[Refactor] Refactor Plan

sxzz:
# Refactor

## Summary

- Improve all project typings.
- Drop deprecated APIs.

## Contribute Guide

- Create an PR. e.g `refactor(components): refactor button`
  - Please at`@` me.
  - Please assign to me if you're Element Plus member.
  - Please add PR link to below `Progress` if you have permission.
  - I will review it ASAP.
- 🛠 ❌ **fix ALL TypeScript errors.**
  - ⚠️ Priority: highest
  - Please refer to https://github.com/element-plus/element-plus/pull/3113#issuecomment-967230194
- 🧲 Extract `.vue` props and emits to `.ts`.
  - Philosophy: Write the logical part into ts file as much as possible.
  - add `InstanceType` for each components #4308
- 🎨 Use setup return render method.
  - Use `setup() { return () => { /* code here */ } }`
- 📁 Structure
  - Move `tokens` to `@element-plus/tokens`.
    - Only context type interface
    - Only injection key
- 🔑 Use `withInstall`.
  - Please refer to completed files if you're confused.
- 🪛 Refactor **ALL** APIs in detail.
- 🧱 Do not use SFC(`.vue`), if there is no template.

For more details, please refer to `Task List` and merged pull request below. Thanks for your contribution!

### ⭐️ Feature

- [x] `el-check-tag` support `v-model:checked`.
  - https://github.com/element-plus/element-plus/pull/3311
- [ ] `el-menu` add `v-model:opened`
  - https://github.com/element-plus/element-plus/issues/3122

### ♻️ Typing

- [x] 🦾 use typescript `strict` mode, increase code reliability.
- `RefElement` -> use the exact element type.
  - e.g `HTMLDivElement`
- `null` -> prefer `undefined`.
  - e.g `ref<HTMLDivElement>()`. **NOT** `ref(null)`
- Try not to use type assertions.
  - e.g ❌ BAD `inject(foo, {} as FooType)`
- [ ] improve and export props and emits type.
- [ ] add `InstanceType` for each components.
  - https://github.com/element-plus/element-plus/issues/4308
- [ ] add `vue-tsc` checker for unit test.

### 🔨 Utils / Hooks

- [ ] `VueUse`
  - https://github.com/element-plus/element-plus/pull/3370
  - https://github.com/element-plus/element-plus/pull/3374
  - [ ] `useEventListener` (remove `on`, `off`)
  - [ ] `onClickOutside` (remove directive)
[...]

PR #4301, opened 2021-11-08
refactor(components): refactor checkbox/checkbox-group/checkbox-button

emojiiii:

#3113


packages/components/checkbox/src/useCheckbox.ts
```diff
@@ -1,48 +1,18 @@
 import { ref, computed, inject, getCurrentInstance, watch } from 'vue'
```
sxzz, line 1: We should rename it to `kebab-case`.
Actually I think `hooks.ts` is better.

packages/components/checkbox/src/checkbox.ts
```diff
@@ -0,0 +1,68 @@
...
+    type: String,
+    default: undefined,
+  },
+  controls: {
+    type: String,
+    default: undefined,
+  },
+  border: Boolean,
+  size: {
+    type: definePropType<ComponentSize>(String),
```
sxzz, line 41: Same here

undefined-degisn: get！

undefined-degisn:
i dont' know why the test often make mistakes

sxzz:
> i dont' know why the test often make mistakes @sxzz 

please just ignore it.

packages/components/checkbox/src/useCheckbox.ts
undefined-degisn: I think hooks are divided according to their functions，so that may be `use-checkbox.hook.ts` will better?

sxzz: Or just rename it to `kebab-case`. How about `use-checkbox.ts`?

undefined-degisn: i think it's a good idea, because the beginning of  `use` already means `hook`

sxzz:
@undefined-degisn Hi, I'm just checking in to see if you are interested in becoming a maintainer of Element Plus since Element Plus is not perfect right now, and we could use any help we can get to make it that way.
If you are interested, please let me know and send your WeChat account to my email (sxzz@sxzz.moe), so that I can invite you into the developing group.

PR #4435, opened 2021-11-17
feat(components): [el-divider] support setting the dividing line style 

### LINKED ISSUE #4362: [Feature Request] [el-divider] can add either a solid line or dashed line property
### Existing Component
Yes

### Component Name
el-divider

### Description
Set solid or dashed lines property

Alanscut:
fix  #4362 

### DIFF
--- a/docs/en-US/component/divider.md
+++ b/docs/en-US/component/divider.md
@@ -47,8 +47,8 @@ divider/vertical-divider
 
 ## Divider Attributes
 
-| Attribute        | Description                               | Type   | Accepted Values                                                                         | Default    |
-| ---------------- | ----------------------------------------- | ------ | --------------------------------------------------------------------------------------- | ---------- |
-| direction        | Set divider's direction                   | string | horizontal / vertical                                                                   | horizontal |
-| border-style     | Set the style of divider                  | string | 请查看[CSS/border-style](https://developer.mozilla.org/zh-CN/docs/Web/CSS/border-style) | solid      |
-| content-position | customize the content on the divider line | String | left / right / center                                                                   | center     |
+| Attribute        | Description                               | Type   | Accepted Values                                                                   | Default    |
+| ---------------- | ----------------------------------------- | ------ | --------------------------------------------------------------------------------- | ---------- |
+| direction        | Set divider's direction                   | string | horizontal / vertical                                                             | horizontal |
+| border-style     | Set the style of divider                  | string | [CSS/border-style](https://developer.mozilla.org/zh-CN/docs/Web/CSS/border-style) | solid      |
+| content-position | customize the content on the divider line | String | left / right / center                                                             | center     |
--- a/docs/en-US/component/divider.md
+++ b/docs/en-US/component/divider.md
@@ -27,6 +27,16 @@ divider/custom-content
 
 :::
 
+## dashed line
+
+You can set the style of divider.
+
+:::demo
+
+divider/line-dashed
+
+:::
+
 ## Vertical divider
 
 :::demo
@@ -37,7 +47,8 @@ divider/vertical-divider
 
 ## Divider Attributes
 
-| Attribute        | Description                               | Type   | Accepted Values       | Default    |
-| ---------------- | ----------------------------------------- | ------ | --------------------- | ---------- |
-| direction        | Set divider's direction                   | string | horizontal / vertical | horizontal |
-| content-position | customize the content on the divider line | String | left / right / center | center     |
+| Attribute        | Description                               | Type   | Accepted Values                                                                         | Default    |
+| ---------------- | ----------------------------------------- | ------ | --------------------------------------------------------------------------------------- | ---------- |
+| direction        | Set divider's direction                   | string | horizontal / vertical                                                                   | horizontal |
+| border-style     | Set the style of divider                  | string | 请查看[CSS/border-style](https://developer.mozilla.org/zh-CN/docs/Web/CSS/border-style) | solid      |
+| content-position | customize the content on the divider line | String | left / right / center                                                                   | center     |
--- a/docs/examples/divider/vertical-divider.vue
+++ b/docs/examples/divider/vertical-divider.vue
@@ -3,7 +3,7 @@
     <span>Rain</span>
     <el-divider direction="vertical"></el-divider>
     <span>Home</span>
-    <el-divider direction="vertical"></el-divider>
+    <el-divider direction="vertical" border-style="dashed"></el-divider>
     <span>Grass</span>
   </div>
 </template>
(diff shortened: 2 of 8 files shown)


emojiiii:
perhaps it is more appropriate for us to use `--border-style` and leave it to users to choose?

Alanscut:
> perhaps it is more appropriate for us to use `--border-style` and leave it to users to choose?

Why `--border-style`  is more appropriate than `border-left(top)` ?

emojiiii:
> Why `--border-style` is more appropriate than `border-left(top)` ?

if user want to support `dotted` or `double` or other attributes, you must continue to change them
but you use `border-style: (--divider-border-style)` and use style to set that can choose by user

Alanscut:
I prefer to add `dotted `and `double `to accepted values. This is more convenient than if the user needs to distinguish horizontal or vertical dividing lines and then set it like `border-style: dashed none none;`.

msidolphin:
Change the line to type or style is more appropriate, just a suggestion, you can ignore

Alanscut:
> Change the line to type or style is more appropriate, just a suggestion, you can ignore

What do you think of changing the line to line-style

YunYouJun:
How about `border-style`, just like tailwindcss, etc. [border-style](https://windicss.org/utilities/borders.html#border-style).

packages/theme-chalk/src/divider.scss
```diff
@@ -19,6 +34,22 @@
     margin: 0 8px;
     vertical-align: middle;
     position: relative;
+
+    @include when(solid) {
```
YunYouJun, line 38: Try [@each](https://sass-lang.com/documentation/at-rules/control/each).

packages/theme-chalk/src/divider.scss
```diff
@@ -19,6 +34,22 @@
...
+    @include when(dashed) {
+      border-left: 1px var(--el-border-color-base) dashed;
+    }
+
+    @include when(dotted) {
+      border-left: 1px var(--el-border-color-base) dotted;
+    }
+
+    @include when(double) {
+      border-left: 4px var(--el-border-color-base) double;
```
YunYouJun, line 51: By the way, i think we can use `border-left: 4px var(--el-border-color-base) var(--el-border-style)` for it.

Then computed style `--el-border-style: dotter` for it in vue sfc.

So user can use property or css var. It also saves extra styles.

emojiiii: yep, i think so, too

Alanscut: @YunYouJun @emojiiii ok,

packages/components/divider/src/divider.ts
```diff
@@ -13,5 +13,9 @@ export const dividerProps = buildProps({
     values: ['left', 'center', 'right'],
     default: 'center',
   },
+  lineStyle: {
```
YunYouJun, line 16: add values like `contentPosition` ['left', 'center', 'right']?

docs/examples/divider/vertical-divider.vue
```diff
@@ -3,7 +3,7 @@
     <span>Rain</span>
     <el-divider direction="vertical"></el-divider>
     <span>Home</span>
-    <el-divider direction="vertical"></el-divider>
+    <el-divider direction="vertical" line-style="dashed"></el-divider>
```
YunYouJun, line 6: `border-style` vs `lint-style`? @JeremyWuuuuu WDYT

packages/components/divider/src/divider.ts
Alanscut: now, users can set lots of style like border-style, 
image
so i delete the values.
do you think we should add them all ,or just add ['solid','doshed','dotted','double'](this will limit user)?

docs/examples/divider/vertical-divider.vue
Alanscut: in my opinion, border have four lines, and divider just have one line.

JeremyWuuuuu: I'd prefer `border-style`

packages/components/divider/src/divider.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <div :class="['el-divider', `el-divider--${direction}`]">
+  <div
+    :class="['el-divider', `el-divider--${direction}`]"
+    :style="{ '--el-border-style': $props.lineStyle }"
```
JeremyWuuuuu, line 4: I think the `$props` prefix is unnecessary since if you don't have the same API returned by setup function, you should always be good to use `lineStyle` in the template directly.

jw-foss (commented):
Let's change the API to `border-style`, also left a nitpick suggestion for you to consider.


Please let me know if you have any questions.

packages/components/divider/src/divider.ts
YunYouJun: emm, if we can use type to define props, we can use `CSSStyleDeclaration['borderStyle']` directly.

@sxzz Is there any better way?

sxzz: Yeah, we can use `definePropType`.
Here is pseudo code
```
type: definePropType<BorderStyle>(String)
```

Alanscut:
@YunYouJun @JeremyWuuuuu 
used `CSSStyleDeclaration['borderStyle']`
and changed `line-style` to `border-style`

docs/en-US/component/divider.md
```diff
@@ -37,7 +47,8 @@ divider/vertical-divider
...
 ## Divider Attributes
 
-| Attribute        | Description                               | Type   | Accepted Values       | Default    |
-| ---------------- | ----------------------------------------- | ------ | --------------------- | ---------- |
-| direction        | Set divider's direction                   | string | horizontal / vertical | horizontal |
-| content-position | customize the content on the divider line | String | left / right / center | center     |
+| Attribute        | Description                               | Type   | Accepted Values                                                    | Default    |
+| ---------------- | ----------------------------------------- | ------ | ------------------------------------------------------------------ | ---------- |
+| direction        | Set divider's direction                   | string | horizontal / vertical                                              | horizontal |
+| border-style     | Set the style of divider                  | string | dotted / dashed / solid / double / groove / ridge / inset / outset | solid      |
```
YunYouJun, line 53: It's ok. Just a suggestion. Maybe we can leave a link directly instead of writing so many values.

https://developer.mozilla.org/zh-CN/docs/Web/CSS/border-style

YunYouJun (approved):

docs/en-US/component/divider.md
Alanscut: ok，

YunYouJun: Allow me to nitpick, because it's an English document by default. I think we don't need to write "请查看".

Alanscut: ok,

Issue #4432, opened 2021-11-17
[Bug Report] Cascader's empty slot is invalid


15998194741:
### Element Plus version
1.2.0-beta.3

### OS/Browsers version
win10

### Vue version
3.2.22

### Reproduction Link
https://codepen.io/15998194741/pen/XWaogwj

### Steps to reproduce
When using the Cascader component, I want to customize the interactive project, so I thought of the empty slot,But it doesn't work.

### What is Expected?
When there is no options, the floating display HTML

### What is actually happening?
Nothing happened

emojiiii:
![impicture_20211117_160213](https://user-images.githubusercontent.com/30518686/142159674-657e6e60-f0e7-4e17-8be8-7e9ec0029484.png)

`empty slot` is used to filter the display when there is no data

Dsaquel:
Fixed on newer versions.

PR #4439, opened 2021-11-17
fix: cascader-empty-slot-is-invalid #4432

xiaojieajie:
I'm here to fix it Cascader's empty slot is invalid #4432 

image

image


xiaojieajie:
What is the community contribution? I'm a newcomer。

PR #4578, opened 2021-11-29
fix(style): fix table use with fixed="right", summary not display

### LINKED ISSUE #4550: [Bug Report] table use with fixed="right",  Summary not display
### Element Plus version
1.2.0-beta.3

### OS/Browsers version
mac/win10

### Vue version
3.2.23

### Reproduction Link
https://codepen.io/jkbopo/pen/PoKMBzV

image

### Steps to reproduce
The official website rendering out of the problem

### What is Expected?
table use with fixed="right"  show Summary 

table use with fixed="right",  Summary not display

Ther-su:
fix table use with fixed="right", summary not display

fix #4550
In the original case,the property, either left or right, will go through the first if, which causes the bug.And the case is showed.

image


msidolphin:
This issue was fixed in the refactor

PR #4593, opened 2021-11-30
fix(components): [el-collapse] watch modelValue invalid

### LINKED ISSUE #4587: [Bug Report] Collapse dont't watch v-model change when the model type of Array
### Element Plus version
1.2.0-beta.4

### OS/Browsers version
chrome 96.0.4664.45

### Vue version
3.2.23

### Reproduction Link
https://codepen.io/liwenda/pen/XWeWYdy

### Steps to reproduce
ref model is array
then onMounted to pushed the model 
in html the model is changed 
but the Collapse don't to change 

### What is Expected?
ref model is array
then onMounted to pushed the model 
in html the model is changed 
and the Collapse to change 

ref model is array
then onMounted to pushed the model 
in html the model is changed 
but the Collapse don't to change

Alanscut:

fix #4587

### DIFF
--- a/packages/components/collapse/src/collapse.vue
+++ b/packages/components/collapse/src/collapse.vue
@@ -57,6 +57,9 @@ export default defineComponent({
       () => props.modelValue,
       () => {
         activeNames.value = [].concat(props.modelValue)
+      },
+      {
+        deep: true,
       }
     )
 
--- a/packages/components/collapse/__tests__/collapse.spec.ts
+++ b/packages/components/collapse/__tests__/collapse.spec.ts
@@ -151,4 +151,43 @@ describe('Collapse.vue', () => {
     expect(collapseItemWrappers[0].vm.isActive).toBe(false)
     expect(vm.activeNames).toEqual(['3'])
   })
+
+  test('deep watch modelValue', async () => {
+    const wrapper = mount({
+      components: {
+        'el-collapse': Collapse,
+        'el-collapse-item': CollapseItem,
+      },
+      data() {
+        return {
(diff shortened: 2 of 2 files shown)


Alanscut:
@msidolphin ok

jagoncalves14 (approved):

packages/components/collapse/src/collapse.vue
```diff
@@ -57,6 +57,9 @@ export default defineComponent({
       () => props.modelValue,
       () => {
         activeNames.value = [].concat(props.modelValue)
+      },
+      {
+        deep: true,
       }
```
JeremyWuuuuu, line 63: Do you think it is possible to watch like this:

```ts
watch(() => props.modelValue.length, handler)
```

instead of adding `deep`?

jw-foss (approved):

JeremyWuuuuu:
I left a comment for you, but note that this is not a blocking comment, so you can merge this by yourself if you don't think the suggestions is good.

packages/components/collapse/src/collapse.vue
Alanscut: Performance of `deep` is indeed not as good as `watch`.

When `length` is not changed but the item changed, it can't be monitoring, so should we watch both `props.modelVaule` and `props.modelValue.length`?

JeremyWuuuuu: Let's use deep here for now, if there is any perf concern we can try to fix it later.

JeremyWuuuuu:
LGTM, merged!

PR #4643, opened 2021-12-02
chore(play): fix dev environment

cokemine:
The original `app.example` file does not conform to the format of prettier and eslint

### DIFF
--- a/play/app.example
+++ b/play/app.example
@@ -1,10 +1,10 @@
 <template>
   <div class="play-container">
     <el-icon color="#409EFC">
-      <CheckIcon />
+      <check-icon />
     </el-icon>
     <el-icon>
-      <MenuIcon />
+      <menu-icon />
     </el-icon>
     <el-button>test</el-button>
     <div v-loading="true"></div>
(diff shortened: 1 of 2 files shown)


play/main.ts
```diff
@@ -1,8 +1,11 @@
 import { createApp } from 'vue'
+import ElementPlus from 'element-plus'
```
sxzz, line 2: We can auto import this without import manually

cokemine: Yes, The modification here was accidentally caused, because `unplugin-vue-components` does not work in my environment, there should be some problem somewhere else

play/app.example
```diff
@@ -1,18 +1,18 @@
 <template>
   <div class="play-container">
     <el-icon color="#409EFC">
-      <check />
+      <CheckIcon />
```
sxzz, line 4:
```suggestion
      <check-icon />
```

play/app.example
```diff
@@ -1,18 +1,18 @@
...
     </el-icon>
     <el-icon>
       <MenuIcon />
```
sxzz, line 7: same here

sxzz (approved):

PR #4793, opened 2021-12-11
refactor(utils)!: refactor global config

sxzz:

⚠️ This is a **BREAKING CHANGE!**


packages/utils/popup-manager.ts
```diff
@@ -88,7 +96,7 @@ const PopupManager: IPopupManager = {
   },
 
   nextZIndex() {
-    return ++PopupManager.zIndex
+    return this.getInitialZIndex() + ++this.zIndex
```
JeremyWuuuuu, line 99:
```suggestion
return `${this.getInitialZIndex()}${this.zIndex}`
```

packages/utils/popup-manager.ts
```diff
@@ -65,10 +67,16 @@ const getModal = function (): HTMLElement {
...
+export const PopupManager: PopupManager = {
   modalFade: true,
   modalDom: undefined,
-  zIndex,
+  globalInitialZIndex: 2000,
+  zIndex: 0,
+
+  getInitialZIndex() {
+    if (!getCurrentInstance()) return this.globalInitialZIndex
+    return useGlobalConfig('zIndex').value ?? this.globalInitialZIndex
```
JeremyWuuuuu, line 78:
```suggestion
let zIndex = this.globalInitialZIndex
if (getCurrentInstance()) {
  zIndex = useGlobalConfig('zIndex').value ?? zIndex
}

return zIndex
```

packages/utils/popup-manager.ts
```diff
@@ -216,7 +211,7 @@ const getTopPopup = function () {
 
 if (!isServer) {
   // handle `esc` key when the popup is shown
-  on(window, 'keydown', function (event: KeyboardEvent) {
+  window.addEventListener('keydown', function (event: KeyboardEvent) {
```
JeremyWuuuuu, line 214: **Question:**  Why?

packages/element-plus/make-installer.ts
```diff
@@ -1,33 +1,14 @@
...
     components.forEach((c) => app.use(c))
-
-    if (option.locale) {
-      const localeProvides = localeProviderMaker(opts.locale)
-      app.provide(localeContextKey, localeProvides)
-    }
-
-    app.config.globalProperties.$ELEMENT = option
-    // app.provide() ? is this better? I think its not that flexible but worth implement
-    setConfig(option)
```
JeremyWuuuuu, line 30: We also need to provide the initial global configs to the application if the user is not using `config-provider` to wrap the application.

packages/utils/popup-manager.ts
sxzz: zindex is a number.

packages/utils/popup-manager.ts
sxzz: Fix ts error

packages/element-plus/make-installer.ts
sxzz: IMO, we should let all user move to `config-provider`.

packages/utils/popup-manager.ts
sxzz: So this is indeed the addition of two numbers, and template strings cannot be used.

packages/utils/popup-manager.ts
sxzz: Why？

JeremyWuuuuu: 1. `useGlobalConfig` can only be used when `getCurrentInstance` returns the current rendering instance, per your code
2. For readability's sake, this function body is only for like less than 5 lines, we don't really need to use such brief way to write the code.

packages/element-plus/make-installer.ts
JeremyWuuuuu: The idea is that user should be able to use this library without `config-provider`, the default value we provided is for the underlying code to work correctly (to avoid potential bugs)

packages/utils/popup-manager.ts
JeremyWuuuuu: May be we should consider enhance the typing for `on` method, instead of use `window.addEventListener` as alternative.

packages/element-plus/make-installer.ts
sxzz: I don't think we need to give two ways to set global configuration, please refer to `naive-ui` and `ant-design-vue`. And the configuration of config-provider is reactive.

packages/utils/popup-manager.ts
sxzz: I don't see any benefits of using `on` and `off`. Why do we need this function? 
Also, I plan to remove `on` and `off` in future PRs.

packages/utils/popup-manager.ts
YunYouJun: I agree.

packages/utils/popup-manager.ts
YunYouJun: `on` and `off` is a shortcut. So we have a extra file `utils/dom.ts`.

After reviewing the current code, i think we can use `addEventListener` by ourselves. This reduces both additional file maintenance and better types of readability.

YunYouJun (approved):
Incidentally, I think it's okay to keep the default configuration option for initialization.

YunYouJun:
We also can use [useEventListener](https://vueuse.org/core/useeventlistener/) directly.

sxzz:
@YunYouJun `useEventListener` is generally used in vue components. Since popup manager is a normal js class. So, we can just use normal `addEventListener`.

packages/hooks/use-global-config/index.ts
```diff
@@ -10,7 +11,9 @@ export function useGlobalConfig(): ConfigProviderContext
 export function useGlobalConfig(key?: keyof ConfigProviderContext) {
   const config = inject(configProviderContextKey, {})
   if (key) {
-    return toRef(config, key)
+    return isObject(config) && hasOwn(config, key)
+      ? toRef(config, key)
+      : ref(undefined)
```
JeremyWuuuuu, line 16: I think `ref()` by default returns `Ref<undefined>`

packages/utils/popup-manager.ts
```diff
@@ -44,10 +47,9 @@ const onModalClick = () => {
 }
 
 let hasModal = false
-let zIndex: number
 
 const getModal = function (): HTMLElement {
-  if (!isClient) return
+  if (!isClient) return undefined as any
```
JeremyWuuuuu, line 52: Should we mark the return type as `HTMLElement | undefined` instead of this?

jw-foss (commented):
Left 2 non-blocking comments for you.

packages/hooks/use-global-config/index.ts
sxzz: ref() returns `Ref<any>`, so we should pass `undefined` to ref()

JeremyWuuuuu: Fair enough, since `ref(undefined)` or `ref<undefined>()` are both workable way, we can accept any of them.

packages/utils/popup-manager.ts
sxzz: In the case of SSR, `getModal` should not be called, so `undefined` will not happen.

PR #4972, opened 2021-12-25
feat(components): [date/time-picker] add id attribute to the inputs

purepear:
Support for adding `id` attribute to the inputs of `el-date-picker`, `el-datetime-picker` and `el-time-picker`.
This is very useful for accessibility in forms - clicking on the labels focuses/opens the pickers.

Examples:

```
<el-form :model="formModel">
  <el-form-item prop="birthday" for="birthday-id" label="Birthday">
    <el-date-picker 
      id=""birthday-id"
      v-model="formModel.birthday"
    />
  </el-form-item>
  
  <el-form-item prop="dateRange" for="range-start-id" label="Date range">
    <el-date-picker 
      :id="['range-start-id', 'range-end-id']"
      v-model="formModel.dateRange"
    />
  </el-form-item>
</el-form>
```


packages/components/time-picker/src/common/props.ts
```diff
@@ -6,6 +6,9 @@ import type { Options } from '@popperjs/core'
 import type { ComponentSize } from '@element-plus/utils/types'
 
 export const timePickerDefaultProps = {
+  id: {
```
purepear, line 9: We are **not** defaulting the id to empty string (`default: ''`) on purpose. We want to only add the id if we specify a value. Having an empty string id as the default will add additional noise to the html output `<input id="" .... />` and will break existing test snapshots(not truly backwards compatible).
I was also tempted to remove the `name` prop's `default: ''` for the same reason (we have empty name attributes by default `<input name="" ... />`) but didn't do it because it will break snapshots as well.

msidolphin (approved):

PR #5063, opened 2021-12-30
feat(components): [el-message] config-provider message max attr

### LINKED ISSUE #4868: [Feature Request] el-message max attribute
### Existing Component
是

### Component Name
el-message

### Description
Add Max attribute for El message and can be set in El Config Provider

btea:

feat #4868


packages/components/config-provider/__tests__/config-provider.spec.ts
```diff
@@ -120,4 +120,38 @@ describe('config-provider', () => {
...
+        `,
+      })
+      await nextTick()
+      wrapper.find('.el-button').trigger('click')
+      wrapper.find('.el-button').trigger('click')
+      wrapper.find('.el-button').trigger('click')
+      wrapper.find('.el-button').trigger('click')
+      expect(document.querySelectorAll('.el-message').length).toBe(4)
+    })
+  })
```
JeremyWuuuuu, line 156: **Questions:** when you have the limitation set to max messages to `3`, why do you assert that there are 4 messages that shows up? I think this test case is completely wrong.

**Suggestion:** you should add `await` keyword before `.trigger([event])` statement.

packages/components/config-provider/src/config-provider.ts
```diff
@@ -30,6 +37,7 @@ export default defineComponent({
 
   setup(props, { slots }) {
     provideLocale()
+    Object.assign(messageConfig, props.message || {})
```
JeremyWuuuuu, line 40: I think you should make sure that this value is reactive, for example:

```ts
watch(() => props.message, (val) => {
  Object.assign(messageConfig, val ?? {})
}, { immediate: true })
```

JeremyWuuuuu: Then you should add test against updating the message maximum value.

packages/components/message/src/message-method.ts
```diff
@@ -16,6 +17,9 @@ let seed = 1
 
 const message: MessageFn & Partial<Message> = function (options = {}) {
   if (!isClient) return { close: () => undefined }
+  if (messageConfig.max && instances.length >= messageConfig.max) {
```
JeremyWuuuuu, line 20: We want to make sure that the value passed to `messageConfig.max` is a number.

Do:

```ts
isNumber(messageConfig.max) && instances.length >= messageConfig.max
```

jw-foss (changes requested):
Left a couple comments for you to check, thanks.

packages/components/config-provider/__tests__/config-provider.spec.ts
btea: > **Questions:** when you have the limitation set to max messages to `3`, why do you assert that there are 4 messages that shows up? I think this test case is completely wrong.

I'm sorry, this should be some conflicts caused by another branch merged with dev. I copied it directly and forgot to modify it. 🤣 
> **Suggestion:** you should add `await` keyword before `.trigger([event])` statement.

Maybe I just need to add an `await nextTick()` after all `.trigger([event])`.

packages/components/config-provider/src/config-provider.ts
btea: You are right.

packages/components/message/src/message-method.ts
btea: Done.

emojiiii:
I think we should discuss how to deal(`discard/queue`) with redundant messages if the maximum is exceeded @btea @JeremyWuuuuu

btea:
IMO, redundant message should be discarded. If multiple messages are placed in the queue and pop up one by one, the user may switch to other pages, but the code logic is still popping up messages one after another, which will cause confusion.

emojiiii:
> IMO, redundant message should be discarded. If multiple messages are placed in the queue and pop up one by one, the user may switch to other pages, but the code logic is still popping up messages one after another, which will cause confusion.

But if they are discarded, some important information may be lost?

btea:
> > IMO, redundant message should be discarded. If multiple messages are placed in the queue and pop up one by one, the user may switch to other pages, but the code logic is still popping up messages one after another, which will cause confusion.
> 
> But if they are discarded, some important information may be lost?

I think it depends on the user. If the user thinks the information is important, he can not set the max, or he can set the max a little larger.

emojiiii:
can you test when you use two `config-provide` and different `max` ? @btea

btea:
Generally, only one `config-provider` will be used. If there are more than one, the latter will overwrite the former, and the other configurations are the same.

emojiiii:
You're right. If you can add another `unit test` for this, it will be very robust

btea:
Do you mean to add multiple `config-provider` and then set different `max` values?

emojiiii:
> Do you mean to add multiple `config-provider` and then set different `max` values?

yes

btea:
Maybe it is better to modify the value of config directly like in the test code?

emojiiii:
> Maybe it is better to modify the value of config directly like in the test code?

I'm just worried about whether multiple `config-provider` can merge configuration items as expected

btea:
Okay, I will add unit test about this.

JeremyWuuuuu:
> > Maybe it is better to modify the value of config directly like in the test code?
> 
> I'm just worried about whether multiple `config-provider` can merge configuration items as expected

IMHO, multiple provider is made by the developer, so that it is at their choice to make the call, so that we cannot handle this because they did what they did, that was completely on their own, so that we can just ignore it. Because the specialty about these global items (as designed this way) they should be able to accept that. Or maybe later we can try to pass the context to the call so that we can get the app context.

jw-foss (approved):
LGTM

PR #5187, opened 2022-01-06
feat(components): make `value-format="x"` emit number value

CarterLi:

element-ui 2.0 的 `date-picker` 组件支持 `value-format="timestamp"`，这个参数对于前后端使用时间戳通信时非常有用。

https://element.eleme.io/#/zh-CN/component/date-picker#ri-qi-ge-shi

而 element-plus 却没有对应的属性。最接近的 `value-format="x"` 虽然会把 value 格式化为时间戳，但是得到的确是字符串类型的时间戳，而非数字类型

### DIFF
--- a/packages/components/date-picker/__tests__/date-picker.spec.ts
+++ b/packages/components/date-picker/__tests__/date-picker.spec.ts
@@ -356,8 +356,8 @@ describe('DatePicker', () => {
       )
       const vm = wrapper.vm as any
       const input = wrapper.find('input')
-      input.trigger('blur')
-      input.trigger('focus')
+      await input.trigger('blur')
+      await input.trigger('focus')
       await nextTick()
       {
         ;(document.querySelector('td.available') as HTMLElement).click()
@@ -369,9 +369,50 @@ describe('DatePicker', () => {
           valueFormat
         ).format(valueFormat)
       )
-      wrapper.find('button').trigger('click')
+      await wrapper.find('button').trigger('click')
       await nextTick()
       expect(wrapper.findComponent(Input).vm.modelValue).toBe('2021-05-31')
+    })
+
+    it('with "x"', async () => {
+      const format = 'YYYY/MM/DD'
+      const dateStr = '2021/05/31'
+      const valueFormat = 'x'
+      const value = Date.now()
+      const wrapper = _mount(
+        `
+        <el-date-picker
+          ref="compo"
+          v-model="value"
+          type="date"
+          format="${format}"
+          value-format="${valueFormat}" />
+        <button @click="changeValue">click</button>
+      `,
+        () => {
+          return {
+            value,
+          }
+        },
+        {
+          methods: {
+            changeValue() {
+              this.value = +new Date(dateStr)
+            },
+          },
+        }
+      )
+      const vm = wrapper.vm as any
+      const input = wrapper.find('input')
+      await input.trigger('blur')
+      await input.trigger('focus')
+      await nextTick()
+      ;(document.querySelector('td.available') as HTMLElement).click()
+      await nextTick()
+      expect(vm.value).toBe(+dayjs().startOf('M'))
+      await wrapper.find('button').trigger('click')
+      await nextTick()
+      expect(wrapper.findComponent(Input).vm.modelValue).toBe(dateStr)
     })
   })
 })
--- a/packages/components/time-picker/src/common/picker.vue
+++ b/packages/components/time-picker/src/common/picker.vue
@@ -212,14 +212,17 @@ const parser = function (
   format: string,
   lang: string
 ): Dayjs {
-  const day = isEmpty(format)
-    ? dayjs(date).locale(lang)
-    : dayjs(date, format).locale(lang)
+  const day =
+    isEmpty(format) || format === 'x'
+      ? dayjs(date).locale(lang)
+      : dayjs(date, format).locale(lang)
   return day.isValid() ? day : undefined
 }
 
 const formatter = function (date: number | Date, format: string, lang: string) {
-  return isEmpty(format) ? date : dayjs(date).locale(lang).format(format)
+  if (isEmpty(format)) return date
+  if (format === 'x') return +date
+  return dayjs(date).locale(lang).format(format)
 }
 
 export default defineComponent({


JeremyWuuuuu:
I personally support this feature, and I think this is very useful. I'm going to review your code, for getting it merged. 🙏 Thank you for the contribution in advance.

packages/components/date-picker/__tests__/date-picker.spec.ts
```diff
@@ -373,6 +373,46 @@ describe('DatePicker', () => {
...
+            changeValue() {
+              this.value = +new Date('2021/05/31')
+            },
+          },
+        }
+      )
+      const vm = wrapper.vm as any
+      const input = wrapper.find('input')
+      input.trigger('blur')
+      input.trigger('focus')
```
JeremyWuuuuu, line 407: Keep in mind that we need to add `await wrapper.trigger()` because this returns `Promise`, we need to wait for the changes to happen

packages/components/date-picker/__tests__/date-picker.spec.ts
```diff
@@ -373,6 +373,46 @@ describe('DatePicker', () => {
...
+      `,
+        () => {
+          return {
+            value,
+          }
+        },
+        {
+          methods: {
+            changeValue() {
+              this.value = +new Date('2021/05/31')
```
JeremyWuuuuu, line 399: Let's move this `'2021/05/31'` to a variable so that we can assert it in the upcoming assertions.

jw-foss (commented):
I left a few suggestion from the code style side for you to consider, let me know if you have any questions.

CarterLi:
> I personally support this feature, and I think this is very useful. I'm going to review your code, for getting it merged. 🙏 Thank you for the contribution in advance.

Big THANKS to you

> I left a few suggestion from the code style side for you to consider, let me know if you have any questions.

Fixed

JeremyWuuuuu:
I've just talked to @iamkun, this whole idea is good, and I have found that dayjs has provided formatter for timestamps: [Link](https://day.js.org/docs/en/parse/string-format). Can we do something accordingly to get this feature integrated? Because this change seems complicating the problem.

CarterLi:
@JeremyWuuuuu 

而 element-plus 却没有对应的属性。最接近的 value-format="x" 虽然会把 value 格式化为时间戳，但是得到的确是字符串类型的时间戳，而非数字类型

English: `value-format="x"` will format the value to a string of timestamp, eg `"1641545341292"`, instead of a number

CarterLi:
From official doc:

image

iamkun:
Hi,

in Day.js, `.format()` will always return a `string` according to its API design.

However, in this case in Element Plus, it is more useful to return a `number` timestamp to get a reasonable result.

Just wondering if it's ok to convert the result from format string `x` and `X`  to a `number ` in Element Plus ?

packages/components/calendar/src/calendar.ts
```diff
@@ -1,10 +1,10 @@
 import { buildProps, definePropType } from '@element-plus/utils/props'
 import { UPDATE_MODEL_EVENT } from '@element-plus/utils/constants'
-import type { ExtractPropTypes } from 'vue'
+import type { ExtractPropTypes, PropType } from 'vue'
 
 export const calendarProps = buildProps({
   modelValue: {
-    type: Date,
+    type: [Date, Number] as PropType<Date | number>,
```
iamkun, line 7: I don't think Calnender should accept a number, and it should always be a Date object.

our user could pass new Date(number) to Calender instead.

JeremyWuuuuu:
@CarterLi I wonder can we handle that in the upstream(user code) because `format` means to some kind of pattern, and for the rest of the API call it returns `string` instead of `number` so we need to keep it that way. So for leveraging the pros and cons, I think we should do that.

For the second part that timestamp in JavaScript is not unix timestamp which contains the milliseconds part. We should provide this ability to the users as you wished, but I think we need to make sure that unix timestamp is available for the users.

Let me know what you think about this. /cc @iamkun

CarterLi:
> @CarterLi I wonder can we handle that in the upstream(user code) because `format` means to some kind of pattern, and for the rest of the API call it returns `string` instead of `number` so we need to keep it that way. So for leveraging the pros and cons, I think we should do that.

You mean users should still write

```vue
<el-date-picker :model-value="myTime" @update:model-value="myTime = +$event" />
```

everywhere?

> 
> For the second part that timestamp in JavaScript is not unix timestamp which contains the milliseconds part. We should provide this ability to the users as you wished, but I think we need to make sure that unix timestamp is available for the users.

[timestamp](https://developer.mozilla.org/zh-CN/docs/Web/API/DOMTimeStamp) means to be milliseconds in JavaScript, which has nothing to do with Unix.

> 
> Let me know what you think about this. /cc @iamkun

CarterLi:
Code updated to use `x` @iamkun

iamkun (approved):
LGTM

wfk007:
@iamkun @CarterLi 
what if I pass through a `string timestamp` with value-format `x`
> For consistent results parsing anything other than ISO 8601 strings, you should use [String + Format](https://day.js.org/docs/en/parse/string-format).

wfk007:
老代码是 1.3.0-beta1 版本，无论传入 string 还是 number 类型的 timestamp，都会 emit string 类型的结果。
版本更新后，string 类型的传值会渲染一个错误的显示效果，而且这种 breaking change 会影响业务代码，排查后才发现这个 PR
image

d

limingzhiguang:
源代码是这个（The source code is this.）
`export const formatter = function (
  date: string | number | Date | Dayjs,
  format: string | undefined,
  lang: string
) {
  if (isEmpty(format)) return date
  if (format === 'x') return +date
  return dayjs(date).locale(lang).format(format)
}`
强烈建议改成这样（This is highly recommended）
`export const formatter = function (
  date: string | number | Date | Dayjs,
  format: string | undefined,
  lang: string
) {
  if (isEmpty(format)) return date
  if (format === 'x' || format==='timestamp') return +date
  return dayjs(date).locale(lang).format(format)
}`
期望：能兼容elementui2的属性,又不失element-plus的更牛的名字**x**,🤣
(Expect: to be compatible with elementui2 properties without losing the better name x for element-plus. 🤣)

Issue #3198, opened 2021-09-02
When entering time in time-picker from the keyboard, with the format "НН: ММ", the time is entered incorrectly


Anastasia226:
we enter the time from the keyboard 1818, 18:08 is added to the input field
the ":" symbol is not taken into account. with input format without ":" there is no such problem.
![photo_2021-09-02_11-43-39](https://user-images.githubusercontent.com/72441267/131812457-394a5c40-a406-4617-a485-12acead23505.jpg)
![photo_2021-09-02_11-43-36](https://user-images.githubusercontent.com/72441267/131812478-27f8c753-aa45-4d0e-8c44-0e356fac1b95.jpg)

Alanscut:
if you specify the format `hh:mm`, then of course you need to enter the `:` between the hours and minutes, 
https://github.com/element-plus/element-plus/blob/01b867896efe3c8fe644238e89db67ae53785d02/packages/components/time-picker/src/common/picker.vue#L308

the raw input value will pass to the `parser` function directly, so you should make sure the input value matches the format of you specified.

Anastasia226:
Is there a way to get around this? write time without the ":" symbol

iamkun:
ref: https://github.com/element-plus/element-plus/pull/5259#issuecomment-1010035646

PR #5259, opened 2022-01-09
Feat/picker

jianjunyuu:

根据 #3198  描述，补充了相关的特性

### DIFF
--- a/packages/components/time-picker/src/time-picker-com/panel-time-pick.vue
+++ b/packages/components/time-picker/src/time-picker-com/panel-time-pick.vue
@@ -184,7 +184,36 @@ export default defineComponent({
 
     const parseUserInput = (value: Dayjs) => {
       if (!value) return null
-      return dayjs(value, props.format).locale(lang.value)
+      const unformattedDayjsValue = dayjs(value, props.format).locale(
+        lang.value
+      )
+
+      function formatValue(value: string, format: string): string {
+        if (value) {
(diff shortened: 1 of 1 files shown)


JeremyWuuuuu:
@mryujianjun @iamkun I don't think this is a good feature, because without the `:` symbol, how do you distinguish which part is minutes, which part is hours, when the digits were like: `113`, how to separate them? We cannot introduce branching here.

mryujianjun:
> @mryujianjun I don't think this is a good feature, because without the `:` symbol, how do you distinguish which part is minutes, which part is hours, when the digits were like: `113`, how to separate them? We cannot introduce branching here. /cc @iamkun

First of all thank you for your feedback.
Based on the situation you proposed: enter "113", the program will replace it according to the format, and the default will be filled with 0.
For example: when the format is HH:mm:ss, the user inputs "113", then the program will first verify whether the user input is a valid time format and whether it is a pure numeric string, if so, it will be processed as 11:03: 00
This is my way of thinking.

iamkun:
I do think it is a bit wired.

To me, maybe, 1212 is not a time string but 12:12 is.

mryujianjun:
> I do think it is a bit wired.
> 
> 
> 
> To me, maybe, 1212 is not a time string but 12:12 is.
> 
> 
> 
> 
> 
> 
> 
> 

ok,I got it, thanks.

PR #5401, opened 2022-01-14
refactor(components)!: refactor form

sxzz:

MINOR BREAKING

- feat: support arrayable prop
- improve types
- refactor to script setup
- fix TS errors
- export component instance type
- refactor to JSX
- CSS namespace
- BREAKING: the algorithm of lodash is used to get the property values. `:` not in conversion to `.`.


packages/components/form/src/form-item.vue
```diff
@@ -119,83 +134,50 @@ export default defineComponent({
...
-      }
-      return ret
-    })
     const fieldValue = computed(() => {
-      const model = elForm.model
-      if (!model || !props.prop) {
-        return
-      }
+      const model = formContext?.model
+      if (!model || !props.prop) return
```
HerringtonDarkholme, line 139: Should we skip `{}` here?

packages/components/form/src/form-item.vue
```diff
@@ -119,83 +134,50 @@ export default defineComponent({
...
-        rules.every((rule) => {
-          if (rule.required) {
-            required = true
-            return false
-          }
-          return true
-        })
-      }
-      return required
+      if (rules.length === 0) return false
```
HerringtonDarkholme, line 164: Empty array `some` always return false. `[].some(() => true) // false`

packages/components/form/src/form-item.vue
```diff
@@ -281,29 +254,47 @@ export default defineComponent({
         .map((rule) => ({ ...rule }))
     }
 
-    const evaluateValidationEnabled = () => {
-      isValidationEnabled.value = !!getRules()?.length
-    }
+    const evaluateValidationEnabled: FormItemContext['evaluateValidationEnabled'] =
+      () => {
+        isValidationEnabled.value = !!getRules()?.length
```
HerringtonDarkholme, line 259: rules cannot be nullable.

packages/components/form/src/form-item.vue
```diff
@@ -281,29 +254,47 @@ export default defineComponent({
...
+      }
 
-    const updateComputedLabelWidth = (width: string | number) => {
-      computedLabelWidth.value = width ? `${width}px` : ''
-    }
+    const updateComputedLabelWidth: FormItemContext['updateComputedLabelWidth'] =
+      (width) => {
+        computedLabelWidth.value = width ? `${width}px` : ''
```
HerringtonDarkholme, line 264: Let's rewrite the update statement here.

packages/tokens/form.ts
```diff
@@ -1,47 +1,38 @@
...
+export type FormItemContext = FormItemProps & {
+  $el: HTMLDivElement | undefined
   prop?: string
   size?: ComponentSize
   validateState: string
-  $el: HTMLDivElement
-  validate(trigger: string, callback?: ValidateFieldCallback): void
-  updateComputedLabelWidth(width: number): void
+  validate(trigger: string, callback?: FormValidateCallback): void
+  updateComputedLabelWidth(width: string | number): void
```
HerringtonDarkholme, line 29: Why adding string here?

packages/tokens/form.ts
```diff
@@ -1,47 +1,38 @@
...
+    // expose
+    addField: (field: FormItemContext) => void
+    removeField: (field: FormItemContext) => void
+    resetFields: () => void
+    clearValidate: (props: string | string[]) => void
+    validateField: (props: string | string[], cb: FormValidateCallback) => void
+  }
 
-export interface ElFormItemContext {
+export type FormItemContext = FormItemProps & {
```
HerringtonDarkholme, line 23: Consider preferring interface over type 
```suggestion
interface FormItemContext extends FormItemProps {
```

packages/components/form/src/label-wrap.ts
```diff
@@ -78,18 +81,18 @@ export default defineComponent({
       )
     })
 
-    function render() {
+    return () => {
```
HerringtonDarkholme, line 84: Consider give the returned function a name here for better stack trace when debugging.

packages/components/form/src/label-wrap.ts
```diff
@@ -27,15 +27,17 @@ export default defineComponent({
     updateAll: Boolean,
   },
   setup(props, { slots }) {
-    const el = ref<Nullable<HTMLElement>>(null)
-    const elForm = inject(elFormKey)
-    const elFormItem = inject(elFormItemKey)
+    const formContext = inject(formContextKey)
```
HerringtonDarkholme, line 30: Is `formContext` optional here?  LabelWrap or form might not work if not used in pair. So throwing an error faster might be a better choice?

packages/components/form/src/form.ts
```diff
@@ -0,0 +1,115 @@
...
+      const index = getLabelWidthIndex(oldVal)
+      potentialLabelWidthArr.value.splice(index, 1, val)
+    } else if (val) {
+      potentialLabelWidthArr.value.push(val)
+    }
+  }
+
+  function deregisterLabelWidth(val: number) {
+    const index = getLabelWidthIndex(val)
+    index > -1 && potentialLabelWidthArr.value.splice(index, 1)
```
HerringtonDarkholme, line 97: Prefer using if else for side-effect statements.

packages/components/form/src/form.ts
```diff
@@ -0,0 +1,115 @@
...
+  const potentialLabelWidthArr = ref<number[]>([])
+
+  const autoLabelWidth = computed(() => {
+    if (!potentialLabelWidthArr.value.length) return '0'
+    const max = Math.max(...potentialLabelWidthArr.value)
+    return max ? `${max}px` : ''
+  })
+
+  function getLabelWidthIndex(width: number) {
+    const index = potentialLabelWidthArr.value.indexOf(width)
```
HerringtonDarkholme, line 79: NIT: using `includes` here?

HerringtonDarkholme (commented):
Comment added. In general LGTM.

packages/components/form/src/form-item.vue
sxzz: I think it might just be a code style. Both of them are acceptable.

packages/components/form/src/form-item.vue
sxzz: I didn't get your point. How to rewrite?

sxzz: I didn't get your point. How to rewrite?

packages/components/form/src/form.ts
```diff
@@ -0,0 +1,61 @@
...
+} as const)
+export type FormProps = ExtractPropTypes<typeof formProps>
+
+export const formEmits = {
+  validate: (prop: FormItemProps['prop'], isValid: boolean, message: string) =>
+    (isArray(prop) || isString(prop)) &&
+    isBoolean(isValid) &&
+    isString(message),
+}
+export type FormEmits = typeof formEmits
```
JeremyWuuuuu, line 59: **Question:** should we include namespace for the type?

packages/components/form/src/form.vue
```diff
@@ -1,258 +1,158 @@
...
+const validateField: FormContext['validateField'] = async (
+  properties = [],
+  callback
+) => {
+  if (callback) {
+    validate()
+      .then(() => callback(true))
+      .catch((fields: ValidateFieldsError) => callback(false, fields))
+    return
+  }
```
JeremyWuuuuu, line 75: **Suggestion:** remove chaining here, use `async/await`

packages/components/form/src/form.vue
```diff
@@ -1,258 +1,158 @@
...
-      const { labelPosition, inline } = props
-      return [
-        prefix,
-        `${prefix}--${formSize.value}`,
-        labelPosition ? `${prefix}--label-${labelPosition}` : '',
-        inline ? `${prefix}--inline` : '',
-      ]
-    })
+  if (!model) {
+    debugWarn(COMPONENT_NAME, 'model is required for validate to work!')
```
JeremyWuuuuu, line 80:
```suggestion
    debugWarn(COMPONENT_NAME, 'model is required for form validation!')
```

packages/components/form/src/form.vue
```diff
@@ -1,258 +1,158 @@
...
+  if (fields.length === 0) {
+    return
+  }
 
-    const addField = (field: FormItemCtx) => {
-      if (field) {
-        fields.push(field)
-      }
-    }
+  const fds = filterFields(fields, properties)
```
JeremyWuuuuu, line 87: **Suggestion:** name this to `filteredFields`, do not shorten the variable name because the bundler will do that for you. These variables are meant to be seen by human so it should be human readable.

packages/components/form/src/form.vue
```diff
@@ -1,258 +1,158 @@
...
+  }
 
-    const removeField = (field: FormItemCtx) => {
-      if (field.prop) {
-        fields.splice(fields.indexOf(field), 1)
-      }
-    }
+  let valid = true
+  let invalidFields: ValidateFieldsError = {}
+  let firstInvalidFields: ValidateFieldsError | undefined = undefined
```
JeremyWuuuuu, line 95: **Nitpick:** I do not think you need to explicitly `undefined` a variable.

packages/components/form/src/form.vue
```diff
@@ -1,258 +1,158 @@
...
-          item.$el.scrollIntoView?.()
-        }
-      })
-    }
+watch(
+  () => props.rules,
+  () => {
+    if (props.validateOnRuleChange) validate()
+  }
+)
```
JeremyWuuuuu, line 128: **Suggestion:** for robust code here I think we should deep watch `props.rules` since `Vue` won't automatically deep watch for us.

packages/components/form/src/form.vue
```diff
@@ -1,258 +1,158 @@
...
-      emit,
-      addField,
-      removeField,
-      ...useFormLabelWidth(),
-    })
+provide(
+  formContextKey,
+  reactive({
+    ...toRefs(props),
+    emit,
```
JeremyWuuuuu, line 134: **Question:** what do you think about change this to `formEmit` to avoid naming issue?

JeremyWuuuuu: It also removes confusion here.

packages/components/rate/src/rate.vue
```diff
@@ -47,14 +47,14 @@
...
-import { elFormKey } from '@element-plus/tokens'
+import { formContextKey } from '@element-plus/tokens'
 import { hasClass } from '@element-plus/utils'
 import { EVENT_CODE, UPDATE_MODEL_EVENT } from '@element-plus/constants'
 import { ElIcon } from '@element-plus/components/icon'
 import { StarFilled, Star } from '@element-plus/icons-vue'
 import { useNamespace, useSize } from '@element-plus/hooks'
 import { rateProps, rateEmits } from './rate'
-import type { ElFormContext } from '@element-plus/tokens'
+import type { FormContext } from '@element-plus/tokens'
```
JeremyWuuuuu, line 57: **Suggestion:** please add an empty line between variables and types.

jw-foss (commented):
@sxzz I have left a few comments for you to consider. Let me know if you have any questions :pray:

Plus:

I think we **SHOULD NOT** ignore the warnings by using `catch(() => undefined)`, instead we should do something like:

```ts
function debugWarning(error) {
  if (process.env.NODE_ENV === 'development') {
      console.error(e)
  }
}

// in our validation:

field.validate().catch((e) => debugWarning(e))
```

packages/components/form/src/form-item.vue
```diff
@@ -50,313 +51,280 @@ import {
...
-        callback(validateMessage.value, errors ? fields : {})
-        elForm.emit?.(
-          'validate',
-          props.prop,
-          !errors,
-          validateMessage.value || null
-        )
-      })
+const vm = getCurrentInstance()!
+const formContext = inject(formContextKey, undefined)
```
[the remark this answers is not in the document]
sxzz: Why? `el-form-item` can be used in outside of `el-form`.

packages/components/form/src/form-item.vue
```diff
@@ -50,313 +51,280 @@ import {
...
+        } else {
+          return rule.trigger === trigger
+        }
       })
-    }
+      // exclude trigger
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      .map(({ trigger, ...rule }): RuleItem => rule)
+  )
+}
```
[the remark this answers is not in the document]
sxzz: No, `getFilteredRule` has a param `trigger`.

packages/components/form/src/form-item.vue
```diff
@@ -50,313 +51,280 @@ import {
...
-    const evaluateValidationEnabled = () => {
-      isValidationEnabled.value = !!getRules()?.length
-    }
+const validate: FormItemContext['validate'] = async (trigger, callback) => {
+  if (callback) {
+    validate(trigger)
+      .then(() => callback(true))
+      .catch((fields: ValidateFieldsError) => callback(false, fields))
+    return
+  }
```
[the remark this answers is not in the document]
sxzz: The code is too hard to read and tooooo long.
```ts
try {
  validate(trigger)
  callback(true)
} catch (err) {
  callback(false, err as ValidateFieldsError)
}
```

packages/components/form/src/form.ts
sxzz: Of course not. Just like `ButtonProps`.

packages/components/form/src/form.vue
sxzz: Same as before

packages/components/form/src/form.vue
sxzz: image

packages/components/form/src/form.vue
sxzz: Since it is `formContextKey`, it must be related to `form`, and I think there is no need to repeat the name.

packages/components/form/src/label-wrap.ts
sxzz: I don't know how this would be good for debug, since most of our code uses arrow functions directly.

packages/components/form/src/form-item.vue
```diff
@@ -50,313 +51,280 @@ import {
...
-  getPropByPath,
+  ensureArray,
+  getProp,
+  isString,
+  isBoolean,
 } from '@element-plus/utils'
-import { elFormItemKey, elFormKey } from '@element-plus/tokens'
-import { useSize } from '@element-plus/hooks'
-import LabelWrap from './label-wrap'
+import { formItemContextKey, formContextKey } from '@element-plus/tokens'
```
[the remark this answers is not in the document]
sxzz: I don't think so, there is nothing benefits by adding `El` as a prefix.

packages/components/form/src/form.vue
JeremyWuuuuu: The reason I commented on this one is because that this is provided by `form` but it was consumed by `form-item` so that we should be considering the downstream in this case `form-item`, `form-item` has its own `emit` method, and we provides `form`'s emit, under this circumstance we should consider renaming the method or we provide a delegated method something like `emitsToParent`

sxzz: Is it really necessary? The variable name here is already explicitly `form`.
```ts
formContext.emit('validate', props.prop, !errors, validateMessage.value)
```

packages/components/form/src/form.vue
JeremyWuuuuu: I was talking about doing:

```ts
let firstInvalidFields: ValidateFieldsError | undefined // no need do assignment like firstInvalidFields = undefined

```

JeremyWuuuuu: If you remove the `undefined` constrain meaning that your variable will always be instantiated but it wasn't in your case, that is why the type checker is complaining about it was been used before initiated.

packages/components/form/src/form.vue
JeremyWuuuuu: Well, since in the consumer side we are using it like `formContext.emit` I think this wouldn't be a problem, resolving this

jw-foss (approved):
Now the code LGTM.

packages/components/form/src/form-item.vue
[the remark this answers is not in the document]
JeremyWuuuuu: I was talking about do we really need the `Context` here? it looks cleaner without it `formItemKey`, since we have marked the type for the key as `InjectionKey<FormItemContext>`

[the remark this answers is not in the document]
sxzz: Because all our tokens end with `ContextKey`.

PR #5438, opened 2022-01-17
fix(components): [el-transfer] can't find clearQuery

### LINKED ISSUE #5431: [Bug Report] Uncaught TypeError: Cannot set properties of null (setting 'query')
### Element Plus version
1.3.0-beta.5

### OS/Browsers version
Chrome 版本 97.0.4692.71（正式版本） (x86_64)

### Vue version
3.2.16

### Reproduction Link
https://codesandbox.io/s/element-plus-spa-forked-9tdpi?file=/src/App.vue

### Steps to reproduce
1、点击Clear按钮，执行clearQuery方法

### What is Expected?
清除左侧输入框里的内容

Uncaught TypeError: Cannot set properties of null (setting 'query')

btea:

fix #5431

### DIFF
--- a/packages/components/transfer/src/index.vue
+++ b/packages/components/transfer/src/index.vue
@@ -189,10 +189,13 @@ export default defineComponent({
     const rightPanel = ref<TransferType>()
 
     const clearQuery = (which: 'left' | 'right') => {
-      if (which === 'left') {
-        leftPanel.value!.query = ''
-      } else if (which === 'right') {
-        rightPanel.value!.query = ''
+      switch (which) {
+        case 'left':
+          leftPanel.value!.query = ''
+          break
+        case 'right':
+          rightPanel.value!.query = ''
+          break
       }
     }
 
--- a/packages/components/transfer/src/index.vue
+++ b/packages/components/transfer/src/index.vue
@@ -185,8 +185,8 @@ export default defineComponent({
       emit
     )
 
-    const leftPanel = ref<TransferType | null>(null)
-    const rightPanel = ref<TransferType | null>(null)
+    const leftPanel = ref<TransferType>()
+    const rightPanel = ref<TransferType>()
 
     const clearQuery = (which: 'left' | 'right') => {
       if (which === 'left') {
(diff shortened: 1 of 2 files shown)


packages/components/transfer/src/index.vue
```diff
@@ -183,14 +185,14 @@ export default defineComponent({
       emit
     )
 
-    const leftPanel = ref(null)
-    const rightPanel = ref(null)
+    const leftPanel = ref<TransferType | null>(null)
+    const rightPanel = ref<TransferType | null>(null)
```
emojiiii, line 189: may be `const left Panel = ref<TransferType>()` will be better

emojiiii (approved):

packages/components/transfer/src/index.vue
btea: Updated

packages/components/transfer/src/index.vue
```diff
@@ -183,14 +185,14 @@ export default defineComponent({
...
+    const leftPanel = ref<TransferType>()
+    const rightPanel = ref<TransferType>()
 
     const clearQuery = (which: 'left' | 'right') => {
       if (which === 'left') {
-        leftPanel.value.query = ''
+        leftPanel.value!.query = ''
       } else if (which === 'right') {
-        rightPanel.value.query = ''
+        rightPanel.value!.query = ''
```
JeremyWuuuuu, line 195: Can we use `switch` in this case?

JeremyWuuuuu: I would also like to ask you to add a test against this, so we can assure this code won't be wrong again.

jw-foss (commented):
I left 2 comments for you to consider, let me know if you have any questions.

packages/components/transfer/src/index.vue
btea: > Can we use `switch` in this case?  

ok.

> I would also like to ask you to add a test against this, so we can assure this code won't be wrong again.  

Good advice.

sxzz (approved):

PR #5451, opened 2022-01-18
    refactor(components): refactor check-tag

sxzz:

### DIFF
--- a/packages/components/check-tag/src/check-tag.vue
+++ b/packages/components/check-tag/src/check-tag.vue
@@ -1,35 +1,22 @@
 <template>
-  <span
-    :class="{
-      'el-check-tag': true,
-      'is-checked': checked,
-    }"
-    @click="onChange"
-  >
+  <span :class="[ns.b(), { 'is-checked': checked }]" @click="onChange">
     <slot></slot>
   </span>
 </template>
 <script lang="ts">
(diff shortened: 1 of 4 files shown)


packages/components/check-tag/src/check-tag.vue
```diff
@@ -1,42 +1,30 @@
 <template>
-  <span
-    :class="{
-      'el-check-tag': true,
-      'is-checked': checked,
-    }"
-    @click="onChange"
-  >
+  <span :class="[ns.b(), { 'is-checked': checked }]" @click="onChange">
```
JeremyWuuuuu, line 2: Can we move this into `setup` so that we can have a cleaner template?

jw-foss (approved):

packages/components/check-tag/src/check-tag.vue
sxzz: I think we should put class and style to **template**. `script` is best to put logic related.

JeremyWuuuuu: Yes I agree, but for here **Computing the classnames is sort of part of the logic here**, I think a cleaner template would have better readability, even though in this case the template is relatively simple, but we kind of need this to follow the code style (making template cleaner)

PR #5454, opened 2022-01-18
refactor(components): [row] add namespace & instance type

sxzz:

### DIFF
--- a/packages/components/row/src/row.ts
+++ b/packages/components/row/src/row.ts
@@ -37,7 +37,7 @@ const Row = defineComponent({
       gutter,
     })
 
-    const style = computed<CSSProperties>(() => {
+    const style = computed(() => {
       const styles: CSSProperties = {
         marginLeft: '',
         marginRight: '',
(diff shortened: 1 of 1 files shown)


packages/components/row/src/row.ts
```diff
@@ -24,34 +25,36 @@ export const rowProps = buildProps({
...
+    const ns = useNamespace('row')
+
     const gutter = computed(() => props.gutter)
     provide('ElRow', {
       gutter,
     })
 
-    const style = computed(() => {
-      const ret = {
+    const style = computed<CSSProperties>(() => {
```
JeremyWuuuuu, line 40: I think the code is a little bit duplicated here since you have marked the return of this computed as `CSSProperties` which means the variable that you returned from the function body is inferred as `CSSProperties`, so that you don't need to explicitly declare that the `styles` is `CSSProperties`

jw-foss (approved):

packages/components/row/src/row.ts
sxzz: Fixed.

jw-foss (approved):

PR #5528, opened 2022-01-21
refactor(components): [el-table] use namespace

msidolphin:

### DIFF
--- a/packages/components/table/src/filter-panel.vue
+++ b/packages/components/table/src/filter-panel.vue
@@ -9,16 +9,16 @@
     append-to-body
     effect="light"
     pure
-    popper-class="el-table-filter"
+    :popper-class="ns.b()"
     persistent
   >
     <template #content>
       <div v-if="multiple">
-        <div class="el-table-filter__content">
-          <el-scrollbar wrap-class="el-table-filter__wrap">
+        <div :class="ns.e('content')">
+          <el-scrollbar :wrap-class="ns.e('wrap')">
             <el-checkbox-group
               v-model="filteredValue"
-              class="el-table-filter__checkbox-group"
+              :class="ns.e('checkbox-group')"
             >
               <el-checkbox
                 v-for="filter in filters"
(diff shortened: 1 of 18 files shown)


packages/components/table/src/filter-panel.vue
```diff
@@ -69,7 +71,10 @@
     <template #default>
       <span
         v-click-outside:[popperPaneRef]="hideFilterPanel"
-        class="el-table__column-filter-trigger el-none-outline"
+        :class="[
+          `${ns.namespace.value}-table__column-filter-trigger`,
```
JeremyWuuuuu, line 75: I don't think we really need this `.value` here since in template it automatically unwraps the refs

packages/components/table/src/table-column/render-helper.ts
```diff
@@ -57,10 +64,17 @@ function useRender<T>(
     const source = cellForced[type] || {}
     Object.keys(source).forEach((prop) => {
       const value = source[prop]
-      if (value !== undefined) {
-        column[prop] = prop === 'className' ? `${column[prop]} ${value}` : value
+      if (prop !== 'className' && value !== undefined) {
+        column[prop] = value
```
JeremyWuuuuu, line 68: This seems like different than the original expression could you please to elaborate here?

packages/components/table/src/table-column/render-helper.ts
```diff
@@ -123,7 +137,7 @@ function useRender<T>(
           style: {},
         }
         if (column.showOverflowTooltip) {
-          props.class += ' el-tooltip'
+          props.class += ` ${unref(ns.namespace)}-tooltip`
```
JeremyWuuuuu, line 140: Can we be consistent here? For instance

```ts
props.class = `${props.class} ${unref(ns.namespace)}-tooltip`
```

packages/components/table/src/filter-panel.vue
msidolphin: > I don't think we really need this `.value` here since in template it automatically unwraps the refs

emmm,  vue not automatically unwraps 

image

jw-foss (approved):

PR #5666, opened 2022-01-27
refactor(components): [el-pagination] optimize component style

msidolphin:

### DIFF
--- a/packages/components/pagination/src/components/sizes.vue
+++ b/packages/components/pagination/src/components/sizes.vue
@@ -4,7 +4,7 @@
       :model-value="innerPageSize"
       :disabled="disabled"
       :popper-class="popperClass"
-      size="small"
+      :size="size"
       @change="handleChange"
     >
       <el-option
@@ -41,6 +41,10 @@ const paginationSizesProps = buildProps({
     default: '',
   },
   disabled: Boolean,
+  size: {
+    type: String,
+    default: 'default',
+  },
 } as const)
 
 export default defineComponent({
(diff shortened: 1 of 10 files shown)


packages/components/pagination/src/components/sizes.vue
```diff
@@ -41,6 +41,10 @@ const paginationSizesProps = buildProps({
     default: '',
   },
   disabled: Boolean,
+  size: {
+    type: String,
+    default: 'default',
+  },
```
JeremyWuuuuu, line 47: Can we use the prop defined in `useSizes`?

packages/components/pagination/src/pagination.ts
```diff
@@ -234,6 +234,15 @@ export default defineComponent({
...
     }
 
+    function addClass(element: any, cls: string) {
+      if (element) {
+        if (!element.props) {
+          element.props = {}
+        }
+        element.props.class = [element.props.class, cls].join(' ')
+      }
+    }
```
JeremyWuuuuu, line 244: I don't think changing the props here is a good idea, can we use `provide/inject` to do the same thing here?

JeremyWuuuuu: Can we do something else here? Exposing API to the parent for mutating the state.

```ts
element.someMethod()
```

jw-foss (approved):
I left some minor comments for you to consider, let me know if you have any questions ;>

packages/components/pagination/src/pagination.ts
msidolphin: This method is modifies the class of the vnode, so I don't think it's necessary to exposing API

packages/components/pagination/src/components/sizes.vue
msidolphin: No, the size only have default and small

JeremyWuuuuu:
Let's roll it

Issue #5612, opened 2022-01-25
[Bug Report] cascader：当同时存在多选和筛选的情况下，placeholder显示异常


frontYang:
### Element Plus version
1.3.0-beta.8

### OS/Browsers version
windows

### Vue version
3.2.29

### Reproduction Link
https://codepen.io/frontyang/pen/bGYbyLX

### Steps to reproduce
1、当同时存在多选和筛选的情况下，placeholder显示异常，有重叠文案
![image](https://user-images.githubusercontent.com/16046879/150902017-6d01d285-fceb-4838-8c97-7e87307ad6f5.png)

2、选中元素后，清空全部选中，placeholder会部分消失
![image](https://user-images.githubusercontent.com/16046879/150902041-056e16b4-1b5a-402f-96d6-1ee1636b9779.png)

### What is Expected?
当同时存在多选和筛选的情况下placeholder能显示正常

### What is actually happening?
1、当同时存在多选和筛选的情况下，placeholder显示异常，有重叠文案
2、选中元素后，清空全部选中，placeholder会部分消失

JeremyWuuuuu:
This cloud be caused by the extra input field that we used in here for providing DOM information.

dingyuanjie:
There are two input overlaps in this area. Just adjust the style。

dingyuanjie:
很少在github提交代码。不太会用。

kooriookami:
It seems be repaired.

PR #5675, opened 2022-01-27
fix: fix that when multiple selection and filtering exist at the same time, the placeholder displays an exception and there are overlapping copies

dingyuanjie:


cn-troy:
link #5612

JeremyWuuuuu:
I'm closing this one since this is duplicated with #5853

PR #5701, opened 2022-01-29
refactor(hooks): refactor popup manager to `useZIndex`

sxzz:


packages/hooks/use-popup-manager/index.ts
```diff
@@ -0,0 +1,30 @@
...
+export const usePopupManager = () => {
+  let globalInitialZIndex = 2000 // TODO: move to @element-plus/constants
+
+  const setGlobalInitialZIndex = (value: number) =>
+    (globalInitialZIndex = value)
+
+  const getInitialZIndex = () => {
+    if (!getCurrentInstance()) return ref(globalInitialZIndex)
+    return useGlobalConfig('zIndex', globalInitialZIndex)
+  }
```
JeremyWuuuuu, line 14: **Question:** is it possible that we use `computed` here instead of invoking the function ourselves?

E.g.

```ts
const globalConfig = useGlobalConfig('zIndex', globalInitialZIndex)

const initialZIndex = computed(() => {
  if (!getCurrentInstance()) return globalInitialZIndex
  return globalConfig.value
})
```

packages/element-plus/make-installer.ts
```diff
@@ -20,7 +19,7 @@ export const makeInstaller = (components: Plugin[] = []) => {
       () => unref(options).zIndex,
       () => {
         const zIndex = unref(options).zIndex
-        if (isNumber(zIndex)) PopupManager.globalInitialZIndex = zIndex
+        if (isNumber(zIndex)) usePopupManager().setGlobalInitialZIndex(zIndex)
```
JeremyWuuuuu, line 22: I don't think that this will work with nested popup manager since the variable `globalInitialZIndex` gets initialized every time when you call `usePopupManager`, which means it is not the same variable that you have changed here, would you mind add a unit test covering this?

packages/hooks/use-popup-manager/index.ts
```diff
@@ -0,0 +1,30 @@
+import { ref, getCurrentInstance } from 'vue'
+import { useGlobalConfig } from '../use-global-config'
+
+let zIndex = 0
+export const usePopupManager = () => {
+  let globalInitialZIndex = 2000 // TODO: move to @element-plus/constants
```
JeremyWuuuuu, line 6: If you really want the `setGlobalInitialZIndex` to work, I think you should move this variable out of this hook body.

```ts
let zIndex = 0
let globalInitialZIndex = 2000

export const setGlobalInitialZIndex = (val) => {
  globalInitialZIndex = value
}
export const usePopupManager = () => {
  // Other contents
}
```

jw-foss (commented):
Left a few comments for you to consider ;)

packages/hooks/use-popup-manager/index.ts
sxzz: Fixed.

sxzz:
I simplified code, PTAL /cc @JeremyWuuuuu

jw-foss (approved):

PR #5765, opened 2022-02-02
fix(component): [pagination] pageSize reactivity bug

### LINKED ISSUE #5700: [Bug Report] el-pagination Page number does not update when page-size changes
### Element Plus version
1.3.0-beta.3

### OS/Browsers version
Windows 10/Google Chrome 97.0.4692.99

### Vue version
3.2.25

### Reproduction Link
https://github.com/element-plus/element-plus

### Steps to reproduce
This problem is also present in the examples from the official documentation.When the size of each page (page-size) changes, the page numbers are not automatically updated.
官方文档的示例中也存在这个问题。在总数量(total)确定的情况下，选择器中修改每页数量后，后面的页码数没有自动更新。

### What is Expected?
when the size of each page (page-size) changes, the page numbers should be automatically updated.
在总数量确定的情况下，选择器中修改每页数量后，后面的页码数应该重新计算，自动更新。

Page numbers have not changed.
实际结果是页码数没有变化，显示的还是修改前的页码数。

neil585456525:
fix #5700

I have edit the part of pageSize logic, to ensure the reactivity. Besides, this change can deal with the "WARNING" part in the [document](https://element-plus.gitee.io/en-US/component/pagination.html#attributes), too. Lets user can use this component less extra effort.

### DIFF
--- a/packages/components/pagination/src/pagination.ts
+++ b/packages/components/pagination/src/pagination.ts
@@ -139,9 +139,9 @@ export default defineComponent({
           // Thus page size listener is required
           // users are account for page size change
           if (!isAbsent(props.pageSize)) {
-            if (!hasPageSizeListener) {
-              return false
-            }
+            // if (!hasPageSizeListener) {
+            //   return false
+            // }
           } else {
             // (else block just for explaination)
             // else page size is controlled by el-pagination internally
--- a/packages/components/pagination/src/pagination.ts
+++ b/packages/components/pagination/src/pagination.ts
@@ -152,20 +152,37 @@ export default defineComponent({
     })
 
     const innerPageSize = ref(
-      isAbsent(props.defaultPageSize) ? 10 : props.defaultPageSize
+      (() => {
+        let pageSize = 10
+        if (!isAbsent(props.defaultPageSize)) {
+          pageSize = props.defaultPageSize
+        }
+        if (!isAbsent(props.pageSize)) {
+          pageSize = props.pageSize
+        }
+        return pageSize
+      })()
     )
+
+    watch(
+      () => props.pageSize,
+      () => {
+        if (!isAbsent(props.pageSize)) {
+          innerPageSize.value = props.pageSize
+        }
+      }
+    )
+
     const innerCurrentPage = ref(
       isAbsent(props.defaultCurrentPage) ? 1 : props.defaultCurrentPage
     )
 
     const pageSizeBridge = computed({
       get() {
-        return isAbsent(props.pageSize) ? innerPageSize.value : props.pageSize
+        return innerPageSize.value
       },
       set(v: number) {
-        if (isAbsent(props.pageSize)) {
-          innerPageSize.value = v
-        }
+        innerPageSize.value = v
         if (hasPageSizeListener) {
           emit('update:page-size', v)
           emit('size-change', v)
--- a/packages/components/pagination/__tests__/pagination.spec.ts
+++ b/packages/components/pagination/__tests__/pagination.spec.ts
@@ -410,5 +410,33 @@ describe('Pagination', () => {
        * expect(style.outline).toBeTruthy()
        */
     })
+
+    test('test the pageSize selected reactivity', async () => {
+      const wrapper = mount({
+        setup() {
+          return () => {
+            return h(Pagination, {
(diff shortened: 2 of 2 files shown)


sam585456525:
Hi, I found that when unit-test show that this must be deprecated. Did you want to fully deprecate the pageSize/defaultPageSize usage, or using my code to keep the compatibility ?

JeremyWuuuuu:
> Hi, I'm sorry I was thoughtless. I found that this must be a deprecated usage. Did you want to fully deprecate the pageSize usage, or using my code to keep the compatibility ? Because of the extra effort which needed to use this component, maybe we can let it more flexible to use in this way. Avoid confusing the users.

For the next major version (`2.0.0`) it still needs to be available, we will deprecate it in `2.1.0`

packages/components/pagination/src/pagination.ts
```diff
@@ -152,20 +152,37 @@ export default defineComponent({
...
+        if (!isAbsent(props.pageSize)) {
+          pageSize = props.pageSize
+        }
+        return pageSize
+      })()
+    )
+
+    watch(
+      () => props.pageSize,
+      () => {
```
JeremyWuuuuu, line 169: I think we can get the value here by:

```ts
watch(() => props.pageSize, (pageSize) => {
  // do all the stuff
})
```

JeremyWuuuuu: **Question:** if I passed a`pageSize` with a valid variable, then I pass another one which makes `isAbsent` true, the `innerPageSize.value` will not be updated, which means, page size is still not reactive, is this intended?

jw-foss (commented):
I left a few comments for you to consider, if you have any question please don't hesitate to reach out ;)

packages/components/pagination/src/pagination.ts
sam585456525: Yeah,I add this logic as the result of the comment block on the isAbsent variable. It told that using the legal input is user's responsibility.

sam585456525: Should we throw a error or a warning message during isAbsent equals to true? 
I am curious about why using the isAbsent check logic instead of using the vue native props type check?

JeremyWuuuuu: I think it wants to make the component work no matter what the user's input is, so it used `isAbsent` to determine whether the attribute has been given, I think maybe you were right, it should always be the users responsibility, not ours. Let's use your way here.

Loop the original author here @ioslh

ioslh:
#5700 should blame on my mistake, very sorry for that; documentation usage is not correct any more after I completely refactored pagination component, will modify soon.


Back to my original refactor idea: 
1. follow vue’s reactive system, take best advantages of double way data-binding. So we encourage users to use v-model double-way binding props like current-page and page-size. 
2. base on 1, try best to reduce users’ effort to adopt pagination component. Take pageSize for example:
    - If user do provide props.pageSize, we respect that prop, and apply it. 
    - Otherwise when (**and only when**) user do not provide props.pageSize, we use internally defined ref value innerPageSize to track internal page size data, make pagination work. In other word, if props.pageSize is provided, we don’t have to read/change `innerPageSize` any more. Please keep this behavior. (`innerCurrentPage` has the same purpose).


`Should we throw an error or a warning message during isAbsent equals to true?`
No. take following code for example. 
```
<el-pagination /*…other-props */ @update:current-page=“onPageChange”  />

const onPageChange = (newPage: number) => {
  // do something with newPage
}
```
currentPage is not provided( isAbsent check will return true ), because user might not want to keep currentPage himself, only want to be notified when currentPage changed. That’s a fair reasonable usage, not an error or warning situation.(In this case, internal `innerCurrentPage` will take over and remember current page, and notify/emit user when changed)


Actually there is one point we should consider more carefully: if user’s usage is not valid(assertValidUsage === false), should we keep return null(current result), or render the pagination(may not work correctly)?  @sam585456525  @JeremyWuuuuu

sam585456525:
In my personal opinion, Maybe we can still render the pagination although it may work incorrectly. The reason is that let pagination fit other components behavior like form, it would render no matter the input if correct.

ioslh:
Documentation fix see #5833

PR #5779, opened 2022-02-03
fix(components): [el-dropdown] use custom attributes for dropdown items

### LINKED ISSUE #5768: [Feature Request] Bind custom attrs to ElDropdownItem
### Existing Component
Yes

### Component Name
ElDropdownItem

### Description
Sometimes we want to bind some custom attributes to ElDropdownItem. In old versions it is possible, but it seems to be removed in the latest version. It would be great, if the attrs binding can be back again. Thanks.

image

mawi1512:
Fix https://github.com/element-plus/element-plus/issues/5768

Maybe the issue should be marked as Bug instead of Feature Request as it worked in previous versions.

### DIFF
--- a/packages/components/dropdown/src/dropdown-item.vue
+++ b/packages/components/dropdown/src/dropdown-item.vue
@@ -5,7 +5,7 @@
   >
     <el-roving-focus-item :focusable="!disabled">
       <el-dropdown-item-impl
-        v-bind="$attrs"
+        v-bind="propsAndAttrs"
         @pointerleave="handlePointerLeave"
         @pointermove="handlePointerMove"
         @click="handleClick"
@@ -44,7 +44,7 @@ export default defineComponent({
   inheritAttrs: false,
   props: dropdownItemProps,
   emits: ['pointermove', 'pointerleave', 'click'],
-  setup(props, { emit }) {
+  setup(props, { emit, attrs }) {
     const { elDropdown } = useDropdown()
     const _instance = getCurrentInstance()
     const itemRef = ref<HTMLElement | null>(null)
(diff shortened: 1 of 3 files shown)


packages/components/dropdown/src/dropdown-item.vue
```diff
@@ -5,7 +5,7 @@
   >
     <el-roving-focus-item :focusable="!disabled">
       <el-dropdown-item-impl
-        v-bind="$props"
+        v-bind="$attrs"
```
JeremyWuuuuu, line 8: I thin we need to do:

```html
v-bind="{...$props, ...$attrs}"
```

Allowing the `$props` to fallthrough.

JeremyWuuuuu: @mawi1512 Would you mind change the code as I commented? I think your code are just fine.

mawi1512: @JeremyWuuuuu You are right, $props was missing. I tried to use v-bind="{...$props, ...$attrs}", but it caused type errors. I hope the alternative solution is also ok. Let me know what you think.

packages/components/dropdown/src/dropdown-item.vue
```diff
@@ -97,11 +98,17 @@ export default defineComponent({
       }
     )
 
+    // direct usage of v-bind={ ...$props, ...$attrs } causes type errors
+    const propsAndAttrs = computed(() => {
+      return { ...props, ...attrs }
+    })
```
JeremyWuuuuu, line 104: **Praise:** I actually want it to be this way.

packages/components/dropdown/src/dropdown-item.vue
JeremyWuuuuu: Looking good, test has been triggered, if everything was fine then we should be able to merge this.

jw-foss (approved):
Let's merge it!

PR #5795, opened 2022-02-06
fix(hooks): [use-namespace] cannot be used in nested hooks

buqiyuan:

### DIFF
--- a/packages/hooks/use-namespace/index.ts
+++ b/packages/hooks/use-namespace/index.ts
@@ -25,9 +25,8 @@ const _bem = (
 }
 
 export const useNamespace = (block: string) => {
-  const namespace = computed(
-    () => useGlobalConfig('namespace').value || defaultNamespace
-  )
+  const globalConfig = useGlobalConfig('namespace')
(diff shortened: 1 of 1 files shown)


packages/hooks/use-namespace/index.ts
```diff
@@ -51,7 +50,7 @@ export const useNamespace = (block: string) => {
       ? _bem(unref(namespace), block, blockSuffix, element, modifier)
       : ''
   const is = (name: string, state = true) =>
-    state ? `${statePrefix}${name}` : ''
+    name && state ? `${statePrefix}${name}` : ''
```
JeremyWuuuuu, line 53: I think the name here is required, we don't need this thing, besides, this method is internal use only, so our developer must know how it works.

sxzz: This feature can be used here. If `name` is a empty string, we just ignore it.

https://github.com/element-plus/element-plus/blob/6ff6270cdccd53507bda14830ccd0e07ce199c78/packages/components/select/src/select.vue#L92

sxzz (approved):

PR #5920, opened 2022-02-10
fix(components): [el-table] avoid scrollbar not display

### LINKED ISSUE #5917: [Bug Report] ElTable 从keepalive中恢复后，垂直滚动条不显示。
### Discussed in https://github.com/element-plus/element-plus/discussions/5908

<div type='discussions-op-text'>

<sup>Originally posted by **zychome** February 10, 2022</sup>
ElTable 从keepalive中恢复后，垂直滚动条不显示。</div>

msidolphin:

### DIFF
--- a/packages/components/scrollbar/src/scrollbar.vue
+++ b/packages/components/scrollbar/src/scrollbar.vue
@@ -115,7 +115,6 @@ export default defineComponent({
 
     const update = () => {
       if (!wrap$.value) return
-
       const offsetHeight = wrap$.value.offsetHeight - GAP
       const offsetWidth = wrap$.value.offsetWidth - GAP
 
@@ -149,6 +148,19 @@ export default defineComponent({
         }
       },
       { immediate: true }
+    )
+
+    watch(
+      () => style.value,
+      () => {
+        if (!props.native)
+          nextTick(() => {
+            update()
+            if (wrap$.value) {
+              barRef.value?.handleScroll(wrap$.value)
+            }
+          })
+      }
     )
 
     provide(
--- a/packages/components/scrollbar/src/scrollbar.vue
+++ b/packages/components/scrollbar/src/scrollbar.vue
@@ -151,7 +151,7 @@ export default defineComponent({
     )
 
     watch(
-      () => style.value,
+      () => [props.maxHeight, props.height],
       () => {
         if (!props.native)
           nextTick(() => {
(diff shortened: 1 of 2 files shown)


packages/components/scrollbar/src/scrollbar.vue
```diff
@@ -151,6 +150,19 @@ export default defineComponent({
       { immediate: true }
     )
 
+    watch(
```
JeremyWuuuuu, line 153: Should we deep watch or we do something like:

```ts
watch(() => {
  const [_, ourStyle] = style.value
  return {...ourStyle}
}, handler)
```

JeremyWuuuuu: I suspect that this watch would take effect since the `style.value.length` has not been changed, `watch` will not deep watch for us in Vue3.

packages/components/scrollbar/src/scrollbar.vue
```diff
@@ -151,6 +150,19 @@ export default defineComponent({
...
+      () => style.value,
+      () => {
+        if (!props.native)
+          nextTick(() => {
+            update()
+            if (wrap$.value) {
+              barRef.value?.handleScroll(wrap$.value)
+            }
```
JeremyWuuuuu, line 161: Is this `nextTick` necessary here? I think we should be a little bit careful with using `nextTick` in our code.

jw-foss (commented):
Left some comments for you to consider ;)

packages/components/scrollbar/src/scrollbar.vue
msidolphin: > I suspect that this watch would take effect since the `style.value.length` has not been changed, `watch` will not deep watch for us in Vue3.

style is a `computed` property, so I think it's fine

packages/components/scrollbar/src/scrollbar.vue
msidolphin: > Is this `nextTick` necessary here? I think we should be a little bit careful with using `nextTick` in our code.

I need to make sure the dom update is complete, otherwise the height will be wrong

packages/components/scrollbar/src/scrollbar.vue
msidolphin: > I suspect that this watch would take effect since the `style.value.length` has not been changed, `watch` will not deep watch for us in Vue3.

I shouldn't listened to the computed property, I changed it to `watch(() => [props.maxHeight, props.height])`

jw-foss (approved):

PR #5933, opened 2022-02-10
refactor(components): refactor button

sxzz:

### DIFF
--- a/packages/components/button/src/button.vue
+++ b/packages/components/button/src/button.vue
@@ -35,8 +35,8 @@
   </button>
 </template>
 
-<script lang="ts">
-import { computed, inject, defineComponent, Text, ref } from 'vue'
+<script lang="ts" setup>
+import { computed, inject, Text, ref, useSlots } from 'vue'
 import { useCssVar } from '@vueuse/core'
 import { TinyColor } from '@ctrl/tinycolor'
 import { ElIcon } from '@element-plus/components/icon'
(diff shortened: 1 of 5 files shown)


packages/components/button/src/button.vue
```diff
@@ -48,120 +48,89 @@ import {
...
+const emit = defineEmits(buttonEmits)
+const slots = useSlots()
+
+const buttonGroupContext = inject(buttonGroupContextKey, undefined)
+const globalConfig = useGlobalConfig('button')
+const ns = useNamespace('button')
+const { form } = useFormItem()
+const _size = useSize(computed(() => buttonGroupContext?.size))
+const _disabled = useDisabled()
+const _ref = ref()
```
JeremyWuuuuu, line 67: Can we add annotation for this one to refer to `HTMLButtonElement`? Which enriches the typing system

packages/hooks/__tests__/use-form-item.spec.ts
```diff
@@ -36,15 +36,15 @@ const mountComponent = (setup = NOOP, options = {}) => {
 
 const getButtonVm = (wrapper: ReturnType<typeof mountComponent>) => {
   return wrapper.findComponent(ElButton).vm as any as {
```
JeremyWuuuuu, line 38: **Question:** why do we need to cast the type here? Shouldn't `ElButton` be type annotated already?

sxzz: Exported variables are not attached to the Button type.

packages/components/button/src/button.vue
sxzz: Fixed.

zhangsanfe (approved):

packages/hooks/__tests__/use-form-item.spec.ts
JeremyWuuuuu: Just as you mentioned it is not attached, should we try test the variable via other options like in the code, `_size` is used here: `n.m(_size)`, shouldn't we test this by finding whether the variable exists or not? For instance:

```ts
expect(wrapper.find('el-button--default').exists()).toBeTrue()
```

jw-foss (approved):

JeremyWuuuuu:
LGTM! Thanks!

PR #5935, opened 2022-02-10
fix(components): [el-tag] add prop hit default value

btea:

image


packages/components/tag/src/tag.ts
```diff
@@ -9,7 +9,10 @@ export const tagProps = buildProps({
     values: ['success', 'info', 'warning', 'danger', ''],
     default: '',
   },
-  hit: Boolean,
+  hit: {
+    type: Boolean,
+    default: false,
```
sxzz, line 14: the default value of boolean prop is `false`. Why should we explicitly define it?

btea: ok.

PR #5936, opened 2022-02-10
refactor: fix `buildProps` usage

sxzz:

### DIFF
--- a/packages/components/dropdown/src/dropdown.ts
+++ b/packages/components/dropdown/src/dropdown.ts
@@ -24,59 +24,56 @@ export interface IElDropdownInstance {
   triggerElm?: ComputedRef<Nullable<HTMLButtonElement>>
 }
 
-export const dropdownProps = {
+export const dropdownProps = buildProps({
   trigger: useTooltipTriggerProps.trigger,
-
   effect: {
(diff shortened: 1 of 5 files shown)


packages/components/popover/src/popover.ts
```diff
@@ -15,12 +15,10 @@ export const usePopoverProps = {
...
-      type: [String, Number],
-      default: 150,
-    },
-  }),
-}
+  width: {
+    type: [String, Number],
+    default: 150,
+  },
+})
```
JeremyWuuuuu, line 24: Seems like you forgot `as const` casting here?

packages/components/dropdown/src/dropdown.ts
```diff
@@ -28,59 +28,56 @@ export interface IElDropdownInstance {
   triggerElm?: ComputedRef<Nullable<HTMLButtonElement>>
 }
 
-export const dropdownProps = {
+export const dropdownProps = buildProps({
```
JeremyWuuuuu, line 31: The reason that I didn't add this method here is that when extending from other component props, `buildProps` tend to complain about the type not matching since the props that we were using is built by `buildProps`.

jw-foss (approved):

PR #6221, opened 2022-02-22
refactor(theme-chalk): use scss list & map generate type css var

YunYouJun:
- add scss function `joinVarName` `getCssVar`
- add scss mixin `css-var-from-global`
- use map & list generate button/tag css vars
- fix button computed can not hot reload when save
- use css var for checkable tag

https://user-images.githubusercontent.com/25154432/155111836-a417bb24-612c-402a-922c-98a69062a47a.mov

close #4856, #5271, #6224

---

Please make sure these boxes are checked before submitting your PR, thank you!

- [x] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).
- [x] Make sure you are merging your commits to `dev` branch.
- [x] Add some descriptions and refer to relative issues for your PR.

PR #6225, opened 2022-02-22
fix(components): [button] style broken inside button-group with type specificed 

### LINKED ISSUE #6224: [Bug Report] Button-group with specific type not working
### Element Plus version
2.0.2

### OS/Browsers version
latest

### Vue version
3.2.31

### Reproduction Link
https://jsfiddle.net/vq6tmuja/

### Steps to reproduce
1. Wrap button with button-group
2. Set specific type for button-group

### What is Expected?
Button should be the type specific in button-group if there is no type specific for it

Button's style is broken if it is inside of button-group with type specific

freedomlang:

fixed #6224 Compute button's style according type in props and injected type

### DIFF
--- a/packages/components/button/src/button.vue
+++ b/packages/components/button/src/button.vue
@@ -85,7 +85,9 @@ const shouldAddSpace = computed(() => {
 })
 
 // calculate hover & active color by color
-const typeColor = computed(() => useCssVar(`--el-color-${props.type}`).value)
+const typeColor = computed(
+  () => useCssVar(`--el-color-${props.type || _type.value}`).value
+)
 const buttonStyle = computed(() => {
   let styles: Record<string, string> = {}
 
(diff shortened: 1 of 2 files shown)


packages/components/button/src/button.vue
```diff
@@ -85,7 +85,9 @@ const shouldAddSpace = computed(() => {
 })
 
 // calculate hover & active color by color
-const typeColor = computed(() => useCssVar(`--el-color-${props.type}`).value)
+const typeColor = computed(
+  () => useCssVar(`--el-color-${props.type || _type.value}`).value
```
JeremyWuuuuu, line 89: **Question:** since`_type.value` is declared:

```ts
const _type = computed(() => props.type || buttonGroupContext?.type || '')
```
Don't you think that only `_type.value` is enough?

freedomlang: Yes, `_type.value` is enough. Thanks for the review, i'll update that solution.

freedomlang:
#6221 do the fixed, no need for this patch

PR #6231, opened 2022-02-23
fix(components): [el-dropdown-menu] adjust html tag

### LINKED ISSUE #6227: [Bug Report] Dropdown component generates invalid HMTL
### Element Plus version
2.0.2

### OS/Browsers version
Irrelevant

### Vue version
3.2.31

### Reproduction Link
https://jsfiddle.net/okdw9atq/1/

### Steps to reproduce
Bug is present in the doc: [https://element-plus.org/en-US/component/dropdown.html#basic-usage](Dropdown Component)

### What is Expected?
A valid HTML is expected:

```
    <div class="el-scrollbar__view el-dropdown__list" style="">
      <ul
        class="el-dropdown-menu el-dropdown-menu--default"
        tabindex="-1"
        role="menu"
        style="outline: none;"
      >
        <!--v-if--><li
          data-el-collection-item=""
          aria-disabled="false"
          class="el-dropdown-menu__item"
          tabindex="-1"
          role="menuitem"
        >
          <!--v-if-->Action 1
        </li><!--v-if--><li
          data-el-collection-item=""
          aria-disabled="false"
          class="el-dropdown-menu__item"
          tabindex="-1"
          role="menuitem"
        >
          <!--v-if-->Action 2
        </li><!--v-if--><li
          data-el-collection-item=""
          aria-disabled="false"
          class="el-dropdown-menu__item"
          tabindex="-1"
          role="menuitem"
        >
          <!--v-if-->Action 3
        </li>
      </ul>
    </div>
```

Generated HTML is:

```
    <ul class="el-scrollbar__view el-dropdown__list" style="">
      <ul
        class="el-dropdown-menu el-dropdown-menu--default"
        tabindex="-1"
        role="menu"
        style="outline: none;"
      >
        <!--v-if--><div
          data-el-collection-item=""
          aria-disabled="false"
          class="el-dropdown-menu__item"
          tabindex="-1"
          role="menuitem"
        >
          <!--v-if-->Action 1
        </div><!--v-if--><div
          data-el-collection-item=""
          aria-disabled="false"
          class="el-dropdown-menu__item"
          tabindex="-1"
          role="menuitem"
        >
          <!--v-if-->Action 2
        </div><!--v-if--><div
          data-el-collection-item=""
          aria-disabled="false"
          class="el-dropdown-menu__item"
          tabindex="-1"
          role="menuitem"
        >
          <!--v-if-->Action 3
        </div>
      </ul>
    </ul>
```

Two nested `ul`, no `li`.

msidolphin:
fix #6227

### DIFF
--- a/packages/components/dropdown/src/dropdown-item-impl.vue
+++ b/packages/components/dropdown/src/dropdown-item-impl.vue
@@ -1,10 +1,10 @@
 <template>
-  <div
+  <li
     v-if="divided"
     :class="ns.bem('menu', 'item', 'divided')"
     v-bind="$attrs"
-  ></div>
-  <div
+  ></li>
+  <li
     :ref="itemRef"
     v-bind="{ ...dataset, ...$attrs }"
     :aria-disabled="disabled"
@@ -20,7 +20,7 @@
   >
     <el-icon v-if="icon"><component :is="icon" /></el-icon>
     <slot />
-  </div>
+  </li>
 </template>
 
 <script lang="ts">
--- a/packages/components/dropdown/src/dropdown.vue
+++ b/packages/components/dropdown/src/dropdown.vue
@@ -27,7 +27,7 @@
         <el-scrollbar
           ref="scrollbar"
           :wrap-style="wrapStyle"
-          tag="ul"
+          tag="div"
           :view-class="ns.e('list')"
         >
           <el-focus-trap trapped @mount-on-focus="onMountOnFocus">


packages/components/dropdown/src/dropdown-item-impl.vue
```diff
@@ -1,10 +1,10 @@
 <template>
-  <div
+  <li
     v-if="divided"
     :class="ns.bem('menu', 'item', 'divided')"
     v-bind="$attrs"
-  ></div>
-  <div
+  ></li>
```
JeremyWuuuuu, line 6: Maybe we should keep the original `div` but add `role="separator"` to it.

msidolphin: > Maybe we should keep the original `div` but add `role="separator"` to it.

But div should not be child element of ul, right?

amaury-hanser: Yes, `div` can't be a child of `ul`.  
See: [\<ul\>: The Unordered List element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul)

> Permitted content [of ul] | Zero or more \<li\>, \<script\> and \<template\> elements.

JeremyWuuuuu: > Yes, div can't be a child of ul.
See: [<ul>: The Unordered List element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul)

I see, should we add `role="separator"` for the `li` element?

amaury-hanser: I don't think that it's needed as a `li` is already a semantic element and will be declared as an item.

In the `<el-dropdown-item>`, there is an *attribute* `divider`, but I'm not sure to understand why it would be needed as it seems to do nothing when I test it in the fiddle.

I'm pretty new to *element+* so it might be interesting to wait for a more experienced opinion.

jw-foss (approved):

PR #6300, opened 2022-02-26
refactor(components): [dialog] refactor

sxzz:
- fix TS type.
- enhance prop type `beforeClose`
- move `token.ts` to `@element-plus/tokens`


packages/components/dialog/__tests__/dialog.spec.ts
```diff
@@ -6,25 +6,14 @@ import { Delete } from '@element-plus/icons-vue'
...
-const _mount = ({ slots, ...rest }: Record<string, any>) => {
-  return mount(Dialog, {
-    slots: {
-      default: AXIOM,
-      ...slots,
-    },
-    ...rest,
-  })
-}
+const defaultSlots = { default: AXIOM }
```
JeremyWuuuuu, line 9:
```suggestion
const defaultSlots = { default: () => AXIOM }
```

sxzz: Why, `AXIOM` is a text node

packages/components/dialog/src/dialog.ts
```diff
@@ -11,7 +11,9 @@ export const dialogProps = buildProps({
     default: false,
   },
   beforeClose: {
-    type: definePropType<(...args: any[]) => void>(Function),
+    type: definePropType<(done: (shouldCancel: boolean) => void) => void>(
```
JeremyWuuuuu, line 14: **Nitpick:** can we extract this into a type so that we can have a nicer format here?
e.g.

```ts
type DoneFn = (cancel: boolean) => void
type BeforeCloseFn = (done: DoneFn) => void
```

sxzz: Updated.

packages/components/dialog/__tests__/dialog.spec.ts
JeremyWuuuuu: But slots should be functions instead of plain objects

sxzz: Both are acceptable.

JeremyWuuuuu: Refer: https://vuejs.org/guide/extras/render-function.html#passing-slots

> Instead of an array, we need to pass either a slot function, or an object of slot functions

JeremyWuuuuu: This is rather a convention instead of a rule because of:

> Passing slots as functions allows them to be invoked lazily by the child component. This leads to the slot's dependencies being tracked by the child instead of the parent, which results in more accurate and efficient updates.

JeremyWuuuuu: I know that your code was not syntactically wrong but shouldn't we follow the convention?

sxzz: Updated.

jw-foss (approved):

PR #6440, opened 2022-03-05
feat(components): [date-picker] add align option

### LINKED ISSUE #6406: [Feature Request] [Enhancement] [date-picker] Implement the "align" option for the date picker component as in element-ui
Feature type: **`Enhancement`**

## Details

- Enhanced component name: `date-picker`

## Description

The element-ui date-picker allowed **align** to be set to `left/center/right` and defaulted to `left`. It doesn't seem like the option is available in element-plus and only aligns as `center`.

```vue
<template>
  <el-date-picker
    v-model="dateRange"
    type="daterange"
    :size="size"
    :clearable="false"
    range-separator="-"
    align="start"
    start-placeholder="Start date"
    end-placeholder="End date"
    placeholder="Select date range"
    :shortcuts="datePickerShortcuts"
    @change="setSelectedRange"
    format="DD MMM YYYY"
  />
</template>
```

### Additional comments

_(empty)_

nieyuyao:
Closes #6406 

- [x] Adds `align` option to the `ElDatePicker` and `ElTimePicker`

----


JeremyWuuuuu:
@nieyuyao First of all, thank you for contributing to Element Plus.

One thing I do want to mention here is that the API seems irrelevant with `align` instead it relates to the popup content, so I think we should give the API a `namespace` to avoid misunderstanding here.

packages/components/time-picker/src/common/picker.vue
```diff
@@ -654,6 +663,25 @@ export default defineComponent({
       props,
     })
 
+    const getRealAlign = (propAlign: string) => {
+      return (
+        ['start', 'middle', 'end'].find((align) => align === propAlign) ??
+        'start'
+      )
+    }
```
JeremyWuuuuu, line 671: Why not make this a `computed`?

```ts
const _alignment = computed(() => {
  return ['start', 'middle', 'end'].find((align) => align === props.align) ?? 'start'
})
```

nieyuyao:
> @nieyuyao First of all, thank you for contributing to Element Plus.
> 
> One thing I do want to mention here is that the API seems irrelevant with `align` instead it relates to the popup content, so I think we should give the API a `namespace` to avoid misunderstanding here.

Okay, I see. For example, we name it `pickerContentAlign`? 🤔

JeremyWuuuuu:
> Okay, I see. For example, we name it pickerContentAlign? 🤔

Yes! Something like that, the main idea is that we do not want to make our user confused with so many APIs. It is always a better option to add a namespace

JeremyWuuuuu:
@nieyuyao Just checking in to see if I can help you with anything, let me know if you do.

packages/components/time-picker/src/common/picker.vue
nieyuyao: done

xiaoxian521:

PR #6447, opened 2022-03-06
refactor(docs): [form] improve docs and typings

sxzz:

### DIFF
--- a/docs/en-US/component/form.md
+++ b/docs/en-US/component/form.md
@@ -13,7 +13,7 @@ The component has been upgraded with a flex layout to replace the old float layo
 
 :::
 
-## Basic form
+## Basic Form
 
 It includes all kinds of input items, such as `input`, `select`, `radio` and `checkbox`.
 
@@ -33,7 +33,7 @@ To prevent this behavior, you can add `@submit.prevent` on `<el-form>`.
 
 :::
 
-## Inline form
+## Inline Form
 
 When the vertical space is limited and the form is relatively simple, you can put it in one line.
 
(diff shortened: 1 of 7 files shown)


docs/en-US/component/form.md
```diff
@@ -111,77 +111,81 @@ form/size-control
...
+| `rules`          | Validation rules of form, see the [following table](#formitemrule), more advanced usage at [async-validator](https://github.com/yiminghe/async-validator).    | `FormItemRule \| FormItemRule[]`  | —           |
+| `error`          | Field error message, set its value and the field will validate error and show this message immediately.                                                       | `string`                          | —           |
+| `show-message`   | Whether to show the error message.                                                                                                                            | `boolean`                         | `true`      |
+| `inline-message` | Inline style validate message.                                                                                                                                | `boolean`                         | `false`     |
+| `size`           | Control the size of components in this form-item.                                                                                                             | `'large' \| 'default' \| 'small'` | `'default'` |
+
+#### FormItemRule
```
JeremyWuuuuu, line 172:
```suggestion
#### FormItem Rule
```

jw-foss (commented):
Left a comment for you

docs/en-US/component/form.md
sxzz: This is TS type name

JeremyWuuuuu: Since this was written as a title, I think we should respect the the way how title was treated in the documentation, BTW, we need to annotate what `arrayable` is for the audience because they don't know about the actual type because it was not a built-in type.

sxzz: image

I think it needs to be consistent with the type in the table. And this is a level 4 heading that will not show up in the TOC.

Do you have a better way to label Arrayable?

jw-foss (approved):

PR #6537, opened 2022-03-11
fix(components): [input]compute error

HammCn:
compute error when Emoji contained:

"😄😄".length = 4
Array.from("😄😄").length = 2

### DIFF
--- a/packages/components/input/src/input.vue
+++ b/packages/components/input/src/input.vue
@@ -247,7 +247,7 @@ export default defineComponent({
         !props.readonly &&
         !props.showPassword
     )
-    const textLength = computed(() => Array.from(nativeInputValue.value).length)
+    const textLength = computed(() => nativeInputValue.value.length)
     const inputExceed = computed(
       () =>
         // show exceed style if length of initial value greater then maxlength


HammCn:
image

50 / 100 but I can't type more word any more.

HerringtonDarkholme:
Please fix the testing and resolve conflicts

JeremyWuuuuu:
@HammCn I would like to suggest that we add an additional option e.g. `emoji` as boolean, when this flag is set to `true`, then we compute the length in `emoji` way, otherwise we do it the same way.

HammCn:
> @HammCn I would like to suggest that we add an additional option e.g. `emoji` as boolean, when this flag is set to `true`, then we compute the length in `emoji` way, otherwise we do it the same way.

I mean that it's ok in the condition of restriction, but the label content displayed is wrong.

eg. https://github.com/element-plus/element-plus/pull/6537#issuecomment-1064887658

i think the label  displayed maybe 100 / 100 is right , or : 

i can type 100 emojis really.

xiaoxian521:

HammCn (commented):
Fixed in #11816

PR #6539, opened 2022-03-11
feat(components): [el-table] exposes scroll methods

msidolphin:
re #6432

### DIFF
--- a/packages/components/table/src/composables/use-scrollbar.ts
+++ b/packages/components/table/src/composables/use-scrollbar.ts
@@ -1,0 +1,32 @@
+import { ref } from 'vue'
+import { isNumber } from '@element-plus/utils'
+
+export const useScrollbar = () => {
+  const scrollBarRef = ref()
+
+  const scrollTo = ({ top, left }: { top?: number; left?: number }) => {
+    setScrollTop(top)
+    setScrollLeft(left)
+  }
+
+  const setScrollTop = (top?: number) => {
+    const scrollbar = scrollBarRef.value
+    if (scrollbar && isNumber(top)) {
+      scrollbar.setScrollTop(top)
+    }
+  }
+
+  const setScrollLeft = (left?: number) => {
+    const scrollbar = scrollBarRef.value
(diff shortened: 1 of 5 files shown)


packages/components/table/src/composables/use-scrollbar.ts
```diff
@@ -0,0 +1,32 @@
...
+      scrollbar.setScrollTop(top)
+    }
+  }
+
+  const setScrollLeft = (left?: number) => {
+    const scrollbar = scrollBarRef.value
+    if (scrollbar && isNumber(left)) {
+      scrollbar.setScrollLeft(left)
+    }
+  }
```
JeremyWuuuuu, line 24: It seems to me that these method are quite similar, can we extract the common part to avoid duplication?

```ts
const setScrollPosition = (position: 'Top' | 'Left', val?: number) => {
  const scrollbar = scrollBarRef.value
    if (scrollbar && isNumber(val) && ['Top', 'Left'].includes(position)) {
      scrollbar[`setScroll${position}`](left)
    }
}

const setScrollTop = (val?: number) => setScrollPosition('Top', val);
const setScrollLeft = (val?: number) => setScrollPosition('Left', val);

```

msidolphin:
@JeremyWuuuuu done

packages/components/table/src/composables/use-scrollbar.ts
```diff
@@ -0,0 +1,28 @@
+import { ref } from 'vue'
+import { isNumber } from '@element-plus/utils'
+
+export const useScrollbar = () => {
+  const scrollBarRef = ref()
+
+  const scrollTo = ({ top, left }: { top?: number; left?: number }) => {
+    setScrollTop(top)
+    setScrollLeft(left)
```
JeremyWuuuuu, line 9: **Suggestion:** let's use `element.scrollTo(options)` to implement this method.

Ref: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo

jw-foss (approved):

PR #6570, opened 2022-03-12
ci(test): add SSR test framework

jw-foss:
- Setup cypress framework for e2e tests


vite.config.cy.mjs
```diff
@@ -0,0 +1,20 @@
+import { defineConfig } from 'vite'
```
sxzz, line 1: How about move this file to `/cypress`?

e2e/button/button.spec.cy.tsx
```diff
@@ -0,0 +1,14 @@
+import { mount } from '@cypress/vue'
```
sxzz, line 1: Why move to top level `/e2e` directory?

JeremyWuuuuu: If we keep that inside `packages/components`, we will need to update `jest.config` for skipping cypress tests in this case I think we should keep it under `e2e`

sxzz: I can help you, could you move it to `packages/components` and I will update `tsconfig.json`

JeremyWuuuuu: It was intended to keep them outside the package. Please just accept it as-is.

Issue #6127, opened 2022-02-17
[Bug Report] Input word limit value not correct 


asdjgfr:
### Element Plus version
2.0.2

### OS/Browsers version
Windows 11
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.102 Safari/537.36

### Vue version
3.2.31

### Reproduction Link
https://codepen.io/asdjgfr/pen/LYOeZGK

### Steps to reproduce
paste or input some emoji such as 👩‍👩‍👧‍👧 or 🤰🏻

### What is Expected?
👩‍👩‍👧‍👧 or 🤰🏻 should be calculated as 1

### What is actually happening?
👩‍👩‍👧‍👧 compute as 7
🤰🏻 compute as 2

![image](https://user-images.githubusercontent.com/23442840/154446919-f735760f-2825-4fcc-901a-f31be2782f74.png)

bastarder:
😄！！！
i try `"👩‍👩‍👧‍👧".length` in chrome , get `11`

asdjgfr:
> 😄！！！ i try `"👩‍👩‍👧‍👧".length` in chrome , get `11`

I have already created a [new pull request](https://github.com/element-plus/element-plus/pull/6130) to fixed and test it .

sxzz:
I think it's by design. `maxlength` attribute also has the same effect.
WDYT /cc @JeremyWuuuuu

asdjgfr:
> I think it's by design. `maxlength` attribute also has the same effect. WDYT /cc @JeremyWuuuuu

When I set maxlength to 10 and show-word-limit, input `👩‍👩‍👧‍👧` *(four people)* but got `👩‍👩‍👧‍` *(three people)* and `6 / 10` ,also only one more character can be inserted. It makes me confused. I saw another library, [ Ant Design Vue](https://next.antdv.com/index) didn't have `show-word-limit`, [Vuetify](https://vuetifyjs.com/en/) all props not supported 😂.

JeremyWuuuuu:
@sxzz @asdjgfr I think we should take @asdjgfr's proposal, since the emoji '👩‍👩‍👧‍' is constructed by
[...]

PR #6588, opened 2022-03-13
fix(components): [el-input] unicode charactors does not work as expected

zt123123:

### 问题描述：
当输入表情符号等unicode特殊字符时，使用原生JS计算字符串长度以及使用DOM Props`maxlength`属性存在如下问题：
* JS计算字符串长度与DOM Props计算长度方式不一致，并且与用户期望有所差异。
* 存在兼容问题，不同浏览器行为表现不一致。

### 解决方案：
抛弃原生DOM Props `maxlength`属性，使用JS逻辑控制最大长度
使用了第三方npm包[`runes`](https://www.npmjs.com/package/runes)做字符串长度的计算以及截取

###  截图：
Chrome：
用户认为输入了8个长度字符串，JS计算长度为9，DOM Props计算长度为10（因为此时无法继续输入了）
image

Safari：
计算方式也存在问题，甚至可以输入超过最大长度的字符串
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,8 @@
     "lodash-es": "^4.17.21",
     "lodash-unified": "^1.0.2",
     "memoize-one": "^6.0.0",
-    "normalize-wheel-es": "^1.1.1"
+    "normalize-wheel-es": "^1.1.1",
+    "runes": "^0.4.3"
   },
   "devDependencies": {
     "@babel/core": "^7.17.5",
--- a/packages/components/input/src/input.vue
+++ b/packages/components/input/src/input.vue
@@ -86,7 +86,7 @@
           </el-icon>
           <span v-if="isWordLimitVisible" :class="nsInput.e('count')">
             <span :class="nsInput.e('count-inner')">
-              {{ textLength }} / {{ attrs.maxlength }}
+              {{ textLength }} / {{ maxlength }}
             </span>
           </span>
         </span>
@@ -127,7 +127,7 @@
         @keydown="handleKeydown"
       />
       <span v-if="isWordLimitVisible" :class="nsInput.e('count')">
-        {{ textLength }} / {{ attrs.maxlength }}
+        {{ textLength }} / {{ maxlength }}
       </span>
     </template>
   </div>
@@ -147,6 +147,7 @@ import {
   useAttrs as useRawAttrs,
   toRef,
 } from 'vue'
+import runes from 'runes'
 import { isClient } from '@vueuse/core'
 import { isNil } from 'lodash-unified'
 import { ElIcon } from '@element-plus/components/icon'
@@ -186,7 +187,11 @@ const instance = getCurrentInstance()!
 const rawAttrs = useRawAttrs()
 const slots = useSlots()
 
-const attrs = useAttrs()
+const originAttrs = useAttrs()
+// use custom maxlength instead of native maxlength
+const { maxlength, ...restAttrs } = originAttrs.value
+const attrs = ref(restAttrs)
+
 const { form, formItem } = useFormItem()
 const inputSize = useSize()
 const inputDisabled = useDisabled()
@@ -237,18 +242,17 @@ const showPwdVisible = computed(
 const isWordLimitVisible = computed(
   () =>
     props.showWordLimit &&
-    !!attrs.value.maxlength &&
+    !!maxlength &&
     (props.type === 'text' || props.type === 'textarea') &&
     !inputDisabled.value &&
     !props.readonly &&
     !props.showPassword
 )
-const textLength = computed(() => Array.from(nativeInputValue.value).length)
+const textLength = computed(() => runes(nativeInputValue.value).length)
 const inputExceed = computed(
   () =>
     // show exceed style if length of initial value greater then maxlength
-    !!isWordLimitVisible.value &&
-    textLength.value > Number(attrs.value.maxlength)
+    !!isWordLimitVisible.value && textLength.value > Number(maxlength)
 )
 const suffixVisible = computed(
   () =>
@@ -310,7 +314,7 @@ const updateIconOffset = () => {
 }
 
 const handleInput = async (event: Event) => {
-  const { value } = event.target as TargetElement
+  let { value } = event.target as TargetElement
 
   // should not emit input during composition
   // see: https://github.com/ElemeFE/element/issues/10516
@@ -319,6 +323,12 @@ const handleInput = async (event: Event) => {
   // hack for https://github.com/ElemeFE/element/issues/8548
   // should remove the following line when we don't support IE
   if (value === nativeInputValue.value) return
+
+  // hack for native maxlength, if the input length greater than maxlength, just cut off the excess
+  if (runes(value).length > Number(maxlength)) {
+    // eslint-disable-next-line unicorn/prefer-string-slice
+    value = runes.substr(value, 0, maxlength)
+  }
 
   emit(UPDATE_MODEL_EVENT, value)
   emit('input', value)
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -79,6 +79,7 @@ importers:
       rollup-plugin-css-only: ^3.1.0
       rollup-plugin-esbuild: ^4.8.2
       rollup-plugin-filesize: ^9.1.2
+      runes: ^0.4.3
       sass: ^1.49.9
       sucrase: ^3.20.3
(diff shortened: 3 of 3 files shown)


zt123123:
I guess @asdjgfr has the same question 😂 https://github.com/element-plus/element-plus/issues/6127

plainheart:
How about using a simpler function?
```js
Array.from('12345😀').length
// output: 6
```

But just for your reference as it can't still cover all emoji characters, such as `👩‍❤️‍💋‍👩`. It contains 8 elements in fact,
['👩', '‍', '❤', '️', '‍', '💋', '‍', '👩'].

And I'm concerned that your database can't store your expected length eventually as although these emojis look like one character, the stored length isn't.

zt123123:
> 
it works well
image

zt123123:
> How about using a simpler function?
> 
> ```js
> Array.from('12345😀').length
> // output: 6
> ```
> 
> But just for your reference as it can't still cover all emoji characters, such as `👩‍❤️‍💋‍👩`. It contains 8 elements in fact, ['👩', '‍', '❤', '️', '‍', '💋', '‍', '👩'].
> 
> And I'm concerned that your database can't store your expected length eventually as although these emojis look like one character, the stored length isn't.
> 
> **Addition**
> 
> Tested with MySQL. It confirmed my concern mentioned above.
> 
> 
> 
> 
> 
> As the screenshots show, the database can't store "**the one character**" with a real length of **5**. So just please follow the behavior in javascript.

你说的有道理，不过你确定数据库和JS解析出来的长度是一致的吗，其他数据库呢？

plainheart:
I haven't tested with other databases yet, but I believe most of them won't regard the length of such an emoji character like `👩‍❤️‍💋‍👩` as **1**, in other words, it's at least bigger than **1**.

plainheart:
Additional reply for the question

> 不过你确定数据库和JS解析出来的长度是一致

It's the same. It can be saved to the database only when I change the field length to be **8**, which is the real length of this emoji character.

zt123123:
> Additional reply for the question
> 
> > 不过你确定数据库和JS解析出来的长度是一致的吗
> 
> It's the same. It can be saved to the database only when I change the field length to be **8**, which is the real length of this emoji character.

哦哦，那就没问题，优先保证后端可以正常存储数据。浏览器厂商实现也不同，可能也有这部分原因。毕竟这种情况也少，不修复也无伤大雅，谢谢大佬指点。

zt123123:
> 👩‍❤️‍💋‍👩
大佬，能给我指点一下吗，我看了源码使用了Array.from去计算字符串的长度，当时就感觉很奇怪，为什么要多此一举，原来另有玄机。我想知道为什么这种方式和`length`属性表现不一致，这是什么原理呢？

zt123123:
> I haven't tested with other databases yet, but I believe most of them won't regard the length of such an emoji character like `👩‍❤️‍💋‍👩` as **1**, in other words, it's at least bigger than **1**.
@plainheart 
大佬，能给我指点一下吗，我看了源码使用了Array.from去计算字符串的长度，当时就感觉很奇怪，为什么要多此一举，原来另有玄机。我想知道为什么这种方式和`length`属性表现不一致，这是什么原理呢？

plainheart:
It's hard to make it clear. There is a relatively detailed [explanation](https://stackoverflow.com/questions/60053160/what-characters-are-grouped-with-array-from) for your information.

JeremyWuuuuu:
> > I haven't tested with other databases yet, but I believe most of them won't regard the length of such an emoji character like `👩‍❤️‍💋‍👩` as **1**, in other words, it's at least bigger than **1**.
> > @plainheart
> > 大佬，能给我指点一下吗，我看了源码使用了Array.from去计算字符串的长度，当时就感觉很奇怪，为什么要多此一举，原来另有玄机。我想知道为什么这种方式和`length`属性表现不一致，这是什么原理呢？

Because some `emojis` are concatenated in with  each other for example 👨‍👦  is consisted of 👨 and 👦.

JeremyWuuuuu:
I would like to suggest that we add an additional option e.g. `emoji` as boolean, when this flag is set to `true`, then we compute the length in emoji way, otherwise we do it the same way.

Something like this: 
```vue
<template>
  <el-input v-model="something" emoji />
<template>
```

zt123123:
> I would like to suggest that we add an additional option e.g. `emoji` as boolean, when this flag is set to `true`, then we compute the length in emoji way, otherwise we do it the same way.
> 
> Something like this:
> 
> ```
> <template>
>   <el-input v-model="something" emoji />
> <template>
> ```

may be it is a good idea!!!

JeremyWuuuuu:
Since you guys all want the feature we intended that this should be something to keep the old way work. When enabling this allows the user to input emojis, BTW, when your user types emoji to the input box, it would be very strange to them when **1** emoji taking up to 7 characters. It would be strange, we should do something to patch that to avoid this situation, which would make this feature much more complicated. 

This is not a simple feature to fix IMHO, we should take it carefully.

plainheart:
Generally, we limit the length to ensure the user input won't exceed the defined max length of the column in the database. Even though we calculated the "real length" of the strings that contain emoji characters, it's still not possible to save this string to the database. Even so, do you still think it's meaningful to add a new option for this?

JeremyWuuuuu:
You see, I don't you whether any of you have used `Slack` or `Twitter` before, making a input box can be really complicated, `Slack` treats all emojis as image so that it always fine to put as much as emojis they want

plainheart:
Yes. To be honest, they finally stored the converted characters according to the defined conversion rule and use corresponding images to show them, instead of the raw emoji. But I don't think such a function is necessary to be implemented by a common framework like element-plus.

asdjgfr:
@JeremyWuuuuu 

  I think we should comply with JavaScript length. Because I test some emoji on `windows 11`、`macOS Monterey` 、`ubuntu 20.4 LTS` 、`Android 12` and `IOS 15.4`, different system has different display.

  Even on same system but app use different font also have different display. In vscode use 微软雅黑 look like `👩‍👩‍👧‍👧` , but in webstorm use `JetBrain Mono`  look like `👩👩👧👧` .

  As for confusing users, I think it's pm need to solve, not ours.
cdec9745e907442f956c4350c84f44ab

I will close my [issuse](https://github.com/element-plus/element-plus/issues/6127) and [pr](https://github.com/element-plus/element-plus/pull/6130), thx.

zt123123:
> I think we should comply with JavaScript length. Because I test some emoji on `windows 11`、`macOS Monterey` 、`ubuntu 20.4 LTS` 、`Android 12` and `IOS 15.4`, different system has different display.
> 
> 我认为我们应该遵守 JavaScript 的长度。因为我在 windows 11、 macOS Monterey、 ubuntu 20.4 LTS、 Android 12和 IOS 15.4上测试了一些表情符号，不同的系统有不同的显示方式。
> 
> Even on same system but app use different font also have different display. In vscode use 微软雅黑 look like `👩‍👩‍👧‍👧` , but in webstorm use `JetBrain Mono` look like `👩👩👧👧` .
> 
> 即使在同一个系统，但应用程序使用不同的字体也有不同的显示。在 vscode 中使用 something something 看起来像，但在 webstorm 中使用 JetBrain Mono 看起来像。
> 
> As for confusing users, I think it's pm need to solve, not ours. cdec9745e907442f956c4350c84f44ab
> 
> 至于让用户困惑的问题，我认为需要解决的是下午，而不是我们。
> 
> I will close my [issuse](https://github.com/element-plus/element-plus/issues/6127) and [pr](https://github.com/element-plus/element-plus/pull/6130), thx.
> 
> 我将关闭我的帐篷和公关，谢谢。

zt123123:
> @JeremyWuuuuu
> 
> I think we should comply with JavaScript length. Because I test some emoji on `windows 11`、`macOS Monterey` 、`ubuntu 20.4 LTS` 、`Android 12` and `IOS 15.4`, different system has different display.
> 
> Even on same system but app use different font also have different display. In vscode use 微软雅黑 look like `👩‍👩‍👧‍👧` , but in webstorm use `JetBrain Mono` look like `👩👩👧👧` .
> 
> As for confusing users, I think it's pm need to solve, not ours. cdec9745e907442f956c4350c84f44ab
> 
> I will close my [issuse](https://github.com/element-plus/element-plus/issues/6127) and [pr](https://github.com/element-plus/element-plus/pull/6130), thx.

老哥说的对，解决不了问题，就解决提出问题的人，让产品去考虑这个问题吧

PR #6615, opened 2022-03-14
fix(components): [table] not scroll empty block when scrolling table

plainheart:

I'm not sure if this behavior is by design but I always feel that the empty shouldn't be placed in the scroll view as it also unexpectedly scrolls together with the content when the table is scrollable. See the comparison below,

| Current | After |
| :----: | :----: |
| current behavior | expected behavior |

Playground is available [here](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICAgIDxlbC10YWJsZSBib3JkZXIgc3R5bGU9XCJ3aWR0aDogNTAwcHhcIj5cbiAgICAgIDxlbC10YWJsZS1jb2x1bW4gcHJvcD1cImRhdGVcIiBsYWJlbD1cIkRhdGVcIiB3aWR0aD1cIjE4MFwiIC8+XG4gICAgICA8ZWwtdGFibGUtY29sdW1uIHByb3A9XCJuYW1lXCIgbGFiZWw9XCJOYW1lXCIgd2lkdGg9XCIyMDBcIiAvPlxuICAgICAgPGVsLXRhYmxlLWNvbHVtbiBwcm9wPVwiYWRkcmVzc1wiIGxhYmVsPVwiQWRkcmVzc1wiIHdpZHRoPVwiNDAwXCIgLz5cbiAgICA8L2VsLXRhYmxlPlxuPC90ZW1wbGF0ZT5cblxuPHNjcmlwdCBsYW5nPVwidHNcIiBzZXR1cD5cblxuPC9zY3JpcHQ+XG5cbjxzdHlsZT5cbmh0bWwsIGJvZHkge1xuICB3aWR0aDogMTAwdnc7XG4gIGhlaWdodDogMTAwdmg7XG4gIG1hcmdpbjogMDtcbn1cbiNhcHAge1xuICBoZWlnaHQ6IDEwMCU7XG4gIGRpc3BsYXk6IGZsZXg7XG4gIGFsaWduLWl0ZW1zOiBjZW50ZXI7XG4gIGp1c3RpZnktY29udGVudDogY2VudGVyO1xuICBmbGV4LWRpcmVjdGlvbjogY29sdW1uO1xufVxuPC9zdHlsZT5cblxuXG5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcImltcG9ydHNcIjp7fX0ifQ==).

### DIFF
--- a/packages/components/table/src/table/style-helper.ts
+++ b/packages/components/table/src/table/style-helper.ts
@@ -43,7 +43,10 @@ function useStyle<T>(
   })
   const isGroup = ref(false)
   const scrollbarViewStyle = {
-    display: 'inline-block',
+    //display: 'inline-block',
+    //verticalAlign: 'middle',
+    display: 'inline-flex',
+    flexDirection: 'column',
   }
 
   watchEffect(() => {
--- a/packages/components/table/src/table/style-helper.ts
+++ b/packages/components/table/src/table/style-helper.ts
@@ -43,10 +43,8 @@ function useStyle<T>(
   })
   const isGroup = ref(false)
   const scrollbarViewStyle = {
-    //display: 'inline-block',
-    //verticalAlign: 'middle',
-    display: 'inline-flex',
-    flexDirection: 'column',
+    display: 'inline-block',
+    verticalAlign: 'middle',
   }
 
   watchEffect(() => {
--- a/packages/components/table/src/table/style-helper.ts
+++ b/packages/components/table/src/table/style-helper.ts
@@ -43,8 +43,7 @@ function useStyle<T>(
   })
   const isGroup = ref(false)
   const scrollbarViewStyle = {
-    display: 'inline-flex',
-    flexDirection: 'column',
+    display: 'inline-block',
   }
 
   watchEffect(() => {
@@ -298,7 +297,7 @@ function useStyle<T>(
       height = `calc(100% - ${layout.appendHeight.value}px)`
     }
     return {
-      width: bodyWidth.value,
+      width: `${resizeState.value.width}px`,
       height,
     }
   })
--- a/packages/theme-chalk/src/table.scss
+++ b/packages/theme-chalk/src/table.scss
@@ -42,6 +42,8 @@
 
   // when data is empty
   @include e(empty-block) {
+    position: sticky;
+    left: 0;
     min-height: 60px;
     text-align: center;
     width: 100%;


JeremyWuuuuu:
@plainheart  looks like you didn't update the tests.

plainheart:
I'm waiting for the response about if this behavior is by design, thus no changes will be applied to the test cases. And if this change is accepted, this broken test case can be removed in fact.

msidolphin:
Could you please add a demo for this🙏? so that we can discuss with the designer

plainheart:
@msidolphin Thank you for your response. There is a simple playground url for this behavior in the bottom of the PR description.

msidolphin:
> @msidolphin Thank you for your response. There is a simple playground url for this behavior in the bottom of the PR description.

I want to verify the changes in the document

plainheart:
@msidolphin How do you hope where the demo should be added?

msidolphin:
> @msidolphin What do you think where the demo should be added?

No need to add it, I found a problem, when the cursor is in the empty area, it can't trigger horizontal scrolling. my suggestion is to implement it with sticky,  empty area is still in the scroll-view

plainheart:
Good catch!

plainheart:
@msidolphin Fixed that scroll event can't be triggered on the empty block. Now it's ready for your one more review.

packages/components/table/src/table/style-helper.ts
```diff
@@ -43,8 +43,7 @@ function useStyle<T>(
   })
   const isGroup = ref(false)
   const scrollbarViewStyle = {
-    display: 'inline-flex',
-    flexDirection: 'column',
+    display: 'inline-block',
```
msidolphin, line 46: Please do not change here, inline-block will make the wrong height of scroll-view

plainheart: This is to fix an unexpected gap when using `inline-flex` & `flex-direction: column`. If you have some better idea to fix the gap, then we can revert this change.

msidolphin: > This is to fix an unexpected gap when using `inline-flex` & `flex-direction: column`. If you have some better idea to fix the gap, then we can revert this change.
> 
> 

It seems to have solved the problem 👍

msidolphin (approved):

plainheart:
@msidolphin Could you take some time to take a look at the changes in #6694? In our projects, we usually define many columns and use the `min-width` to ensure it has a proper width to show the content. Unfortunately, this fix didn't solve such a case so the final result is still not perfect. And I saw another developer reported the same concern as well. Thanks a lot!

PR #6683, opened 2022-03-16
perf(components): [input-number] components review

chenxch:

### DIFF
--- a/packages/components/input-number/src/input-number.vue
+++ b/packages/components/input-number/src/input-number.vue
@@ -111,10 +111,10 @@ export default defineComponent({
     const ns = useNamespace('input-number')
 
     const minDisabled = computed(
-      () => handleCompute(props.modelValue) < props.min
+      () => ensurePrecision(props.modelValue, -1) < props.min
     )
     const maxDisabled = computed(
-      () => handleCompute(props.modelValue, true) > props.max
+      () => ensurePrecision(props.modelValue) > props.max
     )
 
     const numPrecision = computed(() => {
@@ -165,22 +165,22 @@ export default defineComponent({
       }
       return precision
     }
-    const handleCompute = (val: number, isIncrease?: boolean) => {
+    const ensurePrecision = (val: number, coefficient: 0 | 1 | -1 = 1) => {
       if (!isNumber(val)) return data.currentValue
       // Solve the accuracy problem of JS decimal calculation by converting the value to integer.
       val = isNumber(val) ? val : Number.NaN
-      return toPrecision(val + props.step * (isIncrease ? 1 : -1))
+      return toPrecision(val + props.step * coefficient)
     }
     const increase = () => {
       if (inputNumberDisabled.value || maxDisabled.value) return
       const value = props.modelValue || 0
-      const newVal = handleCompute(value, true)
+      const newVal = ensurePrecision(value)
       setCurrentValue(newVal)
     }
     const decrease = () => {
       if (inputNumberDisabled.value || minDisabled.value) return
       const value = props.modelValue || 0
-      const newVal = handleCompute(value)
+      const newVal = ensurePrecision(value, -1)
       setCurrentValue(newVal)
     }
     const verifyValue = (
@@ -198,13 +198,8 @@ export default defineComponent({
         if (props.precision !== undefined) {
           newVal = toPrecision(newVal, props.precision)
         }
-
-        if (newVal > props.max) {
-          newVal = props.max
-          update && emit('update:modelValue', newVal)
-        }
-        if (newVal < props.min) {
-          newVal = props.min
+        if (newVal > props.max || newVal < props.min) {
+          newVal = newVal > props.max ? props.max : props.min
           update && emit('update:modelValue', newVal)
         }
       }
--- a/packages/components/input-number/src/input-number.vue
+++ b/packages/components/input-number/src/input-number.vue
@@ -187,19 +187,20 @@ export default defineComponent({
       value: number | string | undefined,
       update?: boolean
     ): number | undefined => {
+      const { max, min, step, precision, stepStrictly } = props
       let newVal = Number(value)
       if (value === null) {
         newVal = Number.NaN
       }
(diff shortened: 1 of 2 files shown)


packages/components/input-number/src/input-number.vue
```diff
@@ -161,54 +165,55 @@ export default defineComponent({
       }
       return precision
     }
-    const _increase = (val: number) => {
+    const handleCompute = (val: number, isIncrease?: boolean) => {
```
JeremyWuuuuu, line 168: Instead of passing `isIncrease` can we just pass coefficient to the method? Something like:

```ts
const handleCompute = (val: number, coefficient: 1 | -1 = 1) => {
   if (!isNumber(val)) return data.currentValue
   // Solve the accuracy problem of JS decimal calculation by converting the value to integer.
   val = isNumber(val) ? val : Number.NaN
   return toPrecision(val + props.step * coefficient)
}
```

packages/components/input-number/src/input-number.vue
```diff
@@ -161,54 +165,55 @@ export default defineComponent({
...
+        if (props.stepStrictly) {
+          newVal = Math.round(newVal / props.step) * props.step
+        }
+        if (props.precision !== undefined) {
+          newVal = toPrecision(newVal, props.precision)
+        }
+
+        if (newVal > props.max) {
+          newVal = props.max
+          update && emit('update:modelValue', newVal)
```
JeremyWuuuuu, line 204: Line#204 and Line#208 is duplicated

packages/components/input-number/src/input-number.vue
```diff
@@ -161,54 +165,55 @@ export default defineComponent({
...
+    ) => {
+      let newVal = Number(value)
+      if (value === null) {
+        newVal = Number.NaN
       }
-      if (typeof newVal === 'number' && props.precision !== undefined) {
-        newVal = toPrecision(newVal, props.precision)
+      if (!Number.isNaN(newVal)) {
+        if (props.stepStrictly) {
+          newVal = Math.round(newVal / props.step) * props.step
```
JeremyWuuuuu, line 196: We can reuse the encapsulated method here as well

packages/components/input-number/src/input-number.vue
JeremyWuuuuu: we can rename this method to something like: `withPrecision` or `ensurePrecision`

jw-foss (commented):
@chenxch Excellent job! Left a couple comments for you to consider 🏓

packages/components/input-number/src/input-number.vue
```diff
@@ -161,54 +165,50 @@ export default defineComponent({
...
-      if (!isNumber(val)) return data.currentValue
-      const precisionFactor = 10 ** numPrecision.value
-      // Solve the accuracy problem of JS decimal calculation by converting the value to integer.
-      val = isNumber(val) ? val : Number.NaN
-      return toPrecision(
-        (precisionFactor * val + precisionFactor * props.step) / precisionFactor
-      )
-    }
-    const _decrease = (val: number) => {
+    const ensurePrecision = (val: number, coefficient: 0 | 1 | -1 = 1) => {
```
JeremyWuuuuu, line 168: **Question:**  `0` was never used, then why are we adding it in to the type of `coefficient`

packages/components/input-number/src/input-number.vue
```diff
@@ -161,54 +165,50 @@ export default defineComponent({
...
-        const stepPrecision = getPrecision(props.step)
-        const precisionFactor = 10 ** stepPrecision
-        newVal =
-          (Math.round(newVal / props.step) * precisionFactor * props.step) /
-          precisionFactor
+    const verifyValue = (
+      value: number | string | undefined,
+      update?: boolean
+    ) => {
+      let newVal = Number(value)
```
JeremyWuuuuu, line 190: **Suggestion:**

```ts
const { max, min, step, precision } = props // move then at the top of this method to avoid duplication 
```

packages/components/input-number/src/input-number.vue
```diff
@@ -161,54 +165,50 @@ export default defineComponent({
...
+      if (value === null) {
+        newVal = Number.NaN
       }
-      if (typeof newVal === 'number' && props.precision !== undefined) {
-        newVal = toPrecision(newVal, props.precision)
+      if (!Number.isNaN(newVal)) {
+        if (props.stepStrictly) {
+          newVal = Math.round(newVal / props.step) * props.step
+        }
+        if (props.precision !== undefined) {
```
JeremyWuuuuu, line 198: **Suggestion**

```ts
import { isUndefined } from '@element-plus/utils'

if (!isUndefined(precision)) {
  // do stuff
}
```

packages/components/input-number/src/input-number.vue
```diff
@@ -161,54 +165,50 @@ export default defineComponent({
...
-      if (newVal !== undefined && newVal <= props.min) newVal = props.min
+      return newVal
+    }
+    const setCurrentValue = (value: number | string | undefined) => {
+      const oldVal = data.currentValue
+      const newVal = verifyValue(value)
       if (oldVal === newVal) return
-      if (!isNumber(newVal)) {
-        newVal = undefined
-      }
```
JeremyWuuuuu, line 211: **Question:** where did this part go?

jw-foss (commented):
Couple questions, back to you🏓

packages/components/input-number/src/input-number.vue
JeremyWuuuuu: Would you mind resolving this?

packages/components/input-number/src/input-number.vue
JeremyWuuuuu: If we do what I suggested, we wouldn't need to use `newVal = newVal > props.max ? props.max : props.min` it would be much shorter.

chenxch:
@JeremyWuuuuu I'm very sorry, but these were restored in the last commit.

jw-foss (approved):

PR #6703, opened 2022-03-18
refactor(components): [empty] refactor

buqiyuan:

### DIFF
--- a/packages/components/empty/src/empty.vue
+++ b/packages/components/empty/src/empty.vue
@@ -16,37 +16,26 @@
   </div>
 </template>
 
-<script lang="ts">
-import { computed, defineComponent } from 'vue'
+<script lang="ts" setup>
+import { computed } from 'vue'
 import { useLocale, useNamespace } from '@element-plus/hooks'
 import ImgEmpty from './img-empty.vue'
 import { emptyProps } from './empty'
 
 import type { CSSProperties } from 'vue'
 
-export default defineComponent({
+defineOptions({
   name: 'ElEmpty',
-  components: {
-    ImgEmpty,
-  },
-
-  props: emptyProps,
-
-  setup(props) {
-    const { t } = useLocale()
-    const ns = useNamespace('empty')
-    const emptyDescription = computed(
-      () => props.description || t('el.table.emptyText')
-    )
-    const imageStyle = computed<CSSProperties>(() => ({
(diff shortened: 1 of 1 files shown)


sxzz (commented):
Also `packages/components/empty/src/img-empty.vue`

sxzz:
https://github.com/element-plus/element-plus/blob/5ee184fdfdcbb5c3d69725b35499c062317809e8/packages/components/empty/src/img-empty.vue#L123

Does anyone know what the `id` property is for?

/cc @JeremyWuuuuu

sxzz (commented):
I see, maybe we can remove `id` and replace it with random number.

JeremyWuuuuu:
> I see, maybe we can remove `id` and replace it with random number.

I'm not following, where is the `id` you've been talking about?

buqiyuan:
> > I see, maybe we can remove `id` and replace it with random number.
> 
> I'm not following, where is the `id` you've been talking about?

It's [here](https://github.com/element-plus/element-plus/blob/aeee7339f4d7c115aaef02734bf32ab044eeca38/packages/components/empty/src/img-empty.vue#L123)

JeremyWuuuuu:
@buqiyuan @sxzz https://github.com/element-plus/element-plus/pull/1534 here are the reason why this is used.

buqiyuan:
> @buqiyuan @sxzz #1534 here are the reason why this is used.

I saw it, but I think the self-incrementing id here has the same meaning as the random number

buqiyuan:
The reason for using random number is that the `script-setup` syntax does not cache variables

JeremyWuuuuu:
@buqiyuan Random number will disrupt the server side rendering hydration, because each time you generates different id, the vnode will be treated as different which causes hydration mismatch.

JeremyWuuuuu:
The id must be consistent from both server side and client side to complete hydration without error.

jw-foss (changes requested):
We should keep the id consistent, no random id.

buqiyuan:
@JeremyWuuuuu ok, maybe you are right then i revert this change

jw-foss (approved):

JeremyWuuuuu:
LGTM. /cc @sxzz

sxzz (approved):

PR #6717, opened 2022-03-19
chore(components): [table] filter out invalid className

nieyuyao:
Filter out invalid className to avoid unnecessary spaces.

---

### DIFF
--- a/packages/components/table/src/table-body/styles-helper.ts
+++ b/packages/components/table/src/table-body/styles-helper.ts
@@ -98,7 +98,7 @@ function useStyles<T>(props: Partial<TableBodyProps<T>>) {
       )
     }
     classes.push(ns.e('cell'))
-    return classes.filter((className) => !!className).join(' ')
+    return classes.filter((className) => Boolean(className)).join(' ')
   }
   const getSpan = (
     row: T,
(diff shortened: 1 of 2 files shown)


packages/components/table/src/table-body/styles-helper.ts
```diff
@@ -98,7 +98,7 @@ function useStyles<T>(props: Partial<TableBodyProps<T>>) {
       )
     }
     classes.push(ns.e('cell'))
-    return classes.join(' ')
+    return classes.filter((className) => !!className).join(' ')
```
JeremyWuuuuu, line 101: **Suggestion:** 

use `Boolean(val)` instead of `!!val` for converting here.

packages/components/table/src/table-header/style.helper.ts
```diff
@@ -109,7 +109,7 @@ function useStyle<T>(props: TableHeaderProps<T>) {
 
     classes.push(ns.e('cell'))
 
-    return classes.join(' ')
+    return classes.filter((className) => !!className).join(' ')
```
JeremyWuuuuu, line 112: Same here.

jw-foss (commented):
@nieyuyao Left a comment for you to consider, thanks!

packages/components/table/src/table-header/style.helper.ts
nieyuyao: done

packages/components/table/src/table-body/styles-helper.ts
nieyuyao: done

jw-foss (approved):

JeremyWuuuuu:
@nieyuyao Actually, I think this could be simplified as 

```ts
classNames.filter(Boolean).join(' ')
```

PR #6782, opened 2022-03-23
refactor(theme-chalk): remove default size class & simplify button css

### LINKED ISSUE #5204: [Feature Request] [button] type 和 size 的默认值都是defalut，导致type不是default的时候class中还会存在el-button-default
### Existing Component
Yes

### Component Name
ElButton

### Description
button type 和 size 的默认值都是defalut，导致type不是default的时候class中还会存在el-button-default
项目需要动态修改按钮颜色，当使用css覆盖样式的时候发现 当按钮type不是default的时候还是存在el-button-defalut，是因为没有设置size导致颜色不好覆盖。
因为button的颜色变量是在style里，最好官方提供一套动态修改主题的方案
https://codepen.io/bell-1/pen/bGojrbx

YunYouJun:
close #5204

- remove `default` size class
- simplify button css & use `getCssVar`
- fix test canused by removing default size

---


YunYouJun:
I don't seem to see an advantage why we changed from '' to 'default'.
But changing '' to 'default' caused a useless class `--default`.

sxzz:
My point of concern is BC, can we have a way to be compatible with both `''` and `'default'`?
Otherwise we will need to wait until version `2.2`.

YunYouJun:
In fact, I didn't remove the `default` in `componentSize`.

And I don't think it's causing BC because user change or not doesn't make a difference from before.
The user's API and results didn't change.

JeremyWuuuuu:
Can we keep `default` in the code? I don't think it will matter about how we implement the code since the default case is default state and the user won't manually set the `default` state to `default` state. The only thing that could matter is that they might be using some `scss` or `css` overriding the default style, changing this should be considered breaking.

jw-foss (approved):

JeremyWuuuuu:
It made sense to me, let's see if @sxzz has any questions at all.

packages/theme-chalk/src/tag.scss
```diff
@@ -133,7 +134,7 @@ $tag-icon-span-gap: map.merge(
     padding-right: map.get($tag-icon-span-gap, 'default') - $border-width;
   }
 
-  @each $size in (large, default, small) {
```
YunYouJun, line 136: Remove tag default is safe.

Its default value has always existed in the style above.

packages/components/form/src/form.vue
```diff
@@ -37,7 +37,9 @@ const formClasses = computed(() => {
   const { labelPosition, inline } = props
   return [
     ns.b(),
-    ns.m(formSize.value),
+    // todo: in v2.2.0, we can remove default
+    // in fact, remove it doesn't affect the final style
+    ns.m(formSize.value || 'default'),
```
YunYouJun, line 42: Add `default` to prevent the form from changing, in fact it doesn't affect the style at all.
We can remote it when `2.2` or not.

packages/theme-chalk/src/button.scss
```diff
@@ -38,7 +38,6 @@ $button-icon-span-gap: map.merge(
   text-align: center;
   box-sizing: border-box;
   outline: none;
-  margin: 0;
```
YunYouJun, line 44: Browser user agent default is margin 0. (Ant design doesn't have it either.)
If we have it, we can not use `mt-2` to change default margin.

JeremyWuuuuu:
@YunYouJun would you mind force push one more time? Looks like the downgrade of GitHub had some influence on your PR.

sxzz (approved):

Issue #6739, opened 2022-03-21
[Bug Report] [Component] [cascader] Cascader Unexpected display 'No Data'


zhangalexhy:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.26`
- Element Plus Version: `2.0.4`
- Browser / OS: `windows 10`
- Build Tool: `Webpack`

### Related Component

- `el-cascader`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtY2FzY2FkZXIgOnByb3BzPVwicHJvcHNcIiA6b3B0aW9ucz1cIm9wdGlvbnNcIiB2LW1vZGVsPVwidmFsdWVcIi8+XG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IGxhbmc9XCJ0c1wiIHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSAndnVlJzsgXG4gIFxuY29uc3Qgb3B0aW9ucyA9IFt7XG4gIGxhYmVsOicxJyxcbiAgdmFsdWU6JzEnXG59XVxuY29uc3QgdmFsdWUgPSByZWYoJzEnKTtcbmNvbnN0IHByb3BzID0ge1xuICBsYXp5OiB0cnVlLFxuICBsYXp5TG9hZChub2RlLCByZXNvbHZlKSB7XG4gICAgc2V0VGltZW91dCgoKSA9PiB7XG4gICAgICByZXNvbHZlKFt7XG4gIFx0XHRcdGxhYmVsOicyJyxcbiAgXHRcdFx0dmFsdWU6JzInXG5cdFx0XHR9XSlcbiAgICB9LCAyMDApXG4gIH0sXG4gIGNoZWNrU3RyaWN0bHk6IHRydWUsXG4gIGVtaXRQYXRoOiBmYWxzZSxcbn1cbjwvc2NyaXB0PiIsImltcG9ydC1tYXAuanNvbiI6IntcImltcG9ydHNcIjp7fX0ifQ==)

### Steps to reproduce

点击组件
click the cascader box

### What is Expected?

二级菜单在有数据时不显示No Data
the second level box doesn't have 'No Data'

### What is actually happening?

[...]

PR #6815, opened 2022-03-24
fix: Cascader Unexpected display 'No Data'

chenxch:

### DIFF
--- a/packages/components/cascader-panel/src/index.vue
+++ b/packages/components/cascader-panel/src/index.vue
@@ -8,7 +8,7 @@
       :key="index"
       :ref="(item) => (menuList[index] = item)"
       :index="index"
-      :nodes="menu"
+      :nodes="[...menu]"
     />
   </div>
 </template>


chenxch:
fix #6739

JeremyWuuuuu:
@chenxch Would you mind elaborate a little bit about why did you do changes like this? Just some thoughts would be good! Thanks!

chenxch:
You can also see in the discussion on the issue that I have located that the update of `props.nodes` will not trigger `computed`, and traced up to find that adding nodes is added through the `appendChild` method of the `class store`, but I have never known why this would cause a two-way Binding's dependency collection fails, so nodes are passed through a destructuring method.

JeremyWuuuuu:
@chenxch I see, seems like shallow watching issue related, since the `nodes` referencing object did not change it was considered as **untouched** so that the `computed` did not get re-triggered.

jw-foss (approved):

JeremyWuuuuu:
As as temporary fix I think the change looks good, approving, BTW, if we can add some tests for ensuring this won't happen again it'd be just great, you can do that in an additional PR if you'd like that.

chenxch:
I am also trying to read and understand the original design of this component, trying to solve more problems, and from the issue point of view, there are still some problems with this component.

PR #6928, opened 2022-03-31
fix(components): [affix] remove root width (#6921)

huanghong1125:
remove affix rootStyle width

### DIFF
--- a/packages/components/affix/src/affix.vue
+++ b/packages/components/affix/src/affix.vue
@@ -48,7 +48,6 @@ const transform = ref(0)
 const rootStyle = computed<CSSProperties>(() => {
   return {
     height: fixed.value ? `${rootHeight.value}px` : '',
-    width: fixed.value ? `${rootWidth.value}px` : '',
   }
 })
 
--- a/packages/components/affix/src/affix.vue
+++ b/packages/components/affix/src/affix.vue
@@ -103,13 +103,11 @@ const handleScroll = () => {
     fixed: fixed.value,
   })
 }
-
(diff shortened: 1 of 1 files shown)


JeremyWuuuuu:
@huanghong1125 Shouldn't we add resize observer instead of removing the width?

huanghong1125:
It is cannot be observed, because the width is set.

JeremyWuuuuu:
@huanghong1125 We can observe the `clientWidth` like this:

```ts
document.addEventListener('resize', () => {
  update()
})
```

Not watching the element it self.

huanghong1125:
@JeremyWuuuuu The affixStyle width is observe rootWidth. Now root width is auto. 
```
const affixStyle = computed<CSSProperties>(() => {
  if (!fixed.value) return {}

  const offset = props.offset ? `${props.offset}px` : 0
  return {
    height: `${rootHeight.value}px`,
    width: `${rootWidth.value}px`,
    top: props.position === 'top' ? offset : '',
    bottom: props.position === 'bottom' ? offset : '',
    transform: transform.value ? `translateY(${transform.value}px)` : '',
    zIndex: props.zIndex,
  }
})
```

huanghong1125:
@JeremyWuuuuu you can review it.
When affix is fixed, the root element is just take a place.
Because root width is automatic, resize change may affect the height of the content. root height is computed by content height.

huanghong1125:
@JeremyWuuuuu 问题我仔细思考了，也好好的去修复了，你可以看看最终代码还有没有问题

PR #7450, opened 2022-04-30
feat(components): [form-item] auto attach labels to inputs

opengraphica:

Basic Concept:
1. FormItemContext now has `isGroup`, `labelId`, `inputIds`, `addInputId()`, `removeInputId()`.
2. It is an input component's responsibility to call `addInputId()`/`removeInputId()` on mount, unmount, or ID update. Nested inputs (such as a `checkbox` inside a `checkbox-group`) should not call these functions.
3. `form-item` will modify `inputIds` when `addInputId()`/`removeInputId()` is called, and will handle label attachment in the DOM accordingly.

[form-item] label vs group
- If a single input is detected or `label-for` prop is provided, same as today, el-form-item__label will be a `label` element
- Otherwise (assume multiple inputs), el-form-item is `role="group"` and el-form-item__label is a `div` with an `id` referenced by the group's `aria-labelledby`
- An `id` for the label is **always** generated in case it is needed by whatever input is inside the `form-item` (such as checkbox-group aria-labelledby).
- If no label is provided at all for the item, it implements no accessibility roles.

[autocomplete]
- No change. Refer to `input`.

[cascader]
- No change. Refer to `input`.

[checkbox]
- **Warning:** Undocumented property `id` now is added to the `<input>` instead of the `<label>`
- If `label` prop or default slot content is omitted, generates id and attaches to `form-item` label
- Multiple labels aren't allowed, so need to change `<label>` to `<span>` if no label is provided and handle click event manually

[checkbox-button]
- Removed `role` and `aria-` attributes from label, this is not a focusable element and causes screen reader confusion. Checkbox element is already there for screen reader.
- No other changes as this is expected to be used with `checkbox-group`.

[checkbox-group]
- Added `label` prop to control `aria-label`
- If `label` not specified, and there's no other inputs in the `form-item`, `aria-labelledby` is used instead to reference the `form-item` label.

[color-picker]
- Added `id` prop and `useFormItemInputId` hook.
- Added `role="button"` and appropriate aria attributes. More will need to be done later for this component to make it accessible.

[date-picker]
- No change. Refer to `input`.

[input]
- Added `id` prop and `useFormItemInputId` hook.
- Fix missing tabindex/label props referenced in template.

[input-number]
- Added aria labels to buttons.
- Don't add aria-valuemin or aria-valuemax when values are infinite.

[popper]
- Unset the formContextKey/formItemContextKey injection. In examples like color/date picker, inputs inside the popover content should not be counted as form item fields.

[radio]
- Removed role and aria attributes from the label, instead allow screen reader to focus on the actual radio input. Error: "Using a negative tabindex on an element inside an interactive control does not prevent assistive technologies from focusing the element (even with 'aria-hidden=true')"

[radio-group]
- Added `label` prop to control `aria-label`
- If `label` not specified, and there's no other inputs in the `form-item`, `aria-labelledby` is used instead to reference the `form-item` label.

[rate]
- Added `id` prop and `useFormItemInputId` hook.
- Remove aria-valuetext attribute when empty string

[slider]
- Added `id` prop and `useFormItemInputId` hook.

[switch]
- Moved aria attributes to the input. Error: "Interactive controls must not be nested"
- Added `id` prop and `useFormItemInputId` hook.

[time-picker]
- No change. Refer to `input`.

[time-select]
- No change. Refer to `input`.

[tree-select]
- No change. Refer to `input`.


JeremyWuuuuu:
@opengraphica sorry for the late update, seems like we are having this conflict in your PR, would you mind resolving them?

packages/components/color-picker/src/index.vue
```diff
@@ -194,6 +224,14 @@ export default defineComponent({
     const currentColor = computed(() => {
       return !props.modelValue && !showPanelColor.value ? '' : color.value
     })
+    const buttonAriaLabel = computed<string>(() => {
```
JeremyWuuuuu, line 227: For correctness:
```suggestion
    const buttonAriaLabel = computed<string | undefined>(() => {
```

packages/components/color-picker/src/index.vue
```diff
@@ -194,6 +224,14 @@ export default defineComponent({
...
+      return !isLabeledByFormItem.value
+        ? props.label || t('el.colorpicker.defaultLabel')
+        : undefined
+    })
+    const buttonAriaLabelledby = computed<string>(() => {
+      return isLabeledByFormItem.value ? elFormItem.labelId : undefined
+    })
```
JeremyWuuuuu, line 234: Same here.

packages/components/form/src/form-item.vue
```diff
@@ -1,19 +1,26 @@
...
       :is-auto-width="labelStyle.width === 'auto'"
       :update-all="formContext?.labelWidth === 'auto'"
     >
-      <label
-        v-if="label || $slots.label"
+      <component
+        :is="labelFor ? 'label' : 'div'"
+        v-if="hasLabel"
+        :id="labelId"
         :for="labelFor"
```
JeremyWuuuuu, line 16: **Question:** would this `:for` attribute be redundant when the tag is actually `div` instead of `label`?

packages/components/form/src/form-item.vue
```diff
@@ -303,6 +325,16 @@ const resetField: FormItemContext['resetField'] = async () => {
...
 }
 
+const addInputId: FormItemContext['addInputId'] = (id: string) => {
+  if (!inputIds.value.includes(id)) {
+    inputIds.value.push(id)
+  }
+}
+
+const removeInputId: FormItemContext['addInputId'] = (id: string) => {
+  inputIds.value = inputIds.value.filter((listId) => listId !== id)
```
JeremyWuuuuu, line 335: **Praise:** clever!

jw-foss (commented):
I just left a few comments for you to consider 🏓, overall, this PR is just amazing to me 🤩.

Thank you for your hard working 👍.

packages/components/autocomplete/src/autocomplete.ts
```diff
@@ -72,6 +72,12 @@ export const autocompleteProps = buildProps({
     type: Boolean,
     default: undefined,
   },
+  label: {
+    type: String,
+  },
+  tabindex: {
+    type: [Number, String],
+  },
```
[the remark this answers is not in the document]
Giwayume: I removed this, misunderstood how useAttrs() was working.

packages/components/form/src/form-item.vue
opengraphica: Yes, in this case the "labelFor" computed returns undefined which removes the attribute.

packages/components/autocomplete/src/autocomplete.ts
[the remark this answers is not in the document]
opengraphica: I removed this, misunderstood how useAttrs() was working.

jw-foss (approved):

PR #7509, opened 2022-05-05
fix(component): [Date-pick] component edit value problem

zgh035:
Hello, this problem  component is el-date-picker(this attribute type is daterange);
when i had selected the date and it's show format type is 'yyyy-MM-dd',then i change the date by hand, eg:2020-01-01 -> 2020-01.01,yeah, it's a bit strange, It then returns a value that becomes an array ,that's not what I want,is should be a string.finally, i have made the following changes.


xiaoxian521:

PR #7559, opened 2022-05-07
fix(components):[select] placeholder local

gjfei:


packages/components/select/src/select.vue
```diff
@@ -513,16 +514,25 @@ export default defineComponent({
...
+          Array.isArray(props.modelValue) &&
+          props.modelValue.length > 0
+        ) {
+          currentPlaceholder.value = ''
+        }
+      },
+      {
+        immediate: true,
       }
+    )
```
JeremyWuuuuu, line 533: **Suggestion:** put this chunk of code inside `onMounted`

jw-foss (changes requested):
Left a comment for you to consider 🙇

packages/components/select/src/select.vue
gjfei: > **Suggestion:** put this chunk of code inside `onMounted`

可以告诉我原因吗？

JeremyWuuuuu: No specific reason, we should keep the change minimal. 

If you think the watch statement is ok to be put outside `onMounted` hook, you can make another PR for that.

Does this make any sense?

gjfei: 我好像不是很能理解watch放在onMounted里面，如果是用工具去对比变化的话，并没有过多的更改
image

JeremyWuuuuu: If you insist, would you mind adding a test case against your change?

PR #7571, opened 2022-05-09
feat(components): [select] add value-display attr

gezhigang1005:
下面是列举了一些其他组件库的类似功能点。

- `tdesign`[valueDisplay](https://tdesign.tencent.com/vue-next/components/select?tab=api)
-  `Vue-Treeselect`[ value-label](https://vue-treeselect.js.org/)
-  `Ant Design Vue`[tagRender](https://www.antdv.com/components/select#API)

### DIFF
--- a/docs/en-US/component/select.md
+++ b/docs/en-US/component/select.md
@@ -146,6 +146,7 @@ If the binding value of Select is an object, make sure to assign `value-key` as 
 | fit-input-width                   | whether the width of the dropdown is the same as the input                                                                  | boolean                            | —                           | false            |
 | suffix-icon                       | Custom suffix icon component                                                                                                | string / Component                 | —                           | ArrowUp          |
 | tag-type                          | tag type                                                                                                                    | string                             | success/info/warning/danger | info             |
+| value-display                     | Customize the presentation mode of selected items                                                                           | function                           | -                           | -                |
 
 ## Select Events
 
--- a/docs/en-US/component/select.md
+++ b/docs/en-US/component/select.md
@@ -146,7 +146,7 @@ If the binding value of Select is an object, make sure to assign `value-key` as 
 | fit-input-width                   | whether the width of the dropdown is the same as the input                                                                  | boolean                            | —                           | false            |
 | suffix-icon                       | Custom suffix icon component                                                                                                | string / Component                 | —                           | ArrowUp          |
 | tag-type                          | tag type                                                                                                                    | string                             | success/info/warning/danger | info             |
-| value-display                     | Customize the presentation mode of selected items                                                                           | function                           | -                           | -                |
+| format-selection-label                     | Customize the presentation mode of selected items                                                                           | function                           | -                           | -                |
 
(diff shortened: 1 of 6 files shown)


Giwayume:
A few thoughs,

There is a precedent for calling these kinds of functions `format-x`, e.g. named `format-label` instead of `label-display`.
- [slider "format-tooltip"](https://element-plus.org/en-US/component/slider.html#attributes)
- [time-picker "format"](https://element-plus.org/en-US/component/time-picker.html#attributes)
- [transfer "format"](https://element-plus.org/en-US/component/transfer.html#attributes)
- [progress "format](https://element-plus.org/en-US/component/progress.html#attributes)
- [table "formatter](https://element-plus.org/en-US/component/table.html#table-column-attributes)
- [date-picker "value-format"](https://element-plus.org/en-US/component/date-picker.html#attributes)
- [input "formatter"](https://element-plus.org/en-US/component/input.html#input-attributes)

Is handling in the framework necessary? The same thing could be done on the consuming side by modifying the options before passing into the select. This would be more performant for any use case involving select v1, and takes less overall code to achieve.

```
let options = [
    {
        value: '选项1',
        label: '黄金糕'
    },
    {
        value: '选项2',
        label: '双皮奶'
    }
]
options = options.map(option => ({ value: option.value, label: `${option.label}123` }))
```
A case could probably be made for having this attribute with the virtualized select. 
Although, [even mapping 1 million items beforehand](https://jsfiddle.net/3msuk845/) might not take as long as you'd expect.

gezhigang1005:
@Giwayume Thanks, there is indeed a problem with the property name

```JavaScript
let options = [
    {
        value: '选项1',
        label: '黄金糕'
    },
    {
        value: '选项2',
        label: '双皮奶'
    }
]
options = options.map(option => ({ value: option.value, label: `${option.label}123` }))
```
But in this way, the drop-down panel has also changed. not the desired effect


- The role of format-label should be like this
image

Giwayume:
Ok, I see, I misunderstood the intent of the change.

In that case I would name it something along the lines of `format-selection-label`, otherwise `format-label` could be ambiguous as to what label is impacted.

gezhigang1005:
@Giwayume The name has been changed to `format-selection-label`

Giwayume:
Can you add unit a test for `tree-select` since it is also affected?

gezhigang1005:
The problem of adding `format-selection-label` to `tree-select`

```html

<template>
     <el-tree-select v-model="value" :data="[...]" />
</template>

<script setup lang="ts">
import { ref } form 'vue'

const value = ref(1)

// directly change the value
</script>
```

Does not trigger any operation of `tree-select`, so the `node object` cannot be obtained in `format-selection-label`

Do you have any good suggestions, I can't handle it here for the time being

tolking:
Thanks for your contribution.

Now similar slots have been added, you can check the [custom-label](https://element-plus.org/en-US/component/select.html#custom-label)

PR #7592, opened 2022-05-09
fix(components): form select size change with inline style

YunYouJun:
- not add `initialInputHeight` when size same as componentSize to avoid inline style override

https://github.com/chenxch/element-plus/blob/d26f995291964175295141d376462cb708d81d60/packages/components/select/src/useSelect.ts#L179

> https://github.com/element-plus/element-plus/pull/7471

When we watch these, select will always resetInputWidth.
This causes when we switch the form size, the size of the select is overwritten by the inline style.

The solution I thought of, when the size is the same as the default size, don't add the initialInputHeight.

@chenxch Can you have a look?

---

### DIFF
--- a/packages/components/select/src/select.vue
+++ b/packages/components/select/src/select.vue
@@ -263,6 +263,7 @@
 import {
   computed,
   defineComponent,
+  inject,
   nextTick,
   onBeforeUnmount,
   onMounted,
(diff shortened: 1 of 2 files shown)


jw-foss (approved):

chenxch:
I'm here to confirm right away.

chenxch:
```js
input.getBoundingClientRect().height + 2 !== getComponentSize(selectSize.value || elForm.size)
```
If the three heights of 22/30/38 are set, there will be failures.

YunYouJun:
Thanks for the feedback, this approach may not work.
 But now select will not change dynamically with form size, do you have any good way?

chenxch:
Previously, the change of size under `el-config-provider` was fixed. I used the previous demo to adjust it to `el-form`, which is ok.

YunYouJun:
It works in drawer. Because it is reinitialized.

You can try to put it on the home page, at this time dynamic switching fails.

chenxch:
Maybe we should remove `states.initialInputHeight`, the component seems to be working fine now.

YunYouJun:
Do you know what the purpose of `states.initialInputHeight` is? 
I'm afraid it might affect something we don't know about.

YunYouJun:
Would you like to create a pr to try? I'll check again in the preview documentation.

PR #7598, opened 2022-05-09
feat(components): [date-picker] work with modal focus trap; a11y controls and attributes

opengraphica:

**Summary**
- Fixed tabbing issues when datepicker is placed inside dialog with focus trap.
- Focus trap is not implemented for the datepicker popper in this PR. This means the keyboard controls for datepicker popper won't work when placed inside of a dialog. Waiting on #7550 to be merged before implementing this.
- Implemented combobox keyboard controls for the basic datepicker. Some types of the datepicker selection such as range and multi-date are not touched in terms of keyboard in this PR. Future work.

[date-picker]
- Keyboard controls reworked for 'date', 'month', 'year' selection modes. Can now shift+tab inside of popper to get to the top buttons, and arrow key controls only apply when focused on the date selection grid.
- Added home/end/pageUp/pageDown/space keys.
- Keyboard controls for range & multi date selection still left unimplemented as they are today, future work.
- `el-date-picker__prev-btn` and `el-date-picker__next-btn` classes moved to container, because the float right changes the tab order.

[input]
- Added `container-role` prop for controlling role on input container.
- Filter certain aria attrs to container div when using container role `combobox`.

[popper]
- Added support for different types of roles that are not `tooltip`.
- Additional aria attributes added to target and content automatically based on role.

[picker]
- Use role="combobox" on input container.
- When `arrow down` pressed, shows the popper if not already visible, then calls `handleFocusPicker` to notify the picker component that it should bring focus to itself.
- When user triggers popper close it focuses back on the input instead of blurring it.
- Popper is closed automatically when it detects after a certain timeframe that active focus is not inside the input(s) or the popper. Moved the change handling logic away from listening for tab key.

[tooltip]
- Passthrough `role` prop to popper.
- Added function for checking focus inside popper content.


packages/components/date-picker/src/date-picker-com/basic-date-table.vue
```diff
@@ -244,9 +265,24 @@ export default defineComponent({
...
+    watch(
+      () => props.date,
+      async () => {
+        if (tbodyRef.value?.contains(document.activeElement)) {
+          await nextTick()
+          currentCellRef.value?.focus()
+        }
+      },
+      { immediate: true }
+    )
```
JeremyWuuuuu, line 277: **Suggestion:** let's move this section into `onMounted` hook since `tbodyRef` will only be instantiated when `onMounted` is invoked. We can avoid unnecessary initial evaluation.

JeremyWuuuuu: Note that this is not a **Blocking** comment, you may choose to leave it as-is.

jw-foss (approved):
Looking good, left a comment, but I'm approving as-is.

packages/components/date-picker/src/date-picker-com/basic-date-table.vue
Giwayume: I removed immediate which has the same effect, thanks for the catch.

opengraphica: I removed immediate which has the same effect, thanks for the catch.

PR #7946, opened 2022-05-28
feat(components): [dropdown] a11y

### LINKED ISSUE #7624: [Bug Report] [Component] [dropdown, dropdown-item] <el-dropdown> does not meet accessibility requirements
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.31`
- Element Plus Version: `2.1.8`
- Browser / OS: `Chrome 101.0.4951.54`
- Build Tool: `Webpack`

### Related Component

- `el-dropdown`
- `el-dropdown-item`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cCBsYW5nPVwidHNcIj5cbmltcG9ydCB7IHJlZiwgdmVyc2lvbiBhcyB2dWVWZXJzaW9uIH0gZnJvbSAndnVlJ1xuaW1wb3J0IHsgdmVyc2lvbiBhcyBFcFZlcnNpb24gfSBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBBcnJvd0Rvd24gfSBmcm9tICdAZWxlbWVudC1wbHVzL2ljb25zLXZ1ZSdcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxkaXYgY2xhc3M9XCJjb250YWluZXJcIj5cbiAgPGVsLWJ1dHRvbj5DbGljazwvZWwtYnV0dG9uPlxuICAgIDxlbC1kcm9wZG93bj5cbiAgICAgIDxzcGFuIGNsYXNzPVwiZWwtZHJvcGRvd24tbGlua1wiPlxuICAgICAgICBPcHRpb25zIExpc3RcbiAgICAgICAgPGVsLWljb24gY2xhc3M9XCJlbC1pY29uLS1yaWdodFwiPlxuICAgICAgICAgIDxhcnJvdy1kb3duIC8+XG4gICAgICAgIDwvZWwtaWNvbj5cbiAgICAgIDwvc3Bhbj5cbiAgICAgIDx0ZW1wbGF0ZSAjZHJvcGRvd24+XG4gICAgICAgIDxlbC1kcm9wZG93bi1tZW51PlxuICAgICAgICAgIDxlbC1kcm9wZG93bi1pdGVtPk9wdGlvbiAxPC9lbC1kcm9wZG93bi1pdGVtPlxuICAgICAgICAgIDxlbC1kcm9wZG93bi1pdGVtPk9wdGlvbiAyPC9lbC1kcm9wZG93bi1pdGVtPlxuICAgICAgICAgIDxlbC1kcm9wZG93bi1pdGVtPk9wdGlvbiAzPC9lbC1kcm9wZG93bi1pdGVtPlxuICAgICAgICA8L2VsLWRyb3Bkb3duLW1lbnU+XG4gICAgICA8L3RlbXBsYXRlPlxuICAgIDwvZWwtZHJvcGRvd24+XG4gIDwvZGl2PlxuPC90ZW1wbGF0ZT5cblxuPHN0eWxlIHNjb3BlZD5cbi5jb250YWluZXIge1xuICBkaXNwbGF5OiBmbGV4O1xuICBhbGlnbi1pdGVtczogY2VudGVyO1xufVxuLmV4YW1wbGUtc2hvd2Nhc2UgLmVsLWRyb3Bkb3duLWxpbmsge1xuICBjdXJzb3I6IHBvaW50ZXI7XG4gIGNvbG9yOiB2YXIoLS1lbC1jb2xvci1wcmltYXJ5KTtcbiAgZGlzcGxheTogZmxleDtcbiAgYWxpZ24taXRlbXM6IGNlbnRlcjtcbn1cbi5lbC1idXR0b24ge1xuICBtYXJnaW4tcmlnaHQ6IDIwcHg7XG59XG48L3N0eWxlPlxuXG4iLCJpbXBvcnRfbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7fVxufSIsIl9vIjp7fX0=)

### Steps to reproduce

Press 'TAB' key. Only button is accessible using TAB key, whereas the dropdown isn't. You cannot open it using keys, which breaks accessibility.

### What is Expected?

Dropdown should be accessible via TAB + Enter keys
And
Oprions inside the dropdown should be accessible using UP + DOWN keys (i.e. you should be able to choose options using arrow keys).
This is bad accessibility, since dropdown is an interactive element.

Dropdown as well as dropdown options aren't accessibile using keyboard navigation.

### Additional comments

_(empty)_

opengraphica:

Fixes #7624

 https://w3c.github.io/aria-practices/examples/menu-button/menu-button-links.html

[el-dropdown]
- Whatever button/element user provides as dropdown trigger automatically receives "button" role and a tabindex so it can be focused.
- Added "role" prop for controlling dropdown menu role. (not a prop on el-dropdown-menu itself so popper is aware)

[el-dropdown-item]
- Uses "menuitem", "link", or "button" role, depending on the dropdown menu role.

[el-popper]
- Added aria attributes to virtual trigger.
- Added "triggerTargetEl" prop for controlling where the "target" aria attributes point to, if not on the popper content (by default).
- Show/hide events emit the event listener object that caused the show/hide if applicable


packages/components/popper/src/content.vue
```diff
@@ -194,6 +205,44 @@ onMounted(() => {
...
+            !isNil(role)
+              ? el.setAttribute('role', role)
+              : el.removeAttribute('role')
+            !isNil(ariaLabel)
+              ? el.setAttribute('aria-label', ariaLabel)
+              : el.removeAttribute('aria-label')
+            !isNil(ariaModal)
+              ? el.setAttribute('aria-modal', ariaModal)
+              : el.removeAttribute('aria-modal')
+            !isNil(id) ? el.setAttribute('id', id) : el.removeAttribute('id')
```
jw-foss, line 231: I think we can shorten these code into:

```ts
(watches) => {
  ['role', 'aria-label', 'aria-modal', 'id'].forEach((key, idx) => {
      isNil(watches[idx]) ? el.removeAttribute(key) : el.setAttribute(key, watches[idx])
  })
}


```

packages/components/popper/src/content.vue
```diff
@@ -194,6 +205,44 @@ onMounted(() => {
...
+          },
+          { immediate: true }
+        )
+      }
+      if (isElement(prevEl)) {
+        prevEl.removeAttribute('role')
+        prevEl.removeAttribute('aria-label')
+        prevEl.removeAttribute('aria-modal')
+        prevEl.removeAttribute('id')
```
jw-foss, line 240:
```ts
['role', 'aria-label', 'aria-modal', 'id'].forEach((key) => {
  prevEl.removeAttribute(key)
})
```

packages/components/popper/src/trigger.vue
```diff
@@ -96,6 +98,30 @@ onMounted(() => {
...
+              : el.removeAttribute('aria-describedby')
+            !isNil(ariaHaspopup)
+              ? el.setAttribute('aria-haspopup', ariaHaspopup)
+              : el.removeAttribute('aria-haspopup')
+            !isNil(ariaExpanded)
+              ? el.setAttribute('aria-expanded', ariaExpanded)
+              : el.removeAttribute('aria-expanded')
+          },
+          { immediate: true }
+        )
```
jw-foss, line 118:
```ts
(changes) => {
  ['aria-controls', 'aria-describedby', 'aria-haspopup', 'expanded'].forEach((key, idx) => {
      isNil(changes[idx]) ? el.removeAttribute(key) : el.setAttribute(key, changes[idx])
  })
}
```

packages/components/popper/src/trigger.vue
```diff
@@ -96,6 +98,30 @@ onMounted(() => {
...
+      }
+      if (isElement(prevEl)) {
+        prevEl.removeAttribute('aria-controls')
+        prevEl.removeAttribute('aria-describedby')
+        prevEl.removeAttribute('aria-haspopup')
+        prevEl.removeAttribute('aria-expanded')
```
jw-foss, line 124:
```ts
['aria-controls', 'aria-describedby', 'aria-haspopup', 'expanded'].forEach(
  (key) => prevEl.removeAttribute(key)
)
```

packages/components/dropdown/src/dropdown.vue
```diff
@@ -126,11 +138,15 @@ export default defineComponent({
...
     const currentTabId = ref<string | null>(null)
     const isUsingKeyboard = ref(false)
+    const triggerKeys = [EVENT_CODE.enter, EVENT_CODE.space, EVENT_CODE.down]
 
     const wrapStyle = computed<CSSProperties>(() => ({
       maxHeight: addUnit(props.maxHeight),
     }))
     const dropdownTriggerKls = computed(() => [ns.m(dropdownSize.value)])
+    const triggerId = computed<string>(() => {
+      return props.id || useId().value
```
jw-foss, line 148: **Suggestion:** we don't invoke computed inside a computed, which will always return a new computed.

```suggestion
const _id = useId()
  const triggerId = computed(() => {
      return props.id || _id.value
      })
```

jw-foss (commented):
@opengraphica in general it looks pretty good, I left a few comments for you to consider 🏓  back to you.

packages/components/dropdown/src/dropdown.vue
Giwayume: From what I remember it only recomputes if that inner referenced computed changes due to its refs updating, but there's no need to risk that in this case, yeah.

jw-foss: Right, just in case

jw-foss (approved):

packages/components/tooltip/src/trigger.vue
```diff
@@ -89,7 +88,8 @@ export default defineComponent({
       stopWhenControlledOrDisabled,
       (e: KeyboardEvent) => {
         const { code } = e
-        if (code === EVENT_CODE.enter || code === EVENT_CODE.space) {
+        if (props.triggerKeys.includes(code)) {
+          e.preventDefault()
```
tolking, line 92: @opengraphica Excuse me, why use `preventDefault` here, is there any special consideration?

This seems to cause the component contained in the Tooltip to be unable to use the Enter  keys 
and Space keys. [Demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cCBsYW5nPVwidHNcIj5cbmltcG9ydCB7IHJlZiB9IGZyb20gJ3Z1ZSdcbmltcG9ydCB7IEVsTWVzc2FnZSB9IGZyb20gJ2VsZW1lbnQtcGx1cydcblxuY29uc3QgdmFsdWUgPSByZWYoJycpXG5jb25zdCBoYW5kbGVDbGljayA9ICgpID0+IHtcbiAgRWxNZXNzYWdlKHtcbiAgICBtZXNzYWdlOiAnQ29uZ3JhdHMsIHRoaXMgaXMgYSBzdWNjZXNzIG1lc3NhZ2UuJyxcbiAgICB0eXBlOiAnc3VjY2VzcycsXG4gIH0pXG59XG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8ZWwtdG9vbHRpcCBjb250ZW50PVwiY29udGVudFwiPlxuICAgIDxlbC1idXR0b24gQGNsaWNrPVwiaGFuZGxlQ2xpY2tcIj5CdXR0b248L2VsLWJ1dHRvbj5cbiAgPC9lbC10b29sdGlwPlxuICA8ZWwtdG9vbHRpcCBjb250ZW50PVwiY29udGVudFwiPlxuICAgIDxlbC1pbnB1dCB2LW1vZGVsPVwidmFsdWVcIiB0eXBlPVwidGV4dGFyZWFcIiAvPlxuICA8L2VsLXRvb2x0aXA+XG48L3RlbXBsYXRlPlxuIiwiaW1wb3J0X21hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge31cbn0iLCJfbyI6e319)

Giwayume: That code is working as intended, basically "which keys trigger the tooltip to show?"

You shouldn't have a button that runs multiple actions when you press enter, either it controls a tooltip, or it does some other action. That is why the preventDefault is used. If "Enter" should do something other than show the tooltip, `triggerKeys` should be set to an empty array by the developer.

That being said, the default configuration for a tooltip doesn't make much sense. For the `tooltip` role, the tooltip should show and be read out on focus of the button, not on click.

E.g. default behavior from what is typically expected from a tooltip would look like:

```<el-tooltip content="content" trigger="focus" :trigger-keys="[]">```

Problem is changing this default behavior would be a breaking change, that can't be done immediately.

File: packages/components/popper/src/content.vue [hunk 7]
```diff
@@ -194,6 +205,38 @@ onMounted(() => {
     }
   )
 
+  watch(
```
purepear, line 208: This watcher is always triggered when a ElTooltip/ElPopover/ElPopconfirm/etc is shown because it's inside onMounted and it's `{immediate: true}`. The `props.triggerTargetEl` is always `undefined` (except for ElDropdown). 
This ends up removing any `role`, `aria-label`, `aria-modal` and `id` attributes for those components.
I believe this is a regression and i'm not quite sure if there's easy way to fix it.

Issue #7919, opened 2022-05-27
[Bug Report] [Style] [select] el-select在disabled属性变化的第一次高度会增加2px


mghcool:
Bug Type: **`Style`**

## Environment

- Vue Version: `3.2.36`
- Element Plus Version: `2.2.2`
- Browser / OS: `Chrome 90.6.7.4430/Win10 21H2`
- Build Tool: `Vue CLI`

### Related Component

- `el-select`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiXG48dGVtcGxhdGU+XG4gIDxoMT57eyBtc2cgfX08L2gxPlxuICA8ZWwtc2VsZWN0IHYtbW9kZWw9XCJ2YWx1ZVwiIDpkaXNhYmxlZD1cImRpc2FibGVcIiBkZXN0cm95LW9uLWNsb3NlIHBsYWNlaG9sZGVyPVwiU2VsZWN0XCI+XG4gICAgICA8ZWwtb3B0aW9uIGxhYmVsPVwiYWJjXCIgdmFsdWU9XCJhYmNcIiAvPlxuICA8L2VsLXNlbGVjdD5cbiAgICA8ZWwtc3dpdGNoIHYtbW9kZWw9XCJkaXNhYmxlXCIgLz5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgeyByZWYsIH0gZnJvbSAndnVlJ1xuXG5jb25zdCBtc2cgPSByZWYoJ3NlbGVjdOWwhuWcqOesrOS4gOasoeeCueWHu+W8gOWFs+aXtumrmOW6puWinuWKoDJweCcpXG5jb25zdCBkaXNhYmxlID0gcmVmKGZhbHNlKVxuPC9zY3JpcHQ+XG4iLCJpbXBvcnRfbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7fVxufSIsIl9vIjp7fX0=)

### Steps to reproduce

```vue
<template>
  <el-select v-model="value" :disabled="disable" destroy-on-close placeholder="Select">
      <el-option label="abc" value="abc" />
  </el-select>
    <el-switch v-model="disable" />
</template>

<script setup>
import { ref, } from 'vue'

const disable = ref(false)
</script>
```

### What is Expected?

[...]

PR #7958, opened 2022-05-28
fix(components): [select]  input height is 2px more

chenxch:

### DIFF
--- a/packages/components/select/__tests__/select.test.ts
+++ b/packages/components/select/__tests__/select.test.ts
@@ -1962,22 +1962,20 @@ describe('Select', () => {
       size: 'large',
     })
     await nextTick(nextTick)
-    const wrapperEl = wrapper.find('input').element as HTMLDivElement
-    expect(wrapperEl.style.height).toEqual('38px')
+    const inputEl = wrapper.find('input').element as HTMLDivElement
+    const sizeMap: Record<string, number> = {
+      small: 24,
+      default: 32,
+      large: 40,
+    }
 
-    // default size
-    await wrapper.setProps({
-      size: 'default',
-    })
-    await nextTick(nextTick)
-    expect(wrapperEl.style.height).toEqual('30px')
-
-    // small size
-    await wrapper.setProps({
-      size: 'small',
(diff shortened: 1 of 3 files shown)


chenxch:
fix #7919 
image
disabled:
image
2px more

packages/components/select/src/useSelect.ts
```diff
@@ -338,9 +337,7 @@ export const useSelect = (props, states: States, ctx) => {
       ) as HTMLInputElement
       const _tags = tags.value
 
-      const sizeInMap =
-        states.initialInputHeight ||
-        getComponentSize(selectSize.value || elForm.size)
+      const sizeInMap = getComponentSize(selectSize.value || elForm.size) - 2
       input.style.height =
         states.selected.length === 0
           ? `${sizeInMap}px`
```
YunYouJun, line 343: I think we should write `-2` here. And add a comment that it's an inner input so reduce it by 2px.

packages/components/select/__tests__/select.test.ts
```diff
@@ -1963,21 +1963,21 @@ describe('Select', () => {
     })
     await nextTick(nextTick)
     const wrapperEl = wrapper.find('input').element as HTMLDivElement
-    expect(wrapperEl.style.height).toEqual('40px')
+    expect(wrapperEl.style.height).toEqual('38px')
```
YunYouJun, line 1966: I think we can test `.el-input__wrapper` directyly, instead of changing sizes.

chenxch: I tried it. It's `style.htight` an empty string.

YunYouJun: OK. I think we can write it as a loop by `for`.

chenxch: OK, leave it to me.

chenxch: You are so awesome that I took it straight to the test.

YunYouJun (approved):

PR #7962, opened 2022-05-29
improvement(theme-chalk): [form-item] change display

ryuhangyeong:
## Why?

In the case of `</el-form-item>`, I think `display: block` is more appropriate than `display: flex`. Because by default the user thinks that if it's an `<h1>`, `<p>` element, it will take up an entire line, and if using a `<span>`, they usually expect it to be listed sideways. But now all tags are listed sideways.

This makes the user need to add one extra tag to take up the whole line inside `</el-form-item>`.

## Example

```ts
<template>
  <el-form :model="form" label-width="120px">
    <el-form-item label="Activity zone">
      <el-select v-model="form.region" placeholder="please select your zone">
        <el-option label="Zone one" value="shanghai" />
        <el-option label="Zone two" value="beijing" />
      </el-select>
      // here
      <span>hello world</span>
      <h1>
        hello world
      </h1>
    </el-form-item>
  </el-form>
</template>

<script lang="ts" setup>
import { reactive } from 'vue'

const form = reactive({
  region: '',
})
</script>
```

### What is Expected?

image

A mix of `<h1>` and `<span>` tags. According to each characteristic, `<h1>` would be expected to be displayed on a single line, and `<span>` would normally be arranged sideways. But in reality it is not.

image

## Foreseeable problems

Changes in the way layouts are laid out can affect existing layouts. I know, but I think it makes more sense to change the display property, so I'm requesting a Pull Request. thank you.

---

### DIFF
--- a/packages/theme-chalk/src/form.scss
+++ b/packages/theme-chalk/src/form.scss
@@ -160,10 +160,8 @@ $form-item-label-top-margin-bottom: map.merge(
     box-sizing: border-box;
   }
   @include e(content) {
-    display: flex;
-    flex-wrap: wrap;
-    align-items: center;
-    flex: 1;
+    display: block;
+    width: 100%;
     line-height: #{map.get($form-item-line-height, 'default')};
     position: relative;
     font-size: var(--font-size);


jw-foss:
I think there was some context behind updating this to `flex` instead of `block` it was `block` ad the very beginning, I'll include some people here for give more context @kooriookami @msidolphin @YunYouJun

YunYouJun:
You can use `flex-direction: column; # or row` to actively control whether it is on a line or not.

I think the purpose of using flex here is to solve common vertical centering problems. In the case of `block`, we would have to use `line-height` to achieve this, which would be fixed instead of dynamically centered.

And if you use `flex`, although the span parent element may have to be manually set (in special cases), you can still achieve whatever effect you want.

ryuhangyeong:
@YunYouJun Thank you for telling us about the history and explaining the pros and cons in detail. I will close the issue.

PR #7968, opened 2022-05-29
feat(components): [image] support native lazy loading

### LINKED ISSUE #7841: [Feature Request] [Enhancement] [image] HTMLImageElement.loading
Feature type: **`Enhancement`**

## Details

- Enhanced compone nt name: `image`

## Description

Currently `el-image` has lazy loading, but there is also a native loading(`loading="lazy"`) property provided by the browser. How about supporting this?

### Additional comments

https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading

tolking:
Add [native lazy loading](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading) support for Image components. 

close #7841 

- `loading` has higher priority than `lazy`.
- If the current browser supports native lazy loading, the native lazy loading will be used first, otherwise will be implemented through scroll.

### DIFF
--- a/packages/components/image/src/image.vue
+++ b/packages/components/image/src/image.vue
@@ -1,7 +1,7 @@
 <template>
   <div ref="container" :class="[ns.b(), $attrs.class]" :style="containerStyle">
     <img
-      v-if="imageSrc"
+      v-if="imageSrc !== undefined && !hasLoadError"
       v-bind="attrs"
       :src="imageSrc"
       :loading="loading"
@@ -75,7 +75,7 @@ const ns = useNamespace('image')
 const rawAttrs = useRawAttrs()
 const attrs = useAttrs()
 
-const imageSrc = ref('')
+const imageSrc = ref<string | undefined>()
 const hasLoadError = ref(false)
 const isLoading = ref(true)
 const showViewer = ref(false)
--- a/packages/components/image/src/image.vue
+++ b/packages/components/image/src/image.vue
@@ -1,19 +1,23 @@
 <template>
   <div ref="container" :class="[ns.b(), $attrs.class]" :style="containerStyle">
-    <slot v-if="loading" name="placeholder">
-      <div :class="ns.e('placeholder')" />
-    </slot>
-    <slot v-else-if="hasLoadError" name="error">
-      <div :class="ns.e('error')">{{ t('el.image.error') }}</div>
-    </slot>
     <img
-      v-else
+      v-if="imageSrc"
       v-bind="attrs"
-      :src="src"
+      :src="imageSrc"
       :style="imageStyle"
       :class="[ns.e('inner'), preview ? ns.e('preview') : '']"
       @click="clickHandler"
+      @load="handleLoad"
+      @error="handleError"
     />
+    <div v-if="loading || hasLoadError" :class="ns.e('overlay')">
+      <slot v-if="loading" name="placeholder">
+        <div :class="ns.e('placeholder')" />
+      </slot>
+      <slot v-else-if="hasLoadError" name="error">
+        <div :class="ns.e('error')">{{ t('el.image.error') }}</div>
+      </slot>
+    </div>
     <template v-if="preview">
       <image-viewer
         v-if="showViewer"
@@ -72,10 +76,9 @@ const ns = useNamespace('image')
 const rawAttrs = useRawAttrs()
 const attrs = useAttrs()
 
+const imageSrc = ref('')
 const hasLoadError = ref(false)
 const loading = ref(true)
-const imgWidth = ref(0)
-const imgHeight = ref(0)
 const showViewer = ref(false)
 const container = ref<HTMLElement>()
 
@@ -113,37 +116,10 @@ const loadImage = () => {
   // reset status
   loading.value = true
   hasLoadError.value = false
-
-  const img = new Image()
-  const currentImageSrc = props.src
-
-  // load & error callbacks are only responsible for currentImageSrc
-  img.addEventListener('load', (e) => {
-    if (currentImageSrc !== props.src) {
-      return
-    }
-    handleLoad(e, img)
-  })
-  img.addEventListener('error', (e) => {
-    if (currentImageSrc !== props.src) {
-      return
-    }
-    handleError(e)
-  })
-
-  // bind html attrs
-  // so it can behave consistently
-  Object.entries(rawAttrs).forEach(([key, value]) => {
-    // avoid onload to be overwritten
-    if (key.toLowerCase() === 'onload') return
-    img.setAttribute(key, value as string)
-  })
-  img.src = currentImageSrc
+  imageSrc.value = props.src
 }
 
-function handleLoad(e: Event, img: HTMLImageElement) {
-  imgWidth.value = img.width
-  imgHeight.value = img.height
+function handleLoad() {
   loading.value = false
   hasLoadError.value = false
 }
--- a/packages/components/image/src/image.vue
+++ b/packages/components/image/src/image.vue
@@ -44,7 +44,6 @@ import {
   nextTick,
   onMounted,
   ref,
-  toRefs,
   useAttrs as useRawAttrs,
   watch,
 } from 'vue'
@@ -71,7 +70,6 @@ const emit = defineEmits(imageEmits)
 
 let prevOverflow = ''
 
-const { loading } = toRefs(props)
 const { t } = useLocale()
 const ns = useNamespace('image')
 const rawAttrs = useRawAttrs()
@@ -113,8 +111,8 @@ const imageIndex = computed(() => {
 })
 
 const isManual = computed(() => {
-  if (loading?.value === 'eager') return false
-  return (!supportLoading && loading?.value === 'lazy') || props.lazy
+  if (props.loading === 'eager') return false
+  return (!supportLoading && props.loading === 'lazy') || props.lazy
 })
 
 const loadImage = () => {
--- a/packages/components/image/src/image.vue
+++ b/packages/components/image/src/image.vue
@@ -4,20 +4,19 @@
       v-if="imageSrc"
       v-bind="attrs"
       :src="imageSrc"
+      :loading="loading"
       :style="imageStyle"
       :class="[ns.e('inner'), preview ? ns.e('preview') : '']"
       @click="clickHandler"
       @load="handleLoad"
       @error="handleError"
     />
-    <div v-if="loading || hasLoadError" :class="ns.e('overlay')">
-      <slot v-if="loading" name="placeholder">
-        <div :class="ns.e('placeholder')" />
-      </slot>
-      <slot v-else-if="hasLoadError" name="error">
-        <div :class="ns.e('error')">{{ t('el.image.error') }}</div>
-      </slot>
-    </div>
+    <slot v-if="isLoading" name="placeholder">
+      <div :class="ns.e('placeholder')" />
+    </slot>
+    <slot v-else-if="hasLoadError" name="error">
+      <div :class="ns.e('error')">{{ t('el.image.error') }}</div>
+    </slot>
     <template v-if="preview">
       <image-viewer
         v-if="showViewer"
@@ -45,6 +44,7 @@ import {
   nextTick,
   onMounted,
   ref,
+  toRefs,
   useAttrs as useRawAttrs,
   watch,
 } from 'vue'
@@ -71,6 +71,7 @@ const emit = defineEmits(imageEmits)
 
 let prevOverflow = ''
 
+const { loading } = toRefs(props)
 const { t } = useLocale()
 const ns = useNamespace('image')
 const rawAttrs = useRawAttrs()
@@ -78,9 +79,10 @@ const attrs = useAttrs()
 
 const imageSrc = ref('')
 const hasLoadError = ref(false)
-const loading = ref(true)
+const isLoading = ref(true)
 const showViewer = ref(false)
 const container = ref<HTMLElement>()
+const supportLoading = ref(true)
 
 const _scrollContainer = ref<HTMLElement | Window>()
 let stopScrollListener: (() => void) | undefined
@@ -110,22 +112,27 @@ const imageIndex = computed(() => {
   return previewIndex
 })
 
+const isManual = computed(() => {
+  if (loading?.value === 'eager') return false
+  return (!supportLoading.value && loading?.value === 'lazy') || props.lazy
+})
+
 const loadImage = () => {
   if (!isClient) return
 
   // reset status
-  loading.value = true
+  isLoading.value = true
   hasLoadError.value = false
   imageSrc.value = props.src
 }
 
 function handleLoad() {
-  loading.value = false
+  isLoading.value = false
   hasLoadError.value = false
 }
 
 function handleError(event: Event) {
-  loading.value = false
+  isLoading.value = false
   hasLoadError.value = true
   emit('error', event)
 }
@@ -211,9 +218,9 @@ function switchViewer(val: number) {
 watch(
   () => props.src,
   () => {
-    if (props.lazy) {
+    if (isManual.value) {
       // reset status
-      loading.value = true
+      isLoading.value = true
       hasLoadError.value = false
       removeLazyLoadListener()
       addLazyLoadListener()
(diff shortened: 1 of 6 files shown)


tolking:
- [the native lazy loading demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cCBsYW5nPVwidHNcIj5cbmNvbnN0IHVybCA9ICdodHRwczovL3BpY3N1bS5waG90b3MvNTM2LzM1ND9yYW5kb209J1xuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPGVsLWltYWdlXG4gICAgdi1mb3I9XCJuIGluIDIwXCJcbiAgICA6a2V5PVwiblwiXG4gICAgOnNyYz1cInVybCArIG5cIlxuICAgIGxvYWRpbmc9XCJsYXp5XCJcbiAgICBzdHlsZT1cImRpc3BsYXk6IGJsb2NrOyBtYXJnaW4tYm90dG9tOiAxMHB4OyB3aWR0aDogNTM2cHg7IGhlaWdodDogMzU0cHhcIlxuICAvPlxuPC90ZW1wbGF0ZT5cblxuPHN0eWxlPlxuI2FwcCB7XG5cdGhlaWdodDogMTAwdmg7XG59XG48L3N0eWxlPlxuIiwiUGxheWdyb3VuZE1haW4udnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCBBcHAgZnJvbSAnLi9BcHAudnVlJ1xuaW1wb3J0IHsgc2V0dXBFbGVtZW50UGx1cyB9IGZyb20gJy4vZWxlbWVudC1wbHVzLmpzJ1xuc2V0dXBFbGVtZW50UGx1cygpXG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8QXBwIC8+XG48L3RlbXBsYXRlPlxuIiwiaW1wb3J0X21hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge1xuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9wcmV2aWV3LTc5NjgtZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS9pbmRleC5mdWxsLm1pbi5tanNcIixcbiAgICBcImVsZW1lbnQtcGx1cy9cIjogXCJ1bnN1cHBvcnRlZFwiXG4gIH1cbn0iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL3VucGtnLmNvbS9AdnVlL3J1bnRpbWUtZG9tQGxhdGVzdC9kaXN0L3J1bnRpbWUtZG9tLmVzbS1icm93c2VyLmpzXCIsXG4gICAgXCJAdnVlL3NoYXJlZFwiOiBcImh0dHBzOi8vdW5wa2cuY29tL0B2dWUvc2hhcmVkQGxhdGVzdC9kaXN0L3NoYXJlZC5lc20tYnVuZGxlci5qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9wcmV2aWV3LTc5NjgtZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS9pbmRleC5mdWxsLm1pbi5tanNcIixcbiAgICBcImVsZW1lbnQtcGx1cy9cIjogXCJ1bnN1cHBvcnRlZFwiLFxuICAgIFwiQGVsZW1lbnQtcGx1cy9pY29ucy12dWVcIjogXCJodHRwczovL3VucGtnLmNvbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJlbGVtZW50LXBsdXMuanMiOiJpbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5pbXBvcnQgRWxlbWVudFBsdXMgZnJvbSAnZWxlbWVudC1wbHVzJ1xuXG5sZXQgaW5zdGFsbGVkID0gZmFsc2VcbmF3YWl0IGxvYWRTdHlsZSgpXG5cbmV4cG9ydCBmdW5jdGlvbiBzZXR1cEVsZW1lbnRQbHVzKCkge1xuICBpZiAoaW5zdGFsbGVkKSByZXR1cm5cbiAgY29uc3QgaW5zdGFuY2UgPSBnZXRDdXJyZW50SW5zdGFuY2UoKVxuICBpbnN0YW5jZS5hcHBDb250ZXh0LmFwcC51c2UoRWxlbWVudFBsdXMpXG4gIGluc3RhbGxlZCA9IHRydWVcbn1cblxuZXhwb3J0IGZ1bmN0aW9uIGxvYWRTdHlsZSgpIHtcbiAgcmV0dXJuIG5ldyBQcm9taXNlKChyZXNvbHZlLCByZWplY3QpID0+IHtcbiAgICBjb25zdCBsaW5rID0gZG9jdW1lbnQuY3JlYXRlRWxlbWVudCgnbGluaycpXG4gICAgbGluay5yZWwgPSAnc3R5bGVzaGVldCdcbiAgICBsaW5rLmhyZWYgPSAnaHR0cHM6Ly9wcmV2aWV3LTc5NjgtZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS9pbmRleC5jc3MnXG4gICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdsb2FkJywgcmVzb2x2ZSlcbiAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2Vycm9yJywgcmVqZWN0KVxuICAgIGRvY3VtZW50LmJvZHkuYXBwZW5kKGxpbmspXG4gIH0pXG59IiwiX28iOnsic2hvd0hpZGRlbiI6dHJ1ZSwic3R5bGVTb3VyY2UiOiJodHRwczovL3ByZXZpZXctNzk2OC1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2luZGV4LmNzcyJ9fQ==)
- [the lazy demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cCBsYW5nPVwidHNcIj5cbmNvbnN0IHVybCA9ICdodHRwczovL3BpY3N1bS5waG90b3MvNTM2LzM1ND9yYW5kb209J1xuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPGVsLWltYWdlXG4gICAgdi1mb3I9XCJuIGluIDIwXCJcbiAgICA6a2V5PVwiblwiXG4gICAgOnNyYz1cInVybCArIG5cIlxuICAgIGxhenlcbiAgICBzdHlsZT1cImRpc3BsYXk6IGJsb2NrOyBtYXJnaW4tYm90dG9tOiAxMHB4OyB3aWR0aDogNTM2cHg7IGhlaWdodDogMzU0cHhcIlxuICAvPlxuPC90ZW1wbGF0ZT5cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydF9tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcImVsZW1lbnQtcGx1c1wiOiBcImh0dHBzOi8vcHJldmlldy03OTY4LWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwidW5zdXBwb3J0ZWRcIlxuICB9XG59IiwiaW1wb3J0LW1hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge1xuICAgIFwidnVlXCI6IFwiaHR0cHM6Ly91bnBrZy5jb20vQHZ1ZS9ydW50aW1lLWRvbUBsYXRlc3QvZGlzdC9ydW50aW1lLWRvbS5lc20tYnJvd3Nlci5qc1wiLFxuICAgIFwiQHZ1ZS9zaGFyZWRcIjogXCJodHRwczovL3VucGtnLmNvbS9AdnVlL3NoYXJlZEBsYXRlc3QvZGlzdC9zaGFyZWQuZXNtLWJ1bmRsZXIuanNcIixcbiAgICBcImVsZW1lbnQtcGx1c1wiOiBcImh0dHBzOi8vcHJldmlldy03OTY4LWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwidW5zdXBwb3J0ZWRcIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly91bnBrZy5jb20vQGVsZW1lbnQtcGx1cy9pY29ucy12dWVAMi9kaXN0L2luZGV4Lm1pbi5qc1wiXG4gIH0sXG4gIFwic2NvcGVzXCI6IHt9XG59IiwiZWxlbWVudC1wbHVzLmpzIjoiaW1wb3J0IHsgZ2V0Q3VycmVudEluc3RhbmNlIH0gZnJvbSAndnVlJ1xuaW1wb3J0IEVsZW1lbnRQbHVzIGZyb20gJ2VsZW1lbnQtcGx1cydcblxubGV0IGluc3RhbGxlZCA9IGZhbHNlXG5hd2FpdCBsb2FkU3R5bGUoKVxuXG5leHBvcnQgZnVuY3Rpb24gc2V0dXBFbGVtZW50UGx1cygpIHtcbiAgaWYgKGluc3RhbGxlZCkgcmV0dXJuXG4gIGNvbnN0IGluc3RhbmNlID0gZ2V0Q3VycmVudEluc3RhbmNlKClcbiAgaW5zdGFuY2UuYXBwQ29udGV4dC5hcHAudXNlKEVsZW1lbnRQbHVzKVxuICBpbnN0YWxsZWQgPSB0cnVlXG59XG5cbmV4cG9ydCBmdW5jdGlvbiBsb2FkU3R5bGUoKSB7XG4gIHJldHVybiBuZXcgUHJvbWlzZSgocmVzb2x2ZSwgcmVqZWN0KSA9PiB7XG4gICAgY29uc3QgbGluayA9IGRvY3VtZW50LmNyZWF0ZUVsZW1lbnQoJ2xpbmsnKVxuICAgIGxpbmsucmVsID0gJ3N0eWxlc2hlZXQnXG4gICAgbGluay5ocmVmID0gJ2h0dHBzOi8vcHJldmlldy03OTY4LWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvaW5kZXguY3NzJ1xuICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignbG9hZCcsIHJlc29sdmUpXG4gICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdlcnJvcicsIHJlamVjdClcbiAgICBkb2N1bWVudC5ib2R5LmFwcGVuZChsaW5rKVxuICB9KVxufSIsIl9vIjp7InNob3dIaWRkZW4iOnRydWUsInN0eWxlU291cmNlIjoiaHR0cHM6Ly9wcmV2aWV3LTc5NjgtZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS9pbmRleC5jc3MifX0=)

packages/components/image/src/image.vue
```diff
@@ -67,17 +71,18 @@ const emit = defineEmits(imageEmits)
...
 
+const imageSrc = ref('')
 const hasLoadError = ref(false)
-const loading = ref(true)
-const imgWidth = ref(0)
-const imgHeight = ref(0)
+const isLoading = ref(true)
 const showViewer = ref(false)
 const container = ref<HTMLElement>()
+const supportLoading = ref(true)
```
sxzz, line 85: it's a constant, and never change. 
```suggestion
const supportLoading = 'loading' in HTMLImageElement.prototype
```

tolking: Because must make sure to judge in the browser environment, otherwise maybe get an error in server rendering.

packages/components/image/src/image.vue
```diff
@@ -107,49 +112,27 @@ const imageIndex = computed(() => {
   return previewIndex
 })
 
+const isManual = computed(() => {
+  if (loading?.value === 'eager') return false
```
jw-foss, line 116: **Question:** loading should always be a `ref` right? Why are we adding optional chaining here?

packages/components/image/src/image.vue
sxzz: Hmmm you can use `isServer` / `isClient` to handle this.

jw-foss, line 85: Should we do something about this thing? Maybe detect the support of this API

jw-foss (commented):
I think we should consider this a breaking change. Maybe we need to do some feature flag on this thing.

jw-foss:
We can support this for sure, but we need a better way to do so, we need the previous version.

My suggestion is:

1. Extract the logic of loading an image.
2. Create another image with native support.
3. Determine which component should be rendered via the presence of the given props.

packages/components/image/src/image.vue
tolking: loading does not have a default value, it may not be defined.

packages/components/image/src/image.vue
tolking: Tip `lazy` -> `loading="lazy"` ?

I do not recommend to do this. because native lazy loading will load more images than scroll (lazy API), and some older browsers do not support this feature.

Maybe we should not break the lazy API now.

jw-foss:
I wonder, we determine the flag with `isClient` which might cause hydration issue since the server rendered template is different than client rendered template.

tolking:
> I wonder, we determine the flag with `isClient` which might cause hydration issue since the server rendered template is different than client rendered template.

HTMLImageElement does not exist in the SSR/SSG. If you use it directly, it will produce an error.

The `isClient` is false on SSR/SSG, so there will be no errors due to HTMLImageElement.

Only affects the supportLoading, It should not affect the template.

I use my SSG project to test without problems (`isClient && 'loading' in HTMLImageElement.prototype`)

jw-foss:
> > I wonder, we determine the flag with `isClient` which might cause hydration issue since the server rendered template is different than client rendered template.
> 
> HTMLImageElement does not exist in the SSR/SSG. If you use it directly, it will produce an error.
> 
> The `isClient` is false on SSR/SSG, so there will be no errors due to HTMLImageElement.
> 
> Only affects the supportLoading, It should not affect the template.
> 
> I use my SSG project to test without problems (`isClient && 'loading' in HTMLImageElement.prototype`)

I see, thanks for the explanation 💯

packages/components/image/src/image.vue
jw-foss: `loading` is a `Ref<'eager' | 'lazy'>` type right? So that `Ref` will always have `.value` attribute on it. That was my question, you can see the screenshot below.

CleanShot 2022-05-30 at 19 44 13@2x

tolking: toRefs -> Ref<BuildPropType<StringConstructor, "eager" | "lazy", unknown> | undefined> | **undefined**

sxzz (approved):
LGTM

jw-foss (approved):

jw-foss:
@tolking one more thing, can we mention this API in the documentation?

Also, we should add entries in `breakings/2.2.2/` folder

tolking:
> @tolking one more thing, can we mention this API in the documentation?
> 
> Also, we should add entries in `breakings/2.2.2/` folder

If used the loading API with `lazy` in the previous version, this will break the loading of images. 

I don't know what will be merged into which version, so the documentation is not change yet.

Current version 2.2.2, maybe the `breakings/2.2.2/` is error

sxzz:
@tolking I think it will be release as `2.2.3`.

Issue #6114, opened 2022-02-17
[Refactor] Refactor Plan


sxzz:
## Plans

- remove all `@ts-nocheck`
- [`script-setup`](https://github.com/element-plus/element-plus/issues/5709)
   - refactor `state` to `ref`
   - expose all things user need
- [style namespace](https://github.com/element-plus/element-plus/issues/5374)
- [export component instance type](https://github.com/element-plus/element-plus/issues/4308)
- [refactor plan](https://github.com/element-plus/element-plus/pull/3113)
- [JSX](https://github.com/element-plus/element-plus/issues/5708)
   - Unit Test included
- Fix ALL TS errors, and add component path to `build/type-safe.json` and `build/type-safe-stricter.json`.

Please community contributors at me in your PR.

## Progress

- [x] affix
  - #6065
  - #6159
  - #6295
- [x] alert
  - #6066
  - #7973 
- [x] autocomplete
  - #6067
- [x] avatar
  - #6070
- [x] backtop
  - #6069
  - #6160
- [x] badge
  - #6068
  - #6247
- [x] breadcrumb / breadcrumb-item
  - #6073
- [x] button
  - #5933
- [x] button-group
  - #6256
- [x] calendar
  - #6682
- [x] card
  - #6071
- [x] carousel / carousel-item
  - #6681
- [ ] cascader / cascader-panel
- [x] check-tag
  - #6296
- [ ] checkbox / checkbox-button / checkbox-group
  - Assignee: @chenxch 
  - #7794 
- [x] col
  - #6303
  - assignee @HeftyKoo
- [x] collapse / collapse-item
  - #6307
- [x] collapse-transition
  - Assignee: @ryuhangyeong
  - #7953
- [x] collection
  - Assignee: @metanas
  - #7710
- [x] color-picker
  - #9672 
- [x] config-provider
  - Assignee: @holazz 
- [x] container / aside / footer / header / main
  - Assignee: @ryuhangyeong
  - #7725
  - #7707
- [ ] date-picker
  - Assignee: @JeremyWuuuuu 
- [ ] descriptions / descriptions-item
  - #8348
  - #8349
- [x] dialog
  - #6087
  - #6211
  - #6300
- [x] divider
  - #6556
- [ ] drawer [WIP]
  - assignee: @JeremyWuuuuu 
  - #6239
- [ ] dropdown / dropdown-item / dropdown-menu
- [x] empty
  - #6703
- [ ] focus-trap
- [x] form / form-item
[...]

PR #7986, opened 2022-05-30
refactor(components): [scrollbar] switch to script-setup syntax

holazz:
https://github.com/element-plus/element-plus/issues/6114

### DIFF
--- a/packages/components/scrollbar/src/scrollbar.ts
+++ b/packages/components/scrollbar/src/scrollbar.ts
@@ -1,5 +1,6 @@
 import { buildProps, definePropType, isNumber } from '@element-plus/utils'
 import type { ExtractPropTypes, StyleValue } from 'vue'
+import type Scrollbar from './scrollbar.vue'
 
 export const scrollbarProps = buildProps({
   height: {
@@ -10,10 +11,7 @@ export const scrollbarProps = buildProps({
     type: [String, Number],
     default: '',
   },
-  native: {
-    type: Boolean,
-    default: false,
-  },
+  native: Boolean,
   wrapStyle: {
     type: definePropType<StyleValue>([String, Object, Array]),
     default: '',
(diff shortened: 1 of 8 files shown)


holazz:
cc @sxzz

packages/components/scrollbar/src/scrollbar.ts
```diff
@@ -10,10 +11,7 @@ export const scrollbarProps = buildProps({
     type: [String, Number],
     default: '',
   },
-  native: {
-    type: Boolean,
-    default: false,
-  },
+  native: Boolean,
```
sxzz, line 14: Why did you change it? I think we can keep it as-is.

packages/components/scrollbar/src/scrollbar.ts
```diff
@@ -35,16 +33,12 @@ export const scrollbarProps = buildProps({
     type: String,
     default: 'div',
   },
-  always: {
-    type: Boolean,
-    default: false,
-  },
+  always: Boolean,
```
sxzz, line 36: Same here

packages/components/scrollbar/src/scrollbar.ts
holazz: I think the default value of Boolean props is `false`, so I simplified the code. 
refer to: https://github.com/vuejs/core/issues/5863#issuecomment-1118446966

sxzz (approved):

sxzz:
https://github.com/element-plus/element-plus/runs/6662396264?check_suite_focus=true

Could you please check this?

holazz:
> [element-plus/element-plus/runs/6662396264?check_suite_focus=true](https://github.com/element-plus/element-plus/runs/6662396264?check_suite_focus=true)
> 
> Could you please check this?
> 
> **Update: never mind**

I'm debugging, it seems to be caused by removing `scrollbar` from `type-unsafe`, I don't need to care about this now?

holazz:
> I found that the table header does not scroll with the body, maybe the Scrollbar components should expose more values.
> 
> eq: scrollbar, resize$ ...

I found that `wrap$` is used in `table`, I fixed it here https://github.com/element-plus/element-plus/pull/8035

PR #8015, opened 2022-05-31
feat(components): [select-v2] add label key

BarHool:
select v2 add label key


packages/components/select-v2/src/useSelect.ts
```diff
@@ -236,6 +238,10 @@ const useSelect = (props: ExtractPropTypes<typeof SelectProps>, emit) => {
 
   const currentPlaceholder = computed(() => {
     const _placeholder = props.placeholder || t('el.select.placeholder')
+    console.log(
+      '----------',
+      props.multiple ? _placeholder : states.selectedLabel || _placeholder
+    )
```
buqiyuan, line 244: Please remove the `console.log`

BarHool:
> I wonder if this feature has been discussed? i did not find it.

No, I just saw it under some issue, and someone of you suggest to add pr. 

I'm sorry I didn't understand your development process

tolking:
I just do not think it is the best solution, maybe we should be consistent with props API in Tree v2. 

maybe need more discuss.

BarHool:
> I just do not think it is the best solution, maybe we should be consistent with props API in Tree v2.
> 
> maybe need more discuss.

I want to know when I can get the result of the discussion.

Over the past eight months，I use this solution locally. I'm tired of follow select-v2 update to modify my code.

If you can decide to use the solution like props API in Tree v2, I'd glad to and will try my best to modify it.

jw-foss:
@BarHool thank you for contributing in advance! But there are two things we need to take care of here:

1. Please use REBASE instead of MERGE strategy to sync the code from upstream.
2. Consider adding at least **1** test case for covering your code.

iamkun:
LGTM, please add some unit tests, also, please make a clean commit (140 commits in this PR)

BarHool:
> LGTM, please add some unit tests, also, please make a clean commit (140 commits in this PR)

I add unit test, but it can't pass it. I try to found what happen, but I can't find the reason.

And I think i clean commit correctly, but this operation is too strange for me. I hope I did it right.

tolking:
maybe you should use `git rebase dev` and `git push --force`.

BarHool:
> Maybe you should use `git rebase dev` and `git push --force`.

I have just used these two commands, but the number of commits has not decreased

BarHool:
> About API:
> 
> should be consistent with props API in Tree v2.
> 
> like `props: { label: 'label', value: 'value', options: 'options' }`

I will try in this direction. Should I create a new PR to do this?

tolking:
It is up to you, but preferably just one PR.

xiaoxian521:

PR #8039, opened 2022-06-01
fix(components): [space] missing empty string size

HeftyKoo:

### DIFF
--- a/packages/components/space/src/use-space.ts
+++ b/packages/components/space/src/use-space.ts
@@ -8,7 +8,6 @@ import type { ComponentSize } from '@element-plus/constants'
 
 const SIZE_MAP = {
   small: 8,
-  '': 8,
   default: 12,
   large: 16,
 } as Record<ComponentSize, number>
--- a/packages/components/space/src/use-space.ts
+++ b/packages/components/space/src/use-space.ts
@@ -56,7 +56,7 @@ export function useSpace(props: SpaceProps) {
       if (isNumber(size)) {
         val = size
       } else {
-        val = SIZE_MAP[size || 'small']
+        val = SIZE_MAP[size || 'small'] || SIZE_MAP.small
       }
 
       if ((wrap || fill) && dir === 'horizontal') {
--- a/packages/components/space/src/use-space.ts
+++ b/packages/components/space/src/use-space.ts
@@ -57,7 +57,7 @@ export function useSpace(props: SpaceProps) {
       if (isNumber(size)) {
         val = size
       } else {
-        val = SIZE_MAP[size] || SIZE_MAP.small
+        val = SIZE_MAP[size || 'small'] || SIZE_MAP.small
       }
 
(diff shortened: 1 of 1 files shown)


jw-foss:
Syntactically I don't think we need this maybe add fallbacks to `falsy` value might be a better solution?

tolking:
This is indeed a `ts` syntax error, maybe the components size should not accept `''` value.

HeftyKoo:
> Syntactically I don't think we need this maybe adding fallbacks to `falsy` value might be a better solution?

It's more than a typo issue, may be exclude the empty string is better.

packages/components/space/src/use-space.ts
```diff
@@ -6,7 +6,7 @@ import type { SpaceProps } from './space'
 import type { CSSProperties, StyleValue } from 'vue'
 import type { ComponentSize } from '@element-plus/constants'
 
-const SIZE_MAP: Record<ComponentSize, number> = {
+const SIZE_MAP: Record<Exclude<ComponentSize, ''>, number> = {
```
tolking, line 9: There mark a type error on line 60, and you should also modify line 60

tolking (changes requested):

tolking (approved):

packages/components/space/src/use-space.ts
```diff
@@ -57,7 +57,7 @@ export function useSpace(props: SpaceProps) {
       if (isNumber(size)) {
         val = size
       } else {
-        val = SIZE_MAP[size] || SIZE_MAP.small
+        val = SIZE_MAP[size || 'small'] || SIZE_MAP.small
```
tolking, line 60: It seems that both are the same

```diff
- val = SIZE_MAP[size || 'small'] || SIZE_MAP.small
+ val = SIZE_MAP[size || 'small']
```

packages/components/space/src/use-space.ts
```diff
@@ -6,7 +6,7 @@ import type { SpaceProps } from './space'
 import type { CSSProperties, StyleValue } from 'vue'
 import type { ComponentSize } from '@element-plus/constants'
 
-const SIZE_MAP: Record<ComponentSize, number> = {
+const SIZE_MAP: Record<Exclude<ComponentSize, ''>, number> = {
   small: 8,
   default: 12,
   large: 16,
```
sxzz, line 12:
```suggestion
const SIZE_MAP = {
  small: 8,
  default: 12,
  large: 16,
) as const
```

packages/components/space/src/use-space.ts
sxzz, line 60:
```suggestion
        val = SIZE_MAP[size || 'small']
```

sxzz (approved):

tolking (approved):

packages/components/space/src/use-space.ts
sxzz: if `size` is an invalid value, it will fallback to `small`

PR #8301, opened 2022-06-16
fix(components): [tooltip] not disappear when <keep-alive> change

### LINKED ISSUE #8287: [Bug Report] [Component] [tooltip] el-tooltip does not disappear when the reference element is in <keep-alive>
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.37`
- Element Plus Version: `2.1.8`
- Browser / OS: `Chrome/Windows`
- Build Tool: `Vue CLI`

### Related Component

- `el-tooltip`

### Reproduction Link

[CodeSandbox](https://codesandbox.io/s/element-plus-tooltip-problem-vuprfr)

### Steps to reproduce

Hover on "Go to About" button to show the tooltip and click it. Alternatively click directly on button without waiting for tooltip.
The dynamic component in the <keep-alive> should be switched after click.

### What is Expected?

The tooltip disappears after the click on the button, because the button is the reference element and not visible anymore

The tooltip is still visible (at top left corner or at the previous button position) after switching the dynamic component in the <keep-alive>, but the reference element is not visible anymore. Only after clicking on the tooltip it disappears.

### Additional comments

It is also reproducible with the latest version of Element Plus (v2.2.5).

Relevant components/files in Codesandbox example are "App.vue" and "Home.vue" in src folder, nothing else.

BgaSol:
[Bug Report] [Component] [tooltip] el-tooltip does not disappear when the reference element is in
<keep-alive>

closed #8287

### DIFF
--- a/packages/components/tooltip/src/tooltip.vue
+++ b/packages/components/tooltip/src/tooltip.vue
@@ -51,6 +51,7 @@
 import {
   computed,
   defineComponent,
+  onDeactivated,
   provide,
   readonly,
   ref,
@@ -206,6 +207,8 @@ export default defineComponent({
         contentRef.value?.contentRef?.popperContentRef
       return popperContent && popperContent.contains(document.activeElement)
     }
+
+    onDeactivated(() => open.value && hide())
 
     return {
       compatShowAfter,


packages/components/tooltip/src/tooltip.vue
```diff
@@ -207,6 +210,10 @@ export default defineComponent({
       return popperContent && popperContent.contains(document.activeElement)
     }
 
+    const isActive = ref(true)
+    onActivated(() => (isActive.value = true))
+    onDeactivated(() => (isActive.value = false))
```
tolking, line 215: It seems that you only need to judge whether the content exists here, and then call `hide` according to the situation.

sunnyCoCosun: Thank you very much for your ideas。
Component open property is now controlled by the keep-alive hook function

packages/components/tooltip/src/tooltip.vue
```diff
@@ -207,6 +208,12 @@ export default defineComponent({
       return popperContent && popperContent.contains(document.activeElement)
     }
 
+    onDeactivated(() => {
+      if (open.value) {
```
tolking, line 212: Maybe it should not be a simple judgment to open then close it.

Should determine whether the DOM element that triggered the popup still exists.

packages/components/tooltip/src/tooltip.vue
```diff
@@ -207,6 +208,12 @@ export default defineComponent({
...
+        open.value = false
```
tolking, line 213: Use the `hide` function to replace this line, we need to emit closed events.

packages/components/tooltip/src/tooltip.vue
sunnyCoCosun: Why do we need to know if dom exists?
If the onDeactivated function is executed, the component is to unAcvived.
Sorry, it's my first time to pr in element.
Thank you very much for answering my questions.

tolking: It seems that you are right, I was overthinking

sunnyCoCosun: Thanks for your help anyway

tolking (approved):

tolking:
@sunnyCoCosun the header not more than 72 characters

sunnyCoCosun:
@tolking I've changed the title. Don't forget to review it

tolking (approved):

PR #8339, opened 2022-06-18
refactor(components): [timeline] use JSX in Unit test

ryuhangyeong:
## Description

#6114 

---


chenxch (approved):
Next time you can try the `git mv a.ts b.tsx` command so that you can do diffs, which I just learned recently.

holazz:
> Next time you can try the `git mv a.ts b.tsx` command so that you can do diffs, which I just learned recently.

Unfortunately, this doesn't always work. In fact, this behaves the same as manual renaming. We @ryuhangyeong  have discussed this before. It seems that it is impossible always to retain history. 🤔
refer to: 
https://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv/3010099#3010099
https://web.archive.org/web/20150209075907/http://permalink.gmane.org:80/gmane.comp.version-control.git/217

ryuhangyeong:
all Right. We've been researching and checking for this, but haven't been able to find a suitable solution.

In my opinion, it might be better to change all extensions in batches to .ts -> .tsx and then refactor.

chenxch:
image
This is what I said about the effect, but I don't know whether it can be compared like this after pushing.

holazz:
The behavior in the IDE is the same as after pushing, if git can't retain the history, what you see in the IDE will be the new file.

chenxch:
No, for git, the operation of the same name is to delete the old file and add the renamed file. These two cannot be directly compared by `diff`, but `mv` can indeed. You can take a look at this [demo](https://github.com/chenxch/easy-work/pull/1), it has only one file in the change log from `ts` to `tsx`.

chenxch:
And I read that your JSX related refactoring pr is also handled with `mv`.

holazz:
Hi, @chenxch I have tested similar behavior, maybe because the changes are relatively small, so the two can be compared by `diff`. And all my refactoring PRs didn't use `git mv`. Try this [demo](https://elementplus-elementplus-7i0g62l5h92.ws-us54.gitpod.io/):

run `git mv packages/components/slider/__tests__/slider.test.ts packages/components/slider/__tests__/slider.test.tsx`

1. Make a small change, such as deleting a line of code. Check out the source control panel and you'll see:

WX20220718-093750@2x

2. Make big changes, copy the contents of file `play/src/slider.test.tsx` to file `packages/components/slider/__tests__/slider.test.tsx`. Check out the source control panel and you'll see:

WX20220718-092256@2x

chenxch:
Thank you so much for such a detailed introduction, I hope the effect is in the processing effect of the first step, because this will keep the history and diff comparison. When you manually rename the file, does git record the changes of 2 files there?

chenxch:
The demo prompts me that I don't have permission.

chenxch:
@holazz I'm here again, thank you so much for giving me new insight.

holazz:
> @holazz I'm here again, thank you so much for giving me new insight.
> 
> This is a translation of the summary of the article I found. `When committing moves, git will provide a similarity threshold, which defaults to 50%, beyond which git will track it during the move. Think of it as a move rather than just a delete and an addition. But this threshold can be configured by ourselves by adding a number after the switch -M.`

Thanks for sharing! I went through the git documentation and I found that `-M` is available in `git diff`. Is there a way we can config it in `git commit`? 🤔

chenxch:
It seems not.😔

PR #8342, opened 2022-06-18
improvement(components): [select] remove duplicate items

ryuhangyeong:
# Suggest

## Description

If collapse-tags-tooltip is used in the current select component, it will be displayed as shown below.

image

You can see that the same element is displayed twice. I think this should be removed.

## Why

There is no reason to show the same element twice, and users would think that the tooltip will contain all but the first element. 

image

And the number indicated by + is different from the number of items displayed in the actual tooltip.

## etc

I think there must have been a good reason why the component was developed like this at the time it was developed. But I thought what I was suggesting would be more intuitive when viewed by users.

Thanks for reading. Any comments would be appreciated!

## video

https://user-images.githubusercontent.com/27342882/174431802-c58da02f-9b82-4635-b8de-93914542bfe5.mp4

---


ryuhangyeong:
#8344 Merge from that PR to close the issue.

Issue #8369, opened 2022-06-20
[Bug Report] [Component] [select-v2] el-select-v2 v-model绑定的值始终string


TLovers:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.37`
- Element Plus Version: `2.2.6`
- Browser / OS: `UserAgent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.0.0 Safari/537.36`
- Build Tool: `Vite`

### Related Component

- `el-select-v2`

### Reproduction Link

[...]

PR #8374, opened 2022-06-20
fix(components): [select-v2] value-key

LinZhanMing:
#8369

### DIFF
--- a/packages/components/select-v2/src/select-dropdown.tsx
+++ b/packages/components/select-v2/src/select-dropdown.tsx
@@ -79,7 +79,11 @@ export default defineComponent({
       if (select.props.multiple) {
         return contains(modelValue, get(target, valueKey))
       }
-      return isEqual(get(modelValue, valueKey), get(target.value, valueKey))
+      if (typeof target.value === 'object') {
+        return isEqual(get(modelValue, valueKey), get(target.value, valueKey))
+      } else {
+        return isEqual(modelValue, target.value)
+      }
     }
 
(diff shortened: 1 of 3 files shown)


packages/components/select-v2/src/useSelect.ts
```diff
@@ -674,7 +680,8 @@ const useSelect = (props: ExtractPropTypes<typeof SelectProps>, emit) => {
         states.previousValue = props.modelValue
         const options = filteredOptions.value
         const selectedItemIndex = options.findIndex(
-          (option) => getValueKey(option) === getValueKey(props.modelValue)
+          (option) =>
+            getValueKey(option.value) === getValueKey(props.modelValue)
```
chenxch, line 684: The `value-key !== 'value'` scenario needs to be considered.

packages/components/select-v2/src/select-dropdown.tsx
```diff
@@ -79,7 +79,11 @@ export default defineComponent({
       if (select.props.multiple) {
         return contains(modelValue, get(target, valueKey))
       }
-      return isEqual(modelValue, get(target, valueKey))
+      if (typeof target.value === 'object') {
```
chenxch, line 82: use
```js
import { isObject } from '@element-plus/utils'

if (isObject(target.value)) {
  //..
}
```

tolking:
same like #9638 

I think there is a problem with the current implementation of `value-key` . There was no problem at first, but then the `value-key` was used to obtain the value in the `options`, so the error occurred.

PR #8380, opened 2022-06-21
Songtianxiang

1328572844:
Hello, the problem with this commit is that when I use dynamic height in the onMounted hook, I find that it will not execute successfully, but is overwritten by the initial value. Of course I could also use setTimeout for wrapping, but that's not what I want. So I made the following modifications.


patienry (commented):
Thank you! This issue should have been fixed by #24118.

PR #8400, opened 2022-06-21
refactor(components)!: [switch] remove value prop

sxzz:
Remove `value` prop and `input` event. Since Vue 3 no longer uses `value` and `input`, but `modelValue` and `update:modelValue`. Also they are not public in the documentation either.

### DIFF
--- a/packages/components/switch/src/switch.vue
+++ b/packages/components/switch/src/switch.vue
@@ -145,7 +145,7 @@ const coreStyle = computed<CSSProperties>(() => ({
   width: addUnit(props.width),
 }))
 
-const toggleSwitch = async () => {
+const toggleSwitch = () => {
   checked.value = !checked.value
   rAF(() => {
     input.value!.checked = checked.value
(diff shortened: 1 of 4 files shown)


tolking (approved):

packages/components/switch/src/switch.vue
```diff
@@ -157,85 +146,37 @@ const coreStyle = computed<CSSProperties>(() => ({
...
-  }
-})
-
-const handleChange = () => {
-  const val = checked.value ? props.inactiveValue : props.activeValue
-  emit(UPDATE_MODEL_EVENT, val)
-  emit(CHANGE_EVENT, val)
-  emit(INPUT_EVENT, val)
-  nextTick(() => {
+const toggleSwitch = async () => {
```
holazz, line 149: Useless `async` keyword.

sxzz: Resolved.

holazz (approved):

PR #8445, opened 2022-06-24
feat(components): [message-box] add autofocus attribute

### LINKED ISSUE #8213: [Feature Request] ElMessageBox confirm button style
### Existing Component
Yes

### Component Name
el-message-box

### Description
Hello, I learned that this is not a bug, but now the product department does not agree, after opening messagebox that confirm button is the focus state of the display, so I want to know whether to give a popover property to control of focusing on the first focusable item when opening?

https://github.com/element-plus/element-plus/issues/6773
and https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html DOCUMENT NOT FOUND

YunYouJun:
close #8213

- add `autofocus` attribute (default: `true`)

> The button in the message-box is no longer activated by default after opening, when set `autofocus: false`.

image

---


docs/examples/message-box/alert.vue
```diff
@@ -8,6 +8,7 @@ import type { Action } from 'element-plus'
 
 const open = () => {
   ElMessageBox.alert('This is a message', 'Title', {
+    // autofocus: true,
```
sxzz, line 11: Why add comments here?

sxzz (commented):
Maybe we should add this option to config provider too.

docs/examples/message-box/alert.vue
YunYouJun: It is the first example of a message-box, and since we changed the behavior, I want this property to be quickly discoverable when the user views the example.

YunYouJun:
> Maybe we should add this option to config provider too.

It is possible to do this, but the message-box may be a little different.
The prompt auto-focuses input, which I think many people will want to turn on, while the default alert will want to turn off.

image

They may behave differently, so I think it's better to leave a dedicated property for this purpose.

packages/components/message-box/src/message-box.type.ts
```diff
@@ -62,6 +63,11 @@ export type Callback =
 
 /** Options used in MessageBox */
 export interface ElMessageBoxOptions {
+  /**
+   * auto focus confirm button when open message-box
+   */
+  autofocus?: boolean
```
sxzz, line 69: IMO, `autofocus` should be `true` by default, avoiding a breaking change.

tolking:
Maybe we should not introduce this feature, remove the autofocus is a bad experience for some users.

Maybe we should guide developers who want to change it like this in the docs to change it through styles.

packages/components/message-box/src/index.vue
```diff
@@ -18,7 +18,7 @@
       >
         <el-focus-trap
           loop
-          :trapped="visible"
+          :trapped="autofocus && visible"
```
Giwayume, line 21: You need to trap irregardless, if autofocus is false set `focus-start-el` to `rootRef` instead of disabling the trap.

Giwayume (changes requested):

packages/components/message-box/src/index.vue
Giwayume: (Or `focusStartRef` to `rootRef`)

This will resolve the issues of the button colors changing due to focus, while still keeping the dialog accessible. It will not be read out to a screen reader otherwise.

sxzz (approved):

Giwayume (approved):
In its current state, `autofocus: false` results in the close x being focused, which turns it blue. You may want to explicitly assign focusStartRef to rootRef when atuofocus is false. This will result in no visual change, and when the user tabs the focus will be trapped between the close and ok buttons.

YunYouJun:
@Giwayume Sorry I'm not familiar with a11y, I made some changes, Is this what you mean?

tolking:
[W3C standard](https://www.w3.org/WAI/ARIA/apg/example-index/dialog-modal/alertdialog.html)

Focus is automatically set to the first focusable element inside the dialog

Giwayume:
> @Giwayume Sorry I'm not familiar with a11y, I made some changes, Is this what you mean?

Yes, looks good!

Issue #8467, opened 2022-06-25
[Component] [tabs] When the tab-pane name is 0, the return type is abnormal


chenxch:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.37`
- Element Plus Version: `2.2.6`
- Browser / OS: `UserAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/99.0.4844.51 Safari/537.36`
- Build Tool: `Vite`

### Related Component

- `el-tabs`

### Reproduction Link

[...]

PR #8469, opened 2022-06-25
fix(components): [tabs] name is number 0 become a string

chenxch:

#8467

### DIFF
--- a/packages/components/tabs/src/tab-nav.tsx
+++ b/packages/components/tabs/src/tab-nav.tsx
@@ -308,7 +308,7 @@ const TabNav = defineComponent({
         : null
 
       const tabs = props.panes.map((pane, index) => {
-        const tabName = pane.props.name || pane.index || `${index}`
+        const tabName = pane.props.name ?? pane.index ?? `${index}`
         const closable: boolean = pane.isClosable || props.editable
         pane.index = `${index}`
 
--- a/packages/components/tabs/src/tabs.tsx
+++ b/packages/components/tabs/src/tabs.tsx
@@ -12,6 +12,7 @@ import {
   definePropType,
   isNumber,
   isString,
+  isUndefined,
 } from '@element-plus/utils'
 import { EVENT_CODE, UPDATE_MODEL_EVENT } from '@element-plus/constants'
 import ElIcon from '@element-plus/components/icon'
@@ -35,13 +36,11 @@ export const tabsProps = buildProps({
   },
   activeName: {
     type: [String, Number],
-    default: '',
   },
   closable: Boolean,
   addable: Boolean,
   modelValue: {
     type: [String, Number],
-    default: '',
   },
   editable: Boolean,
   tabPosition: {
@@ -87,7 +86,9 @@ export default defineComponent({
 
     const nav$ = ref<TabNavInstance>()
     const panes = reactive<Record<number, TabsPaneContext>>({})
-    const currentName = ref(props.modelValue || props.activeName || '0')
+    const currentName = ref<TabPanelName>(
+      props.modelValue ?? props.activeName ?? '0'
+    )
 
     const changeCurrentName = (value: TabPanelName) => {
       currentName.value = value
@@ -95,9 +96,9 @@ export default defineComponent({
       emit('tab-change', value)
     }
 
-    const setCurrentName = async (value: TabPanelName) => {
+    const setCurrentName = async (value?: TabPanelName) => {
       // should do nothing.
-      if (currentName.value === value) return
+      if (currentName.value === value || isUndefined(value)) return
 
       try {
         const canLeave = await props.beforeLeave?.(value, currentName.value)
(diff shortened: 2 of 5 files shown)


packages/components/tabs/src/tab-nav.tsx
```diff
@@ -308,7 +309,9 @@ const TabNav = defineComponent({
         : null
 
       const tabs = props.panes.map((pane, index) => {
-        const tabName = pane.props.name || pane.index || `${index}`
+        const tabName = isNumber(pane.props.name)
+          ? pane.props.name
+          : pane.props.name || pane.index || `${index}`
```
tolking, line 314:
```suggestion
       const tabName = pane.props.name ?? pane.index ?? `${index}`
```

Is it more convenient to use `??`, at the same time the `pane.props.name` should not have a default value `''` (the default value should be `index` of TabPane).

chenxch: Yes, I also used it at the beginning `? ?` , the review coach at the time let me use the `function`, and I will adjust it.
Thanks for your review.

chenxch:
@tolking Processed and needs your review again.

ryuhangyeong:
@chenxch CI is broken. Please check again!

chenxch:
> @chenxch CI is broken. Please check again!

I checked now, the local unit test is passed.

sxzz:
Please check TS error.
image

packages/components/tabs/src/tabs.tsx
```diff
@@ -25,7 +25,7 @@ import type { TabsPaneContext } from '@element-plus/tokens'
 import type { ExtractPropTypes } from 'vue'
 import type { Awaitable } from '@element-plus/utils'
 
-export type TabPanelName = string | number
+export type TabPanelName = string | number | undefined
```
tolking, line 28: This type should not include undefined.

packages/tokens/tabs.ts
```diff
@@ -13,7 +13,7 @@ export type TabsPaneContext = UnwrapRef<{
 
 export interface TabsRootContext {
   props: TabsProps
-  currentName: Ref<string | number>
+  currentName: Ref<string | number | undefined>
```
tolking, line 16: This type should not include undefined.

packages/components/tabs/src/tabs.tsx
chenxch: Can I use `as TabPanelName`?

tolking: I do not recommend you do this

When running the function, these function parameters should all have values, the TS does not know this.

```ts
    const setCurrentName = async (value?: TabPanelName) => {
      if (currentName.value === value || isUndefined(value)) return
```

tolking: just setCurrentName and handleTabRemove function

packages/components/tabs/src/tabs.tsx
```diff
@@ -95,9 +96,9 @@ export default defineComponent({
       emit('tab-change', value)
     }
 
-    const setCurrentName = async (value: TabPanelName) => {
+    const setCurrentName = async (value: TabPanelName | undefined) => {
```
tolking, line 99:
```suggestion
    const setCurrentName = async (value?: TabPanelName) => {
```

Why not?

chenxch: This is taken into account when modifying, but there is no handling of undefined here, so another method is chosen. I will adjust.

packages/components/tabs/src/tabs.tsx
```diff
@@ -71,7 +70,7 @@ export const tabsEmits = {
   'tab-change': (name: TabPanelName) => isPanelName(name),
   edit: (paneName: TabPanelName | undefined, action: 'remove' | 'add') =>
     ['remove', 'add'].includes(action),
-  'tab-remove': (name: TabPanelName) => isPanelName(name),
+  'tab-remove': (name?: TabPanelName) => isPanelName(name),
```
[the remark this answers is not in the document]
chenxch: After canceling, it will cause an error message on the handleTabRemove side.

One of the methods should be used.
```ts
const handleTabRemove = (pane: TabsPaneContext, ev: Event) => {
    if (pane.props.disabled) return
   // if (pane.props.disabled || isUndefined(pane.props.name)) return
    ev.stopPropagation()
    emit('edit', pane.props.name, 'remove')
    emit('tab-remove', pane.props.name)
   // !isUndefined(pane.props.name) && emit('tab-remove', pane.props.name)
  }
```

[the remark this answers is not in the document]
tolking: same with setCurrentName

`if (pane.props.disabled || isUndefined(pane.props.name)) return`

tolking (approved):

chenxch:
@tolking Do I need to do rebase processing?

tolking:
Commit info can be modified by merge PR. If, You need to modify it and you can merge it yourself

chenxch:
OK, I've merged the commits via rebase.

PR #8529, opened 2022-06-29
refactor(components): [popper] fix type error

holazz:

### DIFF
--- a/packages/components/popper/src/arrow.ts
+++ b/packages/components/popper/src/arrow.ts
@@ -12,3 +12,12 @@ export const popperArrowProps = buildProps({
 export type PopperArrowProps = ExtractPropTypes<typeof popperArrowProps>
 
 export type PopperArrowInstance = InstanceType<typeof Arrow>
+
+/** @deprecated use `popperArrowProps` instead, and it will be deprecated in the next major version */
+export const usePopperArrowProps = popperArrowProps
+
+/** @deprecated use `PopperArrowProps` instead, and it will be deprecated in the next major version */
+export type UsePopperArrowProps = PopperArrowProps
+
+/** @deprecated use `PopperArrowInstance` instead, and it will be deprecated in the next major version */
+export type ElPopperArrowInstance = PopperArrowInstance
--- a/packages/components/popper/src/arrow.ts
+++ b/packages/components/popper/src/arrow.ts
@@ -1,12 +1,14 @@
 import { buildProps } from '@element-plus/utils'
 
 import type { ExtractPropTypes } from 'vue'
+import type Arrow from './arrow.vue'
 
-export const usePopperArrowProps = buildProps({
+export const popperArrowProps = buildProps({
   arrowOffset: {
     type: Number,
     default: 5,
   },
 } as const)
+export type PopperArrowProps = ExtractPropTypes<typeof popperArrowProps>
 
-export type UsePopperArrowProps = ExtractPropTypes<typeof usePopperArrowProps>
+export type PopperArrowInstance = InstanceType<typeof Arrow>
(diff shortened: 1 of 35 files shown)


packages/components/popper/src/content.vue
```diff
@@ -218,12 +226,12 @@ onMounted(() => {
       if (isElement(el)) {
         const { ariaLabel, id } = toRefs(props)
         triggerTargetAriaStopWatch = watch(
-          [role, ariaLabel, ariaModal, id],
+          [role, ref(ariaLabel), ariaModal, ref(id)],
```
tolking, line 229: I do not quite understand why `ref(xx)` is used.

tolking, line 229: I do not quite understand why `ref(xx)` is used.

holazz: Because `ariaLabel ` and `id` may be `undefined`, this array is an invalid watch source. Although the result is not wrong, it will cause ts type errors.

holazz: like this:
WX20220703-131325@2x

tolking: I think it maybe caused by the `toRefs`

```ts
const { ariaLabel, id } = toRefs(props)
// id: Ref<string | undefined> | undefined
// props.id: string | undefined
```

holazz: maybe change it to:
```ts
watch(
  [role, () => props.ariaLabel, ariaModal, () => props.id],
  (watches) => {
    ...
  },
  { immediate: true }
)
```
what do you think?

tolking: This should work fine

holazz: updated.

packages/components/popper/src/arrow.ts
```diff
@@ -1,12 +1,14 @@
...
+export const popperArrowProps = buildProps({
   arrowOffset: {
     type: Number,
     default: 5,
   },
 } as const)
+export type PopperArrowProps = ExtractPropTypes<typeof popperArrowProps>
 
-export type UsePopperArrowProps = ExtractPropTypes<typeof usePopperArrowProps>
+export type PopperArrowInstance = InstanceType<typeof Arrow>
```
tolking, line 14: It looks like you modified the export name of the types, I suggest add deprecated to the previous name.

// reference

```ts
export type PopperArrowInstance = InstanceType<typeof Arrow>

/** @deprecated use `PopperArrowInstance` instead, and it will be deprecated in the next major version */
export type ElPopperArrowInstance = PopperArrowInstance
```

packages/components/popper/src/arrow.ts
```diff
@@ -1,12 +1,14 @@
 import { buildProps } from '@element-plus/utils'
 
 import type { ExtractPropTypes } from 'vue'
+import type Arrow from './arrow.vue'
 
-export const usePopperArrowProps = buildProps({
+export const popperArrowProps = buildProps({
```
tolking, line 6: This name change seems to be a major changes

holazz:
@tolking updated, please review again.

packages/components/popper/src/content.vue
```diff
@@ -112,7 +111,15 @@ const ariaModal = computed<string | undefined>(() => {
...
-const createPopperInstance = ({ referenceEl, popperContentEl, arrowEl }) => {
+const createPopperInstance = ({
+  referenceEl,
+  popperContentEl,
+  arrowEl,
+}: {
+  referenceEl: Measurable
+  popperContentEl: HTMLElement
+  arrowEl: HTMLElement | undefined
+}) => {
```
jw-foss, line 122: **Suggestion:** can we move these types into a local type such as `CreatePopperInstanceParams`?

jw-foss (commented):
@holazz in general the changes LGTM, I left a non-blocking comment for you to consider, back to you 🏓

packages/components/popper/src/content.vue
holazz: Updated.

sxzz (approved):

PR #8558, opened 2022-06-30
fix(hooks): [use-attrs] fix type error

holazz:

### DIFF
--- a/packages/hooks/__tests__/use-attrs.test.tsx
+++ b/packages/hooks/__tests__/use-attrs.test.tsx
@@ -20,7 +20,6 @@ const genComp = (
 ) => {
   return defineComponent({
     inheritAttrs,
-    props: {} as Record<string, any>,
     setup() {
       const attrs = useAttrs({ excludeListeners, excludeKeys })
       return () => (
--- a/packages/hooks/__tests__/use-attrs.test.tsx
+++ b/packages/hooks/__tests__/use-attrs.test.tsx
@@ -20,6 +20,7 @@ const genComp = (
 ) => {
   return defineComponent({
     inheritAttrs,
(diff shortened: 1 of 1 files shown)


packages/hooks/__tests__/use-attrs.test.tsx
```diff
@@ -21,7 +20,7 @@ const genComp = (
 ) => {
   return defineComponent({
     inheritAttrs,
-    props: {},
+    props: {} as Record<string, any>,
```
tolking, line 23: Maybe the `props` and `emits` can be written more standardized

holazz: Seems like this line is not needed.
```diff
- props: {} as Record<string, any>,
```

tolking: eg.

```ts
props: {
      [TEST_KEY]: String,
    },
    emits: ['click'],
```

holazz: IMO, in order to test `attrs`, we shouldn't explicitly declare it on component `props` and `emit`. If we do as above, the `[TEST_KEY]` will be treated as `props` instead of `attrs`. Is my understanding correct?

tolking: you are right

tolking (approved):

tolking:
Am I the only one who will show up?

[](https://imgloc.com/image/Czpr)

holazz:
Not sure why I don't have the type error. But when I run it on gitpod, I do get this error. I have fixed it now.

tolking (approved):

PR #8762, opened 2022-07-11
feat(components): [tooltip] adjust controlled mode

chenxch:

`v-model:visible` as a bidirectional parameter does not seem very reasonable to use it as a controlled condition, `:visible` is more suitable.
In the original code, as long as the `:visible` parameter is judged to be used, the controlled mode is turned on. I think `v-model:visible` should not be in the uncontrolled mode. It can coexist with the `trigger` and can be simply used to judge whether the `tooltip` is displayed.

### Related issues
#8696 

### Of course, I think this change is a destructive change. In addition to the change in the controlled mode, the components involved in the tooltip have been adjusted accordingly to keep them under control. The following is the list of components found so far:

- [x] tooltip
- [x] popover
- [x] time-picker
- [x] dropdown

### DIFF
--- a/packages/components/popover/src/popover.ts
+++ b/packages/components/popover/src/popover.ts
@@ -4,7 +4,7 @@ import {
   useTooltipTriggerProps,
 } from '@element-plus/components/tooltip'
 import { dropdownProps } from '@element-plus/components/dropdown'
-import type { ExtractPropTypes } from 'vue'
+import type { ExtractPropTypes, PropType } from 'vue'
 import type Popover from './popover.vue'
 
 export const popoverProps = buildProps({
@@ -56,6 +56,9 @@ export const popoverProps = buildProps({
   persistent: {
     type: Boolean,
     default: true,
+  },
+  'onUpdate:visible': {
+    type: Function as PropType<(visible: boolean) => void>,
   },
 } as const)
 export type PopoverProps = ExtractPropTypes<typeof popoverProps>
--- a/packages/components/popover/src/popover.vue
+++ b/packages/components/popover/src/popover.vue
@@ -23,6 +23,7 @@
     :teleported="teleported"
     :persistent="persistent"
     :gpu-acceleration="gpuAcceleration"
+    @update:visible="onUpdateVisible"
     @before-show="beforeEnter"
     @before-hide="beforeLeave"
     @show="afterEnter"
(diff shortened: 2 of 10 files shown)


YunYouJun (approved):

packages/components/popover/src/popover.ts
```diff
@@ -57,6 +57,9 @@ export const popoverProps = buildProps({
     type: Boolean,
     default: true,
   },
+  'onUpdate:visible': {
+    type: Function as PropType<(visible: boolean) => void>,
+  },
```
buqiyuan, line 62: image
It seems to be repeated here.

packages/components/popover/src/popover.vue
```diff
@@ -57,6 +58,12 @@ defineOptions({
 const props = defineProps(popoverProps)
 const emit = defineEmits(popoverEmits)
 
+const updateEventKeyRaw = `onUpdate:visible` as const
+
+const onUpdateVisible = computed(() => {
+  return props[updateEventKeyRaw]
+})
```
buqiyuan, line 65: 🤔Is it okay to use `emit('update:visible', xxx)` here?

buqiyuan (approved):

packages/components/popover/src/popover.ts
chenxch: Here I read what is written in the `tooltip`, and I will test whether the deletion will affect it.

chenxch: Confirmation is required, otherwise it cannot be judged whether visible is a bidirectional attribute.

buqiyuan:
```ts
import { useAttrs } from 'vue'

const attrs = useAttrs()
```

Maybe you can use `attrs.onUpdate:visible` to determine whether it is passed in

chenxch: let me try.

buqiyuan: You need to remove `onUpdate:visible` in `props` first

chenxch: It is operated on the premise of removal.

chenxch: image

chenxch: image

chenxch:
@buqiyuan Hi, I will close all conversations if there are no more questions.

buqiyuan:
> @buqiyuan Hi, I will close all conversations if there are no more questions.

Okay!

tolking:
This change affects the tooltip that Select component, Maybe all tooltip-based components may have this problem.

chenxch:
> This change affects the tooltip that Select component, Maybe all tooltip-based components have this problem.

I'm so sorry, I neglected to check this part of the component.

chenxch:
I will retrieve the tooltip in the repository later.

exherb:
oh... we update yesterday and encounter this.

chenxch:
> oh... we update yesterday and encounter this.

I am very sorry for affecting your use, please use version `2.2.9` first.

chenxch:
> tomkobenedikt4

what is the problem?

PR #8780, opened 2022-07-12
fix(test-utils): fix type error

holazz:
#6114

### DIFF
--- a/packages/test-utils/make-mount.ts
+++ b/packages/test-utils/make-mount.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 import { mount } from '@vue/test-utils'
 import { merge } from 'lodash'
 
@@ -12,11 +11,11 @@ interface Options {
     [key: string]: any
   }
   methods?: {
-    [key: string]: (...args) => any
+    [key: string]: (...args: any[]) => any
   }
 }
 
-export const makeMountFunc = (defaultOptions) => {
+export const makeMountFunc = (defaultOptions: Record<string, any>) => {
   return (template: string, options: Options) => {
     return mount({
       ...merge({}, defaultOptions, options),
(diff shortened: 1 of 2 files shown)


packages/test-utils/make-mount.ts
```diff
@@ -12,11 +11,11 @@ interface Options {
     [key: string]: any
   }
   methods?: {
-    [key: string]: (...args) => any
+    [key: string]: (...args: any[]) => any
```
tolking, line 14: why not use `unknown`

holazz: Updated. 

Sorry, I didn't think carefully here, because I think this file will be removed in the future.

packages/test-utils/make-mount.ts
```diff
@@ -12,11 +11,11 @@ interface Options {
     [key: string]: any
   }
   methods?: {
-    [key: string]: (...args) => any
+    [key: string]: (...args: unknown[]) => any
   }
 }
 
-export const makeMountFunc = (defaultOptions) => {
+export const makeMountFunc = (defaultOptions: Record<string, any>) => {
```
tolking, line 18:
```suggestion
export const makeMountFunc = <T extends Record<string, unknown>>(defaultOptions: T) => {
```

Maybe use generics

packages/test-utils/make-mount.ts
```diff
@@ -12,11 +11,11 @@ interface Options {
     [key: string]: any
```
tolking, line 11: It can also be change to unknown here

packages/test-utils/make-mount.ts
holazz: Okay!

tolking (approved):

PR #8850, opened 2022-07-17
fix: [timer-pick] automatically skip disabled items

### LINKED ISSUE #8823: [Component] [cascader] el-time-picker 设置了 arrow-control 和 disabled-hours [1,2,4,5] 会导致卡在3不能上下点击
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.25`
- Element Plus Version: `2.2.2`
- Browser / OS: `mac`
- Build Tool: `Vite`

### Related Component

- `el-cascader`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZGl2IGNsYXNzPVwiZXhhbXBsZS1iYXNpY1wiPlxuICAgIDxlbC10aW1lLXBpY2tlclxuICAgICAgdi1tb2RlbD1cInZhbHVlMVwiXG4gICAgICA6ZGlzYWJsZWQtaG91cnM9XCJkaXNhYmxlZEhvdXJzXCJcbiAgICAgIDpkaXNhYmxlZC1taW51dGVzPVwiZGlzYWJsZWRNaW51dGVzXCJcbiAgICAgIDpkaXNhYmxlZC1zZWNvbmRzPVwiZGlzYWJsZWRTZWNvbmRzXCJcbiAgICAgIHBsYWNlaG9sZGVyPVwiQXJiaXRyYXJ5IHRpbWVcIlxuICAgICAgICAgICAgICAgICAgICBhcnJvdy1jb250cm9sXG4gICAgLz5cbiAgPC9kaXY+XG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IGxhbmc9XCJ0c1wiIHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSAndnVlJ1xuXG5jb25zdCB2YWx1ZTEgPSByZWYobmV3IERhdGUoMjAxNiwgOSwgMTAsIDE4LCAzMCkpXG5cbmNvbnN0IG1ha2VSYW5nZSA9IChzdGFydDogbnVtYmVyLCBlbmQ6IG51bWJlcikgPT4ge1xuICBjb25zdCByZXN1bHQ6IG51bWJlcltdID0gW11cbiAgZm9yIChsZXQgaSA9IHN0YXJ0OyBpIDw9IGVuZDsgaSsrKSB7XG4gICAgcmVzdWx0LnB1c2goaSlcbiAgfVxuICByZXR1cm4gcmVzdWx0XG59XG5jb25zdCBkaXNhYmxlZEhvdXJzID0gKCkgPT4ge1xuICByZXR1cm4gbWFrZVJhbmdlKDAsIDE2KS5jb25jYXQobWFrZVJhbmdlKDE5LCAyMykpXG59XG5jb25zdCBkaXNhYmxlZE1pbnV0ZXMgPSAoaG91cjogbnVtYmVyKSA9PiB7XG4gIGlmIChob3VyID09PSAxNykge1xuICAgIHJldHVybiBtYWtlUmFuZ2UoMCwgMjkpXG4gIH1cbiAgaWYgKGhvdXIgPT09IDE4KSB7XG4gICAgcmV0dXJuIG1ha2VSYW5nZSgzMSwgNTkpXG4gIH1cbn1cbmNvbnN0IGRpc2FibGVkU2Vjb25kcyA9IChob3VyOiBudW1iZXIsIG1pbnV0ZTogbnVtYmVyKSA9PiB7XG4gIGlmIChob3VyID09PSAxOCAmJiBtaW51dGUgPT09IDMwKSB7XG4gICAgcmV0dXJuIG1ha2VSYW5nZSgxLCA1OSlcbiAgfVxufVxuPC9zY3JpcHQ+XG5cbjxzdHlsZT5cbi5leGFtcGxlLWJhc2ljIC5lbC1kYXRlLWVkaXRvciB7XG4gIG1hcmdpbjogOHB4O1xufVxuPC9zdHlsZT5cbiIsImltcG9ydF9tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHt9XG59IiwiX28iOnt9fQ==)

### Steps to reproduce

 设置了 arrow-control 和 disabled-hours [1,2,4,5] 

### What is Expected?

点击跳转到下一个不是disable的时间

卡住，不能跳转

### Additional comments

_(empty)_

baiwusanyu-c:
close: #8823 
add prop auto-skip-disabled. When `arrow-control` is equal to `true`, click the arrow to automatically skip the disabled item


baiwusanyu-c:
踢一脚

tolking:
I think the disabled item should be skipped by default instead of adding an attribute to control it. 

How do you like it?

baiwusanyu-c:
> I think the disabled item should be skipped by default instead of adding an attribute to control it.
> 
> How do you like it?

yes i think you're right, if as an attribute it means the user has to use it for the component to work properly; well, i'll fix that later

baiwusanyu-c:
> I think the disabled item should be skipped by default instead of adding an attribute to control it.
> 
> How do you like it?

yes i think you're right, if as an attribute it means the user has to use it for the component to work properly; well, i'll fix that later

baiwusanyu-c:
> I think the disabled item should be skipped by default instead of adding an attribute to control it.
> 
> How do you like it?

I have made changes. Can you review it for me

packages/components/time-picker/src/time-picker-com/basic-time-spinner.vue
```diff
@@ -233,11 +233,38 @@ const scrollDown = (step: number) => {
   const total = currentScrollbar.value === 'hours' ? 24 : 60
   now = (now + step + total) % total
 
+  if (props.arrowControl) {
```
tolking, line 236: Maybe it should not be restricted to only `arrowControl`, it should be the default behavior.

eg, There are still problems with using the arrow keys in the keyboard

packages/components/time-picker/src/time-picker-com/basic-time-spinner.vue
```diff
@@ -233,11 +233,38 @@ const scrollDown = (step: number) => {
...
   nextTick(() => emitSelectRange(label))
 }
-
+const findNextUnDisabled = (
+  type: TimeUnit,
+  value: number,
+  step: number,
+  total: number
+) => {
+  let res: number = value
```
tolking, line 249:
```suggestion
  let res = value
```

Such type declarations can be ignored

tolking (commented):
I left some comments for your reference

baiwusanyu-c:
> I left some comments for your reference

Thanks, I made the revision again

packages/components/time-picker/src/time-picker-com/basic-time-spinner.vue
```diff
@@ -232,12 +232,36 @@ const scrollDown = (step: number) => {
   let now = unref(timePartials)[label]
   const total = currentScrollbar.value === 'hours' ? 24 : 60
   now = (now + step + total) % total
-
+  now = findNextUnDisabled(label, now, step, total)
```
tolking, line 235: Try to avoid assigning `now` twice

packages/components/time-picker/src/time-picker-com/basic-time-spinner.vue
```diff
@@ -232,12 +232,36 @@ const scrollDown = (step: number) => {
...
+  type: TimeUnit,
+  value: number,
+  step: number,
+  total: number
+) => {
+  let res = value
+  const list = unref(timeList)[type]
+  let isDisabled: boolean = list[value]
+  if (isDisabled) {
+    const len: number = list.length
```
tolking, line 250:
```suggestion
```

The `len` seems redundant, it is always the same value as `total`

packages/components/time-picker/src/time-picker-com/basic-time-spinner.vue
```diff
@@ -232,12 +232,36 @@ const scrollDown = (step: number) => {
...
-
+const findNextUnDisabled = (
+  type: TimeUnit,
+  value: number,
+  step: number,
+  total: number
+) => {
+  let res = value
+  const list = unref(timeList)[type]
+  let isDisabled: boolean = list[value]
```
tolking, line 248:
```suggestion
  let isDisabled = list[value]
```

Same as before

packages/components/time-picker/src/time-picker-com/basic-time-spinner.vue
```diff
@@ -232,12 +232,36 @@ const scrollDown = (step: number) => {
...
+  if (isDisabled) {
+    const len: number = list.length
+    while (isDisabled) {
```
tolking, line 251: The judgment conditions are not comprehensive enough. When all options are not selectable, an infinite loop will be entered.

packages/components/time-picker/src/time-picker-com/basic-time-spinner.vue
```diff
@@ -232,12 +232,36 @@ const scrollDown = (step: number) => {
...
+      if (!list[value]) {
+        isDisabled = false
+        res = value
+        break
+      }
+      value = (value + step + total) % total
+      if (value >= len) {
+        value = 0
+      }
```
tolking, line 260:
```suggestion
```

This seems redundant. Because of the `%` operator, the `value` is always less than the `total`.

packages/components/time-picker/src/time-picker-com/basic-time-spinner.vue
tolking: This infinite loop problem still exists

baiwusanyu-c: > This infinite loop problem still exists
image
When all options are disabled, I use set to make a judgment and return directly. Are there other scenarios that will lead to infinite loops?

tolking: Looks like you are overcomplicating things.

Maybe adjust the parameter names to make it easier to understand

```
value -> now
res -> next
```

They are both indexes of the list array, and when they are equal again, it means that all options have been traversed.

packages/components/time-picker/src/time-picker-com/basic-time-spinner.vue
```diff
@@ -231,13 +231,31 @@ const scrollDown = (step: number) => {
...
+  if (isDisabled) {
+    while (isDisabled) {
+      if (!list[next] || next === now) {
+        isDisabled = false
+        break
+      }
+      next = (next + step + total) % total
+    }
+  }
+  return next
```
tolking, line 257:
```suggestion
  const list = unref(timeList)[type]
  let next = (now + step + total) % total

  while (list[next] && next !== now) {
    next = (next + step + total) % total
  }
  return next
```

packages/components/time-picker/src/time-picker-com/basic-time-spinner.vue
```diff
@@ -231,13 +231,31 @@ const scrollDown = (step: number) => {
...
   let now = unref(timePartials)[label]
   const total = currentScrollbar.value === 'hours' ? 24 : 60
-  now = (now + step + total) % total
-
+  now = findNextUnDisabled(label, now, step, total)
   modifyDateField(label, now)
   adjustSpinner(label, now)
   nextTick(() => emitSelectRange(label))
 }
-
```
tolking, line 240: Maybe you can refer to the format of the current file and keep the newlines at both ends of the function.

packages/components/time-picker/src/time-picker-com/basic-time-spinner.vue
```diff
@@ -231,13 +231,31 @@ const scrollDown = (step: number) => {
   const label = currentScrollbar.value!
   let now = unref(timePartials)[label]
   const total = currentScrollbar.value === 'hours' ? 24 : 60
-  now = (now + step + total) % total
-
+  now = findNextUnDisabled(label, now, step, total)
   modifyDateField(label, now)
   adjustSpinner(label, now)
```
tolking, line 236:
``` suggestion
  const now = unref(timePartials)[label]
  const total = currentScrollbar.value === 'hours' ? 24 : 60
  const next = findNextUnDisabled(label, now, step, total)

  modifyDateField(label, next)
  adjustSpinner(label, next)
```

packages/components/time-picker/__tests__/time-picker.test.ts
```diff
@@ -835,4 +836,62 @@ describe('TimePicker(range)', () => {
     expect(startInput.element.value).toBe('')
     expect(endInput.element.value).toBe('')
   })
+
+  it('can auto skip when arrow control', async () => {
```
tolking, line 840: This test probably should not be placed inside `describe('TimePicker(range)')`.

Maybe move it to the `describe('TimePicker')`

packages/components/time-picker/__tests__/time-picker.test.ts
```diff
@@ -835,4 +836,62 @@ describe('TimePicker(range)', () => {
...
+    const hoursElArrowDown = hoursElWrapper.querySelector('.arrow-down')
+
+    const mousedownEvt = new MouseEvent('mousedown')
+    const mouseupEvt = new MouseEvent('mouseup')
+
+    const testTime = 130
+    ;(hoursElArrowDown as any).dispatchEvent(mousedownEvt)
+    ;(hoursElArrowDown as any).dispatchEvent(mouseupEvt)
+    await sleep(testTime)
+    await nextTick()
```
tolking, line 881:
```ts
    await sleep(testTime)
    await nextTick()
```

Why do you need to use both?

packages/components/time-picker/__tests__/time-picker.test.ts
```diff
@@ -835,4 +836,62 @@ describe('TimePicker(range)', () => {
...
+    const list = document.querySelectorAll('.el-time-spinner__list')
+    const hoursEl = list[0]
+    let activeHours = getSpinnerTextAsArray(hoursEl, '.is-active')[0]
+
+    expect(activeHours).toEqual(20)
+    const hoursElWrapperList = document.querySelectorAll(
+      '.el-time-spinner__wrapper'
+    )
+    const hoursElWrapper = hoursElWrapperList[0]
+    const hoursElArrowDown = hoursElWrapper.querySelector('.arrow-down')
```
tolking, line 872: The `hoursElArrowDown` type is `Element | null`, Maybe `any` should be avoided.

tolking (approved):
LGTM

PR #8898, opened 2022-07-19
fix(components): [dialog] fix type error

Chuck-Lau:
#6114 

### DIFF
--- a/packages/components/dialog/src/use-dialog.ts
+++ b/packages/components/dialog/src/use-dialog.ts
@@ -44,13 +44,13 @@ export const useDialog = (
 
   const style = computed<CSSProperties>(() => {
     const style: CSSProperties = {}
-    const varPrefix = `${namespace.value}-dialog`
+    const varPrefix = `--${namespace.value}-dialog` as const
     if (!props.fullscreen) {
       if (props.top) {
-        style[`--${varPrefix}-margin-top`] = props.top
+        style[`${varPrefix}-margin-top`] = props.top
       }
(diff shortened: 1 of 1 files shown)


packages/components/dialog/src/use-dialog.ts
```diff
@@ -45,13 +44,13 @@ export const useDialog = (
 
   const style = computed<CSSProperties>(() => {
     const style: CSSProperties = {}
-    const varPrefix = `--${namespace.value}-dialog`
+    const varPrefix = `${namespace.value}-dialog`
```
holazz, line 47: I think it's better to use the const assertion directly here.

```ts
const varPrefix = `--${namespace.value}-dialog` as const
```

Tsong-LC: What about using `module enhancements`?

Tsong-LC: image

holazz: > I think it's better to use the const assertion directly here.
> 
> ```ts
> const varPrefix = `--${namespace.value}-dialog` as const
> ```

The const assertion here is to narrow the type `string` to ``` `--${string}` ```, what do you mean?

Tsong-LC: Updated. Thanks for your advice

ryuhangyeong (approved):

tolking (approved):

PR #8905, opened 2022-07-19
fix(hooks): [use-namespace] exclude invalid CSS variables

tolking:

Fix style error in SSR/SSG, eg.

```html
<ul role="menubar" style="--el-menu-text-color:;--el-menu-hover-text-color:;--el-menu-bg-color:;--el-menu-hover-bg-color:;--el-menu-active-color:;--el-menu-level:0;" class="el-menu el-menu--vertical pro-menu">
```

### DIFF
--- a/packages/hooks/use-namespace/index.ts
+++ b/packages/hooks/use-namespace/index.ts
@@ -62,7 +62,9 @@ export const useNamespace = (block: string) => {
   const cssVar = (object: Record<string, string>) => {
     const styles: Record<string, string> = {}
     for (const key in object) {
-      styles[`--${namespace.value}-${key}`] = object[key]
+      if (object[key]) {
+        styles[`--${namespace.value}-${key}`] = object[key]
+      }
     }
     return styles
   }
@@ -70,7 +72,9 @@ export const useNamespace = (block: string) => {
   const cssVarBlock = (object: Record<string, string>) => {
     const styles: Record<string, string> = {}
     for (const key in object) {
-      styles[`--${namespace.value}-${block}-${key}`] = object[key]
+      if (object[key]) {
+        styles[`--${namespace.value}-${block}-${key}`] = object[key]
+      }
     }
     return styles
   }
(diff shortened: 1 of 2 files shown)


packages/hooks/use-namespace/index.ts
```diff
@@ -62,15 +62,19 @@ export const useNamespace = (block: string) => {
   const cssVar = (object: Record<string, string>) => {
     const styles: Record<string, string> = {}
     for (const key in object) {
-      styles[`--${namespace.value}-${key}`] = object[key]
+      if (object[key]) {
```
YunYouJun, line 65: I remember that vue automatically determines whether it is empty or not.

Of course it's good to add it.

YunYouJun (approved):

btea (approved):

packages/hooks/use-namespace/index.ts
```diff
@@ -62,15 +62,19 @@ export const useNamespace = (block: string) => {
...
+      }
     }
     return styles
   }
   // with block
   const cssVarBlock = (object: Record<string, string>) => {
     const styles: Record<string, string> = {}
     for (const key in object) {
-      styles[`--${namespace.value}-${block}-${key}`] = object[key]
+      if (object[key]) {
```
plainheart, line 75: Just noticed a potential issue -
Will it work if the value is number `0` rather than string `0`?

tolking: The accepted type is string, and passing the number 0 directly will cause a TS error.

plainheart: It will be compiled to be js file finally.

tolking: The internal use of the component library should not cause problems unless the user calls the function himself.

plainheart: > unless the user calls the function himself.

This is what I'm concerned about. This function has been exported and is public for the users. It may cause unnecessary issues if they use JS rather than TS to develop their projects. We can either fix it or wait for the feedback from the users.

PR #8981, opened 2022-07-23
fix(components): [table] headerWrapper may be undefined

### LINKED ISSUE #8977: [Component] [table] 销毁Table元素后报错 'offsetHeight' of null
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.37`
- Element Plus Version: `2.2.10`
- Browser / OS: `Chrome/103.0.5060.134`
- Build Tool: `Vue CLI`

### Related Component

- `el-table`

### Reproduction Link

[https://element-plus.gitee.io/zh-CN/component/drawer.html#自定义内容](https://element-plus.gitee.io/zh-CN/component/drawer.html#自定义内容)

### Steps to reproduce

官方示例，Drawer开启**destroy-on-close**后显示Table，关闭后即可复现

### What is Expected?

消除报错信息

报错

### Additional comments

_(empty)_

image

holazz:
closes #8977

### DIFF
--- a/packages/components/table/src/table/style-helper.ts
+++ b/packages/components/table/src/table/style-helper.ts
@@ -130,8 +130,7 @@ function useStyle<T>(
     resizeState.value = {
       width: (tableWidth.value = el.offsetWidth),
       height: el.offsetHeight,
-      headerHeight:
-        props.showHeader && tableHeader ? tableHeader.offsetHeight : null,
+      headerHeight: tableHeader?.offsetHeight,
     }
 
     // init filters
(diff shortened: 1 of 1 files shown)


tolking (commented):
It seems unnecessary to execute the resizeListener function after the Table has be destroyed

plainheart:
@tolking @holazz It seems #9011 also fixes the issue that the header wrapper may be undefined. And as per the source code, both `useEventListener` and `useResizeObserver`  will try to destroy themselves when their scope is disposed of. So it may be unnecessary to dispose of them manually. They can be disposed of after the component was destroyed in fact. But one thing to be noticed is that the resize observer will also trigger the callback after its target element was removed, which may be the reason why the NPE occurs.

holazz:
> And as per the source code, both `useEventListener` and `useResizeObserver` will try to destroy themselves when their scope is disposed of. So it may be unnecessary to dispose of them manually. They can be disposed of after the component was destroyed in fact.

Yes, but I haven't figured out why the callback is triggered after the target element was removed.

> But one thing to be noticed is that the resize observer will also trigger the callback after its target element was removed, which may be the reason why the NPE occurs.

And manually stopping the `ResizeObserver` seems to avoid triggering the callback after the target element was removed.

plainheart:
> Yes, but I haven't figured out why the callback is triggered after the target element was removed.

It seems reasonable that the resize callback will be triggered once after the target element was removed, as it has been removed and so its size becomes 0. See also [the description on W3C](https://www.w3.org/TR/resize-observer/#resize-observer-callback:~:text=Observation%20will%20fire%20when%20watched%20Element%20is%20inserted/removed%20from%20DOM.)

holazz:
> > Yes, but I haven't figured out why the callback is triggered after the target element was removed.
> 
> It seems reasonable that the resize callback will be triggered once after the target element was removed, as it has been removed and so its size becomes 0. See also [the description on W3C](https://www.w3.org/TR/resize-observer/#resize-observer-callback:~:text=Observation%20will%20fire%20when%20watched%20Element%20is%20inserted/removed%20from%20DOM.)

I just saw this too, but I'm still confused why this doesn't work for other components that use `useResizeObserver`. such as `el-menu`, `el-select-v2`.(I have tested these two components, and found that the callback isn't triggered after the target element was removed) 🤔

PR #9087, opened 2022-07-28
style(message): Message support adaptive width

kooriookami:
image

### DIFF
--- a/packages/theme-chalk/src/message.scss
+++ b/packages/theme-chalk/src/message.scss
@@ -9,7 +9,6 @@
 }
 
 @include b(message) {
-  width: getCssVar('message', 'width');
   max-width: calc(100% - 32px);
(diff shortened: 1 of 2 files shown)


xiaoxian521 (approved):

packages/theme-chalk/src/message.scss
```diff
@@ -55,7 +53,7 @@
 
       .#{$namespace}-message__content {
         color: getCssVar('message', 'text-color');
-        word-break: break-all;
+        overflow-wrap: anywhere;
```
tolking, line 56: Maybe affect compatibility

[link](https://caniuse.com/?search=overflow-wrap%3A%20anywhere)

YunYouJun:
Adaptive should be an optional option, the default may still be fixed width, otherwise it is a break change.

PR #9151, opened 2022-08-01
refactor(hooks): [use-lockscreen] change from padding to width

chenxch:

Using the paddingright method, it has been found in the document that page jittering will occur, so it is changed to width. The original method is calc(100% - scrollbarWidth), but the user can change the body to a non-100% width, so the calculation of the width of the body is used.

### DIFF
--- a/packages/hooks/use-lockscreen/index.ts
+++ b/packages/hooks/use-lockscreen/index.ts
@@ -38,7 +38,6 @@ export const useLockscreen = (trigger: Ref<boolean>) => {
   let scrollBarWidth = 0
   let withoutHiddenClass = false
   let bodyWidth = '0'
-  let computedBodyWidth = Number.parseInt(getStyle(document.body, 'width'), 10)
 
   const cleanup = () => {
     removeClass(document.body, hiddenCls.value)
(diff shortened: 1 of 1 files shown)


packages/hooks/use-lockscreen/index.ts
```diff
@@ -37,13 +37,13 @@ export const useLockscreen = (trigger: Ref<boolean>) => {
...
-  let bodyPaddingRight = '0'
-  let computedBodyPaddingRight = 0
+  let bodyWidth = '0'
+  let computedBodyWidth = Number.parseInt(getStyle(document.body, 'width'), 10)
 
   const cleanup = () => {
     removeClass(document.body, hiddenCls.value)
     if (withoutHiddenClass) {
-      document.body.style.paddingRight = bodyPaddingRight
+      document.body.style.width = bodyWidth
```
tolking, line 46: A fixed width is probably not a good idea. For example, after open a Dialog, the body width is fixed. At this time, the body width will not change with the width of the browser, but will change when the Dialog closed.

chenxch: Indeed, this causes it to lose its responsiveness, can `calc(100% - scrollbarWidth)` still be optimized. 
e.g. `100%` = (bodyWidth + scrollbarWidth) / document.width

tolking (approved):

PR #9171, opened 2022-08-02
fix(components): [switch] fix type error

Chuck-Lau:
#6114 

### DIFF
--- a/packages/components/switch/__tests__/switch.test.tsx
+++ b/packages/components/switch/__tests__/switch.test.tsx
@@ -1,4 +1,3 @@
-// @ts-nocheck
 import { markRaw, nextTick, ref } from 'vue'
 import { mount } from '@vue/test-utils'
 import { afterEach, describe, expect, test, vi } from 'vitest'
@@ -111,8 +110,8 @@ describe('Switch.vue', () => {
   test('change event', async () => {
     const target = ref(1)
     const value = ref(true)
-    const handleChange = (val: boolean) => {
-      target.value = val
+    const handleChange = (val: string | number | boolean) => {
+      target.value = val as number
     }
     const wrapper = mount(() => (
       <Switch v-model={value.value} onUpdate:modelValue={handleChange} />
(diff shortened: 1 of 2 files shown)


packages/components/switch/__tests__/switch.test.tsx
```diff
@@ -185,7 +184,7 @@ describe('Switch.vue', () => {
     const coreWrapper = wrapper.find('.el-switch__core')
     const switchWrapper: VueWrapper<SwitchInstance> =
       wrapper.findComponent(Switch)
-    const switchVm = switchWrapper.vm
+    const switchVm: any = switchWrapper.vm
```
tolking, line 187: Is it possible to avoid use the any?

Tsong-LC: I've tried, but no way not to use any.

packages/components/switch/__tests__/switch.test.tsx
```diff
@@ -111,8 +110,8 @@ describe('Switch.vue', () => {
   test('change event', async () => {
     const target = ref(1)
     const value = ref(true)
-    const handleChange = (val: boolean) => {
-      target.value = val
+    const handleChange = (val: string | number | boolean) => {
+      target.value = val as number
```
holazz, line 114: I think passing a generic parameter to target ref is better than using type assertion.

```ts
const target = ref<string | number | boolean>(1)
```

packages/components/switch/__tests__/switch.test.tsx
```diff
@@ -238,7 +237,7 @@ describe('Switch.vue', () => {
     const value = ref(true)
     const loading = ref(false)
     const asyncResult = ref('error')
-    const beforeChange = () => {
+    const beforeChange: () => Promise<boolean> | boolean = () => {
```
holazz, line 240: Here we can use a more specific type.

```ts
const beforeChange = () => {
  return new Promise<boolean>((resolve, reject) => {
   	...
  })
}
```

packages/components/switch/__tests__/switch.test.tsx
holazz: Maybe expose `checked` from the `el-switch`?

tolking: It seems that removing `:any` does not cause a type error.

xiaoxian521 (approved):

PR #9184, opened 2022-08-03
fix(components): table fix table empty height

Jeffrey-mu:
当有插槽组件时 empty 高度计算不参考body


tolking:
For no empty slot, the height of the default empty slot also needs to fix.

Jeffrey-mu:
> For no empty slot, the height of the default empty slot also needs to be fixed. Can you fix it at the same time?

No slots, just a line of text ，No need to set the height

tolking:
I do not quite agree with you, it is reasonable behavior to set height for the empty slot when table is fixed height.

PR #9215, opened 2022-08-05
feat(components): [page-header] enhancement

jw-foss:
* Add new slot `extra` for page header.
* Add addtional documentations for page header.

### DIFF
--- a/docs/examples/page-header/breadcrumb.vue
+++ b/docs/examples/page-header/breadcrumb.vue
@@ -1,5 +1,5 @@
 <template>
-  <el-page-header content="detail">
+  <el-page-header>
     <template #breadcrumb>
       <el-breadcrumb separator="/">
         <el-breadcrumb-item :to="{ path: './page-header.html' }"
@@ -10,6 +10,9 @@
         >
         <el-breadcrumb-item>route 2</el-breadcrumb-item>
       </el-breadcrumb>
+    </template>
+    <template #content>
+      <span class="text-large font-600 mr-3"> Title </span>
     </template>
   </el-page-header>
 </template>
--- a/docs/examples/page-header/breadcrumb.vue
+++ b/docs/examples/page-header/breadcrumb.vue
@@ -1,0 +1,15 @@
+<template>
+  <el-page-header content="detail">
+    <template #breadcrumb>
+      <el-breadcrumb separator="/">
+        <el-breadcrumb-item :to="{ path: './page-header.html' }"
+          >homepage</el-breadcrumb-item
+        >
+        <el-breadcrumb-item
+          ><a href="./page-header.html">route 1</a></el-breadcrumb-item
+        >
+        <el-breadcrumb-item>route 2</el-breadcrumb-item>
+      </el-breadcrumb>
+    </template>
+  </el-page-header>
+</template>
--- a/docs/examples/page-header/breadcrumb.vue
+++ b/docs/examples/page-header/breadcrumb.vue
@@ -2,9 +2,9 @@
   <el-page-header>
     <template #breadcrumb>
       <el-breadcrumb separator="/">
-        <el-breadcrumb-item :to="{ path: './page-header.html' }"
-          >homepage</el-breadcrumb-item
-        >
+        <el-breadcrumb-item :to="{ path: './page-header.html' }">
+          homepage
+        </el-breadcrumb-item>
         <el-breadcrumb-item
           ><a href="./page-header.html">route 1</a></el-breadcrumb-item
         >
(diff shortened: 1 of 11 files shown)


jw-foss:
@Giwayume if you are available would you mind helping to check if [this](https://github.com/element-plus/element-plus/pull/9215/commits/037625cf6288bd6fa098043970bb396f1ff627ba) a11y feature is added correctly?

Giwayume:
At the very least, I'd wrap the back text + icon in a `<button>` or `<a>` tag.

image

On the "el-page-header" element, a default of `role="header"` would be good, though it really depends on how the website uses this header. It's possible if it is the only header at the top of the page, `role="banner"` would apply. So leave `role` configurable.

In the examples, the "title" and "subtitle" text should be headings. Like `role="heading" aria-level="1"` or `<h1> <h2>` tags etc. Unfortunately due to the versatility of this component I don't think we can automate this for the developer, only inform of best practices.

tolking:
Previously the back event could be triggered by the text `Back`, but now it does not work.

jw-foss:
> Previously the back event could be triggered by the text `Back`, but now it does not work.

Humm, right, I think I need to add that back, did that just because I cross-checked with other libraries. Thanks for pointing out. :bow:

packages/components/page-header/src/page-header.vue
```diff
@@ -5,22 +5,23 @@
...
-        </div>
-        <div :class="ns.e('title')">
-          <slot name="title">{{ title || t('el.pageHeader.title') }}</slot>
+        <div :class="ns.e('back')" @click="handleClick">
+          <div
+            v-if="icon || $slots.icon"
+            :aria-label="title || t('el.pageHeader.title')"
+            :class="ns.e('icon')"
+            role="button"
+            tabindex="0"
```
tolking, line 14: The `role` and `tabindex` attribute should probably be moved with the click event

docs/examples/page-header/breadcrumb.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <el-page-header>
+    <template #breadcrumb>
+      <el-breadcrumb separator="/">
+        <el-breadcrumb-item :to="{ path: './page-header.html' }"
+          >homepage</el-breadcrumb-item
+        >
+        <el-breadcrumb-item
+          ><a href="./page-header.html">route 1</a></el-breadcrumb-item
+        >
```
tolking, line 10:
```suggestion
        <el-breadcrumb-item :to="{ path: './page-header.html' }">
          homepage
        </el-breadcrumb-item>
        <el-breadcrumb-item>
          <a href="./page-header.html">route 1</a>
        </el-breadcrumb-item>
```

Do you want to optimize the code format?

docs/examples/page-header/complete.vue
```diff
@@ -0,0 +1,69 @@
...
+  <div aria-label="A complete example of page header">
+    <el-page-header @back="onBack">
+      <template #breadcrumb>
+        <el-breadcrumb separator="/">
+          <el-breadcrumb-item :to="{ path: './page-header.html' }"
+            >homepage</el-breadcrumb-item
+          >
+          <el-breadcrumb-item
+            ><a href="./page-header.html">route 1</a></el-breadcrumb-item
+          >
```
tolking, line 11: same

docs/examples/page-header/breadcrumb.vue
jw-foss: I used `pnpm prettier` command here to format the code, I think this might related to the prettier configuration 🤔

tolking: Formatting is sometimes not smart enough, so we need to manually modify the code format.

packages/components/page-header/src/page-header.vue
tolking: This is indeed a Bug. Now, it will not be possible to trigger the back event by pressing Tab and Enter keys.

docs/examples/page-header/breadcrumb.vue
tolking: Formatting is sometimes not smart enough, so we need to manually modify the code format.

docs/examples/page-header/complete.vue
jw-foss: Done, PTAL.

docs/examples/page-header/breadcrumb.vue
jw-foss: Done, PTAL :bow:

packages/components/page-header/src/page-header.vue
jw-foss: I moved this to the parent element so this will be triggered correctly now.

tolking (approved):

PR #9241, opened 2022-08-08
docs: improve example

Hongbusi:
In the project using TS, import the error by the original method.

``` vue
<script lang="ts" setup>
import zhCn from 'element-plus/dist/locale/zh-cn.mjs' 
// Could not find a declaration file for module 'element-plus/dist/locale/zh-cn.mjs'.
</script>
```
Found another way in the [official documentation](https://element-plus.org/en-US/guide/i18n.html#global-configuration), which I think is more reasonable.


sxzz:
Please describe why you changed it, and then re-open this PR.

PR #9322, opened 2022-08-15
Cannot read properties of null (reading 'id')

qq282126990:
If columnKey is not set, execute clearFilter and prompt Cannot read properties of null (reading 'id')
Suggestion: Add judgment to clearly prompt developers


qq282126990:
example: image

btea:
@qq282126990 Hello, please provide a recurrence link

qq282126990:
Hi please check here
 https://github.com/qq282126990/element-plus/blob/dev/packages/components/table/src/util.ts#L120 
 @btea

btea:
Ok, I see.

qq282126990:
@btea  Hi I changed the format of the code and it should now pass the test 
image

qq282126990:
@btea hello please try the test case
image

qq282126990:
@btea  hello please try the test case
image

btea:
In this case, is it acceptable to throw an error directly, what do you think? @element-plus/backers

packages/components/table/src/util.ts
```diff
@@ -115,6 +115,7 @@ export const getColumnByKey = function <T>(
   let column = null
   for (let i = 0; i < table.columns.length; i++) {
     const item = table.columns[i]
+    if (!item.columnKey) throw new Error('column does not have columnKey set')
```
holazz, line 118:
```suggestion
```

```diff
    for (let i = 0; i < table.columns.length; i++) {
      const item = table.columns[i]
-     if (!item.columnKey) throw new Error('column does not have columnKey set')
      if (item.columnKey === columnKey) {
        column = item
        break
      }
    }
+   if (!column) throw new Error(`[Table] No column matching with column-key: ${columnKey}`)
    return column
```

holazz (commented):
I think the error should be thrown after all columns are checked and there is no match. The error message should be something like "No matching column". Because we may have set the `column-key`, but the wrong key was passed in when calling the `clearFilter` method, eg:

```vue
<el-table-column>
  column-key="date"
  ...
</el-table-column>
```

```ts
tableRef.value.clearFilter(['data']) // should be 'date'
```

packages/components/table/src/util.ts
btea: good suggestion 👍 !  Maybe also use `throwError`

https://github.com/element-plus/element-plus/blob/742a37ca58aa4833653b3c0d91d5157610ce4463/packages/utils/error.ts#L10

holazz: That's much better！I didn't notice it before.

btea:
@qq282126990 Hi, can you modify the code as suggested by @holazz ? And please follow the [prompts](https://github.com/element-plus/element-plus/pull/9322#issuecomment-1219074731) to modify the submission information.

btea:
The test seems to have failed.

qq282126990:
@btea  The test case cannot find the file
image
image

btea:
@qq282126990 It seems that there is a problem with the file path, maybe you can try to create a new pr.
image

qq282126990:
@btea Hi I recreated a new pull to go  
https://github.com/element-plus/element-plus/pull/9682

PR #9356, opened 2022-08-18
style(docs): change the example mask and overlay z-index

btea:

image
image

### DIFF
--- a/docs/.vitepress/vitepress/styles/css-vars.scss
+++ b/docs/.vitepress/vitepress/styles/css-vars.scss
@@ -55,7 +55,7 @@
   --content-min-width: 16rem;
   --content-max-width: 48rem;
 
-  --nav-z-index: 10;
+  --nav-z-index: 10000;
   --sub-nav-z-index: 10;
   --sidebar-z-index: 40;
   --overlay-z-index: 30;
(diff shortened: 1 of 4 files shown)


docs/.vitepress/vitepress/styles/css-vars.scss
```diff
@@ -55,7 +55,7 @@
   --content-min-width: 16rem;
   --content-max-width: 48rem;
 
-  --nav-z-index: 10;
+  --nav-z-index: 10000;
```
ryuhangyeong, line 58: If the sidebar is open, an exception might occur!

image

btea: Thanks for your careful. There is also a problem when opening the dialog.

ryuhangyeong: Thinking about it, I thought that changing the z-index of .el-table-v2__overlay in the element-plus documentation might have less side effects. What do you think about it?

Changing the z-index of a component will affect the user.

image

btea: You are right, totally agree.

btea: I have updated it, please review it again, thank you

ryuhangyeong (approved):

ryuhangyeong:
Thank you for your effort!

PR #9380, opened 2022-08-19
fix(components): [space] complete the prop of space

MonsterPi13:
As docs mention, when `class` is in `Array` type , its inner element can be `object` type, but it does not specified in type declaration.
image

### DIFF
--- a/packages/components/space/src/space.ts
+++ b/packages/components/space/src/space.ts
@@ -36,11 +36,9 @@ export const spaceProps = buildProps({
   },
 
   class: {
-    type: definePropType<string | string[] | Record<string, boolean>>([
-      String,
-      Object,
-      Array,
-    ]),
+    type: definePropType<
+      string | Array<Record<string, boolean> | string> | Record<string, boolean>
+    >([String, Object, Array]),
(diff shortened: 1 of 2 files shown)


btea (commented):
LGTM, maybe the document should also be updated

MonsterPi13:
> LGTM, maybe the document should also be updated

I have updated docs, please check.

btea (approved):

packages/components/space/src/space.ts
```diff
@@ -36,11 +36,9 @@ export const spaceProps = buildProps({
   },
 
   class: {
-    type: definePropType<string | string[] | Record<string, boolean>>([
-      String,
-      Object,
-      Array,
-    ]),
+    type: definePropType<
+      string | Array<Record<string, boolean> | string> | Record<string, boolean>
```
tolking, line 40: Maybe consider to use the `Arrayable`

MonsterPi13: Thanks for your advise, please check.

tolking (approved):

PR #9422, opened 2022-08-24
fix(hooks): [use-z-index] update currentZIndex mechanism

HelloChunWei:

---

I found dafault z-index will be 2000 and add by 1 **(nextZIndex)**, But it maybe will make a problem.

For example If I make a dialog component and style it z-index be 3000, At the same time, I also use date picker in my dialog. The date picker is behind the dialog component.

https://user-images.githubusercontent.com/18310281/186333892-12b719fa-9da6-4a83-bc83-646a0fb490ef.mov

I figure out source code of date-picker. and find it use useZIndex.

I also check the component to find which component is used useZIndex.
Here is the component list.

1. tooltip
2. dialog
3. loading
4. message
5. notification
6. image-viewer (for image Preview)
7. Message Box
8. Table (for showing overflow tooltip)

In my option, Those component must be on the top in website. They should not be covered by other component.

So I just Modified **useZIndex**

```javascript
const maximumValueOfZIndexAtDocument = () => {
  const minimumValueOfZIndex = 1
  let result = minimumValueOfZIndex
  // Check if it is server-side rendering or not
  if (typeof window !== 'undefined') {
    // In client side
    result = Math.max(
      minimumValueOfZIndex,
      ...Array.from(window.document.querySelectorAll('*'))
        .map((el) => getComputedStyle(el).zIndex)
        .filter((v) => !Number.isNaN(Number.parseInt(v[1])))
        .map((o) => Number.parseInt(o))
    )
  }
  return result
}
const zIndex = ref(0)

export const useZIndex = () => {
  const initialZIndex = useGlobalConfig('zIndex', 2000) // TODO: move to @element-plus/constants
  const currentZIndex = computed(() =>
    Math.max(initialZIndex.value, zIndex.value)
  )

  const nextZIndex = () => {
    zIndex.value = maximumValueOfZIndexAtDocument()
    return currentZIndex.value
  }

  return {
    initialZIndex,
    currentZIndex,
    nextZIndex,
  }
}

```

It will find the maximum value of z-index at the document, and calculate which value is bigger. (initialZIndex or zIndex)

So that can guarantee the component must be on the top.

https://user-images.githubusercontent.com/18310281/186333907-c23b3ee8-5e44-4375-9143-9029c93326e0.mov

### DIFF
--- a/packages/hooks/use-z-index/index.ts
+++ b/packages/hooks/use-z-index/index.ts
@@ -1,14 +1,32 @@
 import { computed, ref } from 'vue'
 import { useGlobalConfig } from '../use-global-config'
 
+const maximumValueOfZIndexAtDocument = () => {
+  const minimumValueOfZIndex = 1
+  let result = minimumValueOfZIndex
+  // Check if it is server-side rendering or not
+  if (typeof window !== 'undefined') {
+    // In client side
+    result = Math.max(
+      minimumValueOfZIndex,
+      ...Array.from(window.document.querySelectorAll('*'))
+        .map((el) => getComputedStyle(el).zIndex)
+        .filter((v) => !Number.isNaN(Number.parseInt(v[1])))
+        .map((o) => Number.parseInt(o))
+    )
+  }
+  return result
+}
 const zIndex = ref(0)
 
 export const useZIndex = () => {
   const initialZIndex = useGlobalConfig('zIndex', 2000) // TODO: move to @element-plus/constants
-  const currentZIndex = computed(() => initialZIndex.value + zIndex.value)
+  const currentZIndex = computed(() =>
+    Math.max(initialZIndex.value, zIndex.value)
+  )
 
   const nextZIndex = () => {
-    zIndex.value++
+    zIndex.value = maximumValueOfZIndexAtDocument()
     return currentZIndex.value
   }
 


sxzz:
So why do you set `z-index` of dialog to `3000`? It will increase by automatic.

HelloJunWei:
Hi @sxzz

In my options, We should not expect how web application are built. Maybe It is a legacy system which is maintained by some engineers. Furthermore, maybe he or she resigned from the compony.

Then someone maybe set `z-index ` of dialog to 3000 or above. (I know this is a bad code smell, but it is possible to happen. Actually It is my experience now )


In conclusion, I think those pop component which is just mentioned must be dynamic calculate value of `z-index`.
It will be more flexible.  We just don't think about how legacy system is built. can focus on features.

HelloJunWei:
Say more about.  I just view website at [juejin](https://juejin.cn/post/7030570376329429022), and it minimum value of z-index is 99999999.

I finally find out which dom value of z-index is 99999999 (`收藏成功`)

截圖 2022-08-28 上午10 58 14




and this zoom in effect `z-index` is 10000
截圖 2022-08-28 上午10 55 32


It it possible to have a  bug when componet is used at those dialog.

sxzz:
I see, but I don't think Element Plus should handle it as a general UI library. Also, the function found all elements in the document is too tricky and bad for performance. I think we should find a better way to solve this problem.

WDYT? @element-plus/backers

HelloJunWei:
Here is my solution now. Although is not a great solution I think, it work for me.

If anyone has this problem maybe can use this solution. 

What do you think? @sxzz 

截圖 2022-08-28 下午1 34 36

sxzz:
I think we have expose `useZIndex` and have a way to override this hook.

HelloJunWei:
I find out `el-popper` has already exposed `z-index` as a prop.

Maybe we can expose this prop to component in order to make user override hook?

PR #9466, opened 2022-08-25
fix(directives): [repeat-click] Interval time is too short for single clicks

opuu:

-------------
The interval to detect long press (100ms) is too low, so single clicks are also being considered as repeat click. 

For example: Clicking the plus or minus button of the `Input Number` component increases number by 2 instead of the defined step value (1).

multiple-click

This PR fixes the [Input Number](https://element-plus.org/en-US/component/input-number.html) component.


holazz:
Hi, maybe you need also to modify the test cases.

opuu:
> Hi, maybe you need also to modify the test cases.

Yes, making the sleep time 630 ms will fix it. I will do it soon.

opuu:
@holazz fixed the unit test issue, the current failed test does not seems to be related to this PR.

opuu:
Now it should pass all tests 🧪

holazz:
> Now it should pass all tests 🧪

I recommend making the last commit as a separate PR, since it's not relevant to this issue.

opuu:
@holazz It is related to this PR. The up and down arrows are using repeat click directive in [Time Picker](https://element-plus.org/en-US/component/time-picker.html#arbitrary-time-range) which also registers single click as double clicks.

holazz:
> @holazz It is related to this PR. The up and down arrows are using repeat click directive in [Time Picker](https://element-plus.org/en-US/component/time-picker.html#arbitrary-time-range) which also registers single click as double clicks.

You're right! I missed it.😅

holazz:
@opuu Thank you for your contribution. After a little thought, I think a more reasonable way is not to increase the time interval of the long press trigger event, but to delay the trigger of the first event after the long press. I added a commit to this PR. What do you think?

opuu:
> @opuu Thank you for your contribution. After a little thought, I think a more reasonable way is not to increase the time interval of the long press trigger event, but to delay the trigger of the first event after the long press. I added a commit to this PR. What do you think?

Thats even better 👌 This way it will behave more like HTML input with number type.

holazz:
Would you mind taking a look on this? @buqiyuan @tolking

tolking (approved):

xiaoxian521 (approved):

PR #9471, opened 2022-08-26
refactor(components): [popper]redfine the prop role type

MonsterPi13:
As title mentioned, elegant way to define the prop of role.

### DIFF
--- a/packages/components/popper/src/popper.ts
+++ b/packages/components/popper/src/popper.ts
@@ -10,7 +10,7 @@ export const Effect = {
   DARK: 'dark',
 } as const
 
-export const roleType = [
+export const roleTypes = [
   'dialog',
   'grid',
   'listbox',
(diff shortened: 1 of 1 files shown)


holazz:
I've already created a [PR](https://github.com/element-plus/element-plus/pull/8529) to refactor it. Of course, we can merge this first and I'll do rebase later.

packages/components/popper/src/popper.ts
```diff
@@ -1,23 +1,31 @@
...
 const effects = ['light', 'dark'] as const
 const triggers = ['click', 'contextmenu', 'hover', 'focus'] as const
 
 export const Effect = {
   LIGHT: 'light',
   DARK: 'dark',
-}
+} as const
+
+export const roleType = [
```
holazz, line 13: How about replacing `roleType` with `roleTypes` ?

MonsterPi13: LGTM, thanks for your advise. 

Yeah, we all do the same thing, we can merge first.

holazz (approved):

btea (approved):

PR #9993, opened 2022-10-03
fix(components): [form-item] do not validate without `prop` property

### LINKED ISSUE #9885: [Component] [form-item] Implicit required validation rule auto-creation based on "required" prop
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.39`
- Element Plus Version: `2.2.17`
- Browser / OS: `Firefox 105 / Windows 10`
- Build Tool: `Vue CLI`

### Related Component

- `el-form-item`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cCBsYW5nPVwidHNcIj5cbmltcG9ydCB7IHJlZiwgdmVyc2lvbiBhcyB2dWVWZXJzaW9uIH0gZnJvbSAndnVlJ1xuaW1wb3J0IHsgdmVyc2lvbiBhcyBFcFZlcnNpb24gfSBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBFbGVtZW50UGx1cyB9IGZyb20gJ0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlJ1xuXG5jb25zdCBtc2cgPSByZWYoJycpXG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8ZWwtZm9ybT5cbiAgICA8ZWwtZm9ybS1pdGVtIGxhYmVsPSdsYWJlbCcgcmVxdWlyZWQ+XG4gIFx0XHQ8ZWwtaW5wdXQgdi1tb2RlbD1cIm1zZ1wiIC8+XG4gICAgPC9lbC1mb3JtLWl0ZW0+XG4gIDwvZWwtZm9ybT5cbjwvdGVtcGxhdGU+XG4iLCJpbXBvcnRfbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7fVxufSIsIl9vIjp7fX0=)

### Steps to reproduce

```vue
<el-form-item label="label" required><el-input v-model="msg" /></el-form-item>
```
We use backend validation and "required" prop on form-item for visualization purposes only (asterisk).
So "prop" is not set intentionally.

### What is Expected?

No required validation when "prop" is undefined

"is required" validation error is always shown on such form items

### Additional comments

Now the only way to disable unneeded required validation is to set :validate-event="false" on every(!) input on the form.This is very inconvenient. Please add check for prop !== undefined in form-item.vue:200

In Element-UI required prop on form-item worked as expected.

btea:

fix #9885

### DIFF
--- a/packages/components/form/src/form-item.vue
+++ b/packages/components/form/src/form-item.vue
@@ -178,7 +178,7 @@ const fieldValue = computed(() => {
 })
 
 const normalizedRules = computed(() => {
-  const { required } = props
+  const { required, prop } = props
 
   const rules: FormItemRule[] = []
 
@@ -197,7 +197,7 @@ const normalizedRules = computed(() => {
     }
   }
 
-  if (required !== undefined) {
+  if (required !== undefined && prop !== undefined) {
     const requiredRules = rules
       .map((rule, i) => [rule, i] as const)
       .filter(([rule]) => Object.keys(rule).includes('required'))
--- a/packages/components/form/src/form-item.vue
+++ b/packages/components/form/src/form-item.vue
@@ -178,26 +178,26 @@ const fieldValue = computed(() => {
 })
 
 const normalizedRules = computed(() => {
-  const { required, prop } = props
+  const { required } = props
 
   const rules: FormItemRule[] = []
 
-  if (props.rules && prop) {
+  if (props.rules) {
     rules.push(...ensureArray(props.rules))
   }
 
   const formRules = formContext?.rules
-  if (formRules && prop) {
+  if (formRules) {
     const _rules = getProp<Arrayable<FormItemRule> | undefined>(
       formRules,
-      prop
+      props.prop
     ).value
     if (_rules) {
       rules.push(...ensureArray(_rules))
     }
   }
 
-  if (required !== undefined && prop !== undefined) {
+  if (required !== undefined) {
     const requiredRules = rules
       .map((rule, i) => [rule, i] as const)
       .filter(([rule]) => Object.keys(rule).includes('required'))
--- a/packages/components/form/src/form-item.vue
+++ b/packages/components/form/src/form-item.vue
@@ -182,15 +182,15 @@ const normalizedRules = computed(() => {
 
   const rules: FormItemRule[] = []
 
-  if (props.rules) {
+  if (props.rules && prop) {
     rules.push(...ensureArray(props.rules))
   }
 
   const formRules = formContext?.rules
-  if (formRules && props.prop) {
+  if (formRules && prop) {
     const _rules = getProp<Arrayable<FormItemRule> | undefined>(
       formRules,
-      props.prop
+      prop
     ).value
     if (_rules) {
       rules.push(...ensureArray(_rules))
--- a/packages/components/form/src/form-item.vue
+++ b/packages/components/form/src/form-item.vue
@@ -187,7 +187,7 @@ const normalizedRules = computed(() => {
   }
 
   const formRules = formContext?.rules
-  if (formRules) {
+  if (formRules && props.prop) {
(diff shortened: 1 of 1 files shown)


ryuhangyeong (approved):

holazz:
I think that if the prop is not passed, the validation can be skipped without going through the validation logic. Like this:

```ts
const validate: FormItemContext['validate'] = async (trigger, callback) => {
  if (!props.prop) return false
  ...
}
```

This will be consistent with ElementUI. WDYT?

btea:
Maybe I can modify it here
https://github.com/element-plus/element-plus/blob/268d87a03bbf5865d49f6177f2868018614c15a5/packages/components/form/src/form-item.vue#L185-L187

```ts
if (props.rules && prop) {
  rules.push(...ensureArray(props.rules))
}
```

holazz:
> Maybe I can modify it here
> 
> https://github.com/element-plus/element-plus/blob/268d87a03bbf5865d49f6177f2868018614c15a5/packages/components/form/src/form-item.vue#L185-L187
> 
> ```ts
> if (props.rules && prop) {
>   rules.push(...ensureArray(props.rules))
> }
> ```

Changing here also seems to execute the entire validation process. IMO, there is no need to execute the rest of the logic in the `validate` function when the `prop` is `undefined`. Or am I missing something? 🤔

btea:
Oops, maybe I misunderstood what you meant, the advice above is great, I'll update it 😅

packages/components/form/src/form-item.vue
```diff
@@ -178,26 +178,26 @@ const fieldValue = computed(() => {
 })
 
 const normalizedRules = computed(() => {
-  const { required } = props
+  const { required, prop } = props
 
   const rules: FormItemRule[] = []
 
-  if (props.rules) {
+  if (props.rules && prop) {
```
holazz, line 185: Maybe there is no need to add extra conditions because the `prop` here must not be `undefined`.

packages/components/form/src/form-item.vue
```diff
@@ -178,26 +178,26 @@ const fieldValue = computed(() => {
...
     rules.push(...ensureArray(props.rules))
   }
 
   const formRules = formContext?.rules
-  if (formRules && props.prop) {
+  if (formRules && prop) {
```
holazz, line 190: Same here.

packages/components/form/src/form-item.vue
```diff
@@ -178,26 +178,26 @@ const fieldValue = computed(() => {
...
-      props.prop
+      prop
     ).value
     if (_rules) {
       rules.push(...ensureArray(_rules))
     }
   }
 
-  if (required !== undefined) {
+  if (required !== undefined && prop !== undefined) {
```
holazz, line 200: Same here.

packages/components/form/src/form-item.vue
btea: Done.

holazz:
Based on the changes now, the `&& props.prop` on line#190 doesn't seem to be needed either.

btea:
If `&& props.prop` is removed, then line 193 should be changed to `props.prop || ''`, otherwise there will be a ts error, so I think it's okay to keep the status quo, what do you think? 🤔

holazz:
> If `&& props.prop` is removed, then line 193 should be changed to `props.prop || ''`, otherwise there will be a ts error, so I think it's okay to keep the status quo, what do you think? 🤔

I feel like it's possible to use a non-null assertion in line 193: `props.prop!`.

holazz (approved):

btea:
@holazz Thanks for your review

Issue #9981, opened 2022-10-01
[Component] [checkbox] Checkbox is not working inside label tag


MonkeyMaster:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.40`
- Element Plus Version: `2.2.17`
- Browser / OS: `Chrome 105, Firefox 105 / Windows 10`
- Build Tool: `Vue CLI`

### Related Component

- `el-checkbox`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cCBsYW5nPVwidHNcIj5cbmltcG9ydCB7IHJlZiB9IGZyb20gJ3Z1ZSdcbmltcG9ydCB7IEVsZW1lbnRQbHVzIH0gZnJvbSAnQGVsZW1lbnQtcGx1cy9pY29ucy12dWUnXG5cbiAgY29uc3QgdmFsdWUgPSByZWYodHJ1ZSlcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxlbC1mb3JtPlxuICBcdDxlbC1mb3JtLWl0ZW0+XG4gIFx0XHQ8bGFiZWw+XG4gICAgXHRcdDxlbC1jaGVja2JveCB2LW1vZGVsPVwidmFsdWVcIi8+XG4gICAgXHQ8L2xhYmVsPlxuICBcdDwvZWwtZm9ybS1pdGVtPlxuICA8L2VsLWZvcm0+XG48L3RlbXBsYXRlPlxuIiwiaW1wb3J0X21hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge31cbn0iLCJfbyI6e319)

### Steps to reproduce

When el-checkbox with no label prop is wrapped with label tag it's not possible to change it's state

### What is Expected?

checkbox change it's state on click

### What is actually happening?

checkbox ignores clicks

### Additional comments

_(empty)_

PR #9883, opened 2022-09-24
fix(components): [checkbox] bind events to input

holazz:
Move events and custom attributes to input, closes #9864

[Before](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICBlbC1jaGVja2JveFxuICA8ZWwtY2hlY2tib3ggdi1tb2RlbD1cImNoZWNrZWRcIiBAY2xpY2s9XCJoYW5kbGVDbGlja1wiIGxhYmVsPVwiT3B0aW9uIDFcIiAvPlxuXG4gIDxkaXY+XG4gICAgbmF0aXZlIGNoZWNrYm94XG4gICAgPGlucHV0IEBjbGljaz1cImhhbmRsZUNsaWNrXCIgdHlwZT1cImNoZWNrYm94XCIgaWQ9XCJPcHRpb24gMVwiIG5hbWU9XCJPcHRpb24gMVwiIC8+XG4gICAgPGxhYmVsIGZvcj1cIk9wdGlvbiAxXCI+T3B0aW9uIDE8L2xhYmVsPlxuICA8L2Rpdj5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5pbXBvcnQgeyByZWYgfSBmcm9tICd2dWUnXG5cbmNvbnN0IGNoZWNrZWQgPSByZWYodHJ1ZSlcblxuZnVuY3Rpb24gaGFuZGxlQ2xpY2soZSkge1xuICBjb25zb2xlLmxvZyhlLnRhcmdldClcbn1cbjwvc2NyaXB0PlxuIiwiaW1wb3J0X21hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge31cbn0iLCJfbyI6e319)

[...]

Issue #9864, opened 2022-09-22
[Component] [checkbox] el-checkbox的@click事件会被触发两次（The @click event of the <el-checkbox> will be triggered twice）


yuyulangqi:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.39`
- Element Plus Version: `2.2.17`
- Browser / OS: `Chrome/104.0.0.0`
- Build Tool: `Vite`

### Related Component

- `el-checkbox`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZGl2PlxuICAgIDxlbC1jaGVja2JveCB2LW1vZGVsPVwiY2hlY2tlZDFcIiBAY2xpY2s9XCJ0ZXN0XCIgbGFiZWw9XCJPcHRpb24gMVwiIHNpemU9XCJsYXJnZVwiIC8+XG4gIDwvZGl2PlxuXG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IGxhbmc9XCJ0c1wiIHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSAndnVlJ1xuICBmdW5jdGlvbiB0ZXN0KCl7XG4gICAgYWxlcnQoJzExMScpXG4gIH1cblxuY29uc3QgY2hlY2tlZDEgPSByZWYodHJ1ZSlcbjwvc2NyaXB0PlxuIiwiaW1wb3J0X21hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge31cbn0iLCJfbyI6e319)

### Steps to reproduce

点击页面复选框会有两个弹窗

### What is Expected?

修复bug

### What is actually happening?

修复bug

### Additional comments

_(empty)_

gjfei:
https://developer.aliyun.com/article/263755

PR #10003, opened 2022-10-04
fix(components): [checkbox] only propagate click event of checkbox input

makedopamine:
I referred to pr #9883 and #9991. And  I am aware of two points.

- In general, the checkbox input is wrapped with label,  we just need to stop click event propagation of other element.
- It is possible that the check input isn't wrapped with label and we need to manually trigger click event  of it.

Let me know if there is any problem.

### DIFF
--- a/packages/components/checkbox/__tests__/checkbox.test.tsx
+++ b/packages/components/checkbox/__tests__/checkbox.test.tsx
@@ -83,7 +83,7 @@ describe('Checkbox', () => {
         </ElFormItem>
       ))
 
-      await wrapper.findComponent(Checkbox).trigger('click')
+      await wrapper.find('.el-checkbox__inner').trigger('click')
       expect(data.value).toBe(true)
     })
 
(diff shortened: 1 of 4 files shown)


makedopamine:
Why test of `avatar` component is failed? I have no idea to solve it.  Any suggestions will be appreciated.😊

iDestin:
I think it doesn't make it simpler, it makes the logic more complicated

makedopamine:
This pr isn't just to fix #9981 but also #9864. Actually, they were both caused by checkbox component fired click event twice. So we must solve the root problem. And do you have a better solution?

makedopamine:
@holazz @chenxch Could you please take a look?

holazz:
> This pr isn't just to fix #9981 but also #9864. Actually, they were both caused by checkbox component fired click event twice. So we must solve the root problem. And do you have a better solution? @iDestin

Hi @makedopamine, I think the two issues you mentioned should be dealt with in separate PRs. For #9981, I am more inclined to @iDestin's solution. 🤔

chenxch:
@makedopamine Thank you very much for the pr you provided for ep when you are busy. I may choose the repair plan of #9991 relatively conservatively. This does not mean that your plan is not good. I hope you can understand and hope that this will not affect you. Passion for ep.🤝

makedopamine:
There may a better way to deal with it, I will just close this PR.

PR #10163, opened 2022-10-21
refactor(components): [calendar]

jw-foss:
* Eliminate long function by extracting code as function calls.

### DIFF
--- a/packages/components/calendar/src/calendar.vue
+++ b/packages/components/calendar/src/calendar.vue
@@ -52,8 +52,9 @@
 </template>
 
 <script lang="ts" setup>
+import { computed } from 'vue'
 import { ElButton, ElButtonGroup } from '@element-plus/components/button'
-import { useNamespace } from '@element-plus/hooks'
+import { useLocale, useNamespace } from '@element-plus/hooks'
 
 import DateTable from './date-table.vue'
 import { useCalendar } from './use-calendar'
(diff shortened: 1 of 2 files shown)


emojiiii (approved):

holazz (approved):

packages/components/calendar/src/calendar.vue
```diff
@@ -75,11 +76,16 @@ const {
...
   validatedRange,
 } = useCalendar(props, emit, COMPONENT_NAME)
 
+const { t } = useLocale()
+
+const i18nDate = computed(() => {
+  const pickedMonth = `el.datepicker.month${date.value.format('M')}`
+  return `${date.value.year()} ${t('el.datepicker.year')} ${t(pickedMonth)}`
+})
+
```
ryuhangyeong, line 88: Question

Why is that logic separated from hooks? Curious about the standards!

jw-foss: @ryuhangyeong the reason I extracted this out of the hook is that this computed member is only used for the template rendering, for instance styles, classes getters should also be extracted from the hooks.

```ts
const style = computed(() => ({ someStyle: 'value' }))
const kls = computed(() => (['kls1', { kls2: true }]))
```

ryuhangyeong: @jw-foss  Thanks~ I will refer to your opinion.

ryuhangyeong (approved):

Issue #10080, opened 2022-10-13
[Component] [table] [Component] [table] 在table中展示树形数据，且添加多选框。在选择数据时，父级数据和子级数据之间没有联动


Fate-ui:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.40`
- Element Plus Version: `2.2.17`
- Browser / OS: `Chrome/106 / Windows NT 10.0`
- Build Tool: `Vite`

### Related Component

- `el-table`

### Reproduction Link

[...]

PR #10221, opened 2022-10-24
fix(component): [table]  can't select row children

faga295:
#10080 


packages/components/table/src/util.ts
```diff
@@ -251,10 +251,20 @@ export function toggleRowStatus<T>(
 
   const addRow = () => {
     statusArr.push(row)
+    if (row.children) {
```
tolking, line 254: It does not seem to work on multi-layer tree

faga1: now it works

faga1:
I moved toggleRowStatus into usewatcher because toggleRowSelection should be trigger inside toggleRowStatus

packages/components/table/src/store/watcher.ts
```diff
@@ -202,6 +201,49 @@ function useWatcher<T>() {
...
+      statusArr.splice(index, 1)
+      if (row.children) {
+        row.children.forEach((item) => {
+          toggleRowSelection(item, newVal ?? !included)
+        })
+      }
+      changed = true
+    }
+
+    if (typeof newVal === 'boolean') {
```
tolking, line 232:
```suggestion
    if (isBoolean(newVal)) {
```

packages/components/table/src/store/watcher.ts
```diff
@@ -202,6 +201,49 @@ function useWatcher<T>() {
...
+    row: T,
+    newVal: boolean
+  ): boolean => {
+    let changed = false
+    const index = statusArr.indexOf(row)
+    const included = index !== -1
+
+    const addRow = () => {
+      statusArr.push(row)
+      if (row.children) {
```
tolking, line 215: IMO, the children may not be an array. Maybe it should be judged. What do you think?

```suggestion
      if (isArray(row.children)) {
```

faga1: yes, the children type is totally  depend on user

packages/components/table/src/store/watcher.ts
```diff
@@ -202,6 +201,49 @@ function useWatcher<T>() {
...
+      if (isArray(row.children)) {
+        row.children.forEach((item) => {
+          toggleRowSelection(item, newVal ?? !included)
+        })
+      }
+      changed = true
+    }
+    const removeRow = () => {
+      statusArr.splice(index, 1)
+      if (row.children) {
```
tolking, line 224: Same here

packages/components/table/src/store/watcher.ts
```diff
@@ -202,6 +201,49 @@ function useWatcher<T>() {
...
+    let changed = false
+    const index = statusArr.indexOf(row)
+    const included = index !== -1
+
+    const addRow = () => {
+      statusArr.push(row)
+      if (isArray(row.children)) {
+        row.children.forEach((item) => {
+          toggleRowSelection(item, newVal ?? !included)
+        })
```
tolking, line 218: This causes the `select` and `selection-change` event to trigger several times

faga1: so if a row having children was selected, the `select` and `selection-change` is expected to trigger only once?I do some surplus works because of it,it triggers only once at the beginning

faga1: now it will trigger only once

packages/components/table/src/util.ts
```diff
@@ -249,36 +249,31 @@ export function toggleRowStatus<T>(
...
-      removeRow()
+      toggleStatus('remove')
     }
   } else {
     if (included) {
-      removeRow()
+      toggleStatus('remove')
     } else {
-      addRow()
+      toggleStatus('add')
```
tolking, line 276:
```suggestion
toggleStatus(included ? 'remove' : 'add')
```

packages/components/table/src/util.ts
```diff
@@ -249,36 +249,31 @@ export function toggleRowStatus<T>(
   const index = statusArr.indexOf(row)
   const included = index !== -1
 
-  const addRow = () => {
-    statusArr.push(row)
-    if (isArray(row.children)) {
-      row.children.forEach((item) => {
-        toggleRowStatus(statusArr, item, newVal ?? !included)
-      })
+  const toggleStatus = (type: 'add' | 'remove'): boolean => {
```
tolking, line 252:
```suggestion
  const toggleStatus = (type: 'add' | 'remove') => {
```

This function should have no return value

faga1: thank you for your patient correction

packages/components/table/src/util.ts
```diff
@@ -269,13 +269,7 @@ export function toggleRowStatus<T>(
...
       toggleStatus('remove')
     }
-  } else {
-    if (included) {
-      toggleStatus('remove')
-    } else {
-      toggleStatus('add')
-    }
-  }
+  } else included ? toggleStatus('remove') : toggleStatus('add')
```
tolking: Do you think this is simpler and easier to understand than before?

faga1: yes, more understandable

tolking: I searched the project and found only three cases where brackets were omitted when using `if else` (there are only simple calls internally). It is weird for you to use it like this.

faga1: maybe it should included in eslint

AxelConceicao:
Thanks for this awesome feature|fix

tolking (approved):

jw-foss (approved):

wubianluoye:
当使用tree-props重新定义children为其他非children字符串时，如 :tree-props="{children: 'list'}"点击父节点无法选中子节点

tolking:
@jw-foss I suddenly realized that there might be a problem with this PR.

Each data in the table is independent of each other, and the `select` should be same. consistent with element ui

What do you think?

PR #10291, opened 2022-10-27
refactor(components): useSelect

Simon-He95:


packages/components/select/src/useSelect.ts
```diff
@@ -661,15 +655,9 @@ export const useSelect = (props, states: States, ctx) => {
...
-      if (toRaw(get(item, valueKey)) === get(value, valueKey)) {
-        index = i
-        return true
-      }
-      return false
-    })
-    return index
+    return arr.findIndex(
+      (item) => toRaw(get(item, valueKey)) === get(value, valueKey)
+    )
```
jw-foss, line 660: **Praise:** good catch!

packages/components/select/src/useSelect.ts
```diff
@@ -802,36 +781,30 @@ export const useSelect = (props, states: States, ctx) => {
   }
 
   const toggleMenu = (e?: PointerEvent) => {
-    if (e && !states.mouseEnter) {
+    if ((e && !states.mouseEnter) || selectDisabled.value) {
```
jw-foss, line 784: **nit:** perhaps we could remove the curly bracket and just append `return` statement after `)`, WDTY?

packages/components/select/src/useSelect.ts
```diff
@@ -844,31 +817,34 @@ export const useSelect = (props, states: States, ctx) => {
...
-        option.disabled === true ||
-        option.states.groupDisabled === true ||
-        !option.visible
-      ) {
-        navigateOptions(direction)
+    } else if (direction === 'prev') {
+      states.hoverIndex--
+      if (states.hoverIndex < 0) {
+        states.hoverIndex = states.options.size - 1
       }
```
jw-foss, line 837: **Question:** WDTY about using switch cases here instead of `if else`?

jw-foss (commented):
@Simon-He95 in general these changes LGTM, I got a few comments for your consideration, let me know if you have any questions, back to you 🏈

packages/components/select/src/useSelect.ts
```diff
@@ -811,30 +809,24 @@ export const useSelect = (props, states: States, ctx) => {
       .filter((option) => option.visible)
       .every((option) => option.disabled)
   )
-
-  const navigateOptions = (direction) => {
```
Simon-He95, line 842: This condition of `!states.visible` was merged by me into the following if condition. Do you think it would work?

packages/components/select/src/useSelect.ts
```diff
@@ -811,30 +809,24 @@ export const useSelect = (props, states: States, ctx) => {
...
-    if (!states.visible) {
-      states.visible = true
-      return
-    }
+  const navigateOptions = (direction: 'prev' | 'next') => {
+    states.visible = true
     if (
```
Simon-He95, line 814: This condition of !states.visible was merged by me into the following if condition. Do you think it would work?

packages/components/select/src/useSelect.ts
```diff
@@ -811,30 +809,24 @@ export const useSelect = (props, states: States, ctx) => {
...
-      states.hoverIndex++
-      if (states.hoverIndex === states.options.size) {
-        states.hoverIndex = 0
-      }
-    } else if (direction === 'prev') {
-      states.hoverIndex--
-      if (states.hoverIndex < 0) {
-        states.hoverIndex = states.options.size - 1
-      }
+    states.hoverIndex =
```
Simon-He95, line 823: If the content of the nested if, I extracted to the outside.
I think the direction just has next or prev that use `? :` to replace if else

jw-foss:
@Simon-He95 humm. looks like we are having some failure tests after the changes, I think you should look into it.

Simon-He95:
> I think you should look into it.
I modified it and checked the docs and there is no error

jw-foss:
@Simon-He95 we are still having unit testing issues, please fix them. https://github.com/element-plus/element-plus/actions/runs/3343503071/jobs/5536788728

Simon-He95:
@jw-foss I have fixed it.
I always thought that many of the ticked options in the front of the build were passed the vitest test. I didn't pay attention when I saw that the build was ok... Do you think we should add test to the build script?

patienry (commented):
Rel [#10376 (comment)](https://github.com/element-plus/element-plus/pull/10376#pullrequestreview-5073721249)

PR #10324, opened 2022-10-31
fix(dev): fix the new developer startup service times vite. create filter is not a function

jiadaoyun:
新贡献者在安装完依赖以后，项目跑不起来，排查后发现是vite版本的问题


tolking:
Thanks for your contribution.

PR #10405, opened 2022-11-04
docs(components): [autocomplete]

jw-foss:
* Add new API component `ref-type`.
* Update doc API name font-family.
* Update API popup style.
* Update Autocomplete documentations.

### Screenshot

CleanShot 2022-11-04 at 15 05 42@2x


chenxch (approved):
image
@jw-foss 
I have two reference suggestions:
1. In the dark mode, the color of the prompt text seems to be brighter.
2. Enumeration type, I was wondering if it would be better to show the primitive type.

jw-foss:
> 2. Enumeration type, I was wondering if it would be better to show the primitive type.

Not so sure about what the **primitive** type means here. I think enum type for those enumerable values which are the perfect use case here.

chenxch:
> > 2. Enumeration type, I was wondering if it would be better to show the primitive type.
> 
> Not so sure about what the **primitive** type means here. I think enum type for those enumerable values which are the perfect use case here.

All right!

PR #10484, opened 2022-11-07
docs: handle `currentRelease` undefined

holazz:


jw-foss:
@holazz maybe we should just rendering the error here when API limit reached we should just show the error.

holazz:
> @holazz maybe we should just rendering the error here when API limit reached we should just show the error.

@jw-foss Yes, the API limit console error is still showing, and what I fixed is this error:

image

jw-foss:
@holazz I meant that by rendering the error to the UI such as showing up a text to the users which explains the API limit has been reached they should try it later or watch if from the source.

holazz:
> @holazz I meant that by rendering the error to the UI such as showing up a text to the users which explains the API limit has been reached they should try it later or watch if from the source.

Gotcha!

PR #10369, opened 2022-11-02
fix(components): [time-picker] prefix trigger date-pane method

chenxch:
- close #10367 

before:
date1

The difference between clicking the prefix icon and clicking the text box is that the former does not focus on the component. I think the performance of the two should be consistent, after all, it is a whole.

after:
date2

Through this component I found some cases, for example, in the `el-input` component, clicking the `prefix` icon or the edge of the component cannot `focus` the component, but they are a whole, because we only give the `focus` time to the `native`  input tag.

image

Click to focus the event only in the blue area.

- fix inputRef type

Please make sure these boxes are checked before submitting your PR, thank you!

- [ ] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).
- [ ] Make sure you are merging your commits to `dev` branch.
- [ ] Add some descriptions and refer to relative issues for your PR.

PR #10710, opened 2022-11-21
feat(components): [time-picker] focus and blur input

Mxm13620:
[time-picker] focus and blur input onpick and on show

Issue: 
When pick a date, input is still focus.
When open calendar calendar is not focus


chenxch:
When pick a date, input is still focus.
=>
This is the correct interaction.

When open calendar calendar is not focus
=>
You can follow #10674 and #10369, the plan is to use the input prefix and suffix. 
So I will close this pr. 
Thank you for your contribution and look forward to your next one.

PR #10796, opened 2022-11-28
fix(components): [notification] option zIndex for notification can't effect

acfinity:
File:`/packages/components/notification/src/notify.ts`
```javascript
const props = {
  ...options,
  zIndex: nextZIndex(),
  offset: verticalOffset,
  id,
  onClose: () => {
    close(id, position, userOnClose);
  }
};
```

`options.zIndex` will not effective since it is write before `zIndex: nextZIndex()`, it will be overode.


tolking (approved):

btea (approved):
Thank you for your contribution  ❤️

PR #10881, opened 2022-12-02
docs(components): [config-provider]

chenxch:
* Adjust API style
before:
image

after:
image


chenxch:
1. Is the type writing of `locale` feasible?
2. The a11y and keyboardNavigation in the attributes seem to have no effect at present, I have not written them into the document, do you want to write them?
/cc @jw-foss

docs/en-US/component/config-provider.md
```diff
@@ -37,34 +37,36 @@ config-provider/message
...
+| namespace             | global component className prefix (cooperated with [$namespace](https://github.com/element-plus/element-plus/blob/dev/packages/theme-chalk/src/mixins/config.scss#L1)) | ^[string]                                                                                                                                                                                                                                                      | el                                                                                     |
+| button                | button related configuration, [see the following table](#button-attributes)                                                                                            | ^[object]`{autoInsertSpace?: boolean}`                                                                                                                                                                                                                         | see the following table                                                                |
```
holazz, line 52: I found that the type here is detailed in the Button Attribute table below, does the definition here seem repetitive? Is it better to use the previous `ButtonGlobalConfig `? 🤔

image

chenxch: Or can we directly remove the attribute description below and change it to link to button attributes， WDYT?

holazz: I think the idea is good, but for `message`, there seems to be nowhere to link.

image

chenxch: It seems that the corresponding attribute description cannot be canceled. This is a global attribute, and perhaps it can only be kept like this. The above tooltip is just for quick viewing of attributes.

holazz: Well, it seems so.

jw-foss (approved):

jw-foss:
LGTM.

PR #10892, opened 2022-12-04
fix(utils):Optimize the operation mode of animation.ts files

gjl-0810:
优化utils文件中animation.ts中的运算方式

### DIFF
--- a/packages/utils/animation.ts
+++ b/packages/utils/animation.ts
@@ -3,4 +3,4 @@ export const cubic = (value: number): number => {
 }
 
 export const easeInOutCubic = (value: number): number =>
-  value < 0.5 ? cubic(value * 2) / 2 : 1 - cubic((1 - value) * 2) / 2
+  value < 0.5 ? cubic(value << 1) >> 1 : 1 - (cubic((1 - value) << 1) >> 1)


packages/utils/animation.ts
```diff
@@ -3,4 +3,4 @@ export const cubic = (value: number): number => {
 }
 
 export const easeInOutCubic = (value: number): number =>
-  value < 0.5 ? cubic(value * 2) / 2 : 1 - cubic((1 - value) * 2) / 2
+  value < 0.5 ? cubic(value << 1) >> 1 : 1 - (cubic((1 - value) << 1) >> 1)
```
tolking, line 6: It seems that it is not as easy to understand as before

gjl-0810: yes，But now it can have higher performance, and bit operation should be the basis. As a public module, it should have better performance?

gjl-0810: For readability, consider adding comments?

tolking: If the `value` is a decimal (eg, 0.5), the calculation result is different from before.

chenxch:
In the case of no performance advantage, it is not recommended to change it, and I think readability is more important than it.

packages/utils/animation.ts
gjl-0810: yes,I was careless， you`re right

tolking: If you can not fix this, you can close this PR by yourself.

PR #10963, opened 2022-12-12
fix(components): [divider] Fixed `border: 1px` breaking `border-style="double"`

gaoxiaoduan:
Fixed `border: 1px` breaking `border-style="double"`

image

### DIFF
--- a/packages/components/divider/__tests__/divider.test.tsx
+++ b/packages/components/divider/__tests__/divider.test.tsx
@@ -55,4 +55,19 @@ describe('Divider.vue', () => {
       )
     ).toBe('solid')
   })
+
+  test('line-double', () => {
+    const wrapper = mount(() => <Divider borderStyle="double" />)
+    expect(
+      getComputedStyle(wrapper.element, null).getPropertyValue(
+        '--el-border-style'
+      )
+    ).toBe('double')
+
+    expect(
+      getComputedStyle(wrapper.element, null).getPropertyValue(
+        '--el-border-width'
+      )
+    ).toBe('2px')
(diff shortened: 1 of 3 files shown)


tolking:
Thank you for your contribution, but I suggest you understand the [relevant standards](https://developer.mozilla.org/en-US/docs/Web/CSS/border-style#values)

gaoxiaoduan:
@tolking Thank you for your reply, I have read the standard documentation and I know that this is a feature of 'double'. this shouldn't be a bug, it should be an optimization point, I found that border-style="double" has no response in the process of use. It is also necessary to set border-width="2px" separately. From the user's point of view, maybe there should be a direct response😂

tolking:
The `border-style="double"` and `border-width="2px"` are not necessarily related. 

[Demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtZGl2aWRlciBib3JkZXItc3R5bGU9XCJkb3VibGVcIiBzdHlsZT1cImJvcmRlci13aWR0aDo0cHhcIiAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydF9tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHt9XG59IiwiX28iOnt9fQ==)

gaoxiaoduan:
Yes, `border-style="double"` might also have something to do with the screen resolution, as different resolutions require different widths

image

image

gaoxiaoduan:
All right, let the user configure it.

Issue #11410, opened 2023-02-03
[Component] [input-number] input-number does not set value if the value matches the value when focus was gained


robert-wloch-iits:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.37`
- Element Plus Version: `2.2.28`
- Browser / OS: `UserAgent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36 OPR/94.0.0.0`
- Build Tool: `Vite`

### Related Component

- `el-input-number`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtaW5wdXQtbnVtYmVyIHYtbW9kZWw9XCJudW1cIiA6bWluPVwiMVwiIDptYXg9XCIyMFwiIC8+XG4gIDxkaXY+bnVtOiB7eyBudW0gfX08L2Rpdj5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5pbXBvcnQgeyByZWYgfSBmcm9tICd2dWUnXG5cbmNvbnN0IG51bSA9IHJlZigxNSlcbjwvc2NyaXB0PlxuIiwiaW1wb3J0X21hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge31cbn0iLCJfbyI6e319)

### Steps to reproduce

Using the minimal reproduction link (source "Basic Usage" in input number docu https://element-plus.org/en-US/component/input-number.html#basic-usage)

1. Observe initial state: input number shows 15, num ref prints 15
2. Double click the 15 in in the input number
3. input number is focussed, 15 is selected
4. type 14
5. Observe state: input number shows 14, num ref prints 14
6. While input number still has focus:
6.a. Select all characters of 14 using the keyboard
6.b. Double click the 14
7. type 15
8. Observe state: input number shows 15, num ref prints 1

### What is Expected?

8. Observe state: input number shows 15, num ref prints 15

### What is actually happening?

See step 8 in Steps to reproduce.

Analysis:
Input-number is storing the value when it gained focus and prevents accepting the value during multiple edits of the input while it still has the focus.
Clicking + or - blurs the focus from the input. So that only is an issue when inside the text field.

### Additional comments

[...]

PR #11418, opened 2023-02-03
[input-number] UPDATE_MODEL_EVENT can be emitted correctly when the input value changes from 1 to 10 and back to 1 in the continuous focus input state

### LINKED ISSUE #11417: [Component] [input-number] update:modelValue event cannot be emitted when the value of input-number changes from 1 to 10 and back to 1 in the continuous focus input state
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.45`
- Element Plus Version: `2.2.26`
- Browser / OS: `Chrome 107.0.5304.110 / MacOS BigSur 11.4 `
- Build Tool: `Vite`

### Related Component

- `el-input-number`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtaW5wdXQtbnVtYmVyIHYtbW9kZWw9XCJudW1cIiA6bWluPVwiMVwiIDptYXg9XCIyMDBcIiBAY2hhbmdlPVwiaGFuZGxlQ2hhbmdlXCIgLz5cbiAgPGRpdj5jb3VudDoge3sgbnVtIH19PC9kaXY+XG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IGxhbmc9XCJ0c1wiIHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSAndnVlJ1xuXG5jb25zdCBudW0gPSByZWYoMSlcbjwvc2NyaXB0PlxuIiwiaW1wb3J0X21hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge31cbn0iLCJfbyI6e319)

### Steps to reproduce

1. Focus on the InputNumber component.(The following steps need to always be in the focus state)
2. input 0 (the value of InputNumber change from `1` to `10`)
3. delete 0 (the value of InputNumber change from `10` to `1`)

### What is Expected?

the count of the div is `1` after step 3

the count of the div is `10` after step 3

### Additional comments

_(empty)_

Feb-03-2023 19-19-06

zh-lx:
When the `input` event is continuously triggered by a focus, the `data.currentValue` property will never be changed. Therefore, in some cases, even if the input value changes, the `data.currentValue` is equal to the `newVal`, resulting in the `update:modelValue` event will not be emitted. It should replace `oldVal === newVal` with `props.modelValue === newVal` to ensure that subsequent events can be triggered normally.

This request fix #11417 and #11410


fratzinger:
Thanks for the fix! I used this to finally get a working `ElInputNumber` component.

While it fixed one issue, it introduced another bug for me. When the input loses focus, the `data.currentValue` is unset and the `data.userInput` is shown in the input. So the `modelValue` diverges from the displayed value. @zh-lx can you reproduce this?

Based on this, my current solution looks like this:

```ts
const newVal = verifyValue(value)
if (props.modelValue === newVal) {
  if (data.currentValue !== newVal) {
    data.currentValue = newVal
  }
  return
}
if (!emitChange) {
```

zh-lx:
> 

You're right, solution about this problem #11287 is better so I could close this PR.

PR #11460, opened 2023-02-07
improvement: components/cascader - suffix/prefix icon

SimboDu123:

### DIFF
--- a/packages/components/input/src/input.vue
+++ b/packages/components/input/src/input.vue
@@ -1,6 +1,5 @@
 <template>
   <div
-    v-show="type !== 'hidden'"
     v-bind="containerAttrs"
     :class="containerKls"
     :style="containerStyle"
@@ -223,6 +222,7 @@ const containerKls = computed(() => [
       slots.suffix || props.suffixIcon || props.clearable || props.showPassword,
     [nsInput.bm('suffix', 'password-clear')]:
       showClear.value && showPwdVisible.value,
+    [nsInput.b('hidden')]: props.type === 'hidden',
   },
   rawAttrs.class,
 ])
--- a/packages/theme-chalk/src/input.scss
+++ b/packages/theme-chalk/src/input.scss
@@ -464,3 +464,8 @@
     }
   }
 }
+
+
+@include b(input-hidden) {
+  display: none !important;
+}


packages/components/cascader/src/cascader.vue
```diff
@@ -60,10 +72,11 @@
             <el-icon
               v-else
               key="arrow-down"
-              :class="cascaderIconKls"
+              :class="[cascaderIconKls]"
```
jw-foss, line 75: **Question:** why add this bracket though? Since this is already an array why would we do this?

packages/components/cascader/src/cascader.vue
```diff
@@ -60,10 +72,11 @@
...
               @click.stop="togglePopperVisible()"
             >
-              <arrow-down />
+              <component :is="iconComponent" />
+              <!-- <arrow-down /> -->
```
jw-foss, line 79: **Suggestion:** I think we can remove this part.

packages/components/cascader/src/cascader.vue
```diff
@@ -342,10 +355,14 @@ const cascaderKls = computed(() => {
...
 
+const iconComponent = computed(() =>
+  props.suffixIcon ? props.suffixIcon : ArrowDown
+)
+
 const cascaderIconKls = computed(() => {
   return [
     nsInput.e('icon'),
-    'icon-arrow-down',
+    'icon-down',
```
jw-foss, line 365: I would suggest that we add a redundant class for this instead of removing it, removing this is considered a breaking change which is not what we want right now.

packages/components/cascader/src/cascader.vue
```diff
@@ -48,6 +48,18 @@
           @blur="handleBlur"
           @input="handleInput"
         >
+          <template v-if="$slots.prefix" #prefix>
+            <div
+              style="
```
jw-foss, line 53: **Question:** why do we use inline styling here? Couldn't we put this into `*.scss` file?

packages/components/cascader/src/cascader.ts
```diff
@@ -55,6 +61,10 @@ export const cascaderProps = buildProps({
     type: Boolean,
     default: true,
   },
+  suffixIcon: {
+    type: iconPropType,
+    default: ArrowDown,
```
jw-foss, line 66: **Suggestion:** since we have already made a fallback in the code, I think we can just leave this as `undefined`.

docs/en-US/component/cascader.md
```diff
@@ -144,28 +144,29 @@ cascader/panel
...
+| teleported                        | whether cascader popup is teleported                                                                                                                                             | boolean                 | —                           | true      |
+| popper-append-to-body(deprecated) | whether to append the popper menu to body. If the positioning of the popper is wrong, you can try to set this prop to false                                                      | boolean                 | —                           | true      |
+| tag-type                          | tag type                                                                                                                                                                         | string                  | success/info/warning/danger | info      |
+| suffix-icon                       | Custom suffix icon component                                                                                                                                                     | `string \| Component`   | —                           | ArrowDown |
```
jw-foss, line 168: **Suggestion:** I think we should put this line after `validate-event` to keep the change as small as possible, WDYT?

jw-foss (commented):
@xinbowang left a couple comments for you to consider, let me know if you have any questions.

docs/en-US/component/cascader.md
xinbowang: OK

PR #11492, opened 2023-02-10
fix(components): [select]

web-farmer:
multiple select enables filtering,can't use backspace to delete tag

### DIFF
--- a/packages/components/select/src/useSelect.ts
+++ b/packages/components/select/src/useSelect.ts
@@ -730,7 +730,9 @@ export const useSelect = (props, states: States, ctx) => {
     if (!Array.isArray(states.selected)) return
     const option = states.selected[states.selected.length - 1]
     if (!option) return
-
+    if (option.isDisabled) {
+      return option.isDisabled
+    }
     if (hit === true || hit === false) {
       option.hitState = hit
       return hit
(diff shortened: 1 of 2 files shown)


packages/components/select/src/useSelect.ts
```diff
@@ -730,7 +730,9 @@ export const useSelect = (props, states: States, ctx) => {
     if (!Array.isArray(states.selected)) return
     const option = states.selected[states.selected.length - 1]
     if (!option) return
-
+    if (option.isDisabled) {
+      return option.isDisabled
+    }
```
ryuhangyeong, line 735: ## Opinion

I think this fix introduces another problem where you can't delete a disabled item, but you can't access the old one for a disabled item by default.

image

Unable to access with delete key (backspace)

wforguo: disabled option can't  clearable, it also can't delete by backspace key when enables filtering

wforguo: when I set the default value, it cannot be clearable, but it can be bankspace, which is unreasonable

ryuhangyeong: Yes, as you said, disabled options are not accessible. However, in the current PR situation, option 1 appears to be inaccessible (option 1 is not disabled).

I'm not sure how this works for the user experience.

PR #11522, opened 2023-02-11
feat(components): [tree-v2] add item-size attribute

### LINKED ISSUE #11501: [Component] [tree-v2] treeV2 属性itemSize 设置无效,期望可以自定义node的高度
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.47`
- Element Plus Version: `2.2.29`
- Browser / OS: `chrome`
- Build Tool: `Webpack`

### Related Component

- `el-tree-v2`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtdHJlZS12MiA6ZGF0YT1cImRhdGFcIiA6cHJvcHM9XCJwcm9wc1wiIDpoZWlnaHQ9XCIyMDhcIiA6aXRlbS1zaXplPVwiNDBcIiAvPlxuPC90ZW1wbGF0ZT5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5pbnRlcmZhY2UgVHJlZSB7XG4gIGlkOiBzdHJpbmdcbiAgbGFiZWw6IHN0cmluZ1xuICBjaGlsZHJlbj86IFRyZWVbXVxufVxuXG5jb25zdCBnZXRLZXkgPSAocHJlZml4OiBzdHJpbmcsIGlkOiBudW1iZXIpID0+IHtcbiAgcmV0dXJuIGAke3ByZWZpeH0tJHtpZH1gXG59XG5cbmNvbnN0IGNyZWF0ZURhdGEgPSAoXG4gIG1heERlZXA6IG51bWJlcixcbiAgbWF4Q2hpbGRyZW46IG51bWJlcixcbiAgbWluTm9kZXNOdW1iZXI6IG51bWJlcixcbiAgZGVlcCA9IDEsXG4gIGtleSA9ICdub2RlJ1xuKTogVHJlZVtdID0+IHtcbiAgbGV0IGlkID0gMFxuICByZXR1cm4gQXJyYXkuZnJvbSh7IGxlbmd0aDogbWluTm9kZXNOdW1iZXIgfSlcbiAgICAuZmlsbChkZWVwKVxuICAgIC5tYXAoKCkgPT4ge1xuICAgICAgY29uc3QgY2hpbGRyZW5OdW1iZXIgPVxuICAgICAgICBkZWVwID09PSBtYXhEZWVwID8gMCA6IE1hdGgucm91bmQoTWF0aC5yYW5kb20oKSAqIG1heENoaWxkcmVuKVxuICAgICAgY29uc3Qgbm9kZUtleSA9IGdldEtleShrZXksICsraWQpXG4gICAgICByZXR1cm4ge1xuICAgICAgICBpZDogbm9kZUtleSxcbiAgICAgICAgbGFiZWw6IG5vZGVLZXksXG4gICAgICAgIGNoaWxkcmVuOiBjaGlsZHJlbk51bWJlclxuICAgICAgICAgID8gY3JlYXRlRGF0YShtYXhEZWVwLCBtYXhDaGlsZHJlbiwgY2hpbGRyZW5OdW1iZXIsIGRlZXAgKyAxLCBub2RlS2V5KVxuICAgICAgICAgIDogdW5kZWZpbmVkLFxuICAgICAgfVxuICAgIH0pXG59XG5cbmNvbnN0IHByb3BzID0ge1xuICB2YWx1ZTogJ2lkJyxcbiAgbGFiZWw6ICdsYWJlbCcsXG4gIGNoaWxkcmVuOiAnY2hpbGRyZW4nLFxufVxuY29uc3QgZGF0YSA9IGNyZWF0ZURhdGEoNCwgMzAsIDQwKVxuPC9zY3JpcHQ+XG4iLCJpbXBvcnRfbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7fVxufSIsIl9vIjp7fX0=)

### Steps to reproduce

`<el-tree-v2 :data="data" :props="props" :height="208" :item-size="40" />`
设置itemSize无用

### What is Expected?

期望 tree-v2 可以自定义node的高度

源代码中使用了常量26px

### Additional comments

_(empty)_

btea:

close #11501

### DIFF
--- a/packages/components/tree-v2/src/tree-node.vue
+++ b/packages/components/tree-v2/src/tree-node.vue
@@ -19,7 +19,10 @@
   >
     <div
       :class="ns.be('node', 'content')"
-      :style="{ paddingLeft: `${(node.level - 1) * indent}px` }"
+      :style="{
+        paddingLeft: `${(node.level - 1) * indent}px`,
+        height: itemSize + 'px',
+      }"
     >
       <el-icon
         v-if="icon"
--- a/packages/components/tree-v2/src/tree.vue
+++ b/packages/components/tree-v2/src/tree.vue
@@ -9,7 +9,7 @@
       :data="flattenTree"
       :total="flattenTree.length"
       :height="height"
-      :item-size="itemSize"
+      :item-size="treeNodeSize"
       :perf-mode="perfMode"
     >
       <template #default="{ data, index, style }">
@@ -21,7 +21,7 @@
           :show-checkbox="showCheckbox"
           :checked="isChecked(data[index])"
           :indeterminate="isIndeterminate(data[index])"
-          :item-size="itemSize"
+          :item-size="treeNodeSize"
           :disabled="isDisabled(data[index])"
           :current="isCurrent(data[index])"
           :hidden-expand-icon="isForceHiddenExpandIcon(data[index])"
@@ -57,7 +57,7 @@ const emit = defineEmits(treeEmits)
 
 const slots = useSlots()
 
-const itemSize = computed(() => props.itemSize)
+const treeNodeSize = computed(() => props.itemSize)
 
 provide(ROOT_TREE_INJECTION_KEY, {
   ctx: {
--- a/packages/components/tree-v2/src/tree.vue
+++ b/packages/components/tree-v2/src/tree.vue
@@ -21,6 +21,7 @@
           :show-checkbox="showCheckbox"
           :checked="isChecked(data[index])"
           :indeterminate="isIndeterminate(data[index])"
+          :item-size="itemSize"
           :disabled="isDisabled(data[index])"
           :current="isCurrent(data[index])"
           :hidden-expand-icon="isForceHiddenExpandIcon(data[index])"
@@ -39,7 +40,7 @@
 </template>
 
 <script lang="ts" setup>
-import { getCurrentInstance, provide, useSlots } from 'vue'
+import { computed, getCurrentInstance, provide, useSlots } from 'vue'
 import { useLocale, useNamespace } from '@element-plus/hooks'
 import { formItemContextKey } from '@element-plus/tokens'
 import { FixedSizeList } from '@element-plus/components/virtual-list'
@@ -56,7 +57,7 @@ const emit = defineEmits(treeEmits)
 
 const slots = useSlots()
 
-const itemSize = 26
+const itemSize = computed(() => props.itemSize)
 
 provide(ROOT_TREE_INJECTION_KEY, {
   ctx: {
--- a/packages/components/tree-v2/src/virtual-tree.ts
+++ b/packages/components/tree-v2/src/virtual-tree.ts
@@ -38,6 +38,11 @@ export const enum SetOperationEnum {
   DELETE = 'delete',
 }
 
+const itemSize = {
+  type: Number,
+  default: 26,
+}
+
 // props
 export const treeProps = buildProps({
   data: {
@@ -87,10 +92,7 @@ export const treeProps = buildProps({
     type: Number,
     default: 16,
   },
-  itemSize: {
-    type: Number,
-    default: 26,
-  },
+  itemSize,
   icon: {
     type: iconPropType,
   },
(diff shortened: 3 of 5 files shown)


packages/components/tree-v2/src/virtual-tree.ts
```diff
@@ -87,6 +87,10 @@ export const treeProps = buildProps({
     type: Number,
     default: 16,
   },
+  itemSize: {
```
jw-foss, line 90: **Question:** why were we adding this twice?

packages/components/tree-v2/src/tree.vue
```diff
@@ -56,7 +57,7 @@ const emit = defineEmits(treeEmits)
 
 const slots = useSlots()
 
-const itemSize = 26
+const itemSize = computed(() => props.itemSize)
```
jw-foss, line 60: **Suggestion:** I suggest that we should avoid naming internal variables as same as `props`

jw-foss (approved):

packages/components/tree-v2/src/tree-node.vue
```diff
@@ -19,7 +19,10 @@
   >
     <div
       :class="ns.be('node', 'content')"
-      :style="{ paddingLeft: `${(node.level - 1) * indent}px` }"
+      :style="{
+        paddingLeft: `${(node.level - 1) * indent}px`,
+        height: itemSize + 'px',
+      }"
```
jw-foss, line 25: **Suggestion:** we do not really need this since we have style being passed into this component, and it contains the height we needed here, see this screenshot 👇 

CleanShot 2023-02-12 at 00 14 40@2x

CleanShot 2023-02-12 at 00 15 57@2x

btea: It should still be necessary here, the code in the figure will set the height of `el-tree-node`, but the height of `el-tree-node__content` is set to `26px`.
https://github.com/element-plus/element-plus/blob/06126ea57123ed6cba058a7465123e04865d396f/packages/theme-chalk/src/tree.scss#L63

packages/components/tree-v2/src/tree.vue
btea: Updated.

packages/components/tree-v2/src/virtual-tree.ts
jw-foss: Oh, my mistake, they were added to two different props, perhaps we can do something like this?

```ts
const itemSize = {
  type: Number,
  default: 26,
}

const treeNodeProps = {
  ...others,
  itemSize,
}
```

packages/components/tree-v2/src/tree-node.vue
jw-foss: Maybe, `height: inherit` here is more appropriate?

jw-foss:
@btea in addition to the comments above, do you think we can introduce a test case to cover this feature?

packages/components/tree-v2/src/tree-node.vue
ryuhangyeong: image

Since the height is already set in the parent, there seems to be no problem if you set the height to inherit(`height: inherit`) as @jw-foss said!

packages/components/tree-v2/src/virtual-tree.ts
btea: Looks good.

packages/components/tree-v2/src/tree-node.vue
btea: `tree` and `tree-v2` use the same style file, such modification may affect `tree`.

btea:
> @btea in addition to the comments above, do you think we can introduce a test case to cover this feature?

Yes, I will add a test case and example.

packages/components/tree-v2/src/tree-node.vue
btea: @ryuhangyeong Yes, as shown in your screenshot, `el-tree-node` has set the height, but it is not set in the `el-tree` component. Therefore, I cannot directly modify the `tree.scss` file.  If setting inline styles like I do now, I think it might be more intuitive to use `itemSize`?

ryuhangyeong: @btea oh my mistake. I agree with you!

ryuhangyeong (approved):

btea:
@jw-foss Please review to see if there are other issues, if not I will merge it.

PR #11563, opened 2023-02-13
chore(components): [drawer] use restore active

acyza:
Please make sure these boxes are checked before submitting your PR, thank you!

- [x] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).
- [x] Make sure you are merging your commits to `dev` branch.
- [x] Add some descriptions and refer to relative issues for your PR.

drawer关闭，drawer中内容被隐藏后不该有焦点停留，使用`useRestoreActive`把焦点移出。
似乎elFocusTrap也能移除焦点，但它似乎有条件。https://github.com/element-plus/element-plus/blob/dev/packages/components/focus-trap/src/focus-trap.vue#L290
useRestoreActive和elFocusTrap同时存在应该没啥问题，至少ElMessageBox是这么干的。

协助#11524 修复#10893 
fix #10893 close #10992

PR #9941, opened 2022-09-27
fix(components): [time-picker] fix popup not showing

Giwayume:
Please make sure these boxes are checked before submitting your PR, thank you!

- [x] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).
- [x] Make sure you are merging your commits to `dev` branch.
- [x] Add some descriptions and refer to relative issues for your PR.

**SUMMARY**

The timepicker popup doesn't always show when you click on it.

Fixes https://github.com/element-plus/element-plus/issues/9842

Broken in this PR:
- https://github.com/element-plus/element-plus/pull/9669

Previous issues that lead to the breaking change:
- https://github.com/element-plus/element-plus/issues/9460
- https://github.com/element-plus/element-plus/issues/8501

Additional focus trap changes:
- 'focusout-prevented' event is emitted when focusout of trap container is detected (previously only emitted when reaching the edges of the container by tabbing with the keyboard)
- 'focusout-prevented' can be canceled, to stop the default prevention behavior. It also provides info on whether pointer (mouse/touch) or keyboard triggered the focusout
- Previously el-focus-trap relied on document.activeElement at trap start to be the return element. Setting this manually in the popper content caused issues where the browser refused to change focus. So instead, the 'focusin' event tracks the return element, and that's used if document.activeElement is already inside the focus trap when the trap starts.

Issue #9837, opened 2022-09-20
[Component] [select, dropdown] 同时存在el-select(filterable=true)和el-dropdown时，再次选中当前select的option后，悬停到dropdown并下移光标，select会自动展开


Ryaningli:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.37`
- Element Plus Version: `2.2.14`
- Browser / OS: `Chorme 105.0.5195.102 / Windows 11 22000.978`
- Build Tool: `Vite`

### Related Component

- `el-select`
- `el-dropdown`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICAgICAgPGVsLXNlbGVjdCB2LW1vZGVsPVwidlwiIGZpbHRlcmFibGVcbiAgICAgID48ZWwtb3B0aW9uIHZhbHVlPVwiMVwiIC8+XG4gICAgPC9lbC1zZWxlY3Q+XG4gIDxlbC1kcm9wZG93bj5cbiAgICAgIERyb3Bkb3duIExpc3RcbiAgICA8dGVtcGxhdGUgI2Ryb3Bkb3duPlxuICAgICAgPGVsLWRyb3Bkb3duLW1lbnU+XG4gICAgICAgIDxlbC1kcm9wZG93bi1pdGVtPkFjdGlvbiAxPC9lbC1kcm9wZG93bi1pdGVtPlxuICAgICAgPC9lbC1kcm9wZG93bi1tZW51PlxuICAgIDwvdGVtcGxhdGU+XG4gIDwvZWwtZHJvcGRvd24+XG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IGxhbmc9XCJ0c1wiIHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSBcInZ1ZVwiXG5jb25zdCB2PSByZWYoMSlcbjwvc2NyaXB0PlxuPHN0eWxlIHNjb3BlZD5cbjwvc3R5bGU+XG4iLCJpbXBvcnRfbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7fVxufSIsIl9vIjp7fX0=)

### Steps to reproduce

1.选中选择器中的选项
2.再次展开选择器并选择原选项（此时选择器为入焦且文字被选中的状态）
[...]

PR #11586, opened 2023-02-15
fix(components): [dropdown] fix focus after close

Giwayume:

Continuation of #9941

"Since version 2.2.18, When using the mouse to close the Dialog, the focus will not return to the previous button. It is works normally in 2.2.17. [demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtYnV0dG9uIHRleHQgQGNsaWNrPVwiZGlhbG9nVmlzaWJsZSA9IHRydWVcIj5cbiAgICBjbGljayBtZVxuICA8L2VsLWJ1dHRvbj5cbiAgPGVsLWRpYWxvZyB2LW1vZGVsPVwiZGlhbG9nVmlzaWJsZVwiPlxuICAgIDxzcGFuPkNsaWNrIHRoZSBDb25maXJtIGJ1dHRvbiB3aXRoIHRoZSBtb3VzZTwvc3Bhbj5cbiAgICA8dGVtcGxhdGUgI2Zvb3Rlcj5cbiAgICAgIDxzcGFuIGNsYXNzPVwiZGlhbG9nLWZvb3RlclwiPlxuICAgICAgICA8ZWwtYnV0dG9uIEBjbGljaz1cImRpYWxvZ1Zpc2libGUgPSBmYWxzZVwiPkNhbmNlbDwvZWwtYnV0dG9uPlxuICAgICAgICA8ZWwtYnV0dG9uIHR5cGU9XCJwcmltYXJ5XCIgQGNsaWNrPVwiZGlhbG9nVmlzaWJsZSA9IGZhbHNlXCI+XG4gICAgICAgICAgQ29uZmlybVxuICAgICAgICA8L2VsLWJ1dHRvbj5cbiAgICAgIDwvc3Bhbj5cbiAgICA8L3RlbXBsYXRlPlxuICA8L2VsLWRpYWxvZz5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5pbXBvcnQgeyByZWYgfSBmcm9tICd2dWUnXG5cbmNvbnN0IGRpYWxvZ1Zpc2libGUgPSByZWYoZmFsc2UpXG48L3NjcmlwdD5cbiIsImltcG9ydF9tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHt9XG59IiwiX28iOnt9fQ==)"

**Focus Trap**

Added `trapContainer.contains(document.activeElement)` condition to check if you are still focused inside the focus trap when it is closed, and go ahead and focus on last element focused before the trap was opened if so.

I am weary of this change, but I cannot find a previous test case that it breaks... so...

**Dropdown**

I have implemented a quick fix for https://github.com/element-plus/element-plus/issues/9837 here. Ultimately, it would be good to remove any sort of `.focus()` calls that occur during pointer enter/leave, as this is unnecessary. But it looks like that's more of a significant change I don't want to tackle and risk breaking things right now.

I ended up changing the code so if the user hovers over the tooltip target (that opens the dropdown), that target is focused. This works in-line with each item in the dropdown receiving focus during hover, and gives the focus trap a sane element to return to when the dropdown closes.

### DIFF
--- a/packages/components/dropdown/src/dropdown.vue
+++ b/packages/components/dropdown/src/dropdown.vue
@@ -49,7 +49,12 @@
         </el-scrollbar>
       </template>
       <template v-if="!splitButton" #default>
-        <el-only-child :id="triggerId" role="button" :tabindex="tabindex">
+        <el-only-child
+          :id="triggerId"
+          ref="triggeringElementRef"
+          role="button"
+          :tabindex="tabindex"
+        >
           <slot name="default" />
         </el-only-child>
       </template>
@@ -95,6 +100,7 @@ import {
   ref,
   toRef,
   unref,
+  watch,
 } from 'vue'
 import ElButton from '@element-plus/components/button'
 import ElTooltip from '@element-plus/components/tooltip'
@@ -102,7 +108,7 @@ import ElScrollbar from '@element-plus/components/scrollbar'
 import ElIcon from '@element-plus/components/icon'
 import ElRovingFocusGroup from '@element-plus/components/roving-focus-group'
 import { ElOnlyChild } from '@element-plus/components/slot'
-import { addUnit } from '@element-plus/utils'
+import { addUnit, isArray } from '@element-plus/utils'
 import { ArrowDown } from '@element-plus/icons-vue'
 import { EVENT_CODE } from '@element-plus/constants'
 import { useId, useLocale, useNamespace, useSize } from '@element-plus/hooks'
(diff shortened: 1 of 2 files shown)


acyza:
close #11563

tolking:
It is indeed unreasonable to trigger the exit of FocusTrap through focus in components.

Maybe we should revise it step by step.

jw-foss (approved):

jw-foss:
LGTM 👍 thank you for following this up.

packages/components/dropdown/src/dropdown.vue
```diff
@@ -152,6 +158,39 @@ export default defineComponent({
...
+            onAutofocusTriggerEnter
+          )
+        }
+        if (
+          triggeringElement?.$el?.addEventListener &&
+          triggerArray.includes('hover')
+        ) {
+          triggeringElement.$el.addEventListener(
+            'pointerenter',
+            onAutofocusTriggerEnter
```
tolking, line 187: Do we need to remove the onAutofocusTriggerEnter function in onBeforeUnmount?

Giwayume: Is there a scenario where dropdown is unmounted and its slot content isn't discarded?

tolking (approved):

packages/components/dropdown/src/dropdown.vue
Giwayume: I added it in case the virtual dom diffing keeps that node around. Better safe than sorry I suppose.

tolking (approved):

PR #11649, opened 2023-02-18
fix(components): [dropdown] unmounted error

chenxch:
1. open [dropdown](https://element-plus.org/en-US/component/dropdown.html)
2. switch other component
3. view console

The reason is that when unloading, the used object does not exist.

### DIFF
--- a/packages/components/dropdown/src/dropdown.vue
+++ b/packages/components/dropdown/src/dropdown.vue
@@ -96,7 +96,7 @@ import {
   computed,
   defineComponent,
   getCurrentInstance,
-  onUnmounted,
+  onBeforeUnmount,
   provide,
   ref,
   toRef,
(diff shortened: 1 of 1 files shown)


tolking (approved):

plainheart (changes requested):
This change only avoids the NPE but cannot fundamentally solve the problem. The value of `triggeringElementRef` has been cleaned and become undefined when it turns to the `onUnmounted` hook. Thus I think it's good to use `onBeforeUnmount` instead.

chenxch:
> This change only avoids the NPE but cannot fundamentally solve the problem. The value of `triggeringElementRef` has been cleaned and become undefined when it turns to the `onUnmounted` hook. Thus I think it's good to use `onBeforeUnmount` instead.

You're right.

plainheart (approved):
LGTM :)

Issue #11637, opened 2023-02-17
[Component] [select] el-select绑定对象数组时，如果这个对象在el-options中不存在时会出现空标签


Mr-Nobody-li:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.33`
- Element Plus Version: `2.1.11`
- Browser / OS: `Chrome 96`
- Build Tool: `Vite`

### Related Component

- `el-select`

### Reproduction Link

[...]

PR #11739, opened 2023-02-25
feat(components): [select] add label-key prop (#11637)

eas0nchan:
if the options data is got from server-side and the value is object, maybe the label can not show immediate. In the code, the label-key of the object value is fixed, maybe better to provide a label-key prop to the user to set.

### DIFF
--- a/packages/components/select/src/select.vue
+++ b/packages/components/select/src/select.vue
@@ -366,6 +366,10 @@ export default defineComponent({
       type: String,
       default: 'value',
     },
+    labelKey: {
+      type: String,
+      default: 'label',
+    },
     collapseTags: Boolean,
     collapseTagsTooltip: {
       type: Boolean,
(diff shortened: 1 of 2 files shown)


tolking:
Thank you for your contribution, but I do not think this PR is very reasonable.

When value is an object, it cannot be determined that there is a label in the object. adding the label-key is meaningless.

eas0nchan:
> Thank you for your contribution, but I do not think this PR is very reasonable.
> 
> When value is an object, it cannot be determined that there is a label in the object. adding the label-key is meaningless.

eas0nchan:
> Thank you for your contribution, but I do not think this PR is very reasonable.
> 
> When value is an object, it cannot be determined that there is a label in the object. adding the label-key is meaningless.
> 

Yes, it is a little strange. But the case of issue #11637 is also strange.😂

PR #11772, opened 2023-03-01
feat(input/textarea): improve padding end behaviour

Hazel-Lin:
When we type many lines into textarea and  find no gap at the bottom line.

image

add attribute to solve it

### DIFF
--- a/packages/theme-chalk/src/input.scss
+++ b/packages/theme-chalk/src/input.scss
@@ -49,6 +49,7 @@
     width: 100%;
     font-size: inherit;
     font-family: inherit;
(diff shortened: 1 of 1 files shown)


chenxch:
@Hazel-Lin Thanks for your contribution, but I personally vote against this proposal.

tolking:
Tanks for your contribution. 

IMO, This PR is a good idea. Although `type="textarea"` has a padding style, it cannot affect the scrolling content, which causes no gap between the text and the bottom border. If you delete a line of text, there will be a gap between the text and the bottom. I think this strange behavior can be optimized. cc @element-plus/backers

PR #11814, opened 2023-03-04
fix(components): [table-v2] reduce unnecessary update of cell component

makedopamine:
This PR could solve the performance problem which is mentioned in #11650  to a certain extent.
I removed `style={_columnsStyles[props.column.key]}` because `props.style` and `_columnsStyles[props.column.key]` seem to be the same object.

### DIFF
--- a/packages/components/table-v2/src/table-v2.tsx
+++ b/packages/components/table-v2/src/table-v2.tsx
@@ -273,11 +273,7 @@ const TableV2 = defineComponent({
                     {slots.cell(props)}
                   </Cell>
                 ) : (
-                  <Cell
-                    {...props}
-                    {...tableCellProps}
-                    style={_columnsStyles[props.column.key]}
-                  />
+                  <Cell {...{ ...props, ...tableCellProps }} />
                 ),
             }}
           </Row>


KarolMilewski (commented):
lgtm :) nice work

packages/components/table-v2/src/table-v2.tsx
```diff
@@ -273,11 +273,7 @@ const TableV2 = defineComponent({
                     {slots.cell(props)}
                   </Cell>
                 ) : (
-                  <Cell
-                    {...props}
-                    {...tableCellProps}
-                    style={_columnsStyles[props.column.key]}
-                  />
+                  <Cell {...{ ...props, ...tableCellProps }} />
```
btea, line 276: Maybe we can use `Object.assign` instead, now the code looks a bit weird.
```suggestion
                  <Cell {...Object.assign({}, props, tableCellProps)} />
```

qgao233: I just want to know how this fix will be published in latest version

qgao233:
hurry up and merge quickly plz, then publish!!!

makedopamine:
@btea
In this PR, the issue still exists if there is a `cell` slot. The PR  #12378  address the issue in a better way, can we speed up it? 
I will close this PR later.

PR #11889, opened 2023-03-08
style(theme-chalk): use variable instead of '-'

fishermanxzx:
Sorry,I closed it unexpectedly,PR again .

It does not work if change $common-separator.

so use variable $common-separator instead of '-'.

closed #11796

### DIFF
--- a/packages/theme-chalk/src/mixins/mixins.scss
+++ b/packages/theme-chalk/src/mixins/mixins.scss
@@ -54,7 +54,7 @@
 
 // BEM
 @mixin b($block) {
-  $B: $namespace + '-' + $block !global;
+  $B: $namespace + $common-separator + $block !global;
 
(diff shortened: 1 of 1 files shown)


jw-foss:
@YunYouJun would you mind reviewing this?

ryuhangyeong (approved):

YunYouJun:
That's OK.

But it seems to me that we are complicating some simple things.
Do we need to customize the separator here?

Can you provide a scenario?

fishermanxzx:
@YunYouJun 
Because the place needs to be consistent with the following.
file:https://github.com/element-plus/element-plus/blob/dev/packages/theme-chalk/src/mixins/function.scss    74 lines
if $common-separator change the function bem will change but the mixin b will not。

PR #11976, opened 2023-03-11
fix(components): [text] not support multi-line ellipsis

gimjin:

### DIFF
--- a/docs/en-US/component/text.md
+++ b/docs/en-US/component/text.md
@@ -25,7 +25,7 @@ text/sizes
 
 ## Ellipsis
 
-:::demo Pass the `truncated` prop to render an ellipsis when the text exceeds the width of the viewport or max-width set.
+:::demo Pass the `truncated` prop to render an ellipsis when the text exceeds the width of the viewport or max-width set. `line-clamp` prop to render multiline ellipsis.
 
 text/truncated
 
@@ -51,12 +51,13 @@ text/mixed
 
 ### Attributes
 
-| Name      | Description        | Type                                                               | Default |
-| --------- | ------------------ | ------------------------------------------------------------------ | ------- |
-| type      | text type          | ^[enum]`'primary' \| 'success' \| 'warning' \| 'danger' \| 'info'` | —       |
-| size      | text size          | ^[enum]`'large' \| 'default' \| 'small'`                           | default |
-| truncated | render ellipsis    | ^[boolean]                                                         | false   |
-| tag       | custom element tag | ^[string]                                                          | span    |
+| Name       | Description        | Type                                                               | Default |
+| ---------- | ------------------ | ------------------------------------------------------------------ | ------- |
+| type       | text type          | ^[enum]`'primary' \| 'success' \| 'warning' \| 'danger' \| 'info'` | —       |
+| size       | text size          | ^[enum]`'large' \| 'default' \| 'small'`                           | default |
+| truncated  | render ellipsis    | ^[boolean]                                                         | false   |
+| line-clamp | maximum lines      | ^[string] / ^[number]                                              | -       |
+| tag        | custom element tag | ^[string]                                                          | span    |
 
 ### Slots
 
--- a/docs/en-US/component/text.md
+++ b/docs/en-US/component/text.md
@@ -51,13 +51,13 @@ text/mixed
 
 ### Attributes
 
-| Name       | Description        | Type                                                               | Default |
-| ---------- | ------------------ | ------------------------------------------------------------------ | ------- |
-| type       | text type          | ^[enum]`'primary' \| 'success' \| 'warning' \| 'danger' \| 'info'` | —       |
-| size       | text size          | ^[enum]`'large' \| 'default' \| 'small'`                           | default |
-| truncated  | render ellipsis    | ^[boolean]                                                         | false   |
-| line-clamp | maximum lines      | ^[string] / ^[number]                                              | -       |
-| tag        | custom element tag | ^[string]                                                          | span    |
+| Name                | Description        | Type                                                               | Default |
+| ------------------- | ------------------ | ------------------------------------------------------------------ | ------- |
+| type                | text type          | ^[enum]`'primary' \| 'success' \| 'warning' \| 'danger' \| 'info'` | —       |
+| size                | text size          | ^[enum]`'large' \| 'default' \| 'small'`                           | default |
+| truncated           | render ellipsis    | ^[boolean]                                                         | false   |
+| line-clamp ^(2.4.0) | maximum lines      | ^[string] / ^[number]                                              | -       |
+| tag                 | custom element tag | ^[string]                                                          | span    |
 
 ### Slots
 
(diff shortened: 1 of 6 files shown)


gimjin:
Hi @jw-foss @ryuhangyeong hope to catch the v2.3.0, this PR has ellipsis tooltip and event.

ryuhangyeong (commented):

## My opinion

Is there a reason why tooltips should be embedded in el-text?

this can in fact be done in the following way.

```vue
<template>
  <div>
      <el-tooltip>
        <el-text style="width: 150px;" truncated>
          Self element set width 100px
        </el-text>
        <template #content>
          Custom tooltip content
        </template>
      </el-tooltip>
    <br />
    <br />
    <el-text style="width: 150px;" type="primary">Primary</el-text>
    <el-text style="width: 150px;" type="success">Success</el-text>
    <el-text style="width: 150px;" type="info">Info</el-text>
    <el-text style="width: 150px;" type="warning">Warning</el-text>
    <el-text style="width: 150px;" type="danger">Danger</el-text>
  </div>
</template>
```

gimjin:
`truncated` prop support automatically squeezed by parent element, ellipsis may not always show. So use el-tooltip wrap unable to switch automatically disabled tooltip.

el-text `tooltip` prop appears only if an ellipsis is present. ResizeObserver and watch slot.default() change reset inside el-tooltip `disabled` prop.

packages/components/text/src/text.vue
```diff
@@ -24,5 +40,50 @@ const textKls = computed(() => [
...
   ns.m(textSize.value),
   ns.is('truncated', props.truncated),
+  ns.is('line-clamp', props.lineClamp !== undefined),
 ])
+
+const textRef = ref<HTMLElement>()
+const showEllipsis = ref<boolean>(false)
+
+const tooltipOptions = computed(() => {
+  if (typeof props.tooltip === 'object') {
```
ryuhangyeong, line 50:
```vue
import { isObject } from '@element-plus/utils'

if (isObject(props.tooltip)) {

}
```

packages/components/text/src/text.vue
```diff
@@ -24,5 +40,50 @@ const textKls = computed(() => [
...
+  return {
+    disabled: !props.tooltip || !showEllipsis.value,
+  }
+})
+
+function computeShowEllipsis(target: HTMLElement) {
+  if (props.truncated) {
+    showEllipsis.value = target.scrollWidth > target.clientWidth
+  }
+  if (props.lineClamp !== undefined) {
```
ryuhangyeong, line 66:
```
import { isUndefined } from '@element-plus/utils'
```

ryuhangyeong (commented):
~~Additionally, does this change work in an SSR environment? 🤔~~

packages/components/text/src/text.vue
```diff
@@ -1,11 +1,25 @@
...
+      ref="textRef"
+      :class="textKls"
+      :style="{ '-webkit-line-clamp': lineClamp }"
+      v-bind="$attrs"
+    >
+      <slot />
+    </component>
+    <template #content>
+      <slot v-if="$slots.tooltip" name="tooltip" />
+      <slot v-else />
```
ryuhangyeong, line 14: ## Question

```vue
<slot v-else />
```

What case does this part apply to?

packages/components/text/src/text.vue
```diff
@@ -15,6 +29,8 @@ defineOptions({
 })
 
 const props = defineProps(textProps)
+const slots = useSlots()
+const emits = defineEmits(['ellipsis'])
```
ryuhangyeong, line 33: ## Suggest

It is recommended to separate them.

### reference

https://github.com/element-plus/element-plus/blob/dev/packages/components/avatar/src/avatar.ts#L60-L62

packages/components/text/src/text.vue
gimjin: If no use tooltip slot, else use default slot, show in tooltip content.

jw-foss:
Actually I was wondering if this is overkill for this component, I referenced from https://chakra-ui.com/docs/components/text/usage, I think this component should be UI only, if we do want something that shows a tooltip, our user can do that themselves or maybe we can provide an dedicated component for them for instance we call it `el-ellipse` or `el-overflow-text`. I personally want all the components to stay as simple as possible by in the mean time we can provide more component with much simpler API for the users, in the end the API number will not change, WDYT @gimjin? /cc @ryuhangyeong 

This kind of issue has been existed in this project for a very long time, a component has been used for tons of things and the code is barely readable. With iterations and the maintainers come and go, no one really knows why the code was written in that way.

gimjin:
Yes @jw-foss, we shouldn't wrap each other EP component inside, the component should be simple enough for the users to compose like a lego.

I'm disappointed with a lot of F2E, only writing CRUD pages. Like automatically display tooltip, subconsciously they can't do it, although it's so simple. You woke me up in a nightmare.

I will delete the support for tooltip.

ryuhangyeong:
@jw-foss Thanks for your comments! I also agree.

packages/components/text/__tests__/text.test.tsx
```diff
@@ -5,43 +5,56 @@ import Text from '../src/text.vue'
...
-    expect(vm.$el.classList.contains('el-text')).toEqual(true)
-    expect(wrapper.text()).toEqual(AXIOM)
+    expect(wrapper.classes()).toContain('el-text')
   })
 
   test('type', () => {
-    const wrapper = mount(() => <Text type="success" />)
-    const vm = wrapper.vm
-    expect(vm.$el.classList.contains('el-text--success')).toEqual(true)
+    const wrapper = mount(<Text type="success" />)
```
ryuhangyeong, line 21: ## Suggest

Can you change back to the old way? Most of the test code follows the method below.

```vue
mount(() => <Text />)
```

gimjin: ok, go home from get off work and continue.

ryuhangyeong (approved):

docs/en-US/component/text.md
```diff
@@ -51,12 +51,13 @@ text/mixed
...
-| type      | text type          | ^[enum]`'primary' \| 'success' \| 'warning' \| 'danger' \| 'info'` | —       |
-| size      | text size          | ^[enum]`'large' \| 'default' \| 'small'`                           | default |
-| truncated | render ellipsis    | ^[boolean]                                                         | false   |
-| tag       | custom element tag | ^[string]                                                          | span    |
+| Name       | Description        | Type                                                               | Default |
+| ---------- | ------------------ | ------------------------------------------------------------------ | ------- |
+| type       | text type          | ^[enum]`'primary' \| 'success' \| 'warning' \| 'danger' \| 'info'` | —       |
+| size       | text size          | ^[enum]`'large' \| 'default' \| 'small'`                           | default |
+| truncated  | render ellipsis    | ^[boolean]                                                         | false   |
+| line-clamp | maximum lines      | ^[string] / ^[number]                                              | -       |
```
btea, line 59:
```suggestion
| line-clamp ^(2.3.13) | maximum lines      | ^[string] / ^[number]                                              | -       |
```

tolking (approved):

btea (approved):

gimjin:
🎉 🎉 🎉

PR #12189, opened 2023-03-24
chore: prevent warnings while running test cases

makedopamine:
Some warnings were encountered while executing test cases.

- [x] [Vue warn]: Component is missing template or render function. (mutiple)
- [x] [Vue warn]: inject() can only be used inside setup() or functional components.(mutiple)
- [x] breadcrumb.test.tsx > Breadcrumb.vue > single
    > toRefs() expects a reactive object but received a plain one.
- [x] cascader.test.tsx > Cascader.vue > should be able to trigger togglePopperVisible outside the component
    >  [Vue warn]: Unhandled error during execution of native event handler
    >   at \<Anonymous >
    >   at \<Anonymous ref="VTU_COMPONENT" >
    >   at \<VTUROOT>
    > Error: Uncaught [TypeError: Cannot read properties of undefined (reading 'togglePopperVisible')]


btea:
There seems to be a conflict in the code, can you fix it?

makedopamine:
The conflict has resolved.  But I'm curious about why there is a conflict, I notice that the test file has not changed in `dev` branch. 🤔

packages/components/cascader/__tests__/cascader.test.tsx
```diff
@@ -428,15 +429,18 @@ describe('Cascader.vue', () => {
...
+      cascader.togglePopperVisible()
     }
     const wrapper = _mount(() => (
       <div>
         <Cascader options={OPTIONS} />
         <button onClick={clickFn} />
       </div>
     ))
+
+    cascader = wrapper.findComponent(Cascader).vm
```
btea, line 443: The effect should be the same before and after modification here, I don’t understand the difference. 🤔

btea (approved):

tolking (approved):

PR #12262, opened 2023-03-30
feat(components): [tabs] add tab-pane provide

yuhengshen:

I need to get the current TabPane props/state (such as `active` and the `label`) in the components inside TabPane to perform corresponding operations.

`provide` work well.

### DIFF
--- a/packages/components/tabs/src/constants.ts
+++ b/packages/components/tabs/src/constants.ts
@@ -1,4 +1,11 @@
-import type { ComputedRef, InjectionKey, Ref, Slots, UnwrapRef } from 'vue'
+import type {
+  ComputedRef,
+  DeepReadonly,
+  InjectionKey,
+  Ref,
+  Slots,
+  UnwrapRef,
+} from 'vue'
 import type { TabsProps } from './tabs'
 import type { TabPaneProps } from './tab-pane'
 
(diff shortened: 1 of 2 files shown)


btea:
@yuhengshen  Hi, first of all thank you for your contribution. However, according to your comment description, I am not very clear about your needs. Can you provide a more specific explanation? 🤔

yuhengshen:
@btea  In most management systems, permission management is more common. I need to automatically perform permission management based on the current button text. But in Tabs, there will be many identical buttons, but the permissions required are different. Therefore, I need to get the current label in the child component of tab pane, and combine it with the text to do permission control.

Example diagram
image

btea:
I seem to understand what you mean, so the modification you made is to provide information in the tab-pane component, and then you use the custom component in the tab-pane to inject related information through `tabsPaneContextKey`?

If so, I think you can also provide tab related information where you use the el-tabs component, and then inject it inside the custom component.

yuhengshen:
el-tabs have injected information about itself, but not enough.
https://github.com/element-plus/element-plus/blob/87e9348d8ecbcdd5ae8b6d17a8e9b6b452b8c12d/packages/components/tabs/src/tabs.tsx#L177-L182

yuhengshen:
@btea ok, you are right。I can wrap a custom component of el-pane, el-pane is different from element-ui, it has no strong parent-child relationship with el-tabs, you can close this pr. Thank you

PR #12341, opened 2023-04-06
Revert "fix(components): [popover] default trigger"

chenxch:
Reverts element-plus/element-plus#12321

### DIFF
--- a/packages/components/popover/src/popover.ts
+++ b/packages/components/popover/src/popover.ts
@@ -8,6 +8,7 @@ import type { ExtractPropTypes, PropType } from 'vue'
 import type Popover from './popover.vue'
 
 export const popoverProps = buildProps({
+  trigger: useTooltipTriggerProps.trigger,
   placement: dropdownProps.placement,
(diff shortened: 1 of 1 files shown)


ryuhangyeong:
@chenxch Is there a problem with that code? 🤔

chenxch:
> @chenxch Is there a problem with that code? 🤔

The default trigger method of popover should be hover, not click.

btea:
If so, this revert should be merged with #12342.

MrWeilian:
😅 unlucky

chenxch:
> 😅 unlucky

Don't worry, luck will favor everyone who really loves open source.😬

plainheart (approved):

MrWeilian:
😬hhh. I found that element-ui default trigger is "click" !

chenxch:
> 😬hhh. I found that element-ui default trigger is "click" !

Yes, but we have compared the practices of several mainstream warehouses with popover by default.

PR #12474, opened 2023-04-16
fix(components): [notification] fix props.zIndex lose efficacy bug

Gnalvin:
option zIndex for notification can't effect


packages/components/notification/src/notification.vue
```diff
@@ -84,7 +84,7 @@ const verticalProperty = computed(() =>
 const positionStyle = computed<CSSProperties>(() => {
   return {
     [verticalProperty.value]: `${props.offset}px`,
-    zIndex: currentZIndex.value,
+   zIndex: props.zIndex ? props.zIndex : currentZIndex.value,
```
jtejederabit, line 87: ```zIndex: props.zIndex ?? currentZIndex.value```

jtejederabit (commented):
👍

packages/components/notification/src/notification.vue
```diff
@@ -84,7 +84,7 @@ const verticalProperty = computed(() =>
 const positionStyle = computed<CSSProperties>(() => {
   return {
     [verticalProperty.value]: `${props.offset}px`,
-    zIndex: currentZIndex.value,
+    zIndex: props.zIndex ?? currentZIndex.value,
```
tolking, line 87: The default value of `zIndex` attribute  is `0`, so the `props.zIndex ?? currentZIndex.value` is always is `0`.

The default value of `zIndex` attribute is unreasonable, maybe we should remove it.

packages/components/notification/src/notification.ts
```diff
@@ -107,8 +107,8 @@ export const notificationProps = buildProps({
    * @description initial zIndex
    */
   zIndex: {
-    type: Number,
-    default: 0,
+    type: definePropType<number | null>([Number, null]),
+    default: null,
   },
```
tolking, line 112:
```
zIndex: Number,
```

tolking (approved):

btea (approved):

PR #12549, opened 2023-04-22
fix(components): [form] improve types

ModyQyW:

This PR was started as a discussion at https://github.com/element-plus/element-plus/discussions/12373.

In element-plus@2.3.3, we have `export declare type FormRules = Partial<Record<string, Arrayable<FormItemRule>>>;`, which means the keys of the rules object can be any string and it may causes unexpected errors. An example shows below.

```typescript
// use reactive from vue
const form = reactive({
  id: 0,
  name: '',
});
// or use ref from vue
const form = ref({
  id: 0,
  name: '',
});

// the keys of the rules object can be any string
// even not in the `form` model
const formRules: FormRules = {
  id: { ... },
  name: { ... },
  // no ts error
  code: { ... },
  ...,
};
```

This PR adds an optional generic to the `FormRules` type and brings better type assertion if someone needs it.

```typescript
const formRules: FormRules<typeof form> = {
  id: { ... },
  name: { ... },
  // ts error
  // Type '{ id: {}; name: {}; code: {}; }' is not assignable to type 'Partial<Record<"id" | "name", Arrayable<FormItemRule>>>'.
  // Object literal may only specify known properties, and 'code' does not exist in type 'Partial<Record<"id" | "name", Arrayable<FormItemRule>>>'.ts(2322)
  code: { ... },
  ...,
};
```

Previous codes still work because the generic is optional.

```typescript
const formRules: FormRules = {
  id: { ... },
  name: { ... },
  // no ts error
  code: { ... },
  ...,
};
```

### DIFF
--- a/docs/examples/form/custom-validation.vue
+++ b/docs/examples/form/custom-validation.vue
@@ -73,13 +73,18 @@ const validatePass2 = (rule: any, value: any, callback: any) => {
   }
 }
 
-const ruleForm = reactive({
+interface RuleForm {
+  pass: string
+  checkPass: string
+  age: string
+}
+const ruleForm = reactive<RuleForm>({
   pass: '',
   checkPass: '',
   age: '',
 })
 
-const rules = reactive<FormRules>({
+const rules = reactive<FormRules<RuleForm>>({
   pass: [{ validator: validatePass, trigger: 'blur' }],
   checkPass: [{ validator: validatePass2, trigger: 'blur' }],
   age: [{ validator: checkAge, trigger: 'blur' }],
--- a/docs/examples/form/custom-validation.vue
+++ b/docs/examples/form/custom-validation.vue
@@ -73,18 +73,13 @@ const validatePass2 = (rule: any, value: any, callback: any) => {
   }
 }
 
-interface RuleForm {
-  pass: string
-  checkPass: string
-  age: string
-}
-const ruleForm = reactive<RuleForm>({
+const ruleForm = reactive({
   pass: '',
   checkPass: '',
   age: '',
 })
 
-const rules = reactive<FormRules<RuleForm>>({
+const rules = reactive<FormRules<typeof ruleForm>>({
   pass: [{ validator: validatePass, trigger: 'blur' }],
   checkPass: [{ validator: validatePass2, trigger: 'blur' }],
   age: [{ validator: checkAge, trigger: 'blur' }],
--- a/packages/components/form/src/types.ts
+++ b/packages/components/form/src/types.ts
@@ -20,11 +20,52 @@ export type FormLabelWidthContext = ReturnType<typeof useFormLabelWidth>
 export interface FormItemRule extends RuleItem {
   trigger?: Arrayable<string>
 }
+
+// Reference https://stackoverflow.com/questions/58434389/typescript-deep-keyof-of-a-nested-object/58436959
+// Concatenates two strings with a dot in the middle, unless the last string is empty
+// Join<"a","b"> => "a.b"
+// Join<"a",""> => "a"
+type Join<K, P> = K extends string | number
+  ? P extends string | number
+    ? `${K}${'' extends P ? '' : '.'}${P}`
+    : never
+  : never
+// Limit the recursion to avoid typescript error
+type Prev = [
+  never,
+  0,
+  1,
+  2,
+  3,
+  4,
+  5,
+  6,
+  7,
+  8,
+  9,
+  10,
+  11,
+  12,
+  13,
+  14,
+  15,
+  16,
+  17,
+  18,
+  19,
+  20,
+  ...0[]
+]
+type Leaves<T, D extends number = 10> = [D] extends [never]
+  ? never
+  : T extends object
+  ? { [K in keyof T]-?: Join<K, Leaves<T[K], Prev[D]>> }[keyof T]
+  : ''
 export type FormRules<
   T extends MaybeRef<Record<string, any> | string> = string
 > = Partial<
   Record<
-    UnwrapRef<T> extends string ? UnwrapRef<T> : keyof UnwrapRef<T>,
+    UnwrapRef<T> extends string ? UnwrapRef<T> : Leaves<UnwrapRef<T>>,
     Arrayable<FormItemRule>
   >
 >
--- a/packages/components/form/src/types.ts
+++ b/packages/components/form/src/types.ts
@@ -1,3 +1,4 @@
+import { ref } from 'vue'
 import type { SetupContext, UnwrapRef } from 'vue'
 import type {
   RuleItem,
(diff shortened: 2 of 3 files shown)


tolking:
It's a good idea, but it doesn't work in all cases right now

eg.

```
const form = ref({
  id: 0,
  name: '',
  obj: { a: '' },
  arr: [{ value: '' }],
  //...
});
const formRules = {
  id: {},
  name: {},
  'obj.a': {},
  'arr.0.value': {},
  //...
}
```

ModyQyW:
I think writing `obj.a` is sensible. I will try to cover this case.

However, I think writing`arr.0.value` in `formRules` is kind of weird. It is hard to set rules in types for every array element, and it is easier and better to bind rules on ElFormItem. Should it be supported?

ModyQyW:
@tolking Hi. I have just updated the type. Can you please review it? Thanks!

tolking:
I think we should support all possible cases

ModyQyW:
> It's a good idea, but it doesn't work in all cases right now
> 
> eg.
> 
> ```
> const form = ref({
>   id: 0,
>   name: '',
>   obj: { a: '' },
>   arr: [{ value: '' }],
>   //...
> });
> const formRules = {
>   id: {},
>   name: {},
>   'obj.a': {},
>   'arr.0.value': {},
>   //...
> }
> ```

> I think we should support all possible cases

Hi @tolking. The latest submission should cover the cases you listed. Can you please review it and give me some advice/suggestions? Thanks! 🙏

image

image

packages/components/form/src/types.ts
```diff
@@ -19,7 +20,55 @@ export type FormLabelWidthContext = ReturnType<typeof useFormLabelWidth>
...
+  ? { [K in keyof T]-?: Join<K, Leaves<T[K], Prev[D]>> }[keyof T]
+  : ''
+export type FormRules<
+  T extends MaybeRef<Record<string, any> | string> = string
+> = Partial<
+  Record<
+    UnwrapRef<T> extends string ? UnwrapRef<T> : Leaves<UnwrapRef<T>>,
+    Arrayable<FormItemRule>
+  >
+>
```
tolking, line 71: IMO, Perhaps we should only support simple types, using MaybeRef seems redundant. what do you think?

example

```ts
interface Form {
  a: number
}

const form = ref<Form>() // or reactive<Form>()

const formRules: FormRules<Form> = {
  a: { ... },
}
```

packages/components/form/src/types.ts
```diff
@@ -19,7 +20,55 @@ export type FormLabelWidthContext = ReturnType<typeof useFormLabelWidth>
...
+  19,
+  20,
+  ...0[]
+]
+type Leaves<T, D extends number = 10> = [D] extends [never]
+  ? never
+  : T extends object
+  ? { [K in keyof T]-?: Join<K, Leaves<T[K], Prev[D]>> }[keyof T]
+  : ''
+export type FormRules<
```
tolking, line 64: It would be even better if you added some explanations to the type.

[link](https://github.com/element-plus/element-plus/blob/dev/packages/utils/vue/props/types.ts)

ModyQyW:
@tolking Hi. I have just updated the types, added comments, and fix some type errors.

> Now TS has optimized the processing of recurs, it no longer needs to be written like this.

Thanks! I have removed it.

> IMO, Perhaps we should only support simple types, using `MaybeRef` seems redundant. what do you think?

I think it is a bad idea. It will increase the type burden on the user side - to use this feature, users will have to write the type by hand and could not easily resort to TypeScript type derivation.

> It would be even better if you added some description to the type.

Thanks! I have added some.

packages/components/form/src/types.ts
```diff
@@ -19,7 +21,82 @@ export type FormLabelWidthContext = ReturnType<typeof useFormLabelWidth>
...
+ * @see {@link FieldPath}
+ */
+type Path<T> = T extends ReadonlyArray<infer V>
+  ? IsTuple<T> extends true
+    ? {
+        [K in TupleKey<T>]-?: PathImpl<K & string, T[K]>
+      }[TupleKey<T>] // tuple
+    : PathImpl<ArrayKey, V> // array
+  : {
+      [K in keyof T]-?: PathImpl<K & string, T[K]>
```
tolking, line 81: > PathImpl<K extends string | number, V>

Why use `K & string`? For an object, the number can also be used as a key.

packages/components/form/src/types.ts
```diff
@@ -19,7 +21,82 @@ export type FormLabelWidthContext = ReturnType<typeof useFormLabelWidth>
...
+    }[keyof T] // object
+/**
+ * Type which collects all paths through a type
+ *
+ * 通过一个类型收集所有路径的类型
+ *
+ * @example
+ * FieldPath<{ a: number; b: string; c: { d: number; e: string }; f: [{ value: string }]; g: { value: string }[] }> => 'a' | 'b' | 'c.d' | 'c.e' | 'f.0' | 'f.0.value' | 'g.number' | 'g.number.value'
```
tolking, line 89: I found that the generated type does not contain `'c' ...`, but some components in the component library accept an object, such as ElSelect

packages/components/form/src/types.ts
ModyQyW: Yes you are right. I just missed that.

packages/components/form/src/types.ts
ModyQyW: That's weird. After resolving the issue above, the type seems to contain `c` already in my side. I also update the jsdoc. Can you please check that again?

image

tolking (approved):
Can you update the types in the examples?

ModyQyW:
> Can you update the types in the examples?

Do you mean `element-plus/docs/examples/form`? Of course, I can.

Before I do this, I want to confirm:

1. Should I add a new part after [Number Validate](https://element-plus.org/en-US/component/form.html#number-validate) or somewhere else? Or should I just update all validation examples' types?

2. Which way do you prefer? Use `interface Form` in the code as a recommendation and only mention `typeof form` in the `:::demo` description, Or reverse it?

tolking:
[custom-validation](https://github.com/element-plus/element-plus/blob/dev/docs/examples/form/custom-validation.vue#L82) and [validation](https://github.com/element-plus/element-plus/blob/dev/docs/examples/form/validation.vue#L100)

I recommend use the `interface Form`

ModyQyW:
> [custom-validation](https://github.com/element-plus/element-plus/blob/dev/docs/examples/form/custom-validation.vue?rgh-link-date=2023-05-25T05%3A11%3A03Z#L82) and [validation](https://github.com/element-plus/element-plus/blob/dev/docs/examples/form/validation.vue?rgh-link-date=2023-05-25T05%3A11%3A03Z#L100)
> 
> I recommend use the `interface Form`

I just push commits about examples. If I misunderstand something, please tell me and I will correct them.

tolking (approved):

ModyQyW:
Just rebased.

docs/examples/form/custom-validation.vue
```diff
@@ -73,13 +73,18 @@ const validatePass2 = (rule: any, value: any, callback: any) => {
   }
 }
 
-const ruleForm = reactive({
+interface RuleForm {
+  pass: string
+  checkPass: string
+  age: string
+}
+const ruleForm = reactive<RuleForm>({
```
btea, line 81: 'RuleForm' can probably be removed.

docs/examples/form/custom-validation.vue
```diff
@@ -73,13 +73,18 @@ const validatePass2 = (rule: any, value: any, callback: any) => {
...
   pass: '',
   checkPass: '',
   age: '',
 })
 
-const rules = reactive<FormRules>({
+const rules = reactive<FormRules<RuleForm>>({
```
btea, line 87:
```suggestion
const rules = reactive<FormRules<typeof ruleForm>>({
```

ModyQyW:
@btea Hi. I just update the example. If I misunderstand something, please tell me and I will correct them. Thanks. 🙏

btea:
Your understanding is absolutely correct. 👍

btea (approved):
LGTM

PR #12706, opened 2023-05-08
fix(components): [select-v2] (#12684)

cobratyt:
fix(components): [select-v2] (#12684)


baiwusanyu-c:
You should write a unit test for the problem you fix

Issue #12773, opened 2023-05-11
[Component] [tree-select] 不能按照预期显示label


tl1994:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.3.1`
- Element Plus Version: `2.3.4`
- Browser / OS: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/113.0.0.0 Safari/537.36`
- Build Tool: `Vite`

### Related Component

- `el-tree-select`

### Reproduction Link

[...]

PR #12795, opened 2023-05-12
fix(components): [select-tree]

13725102796:
#12773

参考show-checkbox的处理方案，加上nextTick就正常了


tolking:
Thank for your contribution, but this issues has been fixed in #12826.

PR #12850, opened 2023-05-17
fix(locale): solve compilation halting due to the gen of extra hid files

joacyc:
to solve the issue of compilation halting due to the generation of extra hidden files in some development environments, such as Mac OS where .DS_Store files are generated without intervention, and so errors caused by these files in the "docs /.vitepress/build" directory during traverseDir method may lead to compilation halting.


tolking:

This project has ignored the `.DS_Store` file, and the `.DS_Store` file will not be submitted in the `.vitepress/rowdin` folder. Such modifications may not be necessary.

Issue #11795, opened 2023-03-03
[Component] [tabs] el-tabs的动态增减标签页缺少添加按钮的插槽


PeyFon:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.45`
- Element Plus Version: `2.2.32`
- Browser / OS: `版本 110.0.5481.78（正式版本） （64 位）`
- Build Tool: `Vue CLI`

### Related Component

- `el-tabs`

### Reproduction Link

[...]

PR #12970, opened 2023-05-29
feat(components): [tabs] add addIcon slot

btea:

### <samp>🤖 Generated by Copilot at f6cca1c</samp>

Added a new `addIcon` slot to the Tabs component and updated the documentation accordingly. This slot enables users to customize the appearance of the add button for editable tabs.

## Related Issue

feat #11795 .

### <samp>🤖 Generated by Copilot at f6cca1c</samp>

* Add a new slot `addIcon` to the Tabs component to allow customizing the icon of the add button for editable tabs ([link](https://github.com/element-plus/element-plus/pull/12970/files?diff=unified&w=0#diff-7ab8b00681ca241d9708f4540d16c56af5e98b40b57b6c38acaba203fddfd115L103-R106), [link](https://github.com/element-plus/element-plus/pull/12970/files?diff=unified&w=0#diff-50e3b1682badc3cce7c21a7fe3eca91302c86aefaf2592c1d743cd249a5333e5R189), [link](https://github.com/element-plus/element-plus/pull/12970/files?diff=unified&w=0#diff-50e3b1682badc3cce7c21a7fe3eca91302c86aefaf2592c1d743cd249a5333e5L199-R206))
* Render the `addIcon` slot in the `tabs.tsx` file if it is provided, or fall back to the default plus icon ([link](https://github.com/element-plus/element-plus/pull/12970/files?diff=unified&w=0#diff-50e3b1682badc3cce7c21a7fe3eca91302c86aefaf2592c1d743cd249a5333e5L199-R206))

### DIFF
--- a/docs/en-US/component/tabs.md
+++ b/docs/en-US/component/tabs.md
@@ -67,6 +67,14 @@ tabs/dynamic-tabs
 
 :::
 
+## Customized add button icon
+
+:::demo
+
+tabs/customized-add-button-icon
+
+:::
+
 ## Customized trigger button of new tab
 
 :::demo
--- a/docs/en-US/component/tabs.md
+++ b/docs/en-US/component/tabs.md
@@ -100,9 +100,10 @@ tabs/customized-trigger
 
 ## Tabs Slots
 
-| Name | Description               | Subtags  |
-| ---- | ------------------------- | -------- |
-| -    | customize default content | Tab-pane |
+| Name    | Description               | Subtags  |
+| ------- | ------------------------- | -------- |
+| -       | customize default content | Tab-pane |
+| addIcon | customize add button icon | -        |
 
 ## Tab-pane Attributes
 
--- a/docs/en-US/component/tabs.md
+++ b/docs/en-US/component/tabs.md
@@ -67,7 +67,7 @@ tabs/dynamic-tabs
 
 :::
 
-## Customized add button icon
+## Customized add button icon ^(2.3.15)
 
 :::demo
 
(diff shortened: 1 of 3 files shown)


ryuhangyeong (commented):
Personally, I think the function is useful, but I think it would be better to get a prop rather than a slot.

How about this?

btea:
> Personally, I think the function is useful, but I think it would be better to get a prop rather than a slot.
> 
> How about this?

Do you mean to add a prop similar to `actionIcon`, and then the default value is `Plus`? 

This seems to be possible, but in this case, if the user wants to customize the element structure, it seems to be more troublesome. 🤔

ryuhangyeong:
> > Personally, I think the function is useful, but I think it would be better to get a prop rather than a slot.
> > 
> > How about this?
> 
> Do you mean to add a prop similar to `actionIcon`, and then the default value is `Plus`? 
> 
> This seems to be possible, but in this case, if the user wants to customize the element structure, it seems to be more troublesome. 🤔

my opinion

From the name addicon, I already got the impression that only icons are available.

And since the area is not long, I think it is difficult to use it as a custom area.

I'd love to hear more people's opinions!

btea:
> From the name addicon, I already got the impression that only icons are available.

I just gave an example, maybe you can customize the text, such as `Add`. However, as you said, the area is not long, maybe it is better to only support icons. 🤔

ryuhangyeong (approved):

HeftyKoo (approved):

PR #13013, opened 2023-05-30
refactor: type narrowing

836334258:

## Description

type narrowing

### DIFF
--- a/packages/utils/objects.ts
+++ b/packages/utils/objects.ts
@@ -2,8 +2,10 @@ import { get, set } from 'lodash-unified'
 import type { Entries } from 'type-fest'
 import type { Arrayable } from '.'
 
-export const keysOf = <T>(arr: T) => Object.keys(arr) as Array<keyof T>
-export const entriesOf = <T>(arr: T) => Object.entries(arr) as Entries<T>
+export const keysOf = <T extends object>(arr: T) =>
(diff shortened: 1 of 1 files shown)


btea:
@ryuhangyeong Do you think this is necessary?

ryuhangyeong (commented):
@btea 

It looks like your existing code has type errors and needs to be changed.

But when I change the code like PR, typecheck throws an error.

What do you think?

image

image

836334258:
The reason for my modification is that the editor has a line prompt, so I view it to modify it

btea:
@ryuhangyeong The error reported in the first screenshot may be due to the ts version problem. vscode now defaults to version 5, and the project uses 4.7.4. Switching ts to the workspace version should not report an error.

btea:
@836334258 Can you give a more specific description of the prompt?

836334258:
image
There is this prompt on my computer

836334258:
image
When I write like this, running `pnpm typecheck` will not report an error

btea:
It seems to be the same problem as I said above, the ts version should be switched.

The error disappears after the ts version is switched to the workspace version

image

836334258:
I tried it, and there will be no error when switching the typescript version to the project version, thank you!

PR #13093, opened 2023-06-05
docs(components): [progress] 标记开始使用striped属性的版本

wuyadan:


btea:
@wuyadan As @wzc520pyfm  said, the `striped-flow` attribute should also add the same markup.

PR #13095, opened 2023-06-05
fix(components): [menu] ellipsis mode show correct menu items

chuyuandu:
# fix menu ellipsis mode

# when menu is ellipsis, one extra menu item was hidden

sliceIndex value is used in slice, it should be equal to the target menu index


chuyuandu:
wrong commit

PR #13110, opened 2023-06-06
docs(components): [progress] 标记striped和striped-flow的版本

wuyadan:


btea:
@wuyadan Can you change the commit message to English?

btea:
You don't need to close pr, just modify the commit information directly, and then submit it.

PR #13233, opened 2023-06-15
fix(components): [tag] fix bug #13229

yj-liuzepeng:
fix bug [#13229](https://github.com/element-plus/element-plus/issues/13229)

空span和有内容的span不对齐，通过 vertical-align: middle; 使之对齐

### DIFF
--- a/packages/components/tag/src/tag.vue
+++ b/packages/components/tag/src/tag.vue
@@ -2,7 +2,7 @@
   <span
     v-if="disableTransitions"
     :class="containerKls"
-    :style="{ backgroundColor: color }"
+    :style="{ backgroundColor: color, verticalAlign: 'middle' }"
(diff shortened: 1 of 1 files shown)


btea (commented):
Maybe it's better to modify it here
https://github.com/element-plus/element-plus/blob/e416826fae9c6d605a9625ae0afbbe92874c98e1/packages/theme-chalk/src/tag.scss#L94-L95

yj-liuzepeng:
> Maybe it's better to modify it here
> 
> https://github.com/element-plus/element-plus/blob/e416826fae9c6d605a9625ae0afbbe92874c98e1/packages/theme-chalk/src/tag.scss#L94-L95


great, thank for the suggestion, i resubmit the PR

Issue #13538, opened 2023-07-11
[Style] [row] ElRow align默认值 top，没有起作用


skyshirt:
Bug Type: **`Style`**

## Environment

- Vue Version: `3.3.4`
- Element Plus Version: `2.3.7`
- Browser / OS: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36`
- Build Tool: `Vite`

### Related Component

- `el-row`

### Reproduction Link

[...]

PR #13550, opened 2023-07-12
fix(components): [row] align="top" is not working properly

yj-liuzepeng:

## Description

the default value of align-items is 'stretch'

## Related Issue

Fixes #https://github.com/element-plus/element-plus/issues/13538.

### DIFF
--- a/packages/components/row/src/row.vue
+++ b/packages/components/row/src/row.vue
@@ -37,6 +37,6 @@ const style = computed(() => {
 const rowKls = computed(() => [
   ns.b(),
   ns.is(`justify-${props.justify}`, props.justify !== 'start'),
-  ns.is(`align-${props.align}`, props.align !== 'top'),
+  ns.is(`align-${props.align}`),
 ])
 </script>
--- a/packages/components/row/src/row.vue
+++ b/packages/components/row/src/row.vue
@@ -37,6 +37,6 @@ const style = computed(() => {
 const rowKls = computed(() => [
   ns.b(),
   ns.is(`justify-${props.justify}`, props.justify !== 'start'),
-  ns.is(`align-${props.align}`),
+  ns.is(`align-${props.align}`, !!props.align),
 ])
 </script>
--- a/packages/components/row/src/row.ts
+++ b/packages/components/row/src/row.ts
@@ -42,7 +42,6 @@ export const rowProps = buildProps({
   align: {
     type: String,
     values: RowAlign,
-    default: 'top',
   },
 } as const)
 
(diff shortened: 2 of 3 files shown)


packages/components/row/src/row.vue
```diff
@@ -37,6 +37,6 @@ const style = computed(() => {
 const rowKls = computed(() => [
   ns.b(),
   ns.is(`justify-${props.justify}`, props.justify !== 'start'),
-  ns.is(`align-${props.align}`, props.align !== 'top'),
+  ns.is(`align-${props.align}`),
```
chenxch, line 40: Good job, here I have a suggestion, the current change may cause the user's layout to change, my suggestion is to make the following changes:
first setp:
```
ns.is(`align-${props.align}`, props.align),
```
second setp:
Remove the default value of `props.align`.

`justify-start` is currently missing

chenxch:
/cc @YunYouJun Should we not set the default value of `align` like `element ui`?
image

YunYouJun:
Yes, I think we should not set a default value for it.

In addition with the popularity of tailwindcss and unocss today, I think this attribute is no longer needed. 🤣

packages/components/row/src/row.vue
yj-liuzepeng: Thanks, I changed the code，please review @chenxch

ryuhangyeong (commented):
## Personal Opinion

In my personal opinion, I think it's better to keep the current state because there is no difference in behavior before and after the code change. 🤔

(I think it will have an impact if the `align-items: flex-start;` property is included, which does not already exist.)

yj-liuzepeng:
> ## Personal Opinion
> In my personal opinion, I think it's better to keep the current state because there is no difference in behavior before and after the code change. 🤔
> 
> (I think it will have an impact if the `align-items: flex-start;` property is included, which does not already exist.)

When setting `align="top"`, the effect is different, the previous setting did not take effect. 
`align-items: flex-start;` has been added in row.scss, this property will take effect,  please take a look

ryuhangyeong (approved):

yj-liuzepeng:
> > ## Personal Opinion
> > In my personal opinion, I think it's better to keep the current state because there is no difference in behavior before and after the code change. 🤔
> > (I think it will have an impact if the `align-items: flex-start;` property is included, which does not already exist.)
> 
> When setting `align="top"`, the effect is different, the previous setting did not take effect. `align-items: flex-start;` has been added in row.scss, this property will take effect, please take a look

PTAL @chenxch

btea (approved):

btea:
> Is there anything that does not comply with the regulations, why is it not being merged ？🤔️ @btea

Don't worry, it's fine. I just thought it might take @chenxch  to confirm.

PR #13863, opened 2023-08-03
feat(components): [table] add tabindex to the scrollable table

Karolis-Stoncius:
When the table is scrollable it fails accessibility tests because its content is not focusable, this fixes that issue

### <samp>🤖 Generated by Copilot at e4f5a43</samp>

Added `tabindex` to `table.vue` to enable keyboard navigation. This improves the accessibility of the table component.

## Related Issue

Fixes #\_\_\_.

### <samp>🤖 Generated by Copilot at e4f5a43</samp>

* Make the table element focusable and accessible by keyboard users by adding the `tabindex` attribute ([link](https://github.com/element-plus/element-plus/pull/13863/files?diff=unified&w=0#diff-616bb754e750605914ca6b8e064991f1b1c91c581fd22ce883f77ee438f53c03R77))


Karolis-Stoncius:
@btea could you review this?

btea:
Is this the same problem that #14020 fixes?

Karolis-Stoncius:
> 

if it was `tabIndex: 0` instead of `tabIndex: -1` it would fix this issue

Karolis-Stoncius:
> Is this the same problem that #14020 fixes?

If it was `tabIndex: 0` instead of `tabIndex: -1` it would fix this issue

btea:
To be honest, I don't quite understand why we need to set `tabIndex: 0`? 🤔

Karolis-Stoncius:
> IMO, Focus the table may seem strange, as I haven't seen any relevant documentation before.
> 
> Accessibility of tables is achieved by using native tags or role attribute

At least for other content [here ](https://dequeuniversity.com/rules/axe/4.0/scrollable-region-focusable) it recommends to use `tabindex`. Maybe make it changeable through props so it wouldn't be hardcoded? And by default have it `undefined`

tolking:
Maybe we should make the scroll component focusable

Karolis-Stoncius:
> Maybe we should make the scroll component focusable

So basically this https://github.com/element-plus/element-plus/pull/14020 but with `tabIndex: 0`

packages/components/scrollbar/src/scrollbar.vue
```diff
@@ -6,6 +6,7 @@
         ref="resizeRef"
         :class="resizeKls"
         :style="viewStyle"
+        tabindex="0"
```
Karolis-Stoncius, line 9: @tolking like that?

tolking: IMO, moving `tabindex` to `wrapRef` may be more appropriate,  because we need to improve the `:focus-visible` style.

It should be better as an external attribute, and developers can cover it.

Karolis-Stoncius:
@tolking I added `tabindex` prop for `el-scrollbar` and `bodyTabindex` for `el-table`. I noticed you can't change the style of `el-scrollbar` wrap from the `el-table` component, should I add a `bodyWrapClass` prop for that?

PR #13932, opened 2023-08-10
feat(components): [select] add aria-label to select's input

Karolis-Stoncius:
Select's input causes accessibility issues without aria-label, this commit adds it

### <samp>🤖 Generated by Copilot at 104a237</samp>

Improve accessibility of `el-select` component by adding `aria-label` attributes and a new prop.

## Related Issue

Fixes #\_\_\_.

### <samp>🤖 Generated by Copilot at 104a237</samp>

*  Add `aria-label` attribute to `el-input` and `el-tag` components inside `el-select` component to improve accessibility for screen readers ([link](https://github.com/element-plus/element-plus/pull/13932/files?diff=unified&w=0#diff-9bfd95a8f0051f829d273a71e6ae89d2f5b583a6bdb2d808dd5b2e22709b2540R152), [link](https://github.com/element-plus/element-plus/pull/13932/files?diff=unified&w=0#diff-9bfd95a8f0051f829d273a71e6ae89d2f5b583a6bdb2d808dd5b2e22709b2540R195))
*  Add `ariaLabel` prop to `el-select` component to allow user to pass a custom label for the select component ([link](https://github.com/element-plus/element-plus/pull/13932/files?diff=unified&w=0#diff-9bfd95a8f0051f829d273a71e6ae89d2f5b583a6bdb2d808dd5b2e22709b2540R423))


packages/components/select/src/select.vue
```diff
@@ -418,6 +420,7 @@ export default defineComponent({
       values: placements,
       default: 'bottom-start',
     },
+    ariaLabel: String,
```
btea, line 423: Do we need to add a default value, for example, `select`? 🤔

Karolis-Stoncius: I don't think so. Aria attributes shouldn't be forced

btea (approved):

packages/components/select/src/select.vue
```diff
@@ -191,6 +192,7 @@
             :validate-event="false"
             :class="[nsSelect.is('focus', visible)]"
             :tabindex="multiple && filterable ? -1 : undefined"
+            :aria-label="ariaLabel"
```
tolking, line 195:
```suggestion
            :label="ariaLabel"
```

[ElInput label](https://element-plus.org/en-US/component/input.html#attributes)

At the same time, can you update the document?

Karolis-Stoncius: Simple html `input` doesn't have a `label` attribute, and if I change that - the accessibility issue comes back.

tolking: You can take a look at the ElInput component's documentation or source code

Karolis-Stoncius: if you're talking about [this](https://github.com/element-plus/element-plus/blob/dev/packages/components/input/src/input.vue), `input` has an `aria-label` attribute, not `label`

Karolis-Stoncius: Or are you asking me to just change `label` to `aria-label` in [this documentation](https://element-plus.org/en-US/component/input.html#attributes)? And change nothing in my code

tolking: Due to some historical reasons, the `aria-label` attribute is configured through the `label` attribute in the ElInput component. It cannot be modified in a short period of time. So I think we should use the `label` to insert parameters into the ElInput component here.

tolking: Add a new attribute should be displayed in the documentation

Karolis-Stoncius: [Fixed ](https://github.com/element-plus/element-plus/pull/13932/commits/bcdf7d6f373d112d8c34543d2e98a9eadb7a1f3a)

packages/components/select/src/select.vue
```diff
@@ -528,7 +528,13 @@ export default defineComponent({
       values: placements,
       default: 'bottom-start',
     },
-    ariaLabel: String,
+    /**
+     * @description native input aria-label
+     */
+    label: {
```
tolking, line 534:
```suggestion
    ariaLabel: {
```

Perhaps you misunderstood my meaning. It is more reasonable to use `ariaLabel` for the ElSelect component.

The `label` is an attribute of ElInput

Karolis-Stoncius: I just thought that `label` is used on ElInput so ElSelect should use it as well, for consistency. But I can change it back

tolking (approved):

PR #14011, opened 2023-08-19
refactor(components): [checkbox] set the label default value to `undefined`

### LINKED ISSUE #14009: [Component] [checkbox-group] 使用循环生成el-checkbox时，label为0时没能正常显示
Bug Type: **`Component`**

## Environment

- Vue Version: `3.3.4`
- Element Plus Version: `2.3.9`
- Browser / OS: `UserAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36 Edg/115.0.1901.203`
- Build Tool: `Vite`

### Related Component

- `el-checkbox-group`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJzcmMvQXBwLnZ1ZSI6IjxzY3JpcHQ+XG4gIGV4cG9ydCBkZWZhdWx0IHtcbiAgICBkYXRhKCkge1xuICAgICAgcmV0dXJuIHtcbiAgICAgICAgY2hlY2tib3hMaXN0OiBbXSxcbiAgICAgIH1cbiAgICB9XG4gIH1cbjwvc2NyaXB0PlxuPHRlbXBsYXRlPlxuICBcdDxlbC1jaGVja2JveC1ncm91cCB2LW1vZGVsPVwiY2hlY2tib3hMaXN0XCI+XG5cdFx0XHRcdFx0PGVsLWNoZWNrYm94IHYtZm9yPVwiaXRlbSBpbiAxMFwiIDprZXk9XCJpdGVtIC0gNVwiIDpsYWJlbD1cIml0ZW0gLSA1XCIgLz5cblx0XHQ8L2VsLWNoZWNrYm94LWdyb3VwPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHt9XG59IiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwiYWxsb3dKc1wiOiB0cnVlLFxuICAgIFwiY2hlY2tKc1wiOiB0cnVlLFxuICAgIFwianN4XCI6IFwicHJlc2VydmVcIixcbiAgICBcInRhcmdldFwiOiBcIkVTTmV4dFwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcInR5cGVzXCI6IFtcImVsZW1lbnQtcGx1cy9nbG9iYWwuZC50c1wiXVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIl9vIjp7fX0=)

### Steps to reproduce

```VUE
<script>
  export default {
    data() {
      return {
        checkboxList: [],
      }
    }
  }
</script>
<template>
  	<el-checkbox-group v-model="checkboxList">
					<el-checkbox v-for="item in 10" :key="item - 5" :label="item - 5" />
		</el-checkbox-group>
</template>
```

### What is Expected?

当值为0时，Label能正常显示

当值为0时，Label没显示

### Additional comments

_(empty)_

image

btea:

## Description
**BREAKING CHANGE**

before:
When the label value is set to `false`, the label is not displayed.

after:
When the label value is set to `false`, the label will display the false string.

### <samp>🤖 Generated by Copilot at 1531abe</samp>

Fix checkbox label bug and enhance null checking. Use `isNil` function from `lodash` to check for `null` or `undefined` values in `label` prop of `use-checkbox-status` composable.

## Related Issue

Fixes #14009.

### <samp>🤖 Generated by Copilot at 1531abe</samp>

* Import `isNil` function from `lodash-unified` to check for null or undefined values ([link](https://github.com/element-plus/element-plus/pull/14011/files?diff=unified&w=0#diff-5433b2d3d78b06b76e58a73b076b27d4b48b291acfa41f478dc5a41a83081764L2-R2))
* Fix bug that prevented checkbox from having label of `0` or `false` by using `isNil` instead of `!!` in `hasOwnLabel` computed property ([link](https://github.com/element-plus/element-plus/pull/14011/files?diff=unified&w=0#diff-5433b2d3d78b06b76e58a73b076b27d4b48b291acfa41f478dc5a41a83081764L44-R44))

### DIFF
--- a/packages/components/checkbox/src/composables/use-checkbox-status.ts
+++ b/packages/components/checkbox/src/composables/use-checkbox-status.ts
@@ -1,5 +1,5 @@
 import { computed, inject, ref, toRaw } from 'vue'
-import { isEqual } from 'lodash-unified'
+import { isEqual, isNil } from 'lodash-unified'
 import { useFormSize } from '@element-plus/components/form'
 import { isArray, isBoolean, isObject } from '@element-plus/utils'
 import { checkboxGroupContextKey } from '../constants'
@@ -41,7 +41,7 @@ export const useCheckboxStatus = (
   const checkboxSize = useFormSize(computed(() => checkboxGroup?.size?.value))
 
   const hasOwnLabel = computed<boolean>(() => {
-    return !!(slots.default || props.label)
+    return !!(slots.default || !isNil(props.label))
   })
 
   return {
--- a/packages/components/checkbox/src/composables/use-checkbox-status.ts
+++ b/packages/components/checkbox/src/composables/use-checkbox-status.ts
@@ -41,7 +41,7 @@ export const useCheckboxStatus = (
   const checkboxSize = useFormSize(computed(() => checkboxGroup?.size?.value))
 
   const hasOwnLabel = computed<boolean>(() => {
-    return !!(slots.default || !isNil(props.label))
+    return !!slots.default || (!isNil(props.label) && props.label !== false)
   })
 
   return {
--- a/packages/components/checkbox/src/composables/use-checkbox-status.ts
+++ b/packages/components/checkbox/src/composables/use-checkbox-status.ts
@@ -41,7 +41,7 @@ export const useCheckboxStatus = (
   const checkboxSize = useFormSize(computed(() => checkboxGroup?.size?.value))
 
   const hasOwnLabel = computed<boolean>(() => {
-    return !!slots.default || (!isNil(props.label) && props.label !== false)
+    return !!slots.default || !isNil(props.label)
   })
 
   return {
--- a/packages/components/checkbox/__tests__/checkbox.test.tsx
+++ b/packages/components/checkbox/__tests__/checkbox.test.tsx
@@ -21,6 +21,11 @@ describe('Checkbox', () => {
     expect(wrapper.classes('is-checked')).toBe(false)
   })
 
+  test('label set to number 0', async () => {
+    const wrapper = mount(() => <Checkbox label={0} />)
+    expect(wrapper.find('.el-checkbox__label').text()).toBe('0')
+  })
+
   describe('no v-model', () => {
     test('checkbox without label', async () => {
       const checked = ref(false)
--- a/packages/components/checkbox/src/checkbox.ts
+++ b/packages/components/checkbox/src/checkbox.ts
@@ -20,6 +20,7 @@ export const checkboxProps = {
    */
   label: {
     type: [String, Boolean, Number, Object],
+    default: undefined,
   },
   /**
    * @description Set indeterminate state, only responsible for style control


packages/components/checkbox/src/composables/use-checkbox-status.ts
```diff
@@ -41,7 +41,7 @@ export const useCheckboxStatus = (
   const checkboxSize = useFormSize(computed(() => checkboxGroup?.size?.value))
 
   const hasOwnLabel = computed<boolean>(() => {
-    return !!(slots.default || props.label)
+    return !!slots.default || (!isNil(props.label) && props.label !== false)
```
tolking, line 44:
```suggestion
    return !!slots.default || !isNil(props.label)
```

The same issues exists when the value is false

```vue
<el-checkbox :label="false" />
```

btea: According to the current logic, if the value of `label` is `false`, it should not be displayed.
Initially, I also modified it according to your suggestion, but this will cause the tree test to fail.
https://github.com/element-plus/element-plus/actions/runs/5908255483/job/16028234409

tolking:
```diff
label: {
    type: [String, Boolean, Number, Object],
+    default: undefined,
  },
```

It seems that the problem is caused by the default value of label being false rather than undefined

btea: Even if the default value is set to `undefined`, when the label value is set to `false`, the page will display the string false, which seems unreasonable.

tolking: `label=false` -> the label has a value -> `hasOwnLabel = true`

 show the false should be a reasonable behavior.

btea: I think it's better to be consistent with the previous behavior, otherwise, it will be a `breaking change`,

tolking: I think it can be modified during this major version update

btea: Ok.

tolking (approved):

packages/components/checkbox/src/composables/use-checkbox-status.ts
tolking: I think it's possible to modify the title of PR so that the default value has changed in the changelog

btea: Updated.

tolking: How about this?

refactor(components): [checkbox] set the label default value to undefined

PR #14071, opened 2023-08-24
fix(components): [select] guard value type error

sseinHu:
This is a problem, I found that selected is an object, and the object does not have the slice api

### <samp>🤖 Generated by Copilot at 5a4a917</samp>

Fixed a bug that caused empty tags to appear in `ElSelect` when `collapseTags` was true. Modified `showTagList` and `collapseTagList` in `useSelect.ts` to handle empty selections.

## Related Issue

Fixes #\_\_\_.

### <samp>🤖 Generated by Copilot at 5a4a917</samp>

* Fix a bug related to the `collapseTags` prop of the `ElSelect` component ([link](https://github.com/element-plus/element-plus/pull/14071/files?diff=unified&w=0#diff-ae55c3efc549b49d40b3f5b535ac2fc1572374d9309b6439f8d529b06edb1368L917-R921))

### DIFF
--- a/packages/components/select/src/useSelect.ts
+++ b/packages/components/select/src/useSelect.ts
@@ -914,11 +914,11 @@ export const useSelect = (props, states: States, ctx) => {
   )
 
   const showTagList = computed(() =>
-    states.selected.slice(0, props.maxCollapseTags)
+    states.selected.length ? states.selected.slice(0, props.maxCollapseTags) : []
   )
 
   const collapseTagList = computed(() =>
-    states.selected.slice(props.maxCollapseTags)
+    states.selected.length ? states.selected.slice(props.maxCollapseTags) : []
   )
 
   const navigateOptions = (direction) => {


sseinHu (commented):
这是一个问题，对象没有slice方法

sseinHu:
这是一个问题，对象没有slice方法

sseinHu:
This is a problem, objects do not have a slice method

sseinHu:
This bug has seriously affected my project, please update as soon as possible.

btea:
Can you provide an example of your usage scenario?

sseinHu:
> Can you provide an example of your usage scenario?
Snipaste_2023-08-24_19-01-19
Snipaste_2023-08-24_19-01-39
Snipaste_2023-08-24_19-02-23

Please take a look at these screenshots, states.selected is an object, use slice on the object, can not report an error?

btea:
According to the logic, the above code should only be triggered when multiple selections are made, but the value should be an array when multiple selections are made, so I am curious about your specific usage scenario.

Can you provide an example using [playground](https://element-plus.run/)?

sseinHu:
> According to the logic, the above code should only be triggered when multiple selections are made, but the value should be an array when multiple selections are made, so I am curious about your specific usage scenario.
> 
> Can you provide an example using [playground](https://element-plus.run/)?

My usage scenario is a single selection, not multiple selection, I suggest compatibility, judge the selected data type

sseinHu:
> Sure, compatibility would be better, but I think there may be a problem with your usage, so I would like to take a look at your specific usage.

image
Use is normal radio use, encapsulates a form component, and is configured using json data

sseinHu:
> Sure, compatibility would be better, but I think there may be a problem with your usage, so I would like to take a look at your specific usage.

Could you please repair the compatibility as soon as possible? The online project has been affected, thank you

btea:
There seems to be a lint error.

> Could you please repair the compatibility as soon as possible? The online project has been affected, thank you

Sorry, the process may not be that fast. If you are in a hurry, you can try `pnpm patch`.

sseinHu:
> There seems to be a lint error.
> 
> > Could you please repair the compatibility as soon as possible? The online project has been affected, thank you
> 
> Sorry, the process may not be that fast. If you are in a hurry, you can try `pnpm patch`.

pnpm patch can only be modified locally. Our online projects execute npm shell scripts in containers, so this method may not work. How long will it take you to fix this problem?

btea:
Maybe you should solve the problem of lint error first.

btea:
For your scene, I think you can also set an additional layer of judgment, judge whether to select more, and then render separately.
```vue
<template v-else-if="el.type === 'el-select'">
  <el-select v-if="el.multiple" :multiple="el.multiple">
    ...
  </el-select>
  <el-select v-else>
    ...
  </el-select>
</template>
```

sseinHu:
> ```
> multiple
> ```

I have just tried to write this, but after running the package, TypeError: l.selected.slice is not a function

btea:
> I have just tried to write this, but after running the package, TypeError: l.selected.slice is not a function

Do you mean according to what I said above, judge separately, render according to the condition or will it report an error?

sseinHu:
> > I have just tried to write this, but after running the package, TypeError: l.selected.slice is not a function
> 
> Do you mean according to what I said above, judge separately, render according to the condition or will it report an error?

Yes, there is no better way to solve this problem online than to fix the component library.

btea:
Can you provide your repro code logic? With screenshots, it's hard to tell where the problem is.

sseinHu:
> Can you provide your repro code logic? With screenshots, it's hard to tell where the problem is.

Sorry, I can't provide these, I can solve my problem by modifying your element-plus source code and adding type check compatibility, I think most people have this problem, which may be caused by your code is not strict.

btea:
This is the first time I see this error mentioned.
Since you have not resolved the lint error of this PR, there is no way to proceed further. The version has been released at noon, and new releases will not be made in a short time. Therefore, I want you to provide the reproduction code, and perhaps avoid errors by changing the way of use.

sseinHu:
> This is the first time I see this error mentioned. Since you have not resolved the lint error of this PR, there is no way to proceed further. The version has been released at noon, and new releases will not be made in a short time. Therefore, I want you to provide the reproduction code, and perhaps avoid errors by changing the way of use.

Given the bugs and workarounds, this is really the worst vue component library I've ever used

btea:
I'm sorry for the inconvenience caused.
However, to continue this pr process, you should fix this lint error first.

image

btea:
Ok, let's move our attention there, this pr should be closed.

sseinHu:
> Ok, let's move our attention there, this pr should be closed.

I see this PR: 14074 has been passed by lint

PR #14310, opened 2023-09-16
docs(descriptions): fix the `customed style` example

makedopamine:
before:
image

after:
image

### DIFF
--- a/docs/examples/descriptions/customized-style.vue
+++ b/docs/examples/descriptions/customized-style.vue
@@ -25,10 +25,10 @@
   </el-descriptions>
 </template>
 <style scoped>
-.my-label {
-  background: var(--el-color-success-light-9);
+:deep(.my-label) {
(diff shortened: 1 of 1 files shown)


docs/examples/descriptions/customized-style.vue
```diff
@@ -25,10 +25,10 @@
   </el-descriptions>
 </template>
 <style scoped>
-.my-label {
-  background: var(--el-color-success-light-9);
+:deep(.my-label ){
```
btea, line 28: Maybe you should remove the spaces between letters and brackets.

makedopamine: Yeah, I didn't enable the eslint plugin while I was fixing the example, which is causing the code style problem. I've force-pushed a new commit, please take a look.

btea (approved):
Thanks!

PR #14335, opened 2023-09-19
chore(MessageBox): unused keyframe

zhousg:


packages/theme-chalk/src/message-box.scss
```diff
@@ -244,15 +244,3 @@
...
-  0% {
-    transform: translate3d(0, 0, 0);
-    opacity: 1;
-  }
-
-  100% {
-    transform: translate3d(0, -20px, 0);
-    opacity: 0;
-  }
-}
```
btea, line 258: Perhaps it should not be removed, but applied in line 232.

zhousg: `animation: msgbox-fade-in getCssVar('transition-duration') reverse;`  

But use reverse,  have less code

btea: You are right!

btea (approved):

YunYouJun (approved):

YunYouJun:
Yeah, it's unused.

Just for more information:

We use `msgbox-fade-in reverse` instead of `msgbox-fade-out`.

https://github.com/element-plus/element-plus/blob/e2167dcd55edccc9eec792c40772a040d3f42592/packages/theme-chalk/src/message-box.scss#L232

PR #14382, opened 2023-09-24
feat(components): [slider] Added Tooltip display mode

Liberty-liu:

## Description
When I try to add functionality that always shows the Tooltip, I found that showTooltip cannot add the third situation.
```
 showTooltip: {
    type: Boolean,
    default: true,
  }
``` 
At the same time, I found that there are quite a lot of props for Tooltip, which are inconsistent with the props naming of Tooltip itself, so I added a new tooltip prop.

https://github.com/element-plus/element-plus/blob/d67d6d96dda33744c83d15521e6a89dc661c591b/packages/components/slider/src/composables/use-slider-button.ts#L71-L111
`initTooltipData` handles both old and new props

| Name                  | Description                                                                                              | Type            | Accepted Values                                                                                           | Default |
| --------------------- | -------------------------------------------------------------------------------------------------------- | --------------- | --------------------------------------------------------------------------------------------------------- | ------- |
| tooltip               | configuration options, see the following `SliderTooltipProps` table.                                     | undefined/object          | -                                                                                  | undefined    |

By default, the default value of tooltip is undefined. At this time, the old props is applied. If it is object, the tooltip props is applied.

The following are Tooltip parameters.

| Attribute      | Description                                                                                                                                                                                                                              | Type            | Accepted Values                 | Default |
| -------------- |------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------|-----------------------------------------------------------------------------------------------------------|---------|
| show  | When it is true, Tooltip will show always, when it is false, it will not be displayed even if dragging or hovering, when it is undefined, it will be displayed when moving in and dragging, and it will not be displayed when moving out | boolean/undefined        | true / false / undefined                                                                                                 |    undefined     |
| format  | Same as format-tooltip                                                                                                                                                                                                                   | function(value) | -                                                                                                         | -       |
| placement  | Same as placement                                                                                                                                                                                                                  | string          |      |
| popperClass  | Same as tooltip-class                                                                                                                                                                                                                  | string          | -                                                                                                        |         |

Both the old and new props have passed the test

## Related Issue


## Explanation of Changes


Liberty-liu:
It is hoped that the props passed to the tooltip in the future can use the props of the tooltip itself instead of giving a new name.

Liberty-liu:
I think this feat will be more common. If I think the API is inappropriate, I can correct it.

tolking:

always show the tooltip may obscure other content, which may not be a good idea

PR #14558, opened 2023-10-16
feat(components): [watermark] add font textAlign and textBaseline

IceyWu:
Description
feat(components): [watermark] add font textAlign and textBaseline
fix(components): [watermark] add export


tolking:
I noticed that some commits is not related to the current PR. Would you mind streamlining it?

IceyWu:
sure

PR #15013, opened 2023-11-30
feat(components): [switch] Add custom activeActionIcon and inactiveActionIcon slots

xing403:

## Description
Add custom activeActionIcon and inactiveActionIcon slots to switch component

## Related Issue

Fixes #\_\_\_.

### DIFF
--- a/packages/components/switch/src/switch.vue
+++ b/packages/components/switch/src/switch.vue
@@ -45,18 +45,16 @@
         <el-icon v-if="loading" :class="ns.is('loading')">
           <loading />
         </el-icon>
-        <template v-else-if="checked">
-          <el-icon v-if="activeActionIcon || $slots.activeActionIcon">
-            <component :is="activeActionIcon" v-if="activeActionIcon" />
-            <slot v-else name="activeActionIcon" />
+        <slot v-else-if="checked" name="activeActionIcon">
+          <el-icon v-if="activeActionIcon">
+            <component :is="activeActionIcon" />
           </el-icon>
-        </template>
-        <template v-else-if="!checked">
-          <el-icon v-if="inactiveActionIcon || $slots.inactiveActionIcon">
-            <component :is="inactiveActionIcon" v-if="inactiveActionIcon" />
-            <slot v-else name="inactiveActionIcon" />
+        </slot>
+        <slot v-else-if="!checked" name="inactiveActionIcon">
+          <el-icon v-if="inactiveActionIcon">
+            <component :is="inactiveActionIcon" />
           </el-icon>
-        </template>
+        </slot>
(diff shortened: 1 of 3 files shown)


btea:
It looks similar to this [feature](https://element-plus.org/zh-CN/component/switch.html#%E8%87%AA%E5%AE%9A%E4%B9%89%E5%8A%A8%E4%BD%9C%E5%9B%BE%E6%A0%87).

xing403:
This feature may be different, using slot can be more flexible to use 'icon' or other custom `icon`

packages/components/switch/src/switch.vue
```diff
@@ -45,12 +45,18 @@
...
+          <el-icon v-if="activeActionIcon || $slots.activeActionIcon">
+            <component :is="activeActionIcon" v-if="activeActionIcon" />
+            <slot v-else name="activeActionIcon" />
+          </el-icon>
+        </template>
+        <template v-else-if="!checked">
+          <el-icon v-if="inactiveActionIcon || $slots.inactiveActionIcon">
+            <component :is="inactiveActionIcon" v-if="inactiveActionIcon" />
+            <slot v-else name="inactiveActionIcon" />
+          </el-icon>
```
kooriookami, line 58: This looks weird, there're two icons here.

xing403: Here are two ways: one is to display ICONS using attributes, the other is to display ICONS using slots, and I have tested that the use of these two ways should not display two ICONS at the same time

packages/components/switch/src/switch.vue
```diff
@@ -45,12 +45,18 @@
...
+        </template>
```
kooriookami, line 59:
```suggestion
        <slot v-else-if="checked" name="activeActionIcon">
          <el-icon v-if="activeActionIcon">
            <component :is="activeActionIcon" />
          </el-icon>
        </slot>
        <slot v-else-if="!checked" name="inactiveActionIcon">
          <el-icon v-if="inactiveActionIcon">
            <component :is="inactiveActionIcon" />
          </el-icon>
        </slot>
```
Sometimes users want to add other things like words or images in slots. @btea What do you think?

kooriookami (changes requested):

packages/components/switch/src/switch.vue
xing403: The code has been modified as suggested

kooriookami:
@xing403 Please don't upload `pnpm-lock.yaml`, the test failed.

xing403:
I'm sorry. :flushed:

kooriookami:
The lock file is still exist.

PR #15101, opened 2023-12-11
chore: update pnpm to v8.12.0

inottn:
Since the release of pnpm8, it has become the version of choice for many actively maintained libraries, including Vue. However, when I switch to working on Element Plus from another project, I need to remember to downgrade to pnpm7, which I sometimes forget to do.
Why not upgrade to pnpm8? If there is a reason that I am unaware of, I'd love to hear it.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 9a3fa76</samp>

Updated `packageManager` to use latest pnpm version in `package.json`. This improves development speed and fixes some pnpm-related issues.

## Related Issue

Fixes #\_\_\_.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 9a3fa76</samp>

* Upgrade pnpm to the latest version for better development experience and bug fixes ([link](https://github.com/element-plus/element-plus/pull/15101/files?diff=unified&w=0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L3-R3))


inottn:
@kooriookami Could you please help review the changes of this PR sometime? 

In addition, Node 16 has reached EOL, and some parts of the build process are currently unable to run on higher versions of Node. If needed, I will work on this.

kooriookami:
Only update the pnpm will cause many problems. Did you test all package.json scripts?

inottn:
> Only update the pnpm will cause many problems. Did you test all package.json scripts?

If we are only upgrading the package manager version, it should be relatively smooth, as all the dependency versions remain unchanged. Although not all scripts have been tested, so far it appears that the CI has passed. However, there may be reasons that I am not aware of. If I have any misunderstandings, please feel free to correct me.

chenxch:
If it is just because the switching versions of personal projects are inconsistent, I think this is not a reason to upgrade. On the contrary, upgrading may also cause problems with the original normal switching. My suggestion is to maintain the current version before conducting full testing and evaluation, and postpone it until the upgrade to vite 5 is completed. WDYT?

sxzz:
I think someone can compare the original bundled files (under `dist`) and the files after this PR. If there's no obvious change, then we can accept this PR 

However, it's also necessary to upgrade Node to the LTS version, and fixes unit tests. It can be in another PR.

chenxch:
> I think someone can compare the original bundled files (under `dist`) and the files after this PR. If there's no obvious change, then we can accept this PR
> 
> However, it's also necessary to upgrade Node to the LTS version, and fixes unit tests. It can be in another PR.

Agreed, upgrading node also prepares for other subsequent upgrades.

inottn:
In my opinion, upgrades often become more challenging over time. As an open-source project, this can also potentially dampen contributors' enthusiasm and increase the cost of maintaining the project. Of course, it is important to ensure that existing functionality is not compromised. I have compared the bundled files before and after this PR, and there are no obvious changes. It would be even better if someone could confirm this.

sxzz (approved):

PR #15190, opened 2023-12-19
style(theme-chalk): replace word-break with overflow-wrap

### LINKED ISSUE #15167: [Style] [text] <el-text> component issue
Bug Type: **`Style`**

## Environment

- Vue Version: `3.3.11`
- Element Plus Version: `2.4.4`
- Browser / OS: `Brave`
- Build Tool: `Vite`

### Related Component

- `el-text`

### Reproduction Link

[Link]()

### Steps to reproduce

<el-text>abazazaaza<el-text>

### What is Expected?

the text is not correctly transferred to a new line

the text is not correctly transferred to a new line

### Additional comments

_(empty)_

kooriookami:
closed #15167

Before
image

Atfter
image

## Description

copilot:summary

## Related Issue

Fixes #\_\_\_.

## Explanation of Changes

copilot:walkthrough


packages/theme-chalk/src/date-picker/picker.scss
```diff
@@ -122,7 +122,7 @@
     padding: 0 5px;
     margin: 0;
     font-size: 14px;
-    word-break: keep-all;
+    white-space: nowrap;
```
chenxch, line 125: image

This results in a single line display.

kooriookami: I think it should be in a single line. Range separator won't be to long.

chenxch: Normally the content of this part will not be very long, but it is inevitable that some of it already has multiple characters. This will be a destructive style change and will squeeze the display of the time on the left and right sides. Most of the competing products are line-changed.

kooriookami: Try to use CJK characters, you'll find it also in a single line using word-break: keep-all.
image

ryuhangyeong (commented):
~~## Suggestion~~

~~This can be confusing, so keep the existing default values, but~~

~~How about allowing it to be overridden through CSS variables?~~

On second thought, I hid this suggestion because I thought it would make the situation more complicated.

packages/theme-chalk/src/date-picker/picker.scss
chenxch: I think this is a scenario that can be ignored. Only Chinese and Japanese have this situation of sentence segmentation without spaces. The words on the delimiter in these two languages should be short compared to other languages. The current change will cause the time display position to be occupied in the case of multiple words in foreign languages.

kooriookami: I updated the style.

chenxch (approved):

tolking (approved):

FrontEndDog (approved):

kooriookami:
Anyone clicks the merge button?

PR #15260, opened 2023-12-25
fix(style): [radio] marginRight 30px same as checkbox

taojunnan:

checkbox
radio


ryuhangyeong (commented):
Good Catch!

## Suggestion

clearly the `margin-right` of the radio and checkbox are different.

However, I think the `margin-right` of the radio should be changed.

Because, if you look at **element-ui**, `margin-right` is 30px.

@btea 

What do you think: 32px or 30px?

btea:
I'm not sure what the specific value should be set to. In fact, I looked through the design draft and it seems that there is no clear explanation of what the right margin value should be set to.

Maybe we need to check with the designer.

kooriookami (approved):

tolking (approved):
`margin-right: 30px;` inherits from element ui.

In the element ui, both Radio and Checkbox use the same style, but #5011 changed the style of Radio. I think it might be better to maintain consistent styling now.

PR #15266, opened 2023-12-26
fix(components): [link] add target attribute and relevant docs

yj-liuzepeng:

image

### DIFF
--- a/packages/components/link/src/link.ts
+++ b/packages/components/link/src/link.ts
@@ -1,5 +1,5 @@
 import { buildProps, iconPropType } from '@element-plus/utils'
-import type { ExtractPropTypes } from 'vue'
+import type { ExtractPropTypes, PropType } from 'vue'
 import type Link from './link.vue'
 
 export const linkProps = buildProps({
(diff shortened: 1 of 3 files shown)


ryuhangyeong (commented):
## Question

Is it for better type inference?

yj-liuzepeng:
> ## Question
> Is it for better type inference?

@ryuhangyeong  This is one reason. Additionally, i think that the 'target' attribute should not be used solely on 'el-link' without providing a definition or explanation

packages/components/link/src/link.ts
```diff
@@ -26,6 +26,14 @@ export const linkProps = buildProps({
...
    */
   href: { type: String, default: '' },
+  /**
+   * @description same as native hyperlink's `target`
+   */
+  target: {
+    type: String,
+    values: ['_blank', '_parent', '_self', '_top'],
+    default: '_self',
+  },
```
ryuhangyeong, line 36: ## Suggestion

There should be no `values` ​​attribute.

This is because the `iframe` name can come as the target value.

## Docs

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target

ryuhangyeong (approved):

yj-liuzepeng:
@btea  Please take a look

btea (approved):

PR #15313, opened 2023-12-29
feat(element-plus): Support last 2 versions of browsers

kooriookami:
We're upgrading the supported browser versions to `last 2 versions` in version `2.5.0`. If you want to support the old version browsers, please add Babel  and Polyfill yourself, or use element-plus in version `2.4.x`.

### DIFF
--- a/docs/en-US/guide/installation.md
+++ b/docs/en-US/guide/installation.md
@@ -5,16 +5,17 @@ lang: en-US
 
 # Installation
 
-## Compatibility
+## Compatibility ^(2.5.0)
 
-Element Plus can run on browsers that support [ES2018](https://caniuse.com/?feats=mdn-javascript_builtins_regexp_dotall,mdn-javascript_builtins_regexp_lookbehind_assertion,mdn-javascript_builtins_regexp_named_capture_groups,mdn-javascript_builtins_regexp_property_escapes,mdn-javascript_builtins_symbol_asynciterator,mdn-javascript_functions_method_definitions_async_generator_methods,mdn-javascript_grammar_template_literals_template_literal_revision,mdn-javascript_operators_destructuring_rest_in_objects,mdn-javascript_operators_spread_spread_in_destructuring,promise-finally) and [ResizeObserver](https://caniuse.com/resizeobserver).
+Element Plus can run on browsers that support last 2 versions.
+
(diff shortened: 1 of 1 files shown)


kooriookami:
@chenxch @btea @FrontEndDog What do you think?

FrontEndDog (approved):

chenxch (approved):

kooriookami:
> @kooriookami I see that this PR only modifies the document description. If we want to modify the corresponding version compatibility, do we need to modify the build configuration simultaneously? 🤔

This is a starting point, we can use the new features to write code later on. I'm refactoring select component now.

btea:
Ok, I see. But I still have a question. Currently, the chrome version we support is `64+`, and the latest chrome version is `120`. Is this span too big?

kooriookami:
> Ok, I see. But I still have a question. Currently, the chrome version we support is `64+`, and the latest chrome version is `120`. Is this span too big?

I referenced other component libraries and wrote it the same way. The latest two versions are not mandatory, just a suggestion. It may work fine with lower versions like `90+`.

btea:
Do we have specific standards to determine which browser versions should be supported?

kooriookami:
> Do we have specific standards to determine which browser versions should be supported?

Modern browsers in three years, at least support ES2021.

kooriookami:
@btea Can we merge this PR？It's related to the refactor of select component.

btea:
@kooriookami To be honest, I'm not sure. But if you want to make related adjustments in the future, I think it's okay. 👍

kooriookami:
OK, just do it. :D

TheNovemberRain:
The big reason I would use element plus over other component libraries is that it clearly defines the minimum supported version, and it's the version that the vast majority of computers are running. I think it would be very irresponsible to say "last two versions supported" as other libraries do. Developers don't know what features are not supported, and it's not clear if element plus will work as expected with the addition of Babel and Polyfill. I hope it would be more appropriate to raise the minimum supported version to a specific version number, like 99.

nicooprat:
Since Chrome releases major versions [every 4 weeks](https://support.google.com/chrome/a/answer/9027636?hl=en#:~:text=It's%20updated%20every%202%2D3,4%20weeks%20for%20major%20releases.) for instance, supporting only 2 last versions means 2 month of support, which is impossible to follow for the vast majority of businesses out there. I think supporting 2 **years** of browsers versions would be more serious.

PR #15351, opened 2024-01-04
fix(components): [input] remove useless contants

pzzyf:


pzzyf:
cc @chenxch  @kooriookami

packages/components/input/src/input.vue
```diff
@@ -42,7 +42,7 @@
           :placeholder="placeholder"
           :style="inputStyle"
           :form="props.form"
```
chenxch, line 44: It should be possible to remove it here too.

pzzyf: image
can not remove

chenxch: Okay, I feel that this variable name should be returned to the control, and an alias should be given here, but it doesn't matter, so let's do this for now.

chenxch (approved):

packages/components/input/src/input.vue
kooriookami: > image can not remove

You can do like this:  const { form: elForm, formItem: elFormItem } = useFormItem(), there're some similar code you can find in other compoents.

pzzyf:
@kooriookami

kooriookami (approved):

kooriookami:
LGTM

btea (approved):

PR #15367, opened 2024-01-05
When the type is password, it seems that the show-word-limit attribute is also required.

pzzyf:
before 
image
after
image


pzzyf:
cc @btea @chenxch @kooriookami

chenxch:
@pzzyf Scenarios where the password needs to display the length are still too rare, so it is recommended to use the suffix slot.

pzzyf:
> @pzzyf Scenarios where the password needs to display the length are still too rare, so it is recommended to use the suffix slot.

I think it is still very common to limit the length of passwords when registering an account, what do you think?

chenxch:
> > @pzzyf Scenarios where the password needs to display the length are still too rare, so it is recommended to use the suffix slot.
> 
> I think it is still very common to limit the length of passwords when registering an account, what do you think?

I have only seen requirements for a minimum length, not a maximum length, and even if there is, it will not be displayed.

pzzyf:
> @pzzyf Scenarios where the password needs to display the length are still too rare, so it is recommended to use the suffix slot.



> > > @pzzyf Scenarios where the password needs to display the length are still too rare, so it is recommended to use the suffix slot.
> > 
> > 
> > I think it is still very common to limit the length of passwords when registering an account, what do you think?
> 
> I have only seen requirements for a minimum length, not a maximum length, and even if there is, it will not be displayed.

You are right. When entering a password, limiting the maximum number of digits is equivalent to helping others narrow down the cracking scope. I closed this pr

PR #15478, opened 2024-01-12
perf(components): [select & selectV2] hide `collapse-tags-tooltip` when dropdown

xiaoxian521:
Be consistent with cascader `collapse-tags-tooltip` behavior

image

before:

https://github.com/element-plus/element-plus/assets/44761321/e7fa6dee-aa07-4e7d-98e5-5e37e913f80a

after:

https://github.com/element-plus/element-plus/assets/44761321/fc870561-0818-4cd5-b391-26292ab4a460


kooriookami:
Looks nice!

kooriookami (approved):

PR #15525, opened 2024-01-15
feat(components): [checkbox & radio] Use value instead of label act as value

### LINKED ISSUE #10365: [Component] [radio, radio-button] Why does el-radio use the label attribute instead of the value attribute to specify the value? el-radio 为何用 label 而不是 value 指定值呢？
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.41`
- Element Plus Version: `2.2.19`
- Browser / OS: `UserAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/106.0.0.0 Safari/537.36`
- Build Tool: `Vite`

### Related Component

- `el-radio`
- `el-radio-button`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cCBsYW5nPVwidHNcIj5cbmltcG9ydCB7IHJlZiwgdmVyc2lvbiBhcyB2dWVWZXJzaW9uIH0gZnJvbSAndnVlJ1xuaW1wb3J0IHsgdmVyc2lvbiBhcyBFcFZlcnNpb24gfSBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBFbGVtZW50UGx1cyB9IGZyb20gJ0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlJ1xuXG5jb25zdCBvcHRpb25zID0gW1xuICB7XG4gICAgbGFiZWw6ICfpgInpobkxJyxcbiAgICB2YWx1ZTogMFxuICB9LFxuICB7XG4gICAgbGFiZWw6ICfpgInpobkyJyxcbiAgICB2YWx1ZTogMVxuICB9XG5dXG5jb25zdCB2YWx1ZSA9IHJlZihudWxsKVxuXG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8aDM+5L2g5omA6YCJ5oup55qE5piv77yae3sgdmFsdWUgfX08L2gzPlxuICA8ZWwtcmFkaW8tZ3JvdXAgdi1tb2RlbD1cInZhbHVlXCI+XG4gIFx0PGVsLXJhZGlvXG4gICAgICB2LWZvcj1cInsgbGFiZWwsIHZhbHVlIH0gaW4gb3B0aW9uc1wiXG4gICAgICA6a2V5PVwidmFsdWVcIlxuICAgICAgOmxhYmVsPVwidmFsdWVcIj57e2xhYmVsfX08L2VsLXJhZGlvPlxuICA8L2VsLXJhZGlvLWdyb3VwPlxuXHQ8cD5cbiAgICBgbGFiZWxgIOWxnuaAp+S4gOiIrOS9nOS4uuKAnOagh+etvuaWh+acrOKAne+8jOWFtuS7lue7hOS7tuS5n+Wkp+WkmuS9nOS4uuaWh+acrOivtOaYjueahOeUqOmAlOOAguS9hiBgZWwtcmFkaW9gIOeahCBgbGFiZWxgIOWxnuaAp+aXouS9nOS4uiBsYWJlbO+8jOWPiOS9nOS4uiB2YWx1ZeOAglxuICA8L3A+XG4gIDxwPlxuICAgIOWvueS6juS7peS4iiBvcHRpb25zIOeahOWumuS5ie+8jOS8muWHuueOsOi/meenjeeOsOixoe+8mmBlbC1yYWRpbyg6bGFiZWw9XCJ2YWx1ZVwiKSB7IHsgbGFiZWwgfSB9YOOAglxuICA8L3A+XG4gIDxwPuaDs+S6huino+S4gOS4i+S4uuS9leS4jemAmui/hyB2YWx1ZSDlsZ7mgKfmjIflrpogcmFkaW8g55qE5YC877yf6L+Z5Lmf5pu06LS06L+RIEhUTUwg55qEICZsdDtpbnB1dCB0eXBlPVwicmFkaW9cIiBuYW1lPVwieHhOYW1lXCIgdmFsdWU9XCJ4eFZhbHVlXCImZ3Q7IOagh+etvuOAgjwvcD5cbiAgPHA+XG4gICAgPGVsLWljb24gY29sb3I9XCJ2YXIoLS1lbC1jb2xvci1wcmltYXJ5KVwiPjxFbGVtZW50UGx1cyAvPjwvZWwtaWNvbj5cbiAgICBFbGVtZW50IFBsdXMge3sgRXBWZXJzaW9uIH19ICsgVnVlIHt7IHZ1ZVZlcnNpb24gfX1cbiAgPC9wPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydF9tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHt9XG59IiwiX28iOnt9fQ==)

### Steps to reproduce

```vue
const options = [
  {
    label: '选项1',
    value: 0
  },
  {
    label: '选项2',
    value: 1
  }
]

<el-radio-group v-model="value">
  <el-radio
      v-for="{ label, value } in options"
      :key="value"
      :label="value">{{ label }}</el-radio>
</el-radio-group>
```

`label` 属性一般作为“标签文本”，其他组件也大多作为文本说明的用途。但 `el-radio` 的 `label` 属性既作为 label，又作为 value。

对于以上 options 的定义，会出现这种现象：`el-radio(:label="value") {{ label }}`。

想了解一下为何不通过 value 属性指定 radio 的值？这也更贴近 HTML 的 `<input type="radio" name="xxName" value="xxValue">` 标签。

### What is Expected?

通过 value 属性指定 el-radio 的值。

通过 label 属性指定 el-radio 的值。

### Additional comments

_(empty)_

kooriookami:
## Why Changed 为什么改变

Currently the Checkbox and Radio APIs are inconsistent with components such as Select, and using `label` act as `value` is a legacy issue that confuses users.

目前 Checkbox 与 Radio 的 API 与 Select 等组件不一致，使用 `label` 扮演 `value` 是一个历史遗留问题，会混淆用户。

## What Changed 改变了什么

Props in Checkbox and Radio that using `label` act as `value` are replaced with `value`.

Checkbox 与 Radio 中原来使用 `label` 扮演 `value` 的参数替换为 `value`。

## Any Break Changes 是否有破坏性

The current code is compatible with using `label` act as `label` and `value`, as well as using `label` and `value` separately. These two usages will be retained until version `3.0.0`. Similarly, `true-label` and `false-label` are replaced with `true-value` and `false-value`.

目前的代码兼容了使用 `label` 同时扮演  `label` 与 `value` 的用法，以及分别使用 `label` 与 `value` 的用法。这两个用法会保留至 `3.0.0` 版本。相同地，`true-label` 与 `false-label` 替换为 `true-value` 与 `false-value`。

Closes #10365


docs/en-US/component/radio.md
```diff
@@ -59,14 +67,15 @@ radio/with-borders
...
-| disabled              | whether Radio is disabled            | ^[boolean]                               | false   |
-| border                | whether to add a border around Radio | ^[boolean]                               | false   |
-| size                  | size of the Radio                    | ^[enum]`'large' \| 'default' \| 'small'` | —       |
-| name                  | native `name` attribute              | ^[string]                                | —       |
+| Name                  | Description                                                            | Type                                     | Default |
+|-----------------------|------------------------------------------------------------------------|------------------------------------------|---------|
+| model-value / v-model | binding value                                                          | ^[string] / ^[number] / ^[boolean]       | —       |
+| value ^(2.5.2)        | the value of Radio                                                     | ^[string] / ^[number] / ^[boolean]       | —       |
```
jw-foss, line 73: **Suggestion:** as we are adding new APIs, we might want to make it a minor patch as `^2.6.0`, adding new API is considered breaking.

docs/en-US/component/radio.md
```diff
@@ -112,11 +122,12 @@ radio/with-borders
...
 
-| Name     | Description               | Type                               | Default |
-| -------- | ------------------------- | ---------------------------------- | ------- |
-| label    | the value of Radio        | ^[string] / ^[number] / ^[boolean] | —       |
-| disabled | whether Radio is disabled | ^[boolean]                         | false   |
-| name     | native 'name' attribute   | ^[string]                          | —       |
+| Name           | Description                                                            | Type                               | Default |
+|----------------|------------------------------------------------------------------------|------------------------------------|---------|
+| value ^(2.5.2) | the value of Radio                                                     | ^[string] / ^[number] / ^[boolean] | —       |
+| label          | the label of Radio. If there's no `value`, `label` will act as `value` | ^[string] / ^[number] / ^[boolean] | —       |
```
jw-foss, line 128: **Praise:** 👍 good compatible patch.

packages/components/radio/src/radio-button.vue
```diff
@@ -2,7 +2,7 @@
   <label
     :class="[
       ns.b('button'),
-      ns.is('active', modelValue === label),
+      ns.is('active', modelValue === getValueOrLabel(props)),
```
jw-foss, line 5: **Suggestion:** can we make this a computed value?

Something like the following.

```ts
const actualValue = computed(() => getValueOrLable(props));
```

packages/components/radio/src/utils.ts
```diff
@@ -0,0 +1,12 @@
...
+import type { RadioButtonProps } from './radio-button'
+
+export const getValueOrLabel = (props: RadioProps | RadioButtonProps) => {
+  // In version 2.x, if there's no props.value, props.label will act as props.value
+  // In version 3.x, remove this computed value, use props.value instead.
+  if (!isNil(props.value)) {
+    return props.value
+  }
+  return props.label
+}
```
jw-foss, line 12: **Suggestion:**

Maybe we should avoid redundant code by extracting this into `element-plus/utils` instead of repeating this for every component.

jw-foss: I'd also suggest that we introduce a deprecation `warning` for this API. 

```ts
useDeprecated('value', 'using label as the value will be deprecated in the next minor/major version, please consider giving explicit values for the all the components.')

```

jw-foss (commented):
@kooriookami left couple suggestions for your consideration let me know if you have any questions.

packages/components/radio/src/radio-button.vue
kooriookami: Yes, I tried to make a computed value in useRadio, but some files don't use useRadio. Or should I make `actualValue` in each files?

packages/components/radio/src/utils.ts
kooriookami: Before, I think the getValueOrLabel method is a business method. But it actually the same in Checkbox and Radio. It can be extracted into a public method.

The `value` API is newly added, and the `label` API has it's own function. So useDeprecated is not a good way.

jw-foss: I meant that we need to inform the users that we have introduced a new way to passing values, to encourage them from migrating the code, we need to tell them that the new API is valid, and label will no longer be used `value` in next version. Maybe you can start by introducing a new hook for API renaming or something, to reduce ambiguity, this step is necessary.

kooriookami: OK, I have a idea.

packages/components/checkbox/src/composables/use-checkbox.ts
```diff
@@ -50,7 +39,51 @@ export const useCheckbox = (
...
+
+  useDeprecated(
+    {
+      from: 'label act as value',
+      replacement: 'value',
+      version: '3.0.0',
+      scope: 'el-checkbox',
+      ref: 'https://element-plus.org/en-US/component/checkbox.html',
+    },
+    computed(() => isGroup.value && isNil(props.value))
```
btea, line 63: Maybe we should just use the encapsulated method.
```suggestion
    computed(() => isGroup.value && isPropAbsent(props.value))
```
https://github.com/element-plus/element-plus/blob/4fa54dde2dd99a087e6a0cd579cc137ea581ae53/packages/utils/types.ts#L29-L31

kooriookami: Yes, I'll change it.

btea (approved):

PR #15540, opened 2024-01-16
feat(components): [select & select-v2] Add loading slot

kooriookami:

### DIFF
--- a/packages/components/select-v2/src/select.vue
+++ b/packages/components/select-v2/src/select.vue
@@ -226,12 +226,17 @@
           <template #default="scope">
             <slot v-bind="scope" />
           </template>
-          <template #empty>
-            <slot name="empty">
-              <p :class="nsSelect.be('dropdown', 'empty')">
-                {{ emptyText ? emptyText : '' }}
-              </p>
-            </slot>
+          <template v-if="$slots.loading && loading" #loading>
+            <div :class="nsSelect.be('dropdown', 'loading')">
+              <slot name="loading" />
+            </div>
+          </template>
+          <template v-else-if="loading || filteredOptions.length === 0" #empty>
+            <div :class="nsSelect.be('dropdown', 'empty')">
+              <slot name="empty">
+                <span>{{ emptyText }}</span>
+              </slot>
+            </div>
           </template>
           <template v-if="$slots.footer" #footer>
             <div :class="nsSelect.be('dropdown', 'footer')">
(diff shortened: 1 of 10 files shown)


packages/components/select-v2/src/select.vue
```diff
@@ -226,12 +226,17 @@
           <template #default="scope">
             <slot v-bind="scope" />
           </template>
-          <template #empty>
-            <slot name="empty">
-              <p :class="nsSelect.be('dropdown', 'empty')">
-                {{ emptyText ? emptyText : '' }}
-              </p>
-            </slot>
+          <template v-if="$slots.loading && loading" #loading>
```
xiaoxian521, line 229: IIRC. `#loading` here is actually redundant because the following `<slot name="loading" />` already exists, so it is more appropriate to replace it with the following code. WYT 🤔
```js
  <div v-if="$slots.loading && loading" :class="nsSelect.be('dropdown', 'loading')">
      <slot name="loading" />
   </div>
```

packages/components/select-v2/src/select.vue
```diff
@@ -226,12 +226,17 @@
...
+            <div :class="nsSelect.be('dropdown', 'loading')">
+              <slot name="loading" />
+            </div>
+          </template>
+          <template v-else-if="loading || filteredOptions.length === 0" #empty>
```
xiaoxian521, line 234: ditto

xiaoxian521 (changes requested):

packages/components/select-v2/src/select.vue
kooriookami: Well, the `<template #loading />` is for `el-select-menu`, and the `<slot name="loading" />` is for `el-select-v2`. It can not be removed.

xiaoxian521 (approved):

FrontEndDog (changes requested):
1.Before `loading` appears, `No Data` will appear, I think the `dropdownMenu` should appear after input debounce. 
2.I found a bug, Input 1111111, The front 1 will be covered

https://github.com/element-plus/element-plus/assets/46493087/088addeb-35e3-410a-b7c6-a5d799b3a2e5

kooriookami:
> 1.Before `loading` appears, `No Data` will appear, I think the `dropdownMenu` should appear after input debounce. 2.I found a bug, Input 1111111, The front 1 will be covered
> 
>  2024-01-17.10.25.04.mov

1. Sounds good, maybe ask other people
2. Fixed by #15539

kooriookami:
@FrontEndDog I added the dropdowm menu shows after loading. Look look.

FrontEndDog (approved):
LGTM

PR #15579, opened 2024-01-18
fix(components): [carousel] improve a11y

cloydlau:

- [✓] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).
- [✓] Make sure you are merging your commits to `dev` branch.
- [✓] Add some descriptions and refer to relative issues for your PR.

When running the Lighthouse of Chrome, it throws 'Buttons do not have an accessible name':

image


packages/components/carousel/src/carousel.vue
```diff
@@ -55,7 +57,7 @@
         @mouseenter="throttledIndicatorHover(index)"
         @click.stop="handleIndicatorClick(index)"
       >
-        <button :class="ns.e('button')">
+        <button :class="ns.e('button')" aria-label="Carousel Button">
```
btea, line 60: It may be clearer to change the label content to `carousel indicator`.

cloydlau: Done

btea (approved):
LGTM

packages/components/carousel/src/carousel.vue
```diff
@@ -13,6 +13,7 @@
           "
           type="button"
           :class="[ns.e('arrow'), ns.em('arrow', 'left')]"
+          aria-label="Carousel Arrow Left"
```
tolking, line 16: IMO, This writing style is not conducive to internationalization, as screen readers can only prompt by English. 

Perhaps adding it to `locale` folder would be better. refer #11817

packages/components/carousel/src/carousel.vue
```diff
@@ -55,7 +57,7 @@
         @mouseenter="throttledIndicatorHover(index)"
         @click.stop="handleIndicatorClick(index)"
       >
-        <button :class="ns.e('button')">
+        <button :class="ns.e('button')" aria-label="Carousel Indicator">
```
tolking, line 60: It would be better if you could give a hint as to which carousel to switch to.

packages/components/carousel/src/carousel.vue
cloydlau: Done. Not entirely certain about the naming though.

packages/components/carousel/src/carousel.vue
cloydlau: Done. Not sure whether the index should start from 0 or 1 here.

tolking (approved):

PR #15727, opened 2024-01-31
feat(components): [check-tag] add type prop

kooriookami:

### DIFF
--- a/packages/components/check-tag/src/check-tag.vue
+++ b/packages/components/check-tag/src/check-tag.vue
@@ -17,7 +17,11 @@ const props = defineProps(checkTagProps)
 const emit = defineEmits(checkTagEmits)
 
 const ns = useNamespace('check-tag')
-const containerKls = computed(() => [ns.b(), ns.is('checked', props.checked)])
+const containerKls = computed(() => [
+  ns.b(),
+  ns.is('checked', props.checked),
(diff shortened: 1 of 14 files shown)


packages/components/check-tag/src/check-tag.vue
```diff
@@ -17,7 +17,11 @@ const props = defineProps(checkTagProps)
 const emit = defineEmits(checkTagEmits)
 
 const ns = useNamespace('check-tag')
-const containerKls = computed(() => [ns.b(), ns.is('checked', props.checked)])
+const containerKls = computed(() => [
+  ns.b(),
+  ns.is('checked', props.checked),
+  ns.m(props.type || 'primary'),
```
btea, line 23: `props.type` has a default value, maybe we don't need to set the default value repeatedly?

kooriookami: It's for setting empty string to `type`.

btea: `type` does not seem to allow accepting empty strings.

kooriookami: Yes, It's an insurance operation. If someone set `null`, we need to use the default value.

btea (approved):

btea (approved):

PR #15823, opened 2024-02-05
fix: threshold comparison in scripts/build-table.ts

slothvips:
fix ts(2365) error


scripts/build-table.ts
```diff
@@ -36,7 +36,7 @@ async function main() {
     output = `**Total changed files:** ${fileDiffs.length}
 
 ${
-  fileDiffs.length >= threshold
+  fileDiffs.length >= +threshold
```
btea, line 39: There should be implicit conversion here, so there is no need to force type conversion.

packages/utils/dom/aria.ts
```diff
@@ -121,6 +121,6 @@ export const getSibling = (
 
 export const focusNode = (el: HTMLElement) => {
   if (!el) return
-  el.focus()
+  el.focus?.()
```
btea, line 124: The optional chain here also seems unnecessary.

PR #15825, opened 2024-02-05
fix: focusNode function to handle cases where el.focus is undefined

slothvips:
fix: focusNode function to handle cases where el.focus is undefined


btea:
@pandavips Hello, thank you for your contribution! However, there seems to be a problem with your commit record. Maybe you can create a branch from the latest commit record of the dev branch.

Regarding the problem you mentioned, can you provide a reproduction link?

btea:
It seems duplicate of #15823

PR #15866, opened 2024-02-18
fix(components): [cascader-panel] only render valid render label

### LINKED ISSUE #5428: [Bug Report] Using the dynamic component reference el-cascader, the label for all options are blank
### Element Plus version
1.3.0-beta.5

### OS/Browsers version
Chrome 95.0.4638.54

### Vue version
3.2.26

### Reproduction Link
https://codepen.io/cjq02/pen/JjrzJOJ

### Steps to reproduce
click el-cascader input, then you can see the result

### What is Expected?
Displays the names of all options

the label for all options are blank
image

### LINKED ISSUE #10537: [Component] [cascader] Cascader label disappear when I use jsx
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.41`
- Element Plus Version: `2.2.20`
- Browser / OS: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.127 Safari/537.36`
- Build Tool: `Vite`

### Related Component

- `el-cascader`

### Reproduction Link

[CodeSandbox](https://codesandbox.io/embed/heuristic-jasper-qo01q7?fontsize=14&hidenavigation=1&theme=dark)

### Steps to reproduce

```vue
 setup() {
    const options = [
      {
        label: "label",
        value: "value"
      }
    ];
    const childen = null; //will be parsed into ： ()=>[null]
    const cascader = <ElCascader options={options}>{childen}</ElCascader>;
    return { cascader };
  }
```
cascader label doesn't work.

### What is Expected?

When the value of children is null||undefined, it is not properly done to be a slot.

Whatever the value of children, ElCascader treats them as a slot.Then the label doesn't show it.

### Additional comments

I think this is a parsing deficiency for a vue slot.This is different from when using h.

Hfutsora:
only render valid content

closed #5428
closed #10537
related https://github.com/vuejs/core/pull/2485


xiangshu233:
不地道

btea:
Can you add test case ?

PR #15922, opened 2024-02-21
chore(components): clear deprecated API

kooriookami:
This PR will remove the deprecated API before `2.6.0` (include).

- [x] Wait for next minor version.

### DIFF
--- a/docs/en-US/component/menu.md
+++ b/docs/en-US/component/menu.md
@@ -95,7 +95,7 @@ menu/popper-offset
 ## Menu Events
 
 | Name   | Description                               | Parameters                                                                                                                                                                 |
-| ------ | ----------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+|--------|-------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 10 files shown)


btea:
The corresponding documentation doesn't seem to be updated.

I have another question. After the update is released, users seem to be unable to see older versions of the documentation. So, should we provide detailed descriptions of the discarded slots and props, such as the specific abandoned versions and the replacement writing methods after discarding, etc.

kooriookami:
> The corresponding documentation doesn't seem to be updated.
> 
> I have another question. After the update is released, users seem to be unable to see older versions of the documentation. So, should we provide detailed descriptions of the discarded slots and props, such as the specific abandoned versions and the replacement writing methods after discarding, etc.

There're deprecated tags and description in the doc. They have been deprecated for many versions since added.

image

btea:
Well, what I mean is that the corresponding prompt has been abandoned in the code, so should the corresponding attribute be removed directly from the document?

kooriookami:
> Well, what I mean is that the corresponding prompt has been abandoned in the code, so should the corresponding attribute be removed directly from the document?

Well, I think keep deprecated api in doc is better. We can remove them in `3.0.0`.

btea:
`sub-menu` the deprecated api format is incorrect.
image

btea (approved):

PR #15944, opened 2024-02-24
feat(components): [badge] add show-zero prop

IceyWu:
Add show-zero prop,whether to show badge when valueis zero

### DIFF
--- a/docs/en-US/component/badge.md
+++ b/docs/en-US/component/badge.md
@@ -58,6 +58,7 @@ badge/dot
 | is-dot | if a little dot is displayed.                                                 | ^[boolean]                                                         | false   |
 | hidden | hidden badge.                                                                 | ^[boolean]                                                         | false   |
 | type   | badge type.                                                                   | ^[enum]`'primary' \| 'success' \| 'warning' \| 'danger' \| 'info'` | danger  |
+| show-zero | Whether to show badge when value is zero.                                  | ^[boolean]                                                         | true    |
 
 ### Slots
 
--- a/docs/en-US/component/badge.md
+++ b/docs/en-US/component/badge.md
@@ -58,7 +58,7 @@ badge/dot
 | is-dot | if a little dot is displayed.                                                 | ^[boolean]                                                         | false   |
 | hidden | hidden badge.                                                                 | ^[boolean]                                                         | false   |
 | type   | badge type.                                                                   | ^[enum]`'primary' \| 'success' \| 'warning' \| 'danger' \| 'info'` | danger  |
-| show-zero | Whether to show badge when value is zero.                                  | ^[boolean]                                                         | true    |
+| show-zero  ^(2.6.0) | Whether to show badge when value is zero.                                  | ^[boolean]                                                         | true    |
 
 ### Slots
 
--- a/packages/components/badge/__tests__/badge.test.tsx
+++ b/packages/components/badge/__tests__/badge.test.tsx
@@ -76,7 +76,7 @@ describe('Badge', () => {
     const wrapper2 = mount(() => (
       <Badge showZero={false} value={badgeValue2.value} />
     ))
-    expect(wrapper2.find('.el-badge__content').text()).toEqual('0')
+    expect(wrapper2.find('.el-badge__content').text()).toEqual('')
     badgeValue2.value = 80
     await nextTick()
     expect(wrapper2.find('.el-badge__content').text()).toEqual('80')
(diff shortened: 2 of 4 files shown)


packages/components/badge/src/badge.vue
```diff
@@ -3,7 +3,7 @@
     <slot />
     <transition :name="`${ns.namespace.value}-zoom-in-center`">
       <sup
-        v-show="!hidden && (content || isDot)"
+        v-show="isShowBadge"
```
kooriookami, line 6:
```suggestion
        v-show="!hidden && (content || isDot)"
```
Please keep it, I will explain later.

packages/components/badge/src/badge.vue
```diff
@@ -39,6 +39,11 @@ const content = computed<string>(() => {
   return `${props.value}`
 })
 
+const isShowBadge = computed(() => {
+  const isShowVal = +content.value === 0 ? props.showZero : content.value
+  return !props.hidden && (isShowVal || props.isDot)
+})
```
kooriookami, line 45:
```suggestion
```
You can't tell the `content.value` is a string or number type, please move the judgment to the computed `content`.

```
const content = computed<string>(() => {
  if (props.isDot) return ''

  if (props.value === 0 && !props.showZero) return ''

  if (isNumber(props.value) && isNumber(props.max)) {
    return props.max < props.value ? `${props.max}+` : `${props.value}`
  }
  return `${props.value}`
})
```

kooriookami (changes requested):

docs/en-US/component/badge.md
```diff
@@ -58,6 +58,7 @@ badge/dot
 | is-dot | if a little dot is displayed.                                                 | ^[boolean]                                                         | false   |
 | hidden | hidden badge.                                                                 | ^[boolean]                                                         | false   |
 | type   | badge type.                                                                   | ^[enum]`'primary' \| 'success' \| 'warning' \| 'danger' \| 'info'` | danger  |
+| show-zero | Whether to show badge when value is zero.                                  | ^[boolean]                                                         | true    |
```
kooriookami, line 61:
```suggestion
| show-zero  ^(2.6.0) | Whether to show badge when value is zero.                                  | ^[boolean]                                                         | true    |
```
Please add a version tag.

kooriookami (changes requested):

docs/en-US/component/badge.md
IceyWu: ok

kooriookami (approved):

btea (changes requested):
@IceyWu Could you add a test case ?

IceyWu:
> @IceyWu Could you add a test case ?

OK

packages/components/badge/__tests__/badge.test.tsx
```diff
@@ -62,4 +62,15 @@ describe('Badge', () => {
...
+
+  test('showZero', async () => {
+    const badgeValue = ref(0)
+    const wrapper = mount(() => (
+      <Badge showZero={true} value={badgeValue.value} />
+    ))
+    expect(wrapper.find('.el-badge__content').text()).toEqual('0')
+    badgeValue.value = 80
+    await nextTick()
+    expect(wrapper.find('.el-badge__content').text()).toEqual('80')
```
kooriookami, line 74: When changing `showZero` to false, it's necessary to test the value `0`.

kooriookami (changes requested):

packages/components/badge/__tests__/badge.test.tsx
IceyWu: Should I add the situation where showZero is false？

kooriookami: Yes, set `const showZero= ref(true);` and change the vaule.

kooriookami (approved):

packages/components/badge/src/badge.vue
```diff
@@ -33,6 +33,8 @@ const ns = useNamespace('badge')
 const content = computed<string>(() => {
   if (props.isDot) return ''
 
+  if (props.value === 0 && !props.showZero) return ''
```
btea, line 36: The judgment here should be placed on line 39.
```ts
if (props.max < props.value) {
  return `${props.max}+`
}
return props.value === 0 && !props.showZero ? '' : `${props.value}`
```
For example, when `props.value` is 0, `showZero` is true, but `props.max` is `-1`, the badge should be displayed.

IceyWu: Like this? 
```suggestion
 const content = computed<string>(() => {
  if (props.isDot) return ''
  if (isNumber(props.value) && isNumber(props.max)) {
    if (props.max < props.value) {
      return `${props.max}+`
    }
    return props.value === 0 && !props.showZero ? '' : `${props.value}`
  }
  return `${props.value}`
})
```

IceyWu: Ok

btea (approved):

PR #15952, opened 2024-02-25
fix(components): [table] fix hydration issue in table for ssr apps

debasispanigrahi:
Added tr tag in thead for matching client and server in ssr app.


btea:
Hello, thank you for your contribution! Can you explain why this modification would solve the hydration issue?

PR #16106, opened 2024-03-12
fix(components): [tree] remove duplicate type declaration

inside5545:
there is a duplicated type declaration in <el-tree> TreeOptionProps.class


btea (approved):

PR #16111, opened 2024-03-12
build: use vue-tsc to generate dts

makedopamine:
relates to #12770

There are some huge type declaration files which slow intellisense in element+ package, like `components/cascader/src
/cascader.vue.d.ts` , it has [13462](https://github.com/element-plus/element-plus/assets/93767616/e63d4022-d304-46dd-baf3-b5423bfa0665) lines.

To improve it, this PR does the following:
- **Use `vue-tsc` to generate dts.** Vue-tsc can generate more compact type declaration for SFC, which reduces the time of top time-consuming function  from [1100ms](https://github.com/element-plus/element-plus/assets/93767616/91ddd066-aef9-4704-b75f-55a4460fbb5a) to about [270ms](https://github.com/element-plus/element-plus/assets/93767616/560b2e40-2594-42f1-82b5-db223a283594) on my PC. For `components/cascader/src/cascader.vue.d.ts`, [557](https://github.com/element-plus/element-plus/assets/93767616/f63286f1-ef87-4206-a46e-2a4b85b24006) lines left. 
- **Reduce duplicated type declaration in dts.**  For example, `components/cascader/index.ts` imports `components/cascader/src/index.vue`, however, its dts(`components/cascader/index.d.ts`) inlines the type declaration of `components/cascader/src/index.vue` instead of importing the type declaration. To avoid inlining the type declaration of components, I add explicit type annotation for them. In the result, the time of top time-consuming function is reduced to about [150ms](https://github.com/element-plus/element-plus/assets/93767616/de2a5716-eb73-4517-a90e-503d3c114118). 

Additional:  [a related issue about inline type declaration](https://github.com/microsoft/TypeScript/issues/37151)


btea:
It looks like there is a conflict in the code.

Cheerwhy:
anything news 😳

makedopamine:
The conflicts have been resolved.

package.json
```diff
@@ -118,7 +118,7 @@
     "vitest": "1.6.0",
     "vue": "^3.2.37",
     "vue-router": "^4.0.16",
-    "vue-tsc": "^0.38.2"
+    "vue-tsc": "^1.8.27"
```
tolking, line 121: I noticed that the version of `vue-tsc` has been updated to 2.x, and it is not possible to update to the new version here

internal/build/src/tasks/types-definitions.ts
```diff
@@ -1,158 +0,0 @@
-import process from 'process'
-import path from 'path'
-import { mkdir, readFile, writeFile } from 'fs/promises'
-import consola from 'consola'
-import * as vueCompiler from 'vue/compiler-sfc'
-import glob from 'fast-glob'
-import chalk from 'chalk'
-import { Project } from 'ts-morph'
```
tolking, line 8: It seems the `ts-morph` is no longer needed, perhaps we should remove it.

package.json
makedopamine: `typescript` need to be upgraded first cuz `vue-tsc@2.x` requires `typescript@5.x`. How about creating a new PR for it?

packages/components/input/__tests__/input.test.tsx
```diff
@@ -146,7 +146,11 @@ describe('Input.vue', () => {
   })
 
   test('rows', () => {
-    const wrapper = mount(() => <Input type="textarea" rows={3} />)
+    const wrapper = mount(() => {
+      // type checking failed as `rows` is a fallthrough attribute
+      // @ts-ignore
```
tolking, line 151: This annotation should be able to be removed now

internal/build/gulpfile.ts
```diff
@@ -50,7 +51,26 @@ export default series(
   parallel(
     runTask('buildModules'),
     runTask('buildFullBundle'),
-    runTask('generateTypesDefinitions'),
+    series(
```
tolking, line 54: A task can be separated without using the `series· function, or you can refer to buildThemeChalk for splitting

makedopamine: The task has been extracted into a separate file.

packages/components/input/__tests__/input.test.tsx
makedopamine: Removed

internal/build/src/tasks/types-definitions.ts
makedopamine: Removed

package.json
tolking: ok

packages/components/cascader-panel/src/index.vue
```diff
@@ -78,7 +78,7 @@ export default defineComponent({
     renderLabel: Function as PropType<RenderLabel>,
   },
 
-  emits: [UPDATE_MODEL_EVENT, CHANGE_EVENT, 'close', 'expand-change'],
+  emits: [UPDATE_MODEL_EVENT, CHANGE_EVENT, 'close', 'expandChange'],
```
tolking, line 81: At present, emits mostly uses hyphens, and changing it to camelize may cause some problems

packages/components/select/__tests__/options.test.tsx
```diff
@@ -61,14 +61,11 @@ describe('options', () => {
...
+          <ElOptionGroupStub>
             {{
               default: () =>
                 samples.map((_, j) => (
-                  <ElOptionStub
-                    label={getLabel(`${i}-${j}`)}
-                    value={j}
-                  ></ElOptionStub>
+                  <ElOptionStub label={getLabel(`${i}-${j}`)}></ElOptionStub>
                 )),
```
tolking, line 69: Can you explain why this has been modified in this way?

internal/build/src/tasks/types-definitions.ts
tolking: It seems that the `ts-morph` has not been removed from the package.json

packages/components/select/__tests__/options.test.tsx
makedopamine: I was not aware that the type issue had been fixed by #17020. Will revert the code.

packages/components/cascader-panel/src/index.vue
makedopamine: It only affects the users who use `OnExpand-change` in JSX or render function, and should have no impact on SFC template and in-DOM template theoretically.  IMO, it doesn't break anything. Let me know if I was wrong.

internal/build/src/tasks/types-definitions.ts
makedopamine: Oops😅, I accidentally didn't commit the files. Will push them in the next commit.

packages/components/cascader-panel/src/index.vue
tolking: It's better to avoid such inconsistent behavior as much as possible

tolking (approved):
LGTM.

makedopamine:
Rebased to resolve the conflict of `pnpm-lock.yaml`

btea:
I used this [repository](https://github.com/pure-admin/vue-pure-admin) to test the packaged code, and some ts errors occurred. I'm not sure whether it was caused by the modification here or the problem with the code over there.

makedopamine:
> I used this [repository](https://github.com/pure-admin/vue-pure-admin) to test the packaged code, and some ts errors occurred. I'm not sure whether it was caused by the modification here or the problem with the code over there.

I'll look into this later today.

makedopamine:
I'm looking into this now. Sorry for the late response, I was kinda busy with running other errands. 🤦‍♂️

btea:
Don't worry, deal with it when you have time.

makedopamine:
@btea  Did you use `pnpm link` to create a symlink for the packaged code?

btea:
@makedopamine Yes.

makedopamine:
> @makedopamine Yes.

By default, Typescript resolves modules based on the physical path of symbolic links. Therefore, in the packaged code, like `import("vue")` actually references to its own `vue` package(`vue@3.2.37`), not the `vue` package of `vue-pure-admin`, which leads to mismatch type.
The problem doesn't occur if you install the package from npm or add ` "preserveSymlinks": true` to `tsconfig.json` of `vue-pure-admin`, so it has nothing to do with this PR.

btea:
@makedopamine Thank you for your quick reply and detailed explanation. 👍

btea (approved):

PR #16175, opened 2024-03-19
fix(hooks): SSR hydration error caused by z-index

tolking:

error

```
- rendered on server: style="z-index:2003;position:absolute;left:0;top:0;display:none;"
  - expected on client: style="z-index:2001;position:absolute;left:0;top:0;"
  Note: this mismatch is check-only. The DOM will not be rectified in production due to performance overhead.
  You should fix the source of the mismatch. 
```

### DIFF
--- a/docs/en-US/guide/ssr.md
+++ b/docs/en-US/guide/ssr.md
@@ -33,7 +33,7 @@ app.provide(ID_INJECTION_KEY, {
 
 ## Provide ZIndex
 
-When you using SSR for development, you may encounter hydration errors caused by `z-ndex`. In this case, we recommend injecting an initial value to avoid such errors.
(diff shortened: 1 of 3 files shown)


docs/en-US/guide/ssr.md
```diff
@@ -31,6 +31,20 @@ app.provide(ID_INJECTION_KEY, {
 })
 ```
 
+## Provide ZIndex
+
+When you using SSR for development, you may encounter hydration errors caused by `z-ndex`. In this case, we recommend injecting an initial value to avoid such errors.
```
btea, line 36:
```suggestion
When you using SSR for development, you may encounter hydration errors caused by `z-index`. In this case, we recommend injecting an initial value to avoid such errors.
```

btea (approved):

PR #16214, opened 2024-03-23
feat(components): [message] add plain prop

kooriookami:
image

### DIFF
--- a/docs/examples/message/plain.vue
+++ b/docs/examples/message/plain.vue
@@ -10,7 +10,7 @@ import { ElMessage } from 'element-plus'
 
 const open1 = () => {
   ElMessage({
-    message: 'This is a message.',
+    message: 'Congrats, this is a success message.',
     type: 'success',
     plain: true,
   })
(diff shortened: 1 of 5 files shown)


packages/components/message/src/message.ts
```diff
@@ -111,6 +112,13 @@ export const messageProps = buildProps({
     values: messageTypes,
     default: messageDefaults.type,
   },
+  /**
+   * @description message plain
+   */
+  plain: {
+    type: Boolean,
+    default: false,
```
btea, line 120: The default value can be omitted.

docs/examples/message/plain.vue
```diff
@@ -0,0 +1,39 @@
...
+const open2 = () => {
+  ElMessage({
+    message: 'Warning, this is a warning message.',
+    type: 'warning',
+    plain: true,
+  })
+}
+const open3 = () => {
+  ElMessage({
+    message: 'Congrats, this is a success message.',
```
btea, line 27: To remain consistent with the original, the content here should probably be replaced with success.

xiaoxian521 (approved):

btea (approved):

PR #16275, opened 2024-03-27
feat(components): [input-number] support slot custom icons

selicens:

| native | use Slots |
| ------ | ---------- |
| image |  image |

### DIFF
--- a/docs/en-US/component/input-number.md
+++ b/docs/en-US/component/input-number.md
@@ -79,7 +79,7 @@ input-number/controlled
 
 :::
 
-## Custom Icon
+## Custom Icon ^(2.6.3)
 
 :::demo Use `prepend` and `append` to set custom icons.
 
--- a/docs/en-US/component/input-number.md
+++ b/docs/en-US/component/input-number.md
@@ -113,10 +113,10 @@ input-number/custom
 
 ### Slots
 
-| Name             | Description                  |
-| ---------------- | ---------------------------- |
-| prepend ^(2.6.3) | custom input box front icon  |
-| append ^(2.6.3)  | custom input box post icon   |
+| Name                   | Description                           |
+| ---------------------- | ------------------------------------- |
+| decrease-icon ^(2.6.3) | custom input box button decrease icon |
+| increase-icon ^(2.6.3) | custom input box button increase icon |
 
 ### Events
 
--- a/docs/en-US/component/input-number.md
+++ b/docs/en-US/component/input-number.md
@@ -81,7 +81,7 @@ input-number/controlled
 
 ## Custom Icon ^(2.6.3)
 
-:::demo Use `prepend` and `append` to set custom icons.
+:::demo Use `decrease-icon` and `increase-icon` to set custom icons.
 
 input-number/custom
 
--- a/docs/en-US/component/input-number.md
+++ b/docs/en-US/component/input-number.md
@@ -79,6 +79,14 @@ input-number/controlled
 
 :::
 
+## Custom Icon
+
+:::demo Use `prepend` and `append` to set custom icons.
+
+input-number/custom
+
+:::
+
 ## API
 
 ### Attributes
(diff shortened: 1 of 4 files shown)


docs/en-US/component/input-number.md
```diff
@@ -79,6 +79,14 @@ input-number/controlled
 
 :::
 
+## Custom Icon
```
btea, line 82:
```suggestion
## Custom Icon ^(2.6.3)
```

docs/examples/input-number/custom.vue
```diff
@@ -0,0 +1,36 @@
...
+      <el-icon>
+        <Minus />
+      </el-icon>
+    </template>
+    <template #append>
+      <el-icon>
+        <Plus />
+      </el-icon>
+    </template>
+  </el-input-number>
```
btea, line 28: It seems that the layout here can also be adjusted, it currently looks too compact.

selicens: image

docs/en-US/component/input-number.md
```diff
@@ -103,6 +111,13 @@ input-number/controlled
 | value-on-clear ^(2.2.0) | value should be set when input box is cleared    | ^[number] / ^[null] / ^[enum]`'min' \| 'max'` | —         |
 | validate-event          | whether to trigger form validation               | ^[boolean]                                    | true      |
 
+### Slots
+
+| Name    | Description                  |
+| ------- | ---------------------------- |
+| prepend | custom input box front icon  |
```
btea, line 118:
```suggestion
| prepend ^(2.6.3) | custom input box front icon  |
```

docs/en-US/component/input-number.md
```diff
@@ -103,6 +111,13 @@ input-number/controlled
...
+| append  | custom input box post icon   |
```
btea, line 119: Same as above.

docs/en-US/component/input-number.md
kooriookami: Do you think `decrease` is better then `prepend`?

docs/en-US/component/input-number.md
kooriookami: May be `increase` is good.

docs/en-US/component/input-number.md
selicens: > Do you think is better then ?`decrease``prepend`

It's not impossible either, it's just that following the naming conventions of similar components would be more convenient to remember

kooriookami: I know you refered to el-input. But the function is not the same. The icon slot in el-input-number only used for replacing the decrease and increase icon, It's not universal. @btea What do you think?
image

btea: It makes sense. In fact, there are cases where these two icons are on the right side at the same time. In this case, the name suggested by @kooriookami  seems clearer.

btea: Maybe it can be clearer, such as `increase-icon`/`decrease-icon`.

selicens: okay

packages/components/input-number/src/input-number.vue
```diff
@@ -17,10 +17,12 @@
       :class="[ns.e('decrease'), ns.is('disabled', minDisabled)]"
       @keydown.enter="decrease"
     >
-      <el-icon>
-        <arrow-down v-if="controlsAtRight" />
-        <minus v-else />
-      </el-icon>
+      <slot name="decreaseIcon">
```
btea, line 20:
```suggestion
      <slot name="decrease-icon">
```
Use hyphens to be consistent with other components.

selicens: Thanks to Vue's handling, this does not affect the results

btea: Yes. It’s just that I think it might be better to keep the component styles consistent.

selicens: okay

btea (approved):
LGTM 👍

kooriookami (approved):

Issue #16304, opened 2024-03-28
[Style] [input-number] el-input-number 样式被覆盖问题


chg122345:
Bug Type: **`Style`**

## Environment

- Vue Version: `3.4.21`
- Element Plus Version: `2.6.2`
- Browser / OS: `Chrome 123`
- Build Tool: `Vite`

### Related Component

- `el-input-number`

### Reproduction Link

[...]

PR #16308, opened 2024-03-29
fix(components): [input-number] fix cascading style order

makedopamine:


FrontEndDog (approved):
It seems to have not solved #16304 ,but I think the adjustment of this order is reasonable. Thanks!

makedopamine:
> It seems to have not solved #16304 ,but I think the adjustment of this order is reasonable. Thanks!

I updated the code, just swapping two lines of code, and I think it's a better way to fix the issue.

btea (approved):

PR #16390, opened 2024-04-03
fix(components): [Input] input's row props that are not declared in the typefile and  fix return value Many<T>[] not matching the type of T[]

Henrycheheng:
quewtions no1:
SF1H8_B{8SH9OCLV5I`U84Q
question no2:
Explain problem: 不能将类型“(T & any[]) | Many<T>[]”分配给类型“T[]”。 不能将类型“Many<T>[]”分配给类型“T[]”。 不能将类型“Many<T>”分配给类型“T”。 “T”可以使用与“Many<T>”无关的任意类型进行实例化。

### DIFF
--- a/packages/components/input/src/input.ts
+++ b/packages/components/input/src/input.ts
@@ -58,13 +58,13 @@ export const inputProps = buildProps({
     type: String,
     default: 'text',
   },
-  // /**
-  //  * @description rows of input
-  //  */
-  // rows: {
-  //   type: Number,
-  //   default: 1,
-  // },
+  /**
+   * @description rows of input
+   */
+  rows: {
+    type: Number,
+    default: 1,
+  },
   /**
    * @description control the resizability
    */
--- a/packages/components/input/src/input.ts
+++ b/packages/components/input/src/input.ts
@@ -58,6 +58,13 @@ export const inputProps = buildProps({
     type: String,
     default: 'text',
   },
+  // /**
+  //  * @description rows of input
+  //  */
+  // rows: {
+  //   type: Number,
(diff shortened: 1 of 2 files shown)


packages/components/input/src/input.ts
```diff
@@ -58,6 +58,13 @@ export const inputProps = buildProps({
     type: String,
     default: 'text',
   },
+  /**
+   * @description rows of input
+   */
+  rows: {
+    type: Number,
+    default: 1,
+  },
```
btea, line 67: There is no need to set the rows prop here. The native textarea supports setting rows directly, and attrs is also directly bound in the code.

packages/utils/arrays.ts
```diff
@@ -5,7 +5,7 @@ type Many<T> = T | ReadonlyArray<T>
 /** like `_.castArray`, except falsy value returns empty array. */
 export const castArray = <T>(arr: Many<T>): T[] => {
   if (!arr && (arr as any) !== 0) return []
-  return Array.isArray(arr) ? arr : [arr]
+  return Array.isArray(arr) ? (arr as T[]) : [arr as T]
```
btea, line 8: There seems to be no problem under the ts version that the code currently depends on, and it may not need to be modified.

packages/components/input/src/input.ts
Henrycheheng: tanks a lot. this problem may exist the ide and my version of ts

packages/utils/arrays.ts
Henrycheheng: i learn from it

PR #16393, opened 2024-04-04
fix(components): fix iOS select click event listening

### LINKED ISSUE #5210: [Bug Report] El-Input with clearable attribute had text need click twice to open keyboard on iOS Safari
### Element Plus version
1.3.0-beta.1

### OS/Browsers version
OS X 10.15.6 / Safari 15

### Vue version
3.2.26

### Reproduction Link
https://jsfiddle.net/st9140626/kg986fns/4/

### Steps to reproduce
When el-input with clearable attribute had inserted text, I need click twice to open input keyboard on iOS Safari.

### What is Expected?
When I touch the input, it will be auto focus and open the keyboard on iOS Safari.

I need to click twice for each input with clearable attribute on iOS Safari.

raphaelbernhart:
## Problem / fix
iOS Safari does not handle click events when a mouseenter event is registered and a DOM-change happens in a child.
We use a Vue custom event binding to only register the event on non-iOS devices.

The inputHovering state gets updated now on iOS-devices on click. This is needed as the mouseenter event normally does this.

Apple Developer Documentation: https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/HandlingEvents/HandlingEvents.html
Vue Issue: https://github.com/vuejs/vue/issues/9859

### Before changes
https://github.com/element-plus/element-plus/assets/48283236/c644f302-8764-413b-b975-b2ca758be154

### After changes
https://github.com/element-plus/element-plus/assets/48283236/7f8c5132-9493-49be-833d-5f0db09fc779

closed #5210

### DIFF
--- a/packages/components/select/src/select.vue
+++ b/packages/components/select/src/select.vue
@@ -3,7 +3,7 @@
     ref="selectRef"
     v-click-outside:[popperRef]="handleClickOutside"
     :class="[nsSelect.b(), nsSelect.m(selectSize)]"
-    @mouseenter="states.inputHovering = true"
+    @[mouseEnterEventName]="states.inputHovering = true"
     @mouseleave="states.inputHovering = false"
     @click.prevent.stop="toggleMenu"
   >
--- a/packages/components/select/src/useSelect.ts
+++ b/packages/components/select/src/useSelect.ts
@@ -28,6 +28,7 @@ import {
   debugWarn,
   isClient,
   isFunction,
+  isIOS,
   isNumber,
   isUndefined,
   scrollIntoView,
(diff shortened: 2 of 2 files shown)


kooriookami:
> @kooriookami @tolking Do you have relevant device verification?

No, I have no IOS device.

raphaelbernhart:
@btea Is there any updates here? :)

We fixed it now with a custom installation of element-plus but we would really like to just use the npm package.

warmthsea:
> @btea Is there any updates here? :)
> 
> We fixed it now with a custom installation of element-plus but we would really like to just use the npm package.

I will participate in the verification tomorrow when I have time.

warmthsea (commented):
### https://warmthsea.github.io/ios-select.mp4
### [test repo](https://github.com/warmthsea/test-ios-select)
### [test preview link](https://test-ios-select.vercel.app/)

#### This PR does solve the problem that select(`clearable`) on iOS needs to be clicked twice before it can be opened.


#### (I only tested Safari on iOS Phone.)

warmthsea:
> @btea Is there any updates here? :)
> 
> 
> 
> We fixed it now with a custom installation of element-plus but we would really like to just use the npm package.

You can consider publishing your own element-plus npm package for the time being.

raphaelbernhart:
> > @btea Is there any updates here? :)
> > We fixed it now with a custom installation of element-plus but we would really like to just use the npm package.
> 
> You can consider publishing your own element-plus npm package for the time being.

Yes, tho atm it is ok for us to keep the package locally, it would just be nice to have the upstream package for updating purposes. :)

Thanks for the review!

btea:
@raphaelbernhart  It seems to be related to the system version. I used Safari to simulate an iPhone 15 and there seemed no problem. 🤔

btea:
>Also keep in mind that the problem is just when selecting an option and then again trying to open the options.

Following the steps you described, it actually happened.

btea (approved):
Thanks!

PR #16395, opened 2024-04-05
test(hooks): add test for use-intermediate-render

wzc520pyfm:
Before:
use-intermediate-render-before
After:
use-intermediate-render-after

### DIFF
--- a/packages/hooks/__tests__/use-intermediate-render.test.ts
+++ b/packages/hooks/__tests__/use-intermediate-render.test.ts
@@ -1,0 +1,98 @@
+import { nextTick, ref } from 'vue'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { useDelayedRender } from '../use-intermediate-render'
+import type { UseDelayedRenderProps } from '../use-intermediate-render'
+
+describe('useDelayedRender', () => {
+  let props: UseDelayedRenderProps
+  let indicator: any
+  let intermediateIndicator: any
+  let beforeShow: any
+  let afterShow: any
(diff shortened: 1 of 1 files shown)


packages/hooks/__tests__/use-intermediate-render.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    props = {
+      indicator,
+      intermediateIndicator,
+      shouldSetIntermediate: vi.fn().mockReturnValue(true),
+      beforeShow,
+      afterShow,
+      beforeHide,
+      afterHide,
+    }
```
btea, line 32: Perhaps the initialization assignment here can be placed together with the variable declaration. To prevent interference, just reset it in beforeEach.

wzc520pyfm: > Perhaps the initialization assignment here can be placed together with the variable declaration. To prevent interference, just reset it in beforeEach.

Sorry, I didn't understand. This code need to reset the variables in beforeEach to reset the mock statistics to zero.

btea: Yes, but I think `mockReset` can be called directly to meet the needs.

btea: Indeed, `mockRestore` does not reset the number of calls. Maybe we can change our thinking and not reset the function, but match the number of calls.

btea (approved):
Thanks!

btea (approved):
Thanks!

PR #16504, opened 2024-04-14
fix(test-utils): fix main export error

wzc520pyfm:
Module redirection of `@element-plus/test-utils` cannot be used, I tried to fix it.
pkg-test-utils-main-export-error
```ts
// Before 
import sleep from '@element-plus/test-utils/sleep'
// After
import { sleep } from '@element-plus/test-utils'
```


packages/components/affix/__tests__/affix.test.tsx
```diff
@@ -1,8 +1,7 @@
 import { nextTick } from 'vue'
 import { mount } from '@vue/test-utils'
 import { afterAll, beforeAll, describe, expect, test, vi } from 'vitest'
-import defineGetter from '@element-plus/test-utils/define-getter'
-import makeScroll from '@element-plus/test-utils/make-scroll'
+import { defineGetter, makeScroll } from '@element-plus/test-utils'
```
btea, line 4: Thank you for your contribution, but in my opinion it might be clearer to show the specific file name of the import.

packages/test-utils/tick.ts
```diff
@@ -1,13 +1,11 @@
 import { nextTick } from 'vue'
 
-const tick = async (times: number) => {
+export const tick = async (times: number) => {
```
btea, line 3: There doesn't seem to be any adjustment needed here either.

packages/components/affix/__tests__/affix.test.tsx
wzc520pyfm: > Thank you for your contribution, but in my opinion it might be clearer to show the specific file name of the import.

I agree with you.
However, just like using `@element-plus/hooks`, my first thought was to use module redirection(`import { sleep } from '@element-plus/test-utils'`), In fact, I don't get any errors when i use, but throws error when I run. 

Maybe we can just fix the export problem in pkg.json, and without adjusting the existing import.
image

btea: Yes, there should be some adjustments here.

btea (approved):

PR #16511, opened 2024-04-15
feat(components): [time-select] export TimeSelectProps and Instance

l-x-f:
All other components have been exported props and instance, but timeSelect has not。

```ts
// support
import type { FormItemProps } from 'element-plus'

// not support
import type { TimeSelectProps} from 'element-plus'
```


btea (approved):

kooriookami:
May be we can use withInstall instead. Can refer to https://github.com/element-plus/element-plus/blob/dev/packages/components/affix/index.ts

tolking (approved):

PR #16597, opened 2024-04-20
chore: remove redundant code

makedopamine:
image
As the variable `selectedIndex` only appears in the two lines, it should be redundant.


kooriookami:
Please fix the lint.

makedopamine:
While I was fixing the lint error, I noticed that there are `ts-nocheck` comments in `select-v2`. It makes the component somewhat inefficient to maintain without type checking. I'll try to remove the comments in another PR.

kooriookami (approved):

FrontEndDog (approved):

PR #16598, opened 2024-04-20
feat(hooks): add use-aria hooks

kooriookami:
Add `use-aria` hooks. 
Change native `label` to `aria-label`.


packages/hooks/use-aria/index.ts
```diff
@@ -0,0 +1,19 @@
...
+   * @description native `aria-orientation` attribute
+   */
+  ariaOrientation: {
+    type: String,
+    values: ['horizontal', 'vertical', 'undefined'],
+  },
+  /**
+   * @description native `aria-controls` attribute
+   */
+  ariaControls: String,
```
btea, line 18: `ariaOrientation` and `ariaControls`prop appear to be useless in most components modified above. Perhaps it should be treated differently here.

kooriookami: In my opinion, `aria-xxx` means native props with no ambiguity.. If you try to differentiate by passing a parameter instead, it becomes cumbersome.

btea: For example, the current `el-rate` code shows that the `aria-controls` and `aria-orientation` settings are supported, but even if they are set, there seems to be no logic to add related attributes, which seems wrong. May cause confusion to users.

kooriookami: Just don't write it in the document.

btea: Yes, I noticed. However, ts may prompt you in the code editor. I think it is better to remove it directly.

Fuphoenixes: Maybe we can add a logic to pick props

Fuphoenixes (approved):

packages/hooks/use-aria/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * @description native `aria-label` attribute
+   */
+  ariaLabel: String,
+  /**
+   * @description native `aria-orientation` attribute
+   */
+  ariaOrientation: {
+    type: String,
+    values: ['horizontal', 'vertical', 'undefined'],
```
btea, line 14: Why is `'undefined'` added here? 🤔

kooriookami: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-orientation#values

btea: I see. 👍

btea (approved):

Issue #16622, opened 2024-04-23
[Component] [table-v2] el-table-v2 加入操作列后 onRowHovered 事件报错


DejaVuuuuuuu:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.45`
- Element Plus Version: `2.7.1`
- Browser / OS: `UserAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/122.0.0.0 Safari/537.36 Edg/122.0.0.0`
- Build Tool: `Vite`

### Related Component

- `el-table-v2`

### Reproduction Link

[...]

PR #16635, opened 2024-04-24
docs(components): [table-v2] fix dynamic-height examples error

warmthsea:
-  `querySelectorAll`  not support `:`, before > after, about https://github.com/element-plus/element-plus/issues/16622

screenshots


btea (approved):
LGTM! Thanks!

PR #15352, opened 2024-01-04
refactor(components): [select & select-v2] Refactor components

kooriookami:
## Why Refactor 为什么重构

Currently select and select-v2 styles and dom structure are not consistent, style files can not be reused, resulting in differences in appearance. The internal code of the component is redundant, and there are some bugs that are hard to fix.

目前 select 与 select-v2 样式和 dom 结构不一致，样式文件不能复用，造成外观有差异。组件内部代码冗余，存在一些难以修复的 bug。

## What Changed 改变了什么

1. The default width of the component is 100%, consistent with the input.
组件默认宽度为 100%，与 input 保持一致

2. select-v2 class naming is consistent with select by removing the v2 suffix.
select-v2 class 命名移除 v2 后缀，与 select 保持一致。

3. select and select-v2 adjust the dom structure, class naming remains consistent.
select 与 select-v2 调整 dom 结构，class 命名保持一致。

4. select and select-v2 share the select style file, no need to maintain two styles.
select 与 select-v2 共用 select 样式文件，不必维护两份样式。

5. Optimised select and select-v2 styles according to the latest [UI Kit](https://www.figma.com/community/file/1305760370797950824/element-plus-design-system-ui-kit). Such as multiple mode,  tag will near the border. select-v2 selected values to remove the background color. Single select mode dropdown menu does not need the right check icon and so on.
根据最新的[原型图](https://www.figma.com/community/file/1305760370797950824/element-plus-design-system-ui-kit)，优化了 select 与 select-v2 样式。如多选模式 tag 会贴边，select-v2 选中值去除背景色，单选模式下拉选中不需要右侧勾号等。

6. In single select mode, empty value includes `empty string`, `null`, `undefined`.
单选模式下，空值包含 `空字符串`, `null`, `undefined`。

7. Fix some bugs, such as select `automatic-dropdown` flash, select-v2 `automatic-dropdown` doesn't work.
修复一些 bug，如 select `automatic-dropdown` 闪硕， select-v2 `automatic-dropdown` 不生效等问题。

8. Optimise code, dynamically listen for dom changes and remove hardcoding. Try to make select and select-v2 internal parameters and methods consistent.
优化代码，动态监听 dom 变化，移除硬编码。尽量使 select 与 select-v2 内部参数与方法一致。

9. Optimize the document.
优化文档。

## Any Break Changes 是否有破坏性

If you are just a user, theoretically there is no effect. If you make changes to the component, such as modifying parameters or methods inside the component, then there may be a conflict.

[...]

PR #16650, opened 2024-04-25
docs(components): [select] unify and enhance version tip

warmthsea:
About: https://github.com/element-plus/element-plus/pull/15352

## Before: 
989b657966b35a31900c1b48827c4362

## After:
b585d8f9232eadcf450bbd613790d76d

## Other:
image

### DIFF
--- a/docs/en-US/component/select.md
+++ b/docs/en-US/component/select.md
@@ -9,7 +9,7 @@ When there are plenty of options, use a drop-down menu to display and select des
 
 :::tip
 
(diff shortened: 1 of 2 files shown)


FrontEndDog (commented):
Thanks! You can also make changes here together!
https://github.com/element-plus/element-plus/blob/c46a21a54eb9f521449173c0a6ecb7d3347c9b59/docs/en-US/guide/changelog.md?plain=1#L18

FrontEndDog (approved):
Thanks!

kooriookami (approved):

PR #16733, opened 2024-04-30
docs: change some examples label to aria-label

warmthsea:
Should these be changed to new ones or deleted directly? I think this is an implicit attribute, and many users are still using versions before 2.7.2.

这些应该改成新的还是直接删除，我认为这是个隐式的属性，并且很多用户还在使用 2.7.2之前的版本.

https://github.com/element-plus/element-plus/pull/16598

cc: @kooriookami


btea (approved):
I think it’s better to keep the features consistent with the new version.

kooriookami (approved):

kooriookami:
> I think it’s better to keep the features consistent with the new version.

I agree.

PR #16734, opened 2024-04-30
feat(components): [container] add validation for the `direction` prop

deng-001:
- I found that the orientation attribute of the Container component was not validated.
- So I added enumeration value validation to the component.


btea (approved):
LGTM.

packages/components/container/src/container.vue
```diff
@@ -18,6 +19,7 @@ const props = defineProps({
    */
   direction: {
     type: String,
+    validator: isDirectionType,
```
tolking, line 22:
```suggestion
    validator: (val: string) => ['horizontal', 'vertical'].includes(val),
```

Because the `direction` attribute is `'ltr' | 'rtl' | 'ttb' | 'btt'` in the ElDrawer component, moving the `validator` to the global may not be a good idea.

btea: @chenyuqiua Please adjust your code a bit as @tolking suggested.

chenyuqiua: I have modified it as above @btea

tolking (approved):
Sorry, I forgot about this PR.

PR #16737, opened 2024-05-01
chore: set `package-manager-strict` to false

btea:

After pnpm releases version [9.0.3](https://github.com/pnpm/pnpm/releases/tag/v9.0.3), when the local pnpm version and packageManger version do not match, an error message will appear and the process will exit when running `pnpm dev`.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,4 @@
 shamefully-hoist=true
 strict-peer-dependencies=false
 shell-emulator=true
+package-manager-strict=false


kooriookami:
Will the lock file be changed in different version?

btea:
Yes, the `lockfileVersion` corresponding to the lock file generated by pnpm 9 is `9.0`. We will update `packageManager` to version 9 later.
The problem now is that if the `packageManager` version is set to `9.0.3`, if you install `9.0.4` locally, an error will occur and the process will exit. I think this restriction is too strict.

kooriookami:
What about `pnpm@9` without minor version.

btea:
This looks good, but I'm not sure if this will throw an error, I might try it later.

btea:
> What about `pnpm@9` without minor version.

That does not work.

image

kooriookami:
OK, I see.

kooriookami (approved):

PR #16740, opened 2024-05-02
docs(examples): change el-date-picker label to aria-label

warmthsea:
https://github.com/element-plus/element-plus/blob/8662178484ca80e297bc74de72db9cd3aa724001/packages/components/date-picker/src/props/date-picker.ts#L8

https://github.com/element-plus/element-plus/blob/8662178484ca80e297bc74de72db9cd3aa724001/packages/components/time-picker/src/common/props.ts#L219

https://github.com/element-plus/element-plus/pull/16733

### DIFF
--- a/docs/examples/form/size-control.vue
+++ b/docs/examples/form/size-control.vue
@@ -37,7 +37,7 @@
         <el-date-picker
           v-model="sizeForm.date1"
           type="date"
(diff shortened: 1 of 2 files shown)


btea (approved):
Thanks!

PR #16743, opened 2024-05-02
chore(scripts): fix type error

warmthsea:
## Before:
image

## After:
image

### DIFF
--- a/scripts/build-table.ts
+++ b/scripts/build-table.ts
@@ -2,7 +2,7 @@ import fs from 'fs/promises'
 import path from 'path'
 
 async function main() {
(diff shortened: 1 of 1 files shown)


scripts/build-table.ts
```diff
@@ -2,7 +2,7 @@ import fs from 'fs/promises'
 import path from 'path'
 
 async function main() {
-  const threshold = process.env.THRESHOLD || 40
+  const threshold = +(process.env.THRESHOLD || 40)
```
btea, line 5:
```suggestion
  const threshold = Number(process.env.THRESHOLD) || 40
```

btea (approved):
Thanks!

PR #16744, opened 2024-05-02
chore(dev): add unocss and delete vite vscode config

warmthsea:
This project uses unocss, but vite vscode plugin is of no use.


btea (approved):
LGTM.

PR #16746, opened 2024-05-02
chore: remove ts-nocheck comments in select-v2

makedopamine:


warmthsea:
Found hundreds of  `@ts-nochecks`.  👀

makedopamine:
> Found hundreds of `@ts-nochecks`. 👀

I actually notice the problem what you say, however, its workload is somewhat heavy to address it completely in  a single PR. And I think a large PR is hard for reviewers to review, so I prefer to split the task into separate PRs.
I'll work on `select` component next. It would be nice if you are interested in working on other components.😀

packages/components/select-v2/src/defaults.ts
```diff
@@ -13,8 +19,9 @@ import { defaultProps } from './useProps'
 import type { Option, OptionType } from './select.types'
 import type { Props } from './useProps'
 import type { Options, Placement } from '@element-plus/components/popper'
+import type { EmitFn } from '@element-plus/utils/vue/typescript'
 
-export const SelectProps = buildProps({
+export const selectProps = buildProps({
```
btea, line 24: It may be better to keep the first letter capitalized here, otherwise it will be a breaking change for users who use related values.

packages/components/select-v2/src/defaults.ts
```diff
@@ -248,7 +255,7 @@ export const SelectProps = buildProps({
   ...useAriaProps(['ariaLabel']),
 } as const)
 
-export const OptionProps = buildProps({
+export const optionProps = buildProps({
```
btea, line 258: Same here.

packages/components/select-v2/src/select-dropdown.tsx
```diff
@@ -190,7 +220,7 @@ export default defineComponent({
     }
 
     const onEscOrTab = () => {
-      select.expanded = false
+      select.expanded.value = false
```
btea, line 223: Is there a bug here?

packages/components/select-v2/src/select.types.ts
```diff
@@ -1,3 +1,5 @@
+import type { Nullable } from '@element-plus/utils'
```
btea, line 1: Maybe `Nullable` should no longer be used here, as it will be deprecated later.

packages/components/select-v2/src/token.ts
```diff
@@ -1,12 +1,12 @@
...
 import type { ExtractPropTypes, InjectionKey, Ref } from 'vue'
 import type { Option } from './select.types'
 import type { TooltipInstance } from '@element-plus/components/tooltip'
 
 export interface SelectV2Context {
-  props: ExtractPropTypes<typeof SelectProps>
-  expanded: boolean
-  tooltipRef: Ref<TooltipInstance>
+  props: ExtractPropTypes<typeof selectProps>
+  expanded: Ref<boolean>
```
btea, line 8: Why has the type changed here?

packages/components/select-v2/src/useProps.ts
```diff
@@ -18,7 +18,7 @@ export const defaultProps: Required<Props> = {
   options: 'options',
 }
 
-export function useProps(props: Pick<ISelectV2Props, 'props'>) {
+export function useProps(props: ISelectV2Props) {
```
btea, line 21: It seems that there is no need to change here. Only `props.props` is used below, and only the corresponding types are extracted.

packages/components/select-v2/src/useSelect.ts
```diff
@@ -43,13 +42,16 @@ import { useAllowCreate } from './useAllowCreate'
 import { useInput } from './useInput'
 import { useProps } from './useProps'
 
-import type ElTooltip from '@element-plus/components/tooltip'
-import type { Option, OptionType } from './select.types'
+import type { Option, OptionType, SelectStates } from './select.types'
 import type { ISelectV2Props } from './token'
+import type { SelectEmitFn } from './defaults'
+import type { TooltipInstance } from '@element-plus/components/tooltip'
+import type { Nullable } from '@element-plus/utils/typescript'
```
btea, line 49: `Nullable` should no longer be used.

packages/components/select-v2/src/token.ts
makedopamine: For `select.expanded.value  = false`

packages/components/select-v2/src/select-dropdown.tsx
makedopamine: Yep, it should indeed not be modified in this PR, causing confusion. I will revert it.

packages/components/select-v2/src/defaults.ts
makedopamine: You are right.

packages/components/select-v2/src/useProps.ts
makedopamine: I think the type should match the runtime value. Anyway, it doesn't matter. I'm fine with both.

makedopamine:
Thanks for your review. There's been a modification, please take a look.

packages/components/select-v2/src/defaults.ts
```diff
@@ -256,8 +259,38 @@ export const OptionProps = buildProps({
...
+  visibleChange: (visible: boolean) => true,
+  clear: () => true,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  focus: (event: FocusEvent) => true,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  blur: (event: FocusEvent) => true,
+}
+export const optionEmits = {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  hover: (index: number) => true,
```
btea, line 288:
```suggestion
  hover: (index: number) => isNumber(index),
```

packages/components/select-v2/src/defaults.ts
```diff
@@ -256,8 +259,38 @@ export const OptionProps = buildProps({
...
+  [UPDATE_MODEL_EVENT]: (val: ISelectV2Props['modelValue']) => true,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  [CHANGE_EVENT]: (val: ISelectV2Props['modelValue']) => true,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  removeTag: (val: unknown) => true,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  visibleChange: (visible: boolean) => true,
+  clear: () => true,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  focus: (event: FocusEvent) => true,
```
btea, line 282:
```suggestion
  focus: (event: FocusEvent) => event instanceof FocusEvent,
```

packages/components/select-v2/src/useSelect.ts
```diff
@@ -151,9 +152,11 @@ const useSelect = (props: ISelectV2Props, emit) => {
   )
 
   const validateState = computed(() => elFormItem?.validateState || '')
-  const validateIcon = computed(
-    () => ValidateComponentsMap[validateState.value]
-  )
+  // eslint-disable-next-line vue/return-in-computed-property
+  const validateIcon = computed(() => {
+    if (!validateState.value) return
+    return ValidateComponentsMap[validateState.value]
```
btea, line 158: There seems to be no need to add additional judgment here. If `validateState.value` does not exist, the returned value should also be `undefined`.

packages/components/virtual-list/src/hooks/use-cache.ts
```diff
@@ -4,14 +4,20 @@ import memoOne from 'memoize-one'
...
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    const _getItemStyleCache = (_: any, __: any, ___: any) => ({})
+    const _getItemStyleCache = (
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      _: any,
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      __: any,
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      ___: any
+    ): Record<string, T> => ({})
```
btea, line 20: It seems that there is no need to add so many comments here, using `as` should be fine.
```ts
const _getItemStyleCache = (_: any, __: any, ___: any) =>
 ({} as Record<string, T>)
```

packages/components/select-v2/src/defaults.ts
```diff
@@ -256,8 +259,38 @@ export const OptionProps = buildProps({
     type: definePropType<Option>(Object),
     required: true,
   },
-  index: Number,
+  index: {
+    type: Number,
+    required: true,
```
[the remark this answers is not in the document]
makedopamine: On second thought, it's indeed not required.

packages/components/select-v2/src/select-dropdown.tsx
```diff
@@ -21,35 +21,58 @@ import { useProps } from './useProps'
...
+  ComputedRef,
+  ExtractPropTypes,
+  Ref,
+} from 'vue'
 
+const props = {
+  loading: Boolean,
+  data: {
+    type: Array,
+    required: true as const,
```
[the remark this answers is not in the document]
makedopamine: It's necessary. The type of  `props.data`  will include `undefined` without it.

packages/components/select-v2/src/useSelect.ts
makedopamine: I see what you mean, however,  `Typescript` will report “Element implicitly has an 'any' type” for it.

packages/components/virtual-list/src/hooks/use-cache.ts
makedopamine: That makes sense.🤔

packages/components/select-v2/src/group-item.vue
```diff
@@ -15,18 +15,24 @@
...
     },
-    style: Object,
-    height: Number,
+    style: {
+      type: Object as PropType<CSSProperties>,
+      required: true,
+    },
+    height: {
+      type: Number,
+      required: true,
```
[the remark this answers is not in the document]
makedopamine: Not necessary, modified.

packages/components/select-v2/src/defaults.ts
```diff
@@ -256,8 +264,35 @@ export const OptionProps = buildProps({
...
   },
-  index: Number,
+  index: {
+    type: Number,
+  },
   style: Object,
   selected: Boolean,
   created: Boolean,
 } as const)
+
```
btea, line 274: Multiple repeated rule comment markers can be written in a different way.
```ts
/* eslint-disable @typescript-eslint/no-unused-vars */
......
/* eslint-enable @typescript-eslint/no-unused-vars */
```

makedopamine: That's neat👍

makedopamine:
Locally-run tests pass, but CI tests failed. 🤔 There seems something wrong with CI.

btea:
Independent of your changes, the test cases seem to fail occasionally.

packages/components/select-v2/src/defaults.ts
```diff
@@ -256,8 +264,31 @@ export const OptionProps = buildProps({
     type: definePropType<Option>(Object),
     required: true,
   },
-  index: Number,
+  index: {
+    type: Number,
+  },
```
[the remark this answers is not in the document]
makedopamine: Ah, it's an oversight on my part. 😅 I removed the `required` property and forgot to simplify the code.

packages/components/select-v2/src/useSelect.ts
btea: Should we add a comment here about this kind of writing?

makedopamine: Agree, will do.

btea (approved):

tolking:
Do you have time to resolve the conflicts

makedopamine:
> Do you have time to resolve the conflicts

I'm going to do it right now.

packages/components/select-v2/src/defaults.ts
```diff
@@ -269,3 +277,24 @@ export const OptionProps = buildProps({
...
 } as const)
+
+/* eslint-disable @typescript-eslint/no-unused-vars */
+export const selectEmits = {
+  [UPDATE_MODEL_EVENT]: (val: ISelectV2Props['modelValue']) => true,
+  [CHANGE_EVENT]: (val: ISelectV2Props['modelValue']) => true,
+  'remove-tag': (val: unknown) => true,
+  'visible-change': (visible: boolean) => true,
+  blur: (event: FocusEvent) => true,
+  focus: (event: FocusEvent) => event instanceof FocusEvent,
```
tolking, line 288:
```suggestion
  focus: (evt: FocusEvent) => evt instanceof FocusEvent,
  blur: (evt: FocusEvent) => evt instanceof FocusEvent,
```

Link #17545

packages/components/select-v2/src/defaults.ts
```diff
@@ -269,3 +277,24 @@ export const OptionProps = buildProps({
...
+  'visible-change': (visible: boolean) => true,
+  focus: (evt: FocusEvent) => evt instanceof FocusEvent,
+  blur: (evt: FocusEvent) => evt instanceof FocusEvent,
+  clear: () => true,
+}
+export const optionEmits = {
+  hover: (index?: number) => isNumber(index),
+  select: (val: Option, index?: number) => true,
+}
+/* eslint-enable @typescript-eslint/no-unused-vars */
```
tolking, line 295: This line seems to be useless

makedopamine: It's to follow a recommend eslint rule called [disable-enable-pair](https://eslint-community.github.io/eslint-plugin-eslint-comments/rules/disable-enable-pair.html).

tolking (approved):

tolking (approved):

PR #16752, opened 2024-05-06
chore(dev): add code save auto fix eslint warn

warmthsea:
I think some simple eslint errors can be saved and handled automatically.
Has no impact on other code formats at all.

reference: 
https://github.com/vueuse/vueuse/blob/b7d843870baaa826b01e6822e1d905231c0e02d6/.vscode/settings.json#L35C1-L38C4

## Before:
screenshots
screenshots

## After:
screenshots
screenshots


btea (approved):
LGTM. Thanks.

FrontEndDog (approved):

PR #16818, opened 2024-05-11
fix(types): object utils error

warmthsea:
TSC seems to not have these problems due to version reasons, but type errors do exist.

screenshots


btea:
There should be no error in the currently dependent ts version. You can try switching the ts version in the lower right corner.

warmthsea:
> There should be no error in the currently dependent ts version. You can try switching the ts version in the lower right corner.

It may be due to the dependence of my different branches. I will submit the upgraded PR uniformly later.

PR #16824, opened 2024-05-11
test(components): [select] modify the grouping of several test cases

makedopamine:
I just moved the code without any other modification.

### DIFF
--- a/packages/components/select/__tests__/select.test.ts
+++ b/packages/components/select/__tests__/select.test.ts
@@ -2480,111 +2480,113 @@ describe('Select', () => {
       const formItem = wrapper.find('[data-test-ref="item"]')
       expect(formItem.attributes().role).toBe('group')
     })
-    // fix: 8544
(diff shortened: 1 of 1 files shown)


btea:
It seems that you just adjusted the position of the test cases. Can you explain the reason?

packages/components/select/__tests__/select.test.ts
```diff
@@ -2480,111 +2480,113 @@ describe('Select', () => {
       const formItem = wrapper.find('[data-test-ref="item"]')
       expect(formItem.attributes().role).toBe('group')
     })
-    // fix: 8544
-    it('When props are changed, label can be displayed correctly after selecting operation', async () => {
```
makedopamine, line 2484: like this test case should not be contained in the group `form item accessibility integration`

btea: I see. Thanks!

btea (approved):

PR #7790, opened 2022-05-19
fix(components): [popper-content] don't provide formItemContext when unnecessary

CarterLi:
Please make sure these boxes are checked before submitting your PR, thank you!

- [x] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).
- [x] Make sure you are merging your commits to `dev` branch.
- [x] Add some descriptions and refer to relative issues for your PR.

`el-popper-content` will constantly provide `formItemContext` with two noop functions, even not inside a `el-form-item`, which is problematic on other places which only test  `formItemContext` is null or undefined.

For example:

https://github.com/element-plus/element-plus/blob/dev/packages/components/radio/src/radio-group.vue#L65

Fixes https://github.com/element-plus/element-plus/pull/7450

image

@opengraphica @JeremyWuuuuu

PR #16834, opened 2024-05-13
fix(components): [popper] remove unnecessary conditional judgment

warmthsea:
`()=> void` always `true`
about: https://github.com/element-plus/element-plus/pull/7790/files


btea (approved):
Thanks!

PR #16845, opened 2024-05-13
chore: change @element-plus/build-constants to workspace

warmthsea:
This problem occurred when I installed dependencies on different branches.

about: https://github.com/element-plus/element-plus/pull/16794

image
image


btea:
Duplicate of #16794

PR #16858, opened 2024-05-14
feat(pagination) : add  `size` prop

### LINKED ISSUE #16830: [Component] [pagination]  el-config-provider 设置size="small" 后 el-pagination 尺寸错误
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.37`
- Element Plus Version: `2.7.2`
- Browser / OS: `Microsoft Edge 124`
- Build Tool: `Vite`

### Related Component

- `el-pagination`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJzcmMvQXBwLnZ1ZSI6IjxzY3JpcHQgc2V0dXAgbGFuZz1cInRzXCI+XG5pbXBvcnQgeyByZWYsIHZlcnNpb24gYXMgdnVlVmVyc2lvbiB9IGZyb20gJ3Z1ZSdcbmltcG9ydCB7IHZlcnNpb24gYXMgZXBWZXJzaW9uIH0gZnJvbSAnZWxlbWVudC1wbHVzJ1xuaW1wb3J0IHsgRWxlbWVudFBsdXMgfSBmcm9tICdAZWxlbWVudC1wbHVzL2ljb25zLXZ1ZSdcbmltcG9ydCB7IEVsQ29uZmlnUHJvdmlkZXIgfSBmcm9tIFwiZWxlbWVudC1wbHVzXCI7XG5cbmNvbnN0IG1zZyA9IHJlZignSGVsbG8gV29ybGQhJylcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxoMT57eyBtc2cgfX08L2gxPlxuICA8ZWwtaW5wdXQgdi1tb2RlbD1cIm1zZ1wiIC8+XG4gIDxiciAvPlxuICA8c3Bhbj7lhajlsYDlsI/lsLrlr7g8L3NwYW4+XG4gPGVsLWNvbmZpZy1wcm92aWRlciAgc2l6ZT1cInNtYWxsXCI+XG5cbiA8ZWwtaW5wdXQgdi1tb2RlbD1cIm1zZ1wiIC8+XG4gXG4gPGVsLXBhZ2luYXRpb24gIGxheW91dD1cInRvdGFsLCBzaXplcywgcHJldiwgcGFnZXIsIG5leHQsIGp1bXBlclwiIDp0b3RhbD1cIjEwMFwiICAvPlxuXG4gPC9lbC1jb25maWctcHJvdmlkZXIgPlxuICA8cD5cbiAgICA8ZWwtaWNvbiBjb2xvcj1cInZhcigtLWVsLWNvbG9yLXByaW1hcnkpXCI+PEVsZW1lbnRQbHVzIC8+PC9lbC1pY29uPlxuICAgIEVsZW1lbnQgUGx1cyB7eyBlcFZlcnNpb24gfX0gKyBWdWUge3sgdnVlVmVyc2lvbiB9fVxuICA8L3A+XG48L3RlbXBsYXRlPlxuIiwiaW1wb3J0LW1hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge31cbn0iLCJ0c2NvbmZpZy5qc29uIjoie1xuICBcImNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogXCJFU05leHRcIixcbiAgICBcImpzeFwiOiBcInByZXNlcnZlXCIsXG4gICAgXCJtb2R1bGVcIjogXCJFU05leHRcIixcbiAgICBcIm1vZHVsZVJlc29sdXRpb25cIjogXCJCdW5kbGVyXCIsXG4gICAgXCJ0eXBlc1wiOiBbXCJlbGVtZW50LXBsdXMvZ2xvYmFsLmQudHNcIl0sXG4gICAgXCJhbGxvd0ltcG9ydGluZ1RzRXh0ZW5zaW9uc1wiOiB0cnVlLFxuICAgIFwiYWxsb3dKc1wiOiB0cnVlLFxuICAgIFwiY2hlY2tKc1wiOiB0cnVlXG4gIH0sXG4gIFwidnVlQ29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiAzLjNcbiAgfVxufVxuIiwiX28iOnt9fQ==)

### Steps to reproduce

 <el-config-provider size="small">   <el-pagination /></el-config-provider>

### What is Expected?

el-pagination 显示为小尺寸 即 :small="true"

 el-pagination 显示标准尺寸 即 :small="false"

### Additional comments

麻烦你们了

webvs2:

close #16830


btea:
I think it would also be good to add unified global setting size support to `el-pagination`. What do you think? cc @element-plus/backers

long4488296 (commented):
微信截图_20240515110156-ok

ok!!!

Liao-js:
I think it would be better to use useFormSize to get size for el-pagination.

webvs2:
> I think it would be better to use useFormSize to get size for el-pagination.

The purpose of this function is to get the size of the previous level rather than just relying on the global, and the use is that if the size is set like 'from', it is actually good to use the local t mentioned above, right?

Liao-js:
> The purpose of this function is to get the size of the previous level rather than just relying on the global

useFormSize obtains size in the order of props, formItem, form, globalConfig

webvs2:
> > The purpose of this function is to get the size of the previous level rather than just relying on the global
> 
> useFormSize obtains size in the order of props, formItem, form, globalConfig
Yes, I looked at what the function does so it makes more sense to rely on the parent, which I will adjust

webvs2:
> > The purpose of this function is to get the size of the previous level rather than just relying on the global
> 
> useFormSize obtains size in the order of props, formItem, form, globalConfig

Yes, I looked at what the function does so it makes more sense to rely on the parent, which I will adjust

btea:
We should keep `small` first, otherwise it will be a breaking change for users.

webvs2:
> 我们应该保留在第一位，否则对用户来说将是一个重大变化。`small`

I do not understand

btea:
We can add a new `size` prop, but we should not delete `small`. We can follow certain logic to ensure that users who use the `small` attribute are not affected.

kooriookami:
就是新旧api都先留着，旧的api可以在下个大版本中删除

packages/components/table/src/table-body/styles-helper.ts
```diff
@@ -29,11 +29,10 @@ function useStyles<T>(props: Partial<TableBodyProps<T>>) {
     const classes = [ns.e('row')]
     if (
       parent?.props.highlightCurrentRow &&
-      row === props.store.states.currentRow.value
+      ref(row).value === props.store.states.currentRow.value
```
btea, line 32: Why should we change the content of the tabel component here?

webvs2: This is a mistake, originated from some misunderstanding of the response, I am very sorry

packages/components/table/src/table-body/styles-helper.ts
```diff
@@ -33,7 +33,6 @@ function useStyles<T>(props: Partial<TableBodyProps<T>>) {
     ) {
       classes.push('current-row')
     }
-
```
tolking, line 36: This change seems unnecessary

packages/components/pagination/src/pagination.ts
```diff
@@ -186,6 +189,9 @@ export default defineComponent({
     const { t } = useLocale()
     const ns = useNamespace('pagination')
     const vnodeProps = getCurrentInstance()!.vnode.props || {}
+    const _size = useFormSize(
+      computed(() => (props.small ? 'small' : vnodeProps?.size))
+    )
```
tolking, line 194: The Pagination component should not belong to the form component, and I don't quite understand why it is handled this way

tolking (commented):
I think we can also add reminders for attribute changes

```ts
useDeprecated()
```

packages/components/pagination/src/pagination.ts
webvs2: I think 'useFormSize' can be managed uniformly, and this will also work if the user puts it in 'from'

packages/components/pagination/src/pagination.ts
```diff
@@ -186,6 +194,19 @@ export default defineComponent({
...
+    useDeprecated(
+      {
+        from: 'small',
+        replacement: 'link',
```
kooriookami, line 203:
```suggestion
        replacement: 'size',
```

packages/components/pagination/src/pagination.ts
```diff
@@ -429,7 +450,7 @@ export default defineComponent({
             ns.b(),
             ns.is('background', props.background),
             {
-              [ns.m('small')]: props.small,
+              [ns.m(String(_size!.value))]: true,
             },
```
kooriookami, line 454: ns.m(_size.value),

kooriookami (changes requested):

packages/components/pagination/src/pagination.ts
btea: However, it seems that the situation where `pagination` is placed inside `form` will not occur. 🤔

packages/theme-chalk/src/pagination.scss
```diff
@@ -208,6 +208,25 @@
...
+      font-size: getCssVar('pagination-font-size-large');
+      min-width: getCssVar('pagination-button-width-large');
+    }
+
+    span:not([class*='suffix']),
+    button {
+      font-size: getCssVar('pagination-font-size-large');
+    }
+
+    .#{$namespace}-select .#{$namespace}-input {
```
btea, line 226: I think the width here does not need to be changed, keep it consistent with the above.
What do you think? cc @kooriookami @tolking

docs/en-US/component/pagination.md
```diff
@@ -67,7 +67,8 @@ pagination/more-elements
...
 | ----------------------------------- | ------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------- | ------------------------------------ |
-| small                               | whether to use small pagination                                                                                                 | ^[boolean]                                                                        | false                                |
+| small ^(deprecated)                 | whether to use small pagination                                                                                                 | ^[boolean]                                                                        | false                                |
+| size ^(2.7.4)                       | pagination size                                                                                                                 | ^[enum]`'large'\| 'default'\| 'small'`                                            | 'default'                            |
```
btea, line 71: The version number needs to be updated.

packages/components/pagination/src/pagination.ts
btea: It seems that this hasn't changed. @webvs2

docs/en-US/component/pagination.md
webvs2: ok

kooriookami, line 71:
```suggestion
| size ^(2.7.6)                       | pagination size                                                                                                                 | ^[enum]`'large'\| 'default'\| 'small'`                                            | 'default'                            |
```

packages/components/pagination/src/pagination.ts
```diff
@@ -186,6 +194,19 @@ export default defineComponent({
...
     const ns = useNamespace('pagination')
     const vnodeProps = getCurrentInstance()!.vnode.props || {}
+    const _size = useFormSize(
+      computed(() => (props.small ? 'small' : vnodeProps?.size))
+    )
+    useDeprecated(
+      {
+        from: 'small',
+        replacement: 'size',
+        version: '2.7.4',
```
kooriookami, line 204:
```suggestion
        version: '3.0.0',
```

packages/components/pagination/src/pagination.ts
kooriookami, line 454: Please change line 452 - line 454 to `ns.m(_size.value),`.

packages/theme-chalk/src/pagination.scss
kooriookami: I agree.

kooriookami (changes requested):

packages/components/pagination/src/pagination.ts
```diff
@@ -194,9 +194,7 @@ export default defineComponent({
     const { t } = useLocale()
     const ns = useNamespace('pagination')
     const vnodeProps = getCurrentInstance()!.vnode.props || {}
-    const _size = useFormSize(
-      computed(() => (props.small ? 'small' : vnodeProps?.size))
-    )
+    const _size = props.small ? 'small' : vnodeProps?.size
```
kooriookami, line 196: Why remove `useFormSize`?

btea: https://github.com/element-plus/element-plus/pull/16858#discussion_r1621769747

btea: https://github.com/element-plus/element-plus/pull/16858#discussion_r1621769747

kooriookami: Can we replace vnodeProps?.size with prop.size? Why need to use `vnodeProps`?

packages/components/pagination/src/pagination.ts
```diff
@@ -447,7 +447,7 @@ export default defineComponent({
             ns.b(),
             ns.is('background', props.background),
             {
-              [ns.m(_size.value)]: true,
+              [ns.m(_size)]: true,
             },
```
kooriookami, line 451:
```
           ns.b(),
           ns.is('background', props.background),
           ns.m(_size),
```
`ns.m(_size)` is OK. It's no need to use an object.

kooriookami (changes requested):

packages/components/pagination/src/pagination.ts
webvs2: After discussion, it is generally assumed that this component will not take effect in from🤔

kooriookami: > After discussion, it is generally assumed that this component will not take effect in from🤔

Okay~

kooriookami: > Can we replace vnodeProps?.size with prop.size? Why need to use `vnodeProps`?

What about this?

webvs2: > Can we replace vnodeProps?.size with prop.size? Why need to use `vnodeProps`?

可以的，我的失误，我认为两者是用一引用，所以直接拿去了上文，prop上的获取更符合他的作用

packages/theme-chalk/src/pagination.scss
kooriookami: @webvs2 When size is large, use the default font-size.

kooriookami (changes requested):

packages/theme-chalk/src/pagination.scss
webvs2: ok

webvs2: @kooriookami  The 'pagination-font-size-large' field is removed

packages/components/pagination/src/pagination.ts
```diff
@@ -186,6 +193,17 @@ export default defineComponent({
     const { t } = useLocale()
     const ns = useNamespace('pagination')
     const vnodeProps = getCurrentInstance()!.vnode.props || {}
+    const _size = props.small ? 'small' : props?.size
```
kooriookami, line 196: Please make `_size` computed

kooriookami (changes requested):

packages/theme-chalk/src/common/var.scss
```diff
@@ -923,6 +923,8 @@ $pagination: map.merge(
     'font-size-small': 12px,
     'button-width-small': 24px,
     'button-height-small': 24px,
+    'button-width-large': 38px,
+    'button-height-large': 38px,
```
kooriookami, line 927:
```suggestion
    'button-width-large': 40px,
    'button-height-large': 40px,
```

kooriookami (changes requested):

kooriookami:
image
Why font-size is still 16px? It should be the same as default.

webvs2:
> Why font-size is still 16px? It should be the same as default.

image
b本地调试出来的确实是14，是否有缓存嫌疑呢

packages/theme-chalk/src/common/var.scss
webvs2: 其他的large尺寸也是40吗

webvs2:
> > > Why font-size is still 16px? It should be the same as default.
> > 
> > 
> > image b本地调试出来的确实是14，是否有缓存嫌疑呢
> 
> image
> 
> It's weird. I used a new browser. You can see this link.
> 
> https://preview-16858-element-plus.surge.sh/en-US/component/pagination.html#more-elements

我也在线上链接复现了该问题，这确实奇怪，我在本地也运行了docs:dev,仍然没有本地复现，我将排查问题

packages/theme-chalk/src/common/var.scss
kooriookami: The size system is 24px, 32px, 40px. Sometimes you see sizes - 2px dut to the border.

webvs2:
> > > > Why font-size is still 16px? It should be the same as default.
> > > 
> > > 
> > > image b本地调试出来的确实是14，是否有缓存嫌疑呢
> > 
> > 
> > image
> > It's weird. I used a new browser. You can see this link.
> > https://preview-16858-element-plus.surge.sh/en-US/component/pagination.html#more-elements
> 
> 我也在线上链接复现了该问题，这确实奇怪，我在本地也运行了docs:dev,仍然没有本地复现，我将排查问题

@kooriookami  That problem seems to have gone away

kooriookami (approved):

kooriookami:
I think the small size demo can be removed.

image

btea:
Maybe we can add a new size effect display here.
image

kooriookami:
> Maybe we can add a new size effect display here. image

Already have.

btea:
Oh, I just noticed this.

packages/components/pagination/src/pagination.ts
```diff
@@ -186,6 +193,17 @@ export default defineComponent({
     const { t } = useLocale()
     const ns = useNamespace('pagination')
     const vnodeProps = getCurrentInstance()!.vnode.props || {}
+    const _size = computed(() => (props.small ? 'small' : props?.size))
```
btea, line 196: A default value should be given to ensure that the default `class` is consistent with the previous one.
```suggestion
    const _size = computed(() => (props.small ? 'small' : props.size || 'default'))
```

webvs2: I don't think it's necessary to add this, because I observe that the component's style does not extract 'default' to write, so no setting is the default, you think

docs/examples/pagination/more-elements.vue
```diff
@@ -76,15 +78,15 @@
...
+import type { ComponentSize } from 'element-plus'
 const currentPage1 = ref(5)
 const currentPage2 = ref(5)
 const currentPage3 = ref(5)
 const currentPage4 = ref(4)
 const pageSize2 = ref(100)
 const pageSize3 = ref(100)
 const pageSize4 = ref(100)
-const small = ref(false)
+const size = ref<sizeType>('default')
```
btea, line 89:
```suggestion
const size = ref<ComponentSize>('default')
```

packages/components/pagination/src/pagination.ts
btea: Yes, there is indeed no style for `default`, but there is a corresponding class for `default`. I think it's best to keep it consistent with before.
@kooriookami WDYT?

kooriookami: Other components set the default value to '', e.g. `props.size || ''`.

btea: Well, keep it consistent with the other components.

btea (approved):

webvs2:
Thanking Everyone ；

kwezdj123:
为什么不顺便把 `el-config-provider` 全局设置也做了呢，现在全局设置了`size`，`Pagination` 还是要单独设置`size`才会生效。

webvs2:
> 为什么不顺便把 `el-config-provider` 全局设置也做了呢，现在全局设置了`size`，`Pagination` 还是要单独设置`size`才会生效。

之前的改动丢失了这个特性，#17306

PR #16860, opened 2024-05-14
style(eslint-config): add eslint eqeqeq rules and fix about codes

warmthsea:
## Before:
98d4508f75ce04d1717a3e56d7e7564

## After:
6ea5240cfc7b6d2d1a9be89ed1fd333


warmthsea:
image
Image test seems to be a problem.

PR #16863, opened 2024-05-15
test: fix ci image/avatar test error

warmthsea:
#### Edit Before: dev test*70 ✅

image

#### Edit before: probabilistic occurrence (Is the node version affected?) ❗
https://github.com/element-plus/element-plus/pull/16860#issuecomment-2109960857

image

#### Edit After:
✅


warmthsea:
@btea 
Please ask a question, how can I execute ci unit-test again without changing the code?

warmthsea:
new test error `avatar`

image

btea:
> @btea Please ask a question, how can I execute ci unit-test again without changing the code?

Generally, it will only be run when the code is updated, or the project team members can help click re-run.
Regarding the test case you modified, it sometimes fails, but it should be related to the environment, so there is no need to modify it, just rerun it.

warmthsea:
I only changed the test errors that I encountered with a high probability. (CI)

btea (approved):
I ran several tests and everything seemed to be normal. We can merge this PR first and see if there will be no errors in the future.

warmthsea:
> I ran several tests and everything seemed to be normal. We can merge this PR first and see if there will be no errors in the future.

I added 25 duplicate CIs for testing, and 1 or 2 errors occurred each time.

screenshots

btea:
I looked at some of the error conditions, and it seems that they are not a problem with the use case, but an internal problem with jsdom. This random situation can be ignored.

warmthsea:
> I looked at some of the error conditions, and it seems that they are not a problem with the use case, but an internal problem with jsdom. This random situation can be ignored.

OK.

sleepyShen1989:
It seems that this problem still exists
捕获

https://github.com/element-plus/element-plus/actions/runs/9378167810/job/25820936910

PR #16912, opened 2024-05-20
refactor(components): [select] remove index ts-nocheck comments

warmthsea:
## Remove ts-nocheck before:

image

## After:

image


packages/components/select/src/useSelect.ts
```diff
@@ -752,6 +752,28 @@ export const useSelect = (props: ISelectProps, emit) => {
     return Number.parseFloat(style.gap || '6px')
   }
 
+  const handleKeydown = (event: KeyboardEvent) => {
+    switch (event.key) {
+      case 'ArrowDown':
```
btea, line 757: Here you can use the event alias in `aria.ts` to make judgments.
https://github.com/element-plus/element-plus/blob/4c3c81e3d61ec8fba411b6a4a233de28acaec8b8/packages/constants/aria.ts#L1

packages/components/select/src/useSelect.ts
```diff
@@ -752,6 +752,28 @@ export const useSelect = (props: ISelectProps, emit) => {
...
+      case 'ArrowUp':
+        navigateOptions('prev')
+        break
+      case 'Escape':
+        handleEsc()
+        break
+      case 'Enter':
+        selectOption()
+        break
+      case 'Backspace':
```
btea, line 769: `delete` seems to match both keys. [link](https://vuejs.org/guide/essentials/event-handling.html#event-modifiers#:~:text=.delete%20(captures%20both%20%22Delete%22%20and%20%22Backspace%22%20keys))

packages/components/select/src/select.vue
```diff
@@ -152,11 +152,7 @@
                 aria-haspopup="listbox"
                 @focus="handleFocus"
                 @blur="handleBlur"
-                @keydown.down.stop.prevent="navigateOptions('next')"
-                @keydown.up.stop.prevent="navigateOptions('prev')"
-                @keydown.esc.stop.prevent="handleEsc"
-                @keydown.enter.stop.prevent="selectOption"
-                @keydown.delete.stop="deletePrevTag"
+                @keydown.prevent.stop="handleKeydown"
```
btea, line 155: `@keydown.delete.stop` does not seem to have the `prevent` modifier bound, which may change the original logic. 🤔

packages/components/select/src/useSelect.ts
warmthsea: https://github.com/element-plus/element-plus/pull/12200

warmthsea:
I found that deleting ts-nocheck IDE no longer reported errors. (only remove ts-nocheck)😨

image

btea:
Are you using the ts version that comes with vscode or the version installed in the project?

warmthsea:
> Are you using the ts version that comes with vscode or the version installed in the project?

Vscode.
I have never configured anything related to ts version.

screenshots

btea:
You can switch in the lower right corner and select the version installed by the project.

image

warmthsea:
> You can switch in the lower right corner and select the version installed by the project.
> 
> image

I saw it, thanks.

warmthsea:
4.7.4 Seems to conflict with Volar.

image

warmthsea:
> I found that deleting ts-nocheck IDE no longer reported errors. (only remove ts-nocheck)😨
> 
> image

I think this PR, just removing ts-nocheck is also beneficial (This is a gradual optimization process), since tsc is normal, regarding the above issue it may be due to the native environment.

btea:
> 4.7.4 Seems to conflict with Volar.

It should be that the ts version required by the new version of volar is relatively high.

btea:
> I think this PR, just removing ts-nocheck is also beneficial (This is a gradual optimization process), since tsc is normal, regarding the above issue it may be due to the native environment.

I think we should keep the comments for the time being, the plugin may be incompatible with the current version of ts, resulting in the error shown in the picture above.

We can wait until we upgrade the ts version in the future and then gradually remove it.

warmthsea:
> I think we should keep the comments for the time being, the plugin may be incompatible with the current version of ts, resulting in the error shown in the picture above.
> 
> We can wait until we upgrade the ts version in the future and then gradually remove it.

I agree.

PR #16928, opened 2024-05-21
refactor(components): [tooltip-v2] remove `ts-nocheck` comments

warmthsea:
### Before:

screenshots

### After:
[placement.value](https://github.com/element-plus/element-plus/blob/b1bfb9fd1b52265ffd7737f4e9d81e0ac99c17d6/packages/components/tooltip-v2/src/content.ts#L9C1-L22C11)
 `placement.value.split('-')[0]` is [Enum value](https://github.com/element-plus/element-plus/blob/b1bfb9fd1b52265ffd7737f4e9d81e0ac99c17d6/packages/components/tooltip-v2/src/common.ts#L14C1-L19C2) .


warmthsea (commented):
LGTM, I want to gradually remove `@ts-nocheck`.

packages/components/tooltip-v2/src/content.vue
```diff
@@ -65,7 +66,9 @@ const zIndex = useZIndex().nextZIndex()
 const ns = useNamespace('tooltip-v2')
 
 const side = computed(() => {
-  return placement.value.split('-')[0]
+  return placement.value.split(
+    '-'
+  )[0] as typeof TooltipV2Sides[keyof typeof TooltipV2Sides]
```
Dsaquel, line 71: **Suggestion**:
```suggestion
  )[0] as TooltipV2Sides
```
maybe sufficient ?

Dsaquel (approved):
LGTM

PR #16929, opened 2024-05-21
refactor(components): [virtual-list] remove use-wheel ts-nocheck comments

warmthsea:
## Remove ts-nocheck

screenshots


btea (approved):
Thanks!

PR #17009, opened 2024-05-28
style(components): [select & select-v2] remove split dash

kooriookami:

### DIFF
--- a/packages/theme-chalk/src/common/var.scss
+++ b/packages/theme-chalk/src/common/var.scss
@@ -429,7 +429,7 @@ $select-group: () !default;
 $select-group: map.merge(
   (
     'text-color': getCssVar('color-info'),
-    'height': 30px,
+    'height': 34px,
     'font-size': 12px,
   ),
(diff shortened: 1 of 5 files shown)


packages/theme-chalk/src/common/var.scss
```diff
@@ -429,7 +429,7 @@ $select-group: () !default;
 $select-group: map.merge(
   (
     'text-color': getCssVar('color-info'),
-    'height': 30px,
+    'height': 34px,
```
btea, line 432: This seems to be a breaking change, maybe we should mark it somewhere. 🤔

kooriookami: image
This only effect the group title, do you think it's a big change?

btea: I looked at it again and it seems to have no effect. The height set here is used in line 23 below, but the style set here is overwritten by the inline style.

btea (approved):

PR #17035, opened 2024-05-31
fix: upload component ajax error

pzzyf:

### DIFF
--- a/packages/components/upload/src/ajax.ts
+++ b/packages/components/upload/src/ajax.ts
@@ -70,15 +70,11 @@ export const ajaxUpload: UploadRequestHandler = (option) => {
   const formData = new FormData()
   if (option.data) {
     for (const [key, value] of Object.entries(option.data)) {
-      if (isArray(value) && value.length) {
-        for (const val of value) {
-          formData.append(key, val)
-        }
-      } else {
-        formData.append(key, value)
-      }
+      if (isArray(value) && value.length) formData.append(key, value.join(','))
+      else formData.append(key, value)
     }
   }
+
   formData.append(option.filename, option.file, option.file.name)
 
   xhr.addEventListener('error', () => {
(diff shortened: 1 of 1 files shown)


pzzyf:
cc @btea @FrontEndDog

FrontEndDog (approved):
LGTM!

packages/components/upload/src/ajax.ts
```diff
@@ -70,8 +70,13 @@ export const ajaxUpload: UploadRequestHandler = (option) => {
   const formData = new FormData()
   if (option.data) {
     for (const [key, value] of Object.entries(option.data)) {
-      if (isArray(value) && value.length) formData.append(key, ...value)
-      else formData.append(key, value)
+      if (isArray(value) && value.length) {
+        for (const val of value) {
+          formData.append(key, val)
+        }
```
btea, line 76: The value type of `data` has no restrictions. If the value of a certain attribute is special, then this modification may cause the original behavior that is inconsistent.
```ts
// for example
const blob = new Blob(['text'], { type: 'text/plain' })
const data = {
  test: [blob, 'testname']
}
// origin 
formData.append('test', ...[data.test])

// changed
formData.append('test', data.test[0])
formData.append('test', data.test[1])
```
The above two writing effects seem to be inconsistent. 🤔

pzzyf: you are right, value type is string | Blob, formData object value type is string or Blob
image
image

packages/components/upload/src/ajax.ts
```diff
@@ -70,10 +70,11 @@ export const ajaxUpload: UploadRequestHandler = (option) => {
   const formData = new FormData()
   if (option.data) {
     for (const [key, value] of Object.entries(option.data)) {
-      if (isArray(value) && value.length) formData.append(key, ...value)
+      if (isArray(value) && value.length) formData.append(key, value.join(','))
```
btea, line 73: It seems that the stitching of multiple values ​​into one value also changes the original logic.

PR #17041, opened 2024-05-31
feat(components): [skeleton] `throttle` supports values ​​as object

### LINKED ISSUE #4804: [Bug Report] el-skeleton's :throttle propertity is useless
### Element Plus version
1.2.0-beta.6

### OS/Browsers version
Chrome

### Vue version
3.2.26

### Reproduction Link
https://codesandbox.io/s/element-plus-spa-forked-l0ptw?file=/src/App.vue:109-118

### Steps to reproduce
Just Look

### What is Expected?
The skeleton screen appears continuously according to the throttle time

If the data request is fast, the skeleton screen will flash, and after the throttle is set, the skeleton screen will not be displayed

chenweiyi:

### Solving the problem

When using the `throttle` property of skeleton in the first screen scene, it does not take effect because the current `throttle` property only deals with the *delay display* issue and does not solve the *delay hide* issue. Therefore, I have extended this attribute value to support configuring the *delay hide* attribute.

Now throttle support config object like `{ trailing: 500 }`

### Relate issue
Through config `{ trailing: xxx, initVal: true }`   fix #4804  

### Rendering

image


packages/hooks/use-throttle-render/index.ts
```diff
@@ -1,29 +1,60 @@
...
+    if (
+      typeof throttle === 'object' &&
+      throttle.trailing &&
+      throttle.trailing > 0
+    ) {
+      dispatchThrottling(throttle.trailing)
+    } else {
+      throttled.value = false
+    }
+  }
```
btea, line 47: There is a lot of overlap between these two functions, and perhaps they can be combined into one function.

Some type determination operations can directly reference internal encapsulated tool functions.
https://github.com/element-plus/element-plus/blob/72dd11aac271e788da45e0e92b3ad3dd343c1601/packages/utils/types.ts#L18

chenweiyi: I'm not quite clear about what you mean. The `use-throttle-render/index.ts` mainly exposes a throttle hook, and `utlils/types.ts` contains mostly pure function logic. These two can't be simply merged into one, can they? Could you please elaborate a bit more, and then I can see what I might be able to do?

btea: Sorry, it looks like my description was a bit confusing.

 That some of the code logic of function `leadingDispatch` and function `trailingDispatch` is the same. These two functions can be merged into one function by passing a Boolean value to distinguish the different logics.

The second sentence describes that code like `typeof throttle === 'number'` can be processed by referencing the internal encapsulated function and replaced with `isNumber(throttle)`.

chenweiyi: Thank you for your patient explanation, I have made a revision, please review it again.

packages/hooks/use-throttle-render/index.ts
```diff
@@ -1,31 +1,58 @@
...
+      if (isNumber(throttle)) {
+        dispatchThrottling(throttle)
+      } else {
+        if (throttle?.leading && throttle.leading > 0) {
+          dispatchThrottling(throttle.leading!)
+        }
+      }
+    } else {
+      if (
+        typeof throttle === 'object' &&
```
btea, line 37:
```suggestion
        isObject(throttle) &&
```

docs/en-US/component/skeleton.md
```diff
@@ -77,7 +77,7 @@ skeleton/rendering-with-data
 
 ## Avoiding rendering bouncing.
 
-Sometimes API responds very quickly, when that happens, the skeleton just gets rendered to the DOM then it needs to switch back to real DOM, that causes the sudden flashy. To avoid such thing, you can use the `throttle` attribute.
+Sometimes API responds very quickly, when that happens, the skeleton just gets rendered to the DOM then it needs to switch back to real DOM, that causes the sudden flashy. To avoid such thing, you can use the `throttle` attribute. The `throttle` attribute supports two values: `number` and `object`. When passing a `number`, it is equivalent to `{leading: xxx}`, controlling the throttling of the skeleton screen display. Of course, you can also control the throttling of the skeleton screen disappearance by passing `{trailing: xxx}`. When the initial value of loading is true, you can set `{initVal: true}` to control the immediate display of the initial skeleton screen without throttling.
```
btea, line 80: The content here may be created as a separate subheading, and the supported version number can be conveniently marked.

btea:
Right now, the content appears and disappears a bit abruptly. Maybe we can add a transition effect to make it look softer.
WDYT？@chenweiyi

docs/en-US/component/skeleton.md
chenweiyi: ok

chenweiyi:
> Right now, the content appears and disappears a bit abruptly. Maybe we can add a transition effect to make it look softer. WDYT？@chenweiyi

I create two sperate section for describle the new function. Please take a look at.

packages/hooks/use-throttle-render/index.ts
```diff
@@ -1,31 +1,53 @@
...
 import type { Ref } from 'vue'
 
-export const useThrottleRender = (loading: Ref<boolean>, throttle = 0) => {
+export const useThrottleRender = (
+  loading: Ref<boolean>,
+  throttle: SkeletonThrottle = 0
+) => {
   if (throttle === 0) return loading
-  const throttled = ref(false)
+  const initVal = isObject(throttle) ? throttle.initVal ?? false : false
```
btea, line 12: It seems that there is no need for false to fall back here.
```suggestion
  const initVal = isObject(throttle) ? throttle.initVal : false
```

packages/hooks/use-throttle-render/index.ts
```diff
@@ -1,31 +1,53 @@
...
+    if (type === 'leading') {
+      if (isNumber(throttle)) {
+        dispatchThrottling(throttle)
+      } else {
+        if (throttle?.leading && throttle.leading > 0) {
+          dispatchThrottling(throttle.leading!)
+        }
+      }
+    } else {
+      if (isObject(throttle) && throttle?.trailing && throttle.trailing > 0) {
```
btea, line 35:
```suggestion
      if (isObject(throttle)) {
```

packages/hooks/use-throttle-render/index.ts
```diff
@@ -1,31 +1,53 @@
...
+    }, timer)
   }
-  onMounted(dispatchThrottling)
+
+  const dispatcher = (type: 'leading' | 'trailing') => {
+    if (type === 'leading') {
+      if (isNumber(throttle)) {
+        dispatchThrottling(throttle)
+      } else {
+        if (throttle?.leading && throttle.leading > 0) {
```
btea, line 30: The judgment here can be removed, otherwise, there will be problems when it is set to 0.

docs/en-US/component/skeleton.md
```diff
@@ -79,23 +79,57 @@ skeleton/rendering-with-data
...
+
+:::
+
 :::demo
 
 skeleton/avoiding-rendering-bouncing
 
 :::
 
+## Initial rendering loading
```
btea, line 94:
```suggestion
## Initial rendering loading ^(2.8.7)
```

docs/en-US/component/skeleton.md
```diff
@@ -79,23 +79,57 @@ skeleton/rendering-with-data
...
+
+When the initial value of loading is true, you can set `throttle: {initVal: true, leading: xxx}` to control the immediate display of the initial skeleton screen without throttling.
+
+:::demo
+
+skeleton/initial-rendering-loading
+
+:::
+
+## Toggle show/hide without rending bouncing
```
btea, line 104:
```suggestion
## Toggle show/hide without rending bouncing ^(2.8.7)
```

packages/hooks/use-throttle-render/index.ts
chenweiyi: If remove `throttle.initVal ?? false` , `useThrottleRender` will return `ref<boolean | undefined>`, But I think `useThrottleRender` of return `ref<boolean>` express more clear meaning.

image

image

packages/hooks/use-throttle-render/index.ts
chenweiyi: I think `isObject(throttle) && throttle?.trailing && throttle.trailing > 0` can be changed to `Object(throttle) && throttle?.trailing`. The judgment of `throttle.trailing > 0` is essential, otherwise the parameter `timer` in `dispatchThrottling` may throw an error in `setTimeout`.

packages/hooks/use-throttle-render/index.ts
chenweiyi: Here `isNumber(throttle)` is for compatibility with the previous logic. `throttle?.leading && throttle.leading > 0` is here to determine that the `throttle` passed in by the user is an object, but the `leading` attribute is not passed. Here I think the judgment of `throttle.leading > 0` can also be removed.

btea: Yes, that's what I meant.

packages/hooks/use-throttle-render/index.ts
btea: The second parameter of `setTimeout` will not throw an error even if it accepts `undefined` or a negative number. Of course, you can make a more accurate judgment, but we need to make a backup. When the conditions are not met, perhaps we should give a default value of `0`.

packages/hooks/use-throttle-render/index.ts
chenweiyi: I have thought about modifying it to this `const initVal = isObject(throttle) && Boolean(throttle.initVal)`

packages/hooks/use-throttle-render/index.ts
chenweiyi: I have thought about modifying it to this `const initVal = isObject(throttle) && Boolean(throttle.initVal)`

packages/hooks/use-throttle-render/index.ts
chenweiyi: I have modified the `dispatcher` method again：
```
const dispatcher = (type: 'leading' | 'trailing') => {
    if (type === 'leading') {
      if (isNumber(throttle)) {
        dispatchThrottling(throttle)
      } else {
        dispatchThrottling(throttle.leading)
      }
    } else {
      if (isObject(throttle)) {
        dispatchThrottling(throttle.trailing)
      } else {
        throttled.value = false
      }
    }
  }
```
Then `dispatchThrottling` was modified, adding handling for the `undefined` case as well ：
```
const dispatchThrottling = (timer: number | undefined) => {
    if (isUndefined(timer)) {
      throttled.value = loading.value
      return
    }
    if (timeoutHandle) {
      clearTimeout(timeoutHandle)
    }
    timeoutHandle = setTimeout(() => {
      throttled.value = loading.value
    }, timer)
  }
```

btea (approved):
Thanks!

chenweiyi:
> Thanks!

Thank you for reviewing my code very patiently. I am glad I can contribute.

docs/en-US/component/skeleton.md
```diff
@@ -79,23 +79,57 @@ skeleton/rendering-with-data
...
+
+:::
+
 :::demo
 
 skeleton/avoiding-rendering-bouncing
 
 :::
 
+## Initial rendering loading ^(2.8.7)
```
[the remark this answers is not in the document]
chenweiyi: Ok, Thank you!

docs/en-US/component/skeleton.md
```diff
@@ -79,23 +79,57 @@ skeleton/rendering-with-data
...
+
+When the initial value of loading is true, you can set `throttle: {initVal: true, leading: xxx}` to control the immediate display of the initial skeleton screen without throttling.
+
+:::demo
+
+skeleton/initial-rendering-loading
+
+:::
+
+## Toggle show/hide without rending bouncing ^(2.8.7)
```
btea, line 104:
```suggestion
## Toggle show/hide without rending bouncing ^(2.8.8)
```

packages/hooks/use-throttle-render/index.ts
```diff
@@ -1,31 +1,55 @@
 import { onMounted, ref, watch } from 'vue'
+import { isNumber, isObject, isUndefined } from '@element-plus/utils'
 
+import type { SkeletonThrottle } from '@element-plus/components'
```
tolking, line 4: Avoid import components content in hooks. It is recommended to export the `SkeletonThrottle` type here.

docs/en-US/component/skeleton.md
```diff
@@ -79,23 +79,57 @@ skeleton/rendering-with-data
 
 Sometimes API responds very quickly, when that happens, the skeleton just gets rendered to the DOM then it needs to switch back to real DOM, that causes the sudden flashy. To avoid such thing, you can use the `throttle` attribute.
 
+:::tip
+
+The `throttle` attribute supports two values: `number` and `object`. When passing a `number`, it is equivalent to `{leading: xxx}`, controlling the throttling of the skeleton screen display. Of course, you can also control the throttling of the skeleton screen disappearance by passing `{trailing: xxx}`
```
tolking, line 84: It would be better to add a version hint here. 

eg. Since ^(2.8.8), ...

tolking (commented):
Could you add some test cases for useThrottleRender?

chenweiyi:
> Could you add some test cases for useThrottleRender?

I added 2 test cases, thank you for your review.

packages/hooks/use-throttle-render/index.ts
tolking: This issue does not seem to be resolved

Suggest moving the type declaration of `SkeletonThrottle` to here

chenweiyi: Okay, I understand what you mean, I will also change the `SkeletonThrottle` field to `ThrottleType` to increase generality.

tolking (approved):

PR #17050, opened 2024-06-02
style(components): [button] keep default hover color of link unchanged

### LINKED ISSUE #16979: [Component] [button] Button link variant has invalid color
Bug Type: **`Component`**

## Environment

- Vue Version: `3.4.27`
- Element Plus Version: `2.7.3`
- Browser / OS: `Mac/Chrome`
- Build Tool: `Vite`

### Related Component

- `el-button`

### Reproduction Link

[Github Repo](https://github.com/element-plus/element-plus/blob/4c3c81e3d61ec8fba411b6a4a233de28acaec8b8/packages/theme-chalk/src/common/var.scss#L698)

### Steps to reproduce

Button with "link" type is using "info" color, so in case you have customized "primary-color" this color is ignored and button is still in blue color

### What is Expected?

Use primary color

Use info color

### Additional comments

_(empty)_

btea:

fix #16979

**breaking change**

When the button `type` is not set, the text color of the hover state is the color corresponding to the referenced `info` type before modification. It will not change according to the `info` type after modification.

|     before   |     after    |
|------------|-----------|
| image| image|


packages/theme-chalk/src/common/var.scss
```diff
@@ -695,7 +695,7 @@ $button: map.merge(
     'active-border-color': getCssVar('color-primary'),
     'active-bg-color': getCssVar('button', 'hover-bg-color'),
     'outline-color': getCssVar('color-primary', 'light-5'),
-    'hover-link-text-color': getCssVar('color-info'),
+    'hover-link-text-color': #909399,
```
ryuhangyeong, line 698: I am not sure exactly what the problem is. Can you explain?

The problem seems to be that variables cannot be redefined. But why do you delete the existing variable and put it as a constant?

btea: According to the discussion of #16979, in the default theme, when `el-button` is set to link but not type, the hover state text color uses the text color of type info, which is normal.

However, if the theme color is changed, if the color corresponding to info is set to blue when the type of `el-button` is not specified, the hover state text color becomes blue, which does not seem reasonable.

ryuhangyeong (approved):
I agree with the changes, but I think they need buy-in from more people because the impact is so large.

ryuhangyeong:
@element-plus/backers

warmthsea (commented):
I don't think this is a good idea, all hover colors have changed.

screenshots

btea:
It's been so long that I've almost forgotten it. Are you saying that if the link is set after the modification, all hover styles will have problems?

warmthsea:
I re-tested and it seems to be ok.

Before:

screenshots

After: 

screenshots

warmthsea:
But the default button is also the theme color

screenshots

tolking:
IMO, Using a fixed color may not be user-friendly.

I found that in the [old UI design](https://www.figma.com/design/h6XpYRCR0zX5EjSOC14kuV/Element-Plus-Design-System(Components)-(Community)?node-id=11790-149533&t=OcP3cj7htcW3hoir-1), the color of the link button hover state is `text-color-secondary`. Maybe we can change it.

截屏2024-11-13 14 43 03

tolking (approved):

warmthsea (approved):

PR #17066, opened 2024-06-03
feat(components): [image] add empty value contral

### LINKED ISSUE #17059: [Component] [image] el-image当src=null时不会触发error
Bug Type: **`Component`**

## Environment

- Vue Version: `3.4.27`
- Element Plus Version: `2.7.4`
- Browser / OS: `Chrome 124.0.6367.119 / Windows NT 10`
- Build Tool: `Vite`

### Related Component

- `el-image`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJzcmMvQXBwLnZ1ZSI6IjxzY3JpcHQgc2V0dXAgbGFuZz1cInRzXCI+XG5pbXBvcnQgeyByZWYsIHZlcnNpb24gYXMgdnVlVmVyc2lvbiB9IGZyb20gJ3Z1ZSdcblxuaW1wb3J0IHsgUGljdHVyZSBhcyBJY29uUGljdHVyZSB9IGZyb20gJ0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlJ1xuXG5jb25zdCBzcmM9cmVmKG51bGwpXG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8ZGl2IGNsYXNzPVwiZGVtby1pbWFnZV9fZXJyb3JcIj5cbiAgICA8ZGl2IGNsYXNzPVwiYmxvY2tcIj5cbiAgICAgIDxzcGFuIGNsYXNzPVwiZGVtb25zdHJhdGlvblwiPkN1c3RvbTwvc3Bhbj5cbiAgICAgIDxlbC1pbWFnZT5cbiAgICAgICAgPHRlbXBsYXRlICNlcnJvcj5cbiAgICAgICAgICA8ZGl2IGNsYXNzPVwiaW1hZ2Utc2xvdFwiPlxuICAgICAgICAgICAgPGVsLWljb24+PGljb24tcGljdHVyZSAvPjwvZWwtaWNvbj5cbiAgICAgICAgICA8L2Rpdj5cbiAgICAgICAgPC90ZW1wbGF0ZT5cbiAgICAgIDwvZWwtaW1hZ2U+XG4gICAgPC9kaXY+XG4gICAgPGRpdiBjbGFzcz1cImJsb2NrXCI+XG4gICAgICA8c3BhbiBjbGFzcz1cImRlbW9uc3RyYXRpb25cIj5DdXN0b208L3NwYW4+XG4gICAgICA8ZWwtaW1hZ2UgOnNyYz1cInNyY1wiPlxuICAgICAgICA8dGVtcGxhdGUgI2Vycm9yPlxuICAgICAgICAgIDxkaXYgY2xhc3M9XCJpbWFnZS1zbG90XCI+XG4gICAgICAgICAgICA8ZWwtaWNvbj48aWNvbi1waWN0dXJlIC8+PC9lbC1pY29uPlxuICAgICAgICAgIDwvZGl2PlxuICAgICAgICA8L3RlbXBsYXRlPlxuICAgICAgPC9lbC1pbWFnZT5cbiAgICA8L2Rpdj5cbiAgPC9kaXY+XG48L3RlbXBsYXRlPlxuXG5cbjxzdHlsZSBzY29wZWQ+XG4uZGVtby1pbWFnZV9fZXJyb3IgLmJsb2NrIHtcbiAgcGFkZGluZzogMzBweCAwO1xuICB0ZXh0LWFsaWduOiBjZW50ZXI7XG4gIGJvcmRlci1yaWdodDogc29saWQgMXB4IHZhcigtLWVsLWJvcmRlci1jb2xvcik7XG4gIGRpc3BsYXk6IGlubGluZS1ibG9jaztcbiAgd2lkdGg6IDQ5JTtcbiAgYm94LXNpemluZzogYm9yZGVyLWJveDtcbiAgdmVydGljYWwtYWxpZ246IHRvcDtcbn1cbi5kZW1vLWltYWdlX19lcnJvciAuZGVtb25zdHJhdGlvbiB7XG4gIGRpc3BsYXk6IGJsb2NrO1xuICBjb2xvcjogdmFyKC0tZWwtdGV4dC1jb2xvci1zZWNvbmRhcnkpO1xuICBmb250LXNpemU6IDE0cHg7XG4gIG1hcmdpbi1ib3R0b206IDIwcHg7XG59XG4uZGVtby1pbWFnZV9fZXJyb3IgLmVsLWltYWdlIHtcbiAgcGFkZGluZzogMCA1cHg7XG4gIG1heC13aWR0aDogMzAwcHg7XG4gIG1heC1oZWlnaHQ6IDIwMHB4O1xuICB3aWR0aDogMTAwJTtcbiAgaGVpZ2h0OiAyMDBweDtcbn1cblxuLmRlbW8taW1hZ2VfX2Vycm9yIC5pbWFnZS1zbG90IHtcbiAgZGlzcGxheTogZmxleDtcbiAganVzdGlmeS1jb250ZW50OiBjZW50ZXI7XG4gIGFsaWduLWl0ZW1zOiBjZW50ZXI7XG4gIHdpZHRoOiAxMDAlO1xuICBoZWlnaHQ6IDEwMCU7XG4gIGJhY2tncm91bmQ6IHZhcigtLWVsLWZpbGwtY29sb3ItbGlnaHQpO1xuICBjb2xvcjogdmFyKC0tZWwtdGV4dC1jb2xvci1zZWNvbmRhcnkpO1xuICBmb250LXNpemU6IDMwcHg7XG59XG4uZGVtby1pbWFnZV9fZXJyb3IgLmltYWdlLXNsb3QgLmVsLWljb24ge1xuICBmb250LXNpemU6IDMwcHg7XG59XG48L3N0eWxlPlxuIiwiaW1wb3J0LW1hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge31cbn0iLCJ0c2NvbmZpZy5qc29uIjoie1xuICBcImNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogXCJFU05leHRcIixcbiAgICBcImpzeFwiOiBcInByZXNlcnZlXCIsXG4gICAgXCJtb2R1bGVcIjogXCJFU05leHRcIixcbiAgICBcIm1vZHVsZVJlc29sdXRpb25cIjogXCJCdW5kbGVyXCIsXG4gICAgXCJ0eXBlc1wiOiBbXCJlbGVtZW50LXBsdXMvZ2xvYmFsLmQudHNcIl0sXG4gICAgXCJhbGxvd0ltcG9ydGluZ1RzRXh0ZW5zaW9uc1wiOiB0cnVlLFxuICAgIFwiYWxsb3dKc1wiOiB0cnVlLFxuICAgIFwiY2hlY2tKc1wiOiB0cnVlXG4gIH0sXG4gIFwidnVlQ29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiAzLjNcbiAgfVxufVxuIiwiX28iOnt9fQ==)

### Steps to reproduce

```vue
<script setup lang="ts">
import { ref, version as vueVersion } from 'vue'

import { Picture as IconPicture } from '@element-plus/icons-vue'

const src=ref(null)
</script>

<template>
  <div class="demo-image__error">
    <div class="block">
      <span class="demonstration">Custom</span>
      <el-image>
        <template #error>
          <div class="image-slot">
            <el-icon><icon-picture /></el-icon>
          </div>
        </template>
      </el-image>
    </div>
    <div class="block">
      <span class="demonstration">Custom</span>
      <el-image :src="src">
        <template #error>
          <div class="image-slot">
            <el-icon><icon-picture /></el-icon>
          </div>
        </template>
      </el-image>
    </div>
  </div>
</template>


<style scoped>
.demo-image__error .block {
  padding: 30px 0;
  text-align: center;
  border-right: solid 1px var(--el-border-color);
  display: inline-block;
  width: 49%;
  box-sizing: border-box;
  vertical-align: top;
}
.demo-image__error .demonstration {
  display: block;
  color: var(--el-text-color-secondary);
  font-size: 14px;
  margin-bottom: 20px;
}
.demo-image__error .el-image {
  padding: 0 5px;
  max-width: 300px;
  max-height: 200px;
  width: 100%;
  height: 200px;
}

.demo-image__error .image-slot {
  display: flex;
  justify-content: center;
  align-items: center;
  width: 100%;
  height: 100%;
  background: var(--el-fill-color-light);
  color: var(--el-text-color-secondary);
  font-size: 30px;
}
.demo-image__error .image-slot .el-icon {
  font-size: 30px;
}
</style>
```

### What is Expected?

当src为null时，判断加载失败，加载error slot中的内容

当src为null时，只会显示空白。
不会判断加载失败，不会加载error slot中的内容。

### Additional comments

_(empty)_

warmthsea:
closed https://github.com/element-plus/element-plus/issues/17059, I think this problem should be fixed, even if `img` value is `null` not trigger `@error`

TS user:

image

JS user null No error message.

<hr />

If src value is `null` `undefined`,  control the value of slot display, `hasLoadError.value` directly `true`.

before:

image

after:

image

<hr />

😥 The newly written test case runs without errors in the old code, but the user is actually wrong.

``` vitest
test('image value is null error test', async () => {
    const src = ref(null)
    const wrapper = mount({
      setup() {
        return () => (
          <Image v-model={src.value}>{{ error: () => 'fallback' }}</Image>
        )
      },
    })
    await doubleWait()
    wrapper.emitted('error') && expect(wrapper.emitted('error')).toBeDefined()
    expect(wrapper.find('.el-image__inner').exists()).toBe(false)
    expect(wrapper.find('img').exists()).toBe(false)
    expect(wrapper.text()).toBe('fallback')
})

```

| Tew Test | Test Results | But should |
|-------|-------|-------|
| edit before |  ✅😰    | ❌↘️    |
| edit after   |  ✅    | ✅   |

before:

image

### DIFF
--- a/packages/components/image/__tests__/image.test.tsx
+++ b/packages/components/image/__tests__/image.test.tsx
@@ -62,24 +62,6 @@ describe('Image.vue', () => {
     expect(wrapper.find('.el-image__error').exists()).toBe(true)
   })
 
-  // js user null No error message
-  test('image value is null error test', async () => {
-    const wrapper = mount(
-      <Image
-        src={null}
-        v-slots={{
-          error: () => 'error text',
-        }}
-      />
-    )
-
-    await doubleWait()
-    wrapper.emitted('error') && expect(wrapper.emitted('error')).toBeDefined()
-    expect(wrapper.find('.el-image__inner').exists()).toBe(false)
-    expect(wrapper.find('img').exists()).toBe(false)
-    expect(wrapper.text()).toContain('error text')
-  })
-
   test('image load sequence success test', async () => {
     const wrapper = mount(Image, {
       props: {
--- a/packages/components/image/__tests__/image.test.tsx
+++ b/packages/components/image/__tests__/image.test.tsx
@@ -62,6 +62,24 @@ describe('Image.vue', () => {
     expect(wrapper.find('.el-image__error').exists()).toBe(true)
   })
 
+  // js user null No error message
+  test('image value is null error test', async () => {
+    const wrapper = mount(
+      <Image
+        src={null}
+        v-slots={{
+          error: () => 'error text',
+        }}
+      />
+    )
+
+    await doubleWait()
+    wrapper.emitted('error') && expect(wrapper.emitted('error')).toBeDefined()
+    expect(wrapper.find('.el-image__inner').exists()).toBe(false)
+    expect(wrapper.find('img').exists()).toBe(false)
+    expect(wrapper.text()).toContain('error text')
+  })
+
   test('image load sequence success test', async () => {
     const wrapper = mount(Image, {
       props: {
--- a/packages/components/image/src/image.vue
+++ b/packages/components/image/src/image.vue
@@ -135,8 +135,8 @@ const loadImage = () => {
 
   // reset status
   isLoading.value = true
-  hasLoadError.value = false
-  imageSrc.value = props.src
+  hasLoadError.value = !props.src
+  imageSrc.value = props.src || undefined
 }
 
 function handleLoad(event: Event) {
--- a/packages/components/image/src/image.vue
+++ b/packages/components/image/src/image.vue
@@ -5,6 +5,7 @@
     </slot>
     <template v-else>
       <img
+        v-if="imageSrc !== undefined"
         v-bind="attrs"
         :src="imageSrc"
         :loading="loading"
@@ -83,7 +84,7 @@ const ns = useNamespace('image')
 const rawAttrs = useRawAttrs()
 const attrs = useAttrs()
 
-const imageSrc = ref('')
+const imageSrc = ref<string | undefined>()
 const hasLoadError = ref(false)
 const isLoading = ref(true)
 const showViewer = ref(false)
--- a/packages/components/image/src/image.vue
+++ b/packages/components/image/src/image.vue
@@ -139,7 +139,9 @@ const loadImage = () => {
   hasLoadError.value = false
 
   isEmpty.value = !props.src
-  imageSrc.value = props.src
+  if (props.src) {
+    imageSrc.value = props.src
+  }
 }
 
 function handleLoad(event: Event) {
--- a/packages/components/image/src/image.vue
+++ b/packages/components/image/src/image.vue
@@ -5,7 +5,6 @@
     </slot>
     <template v-else>
       <img
-        v-if="imageSrc !== undefined"
         v-bind="attrs"
         :src="imageSrc"
         :loading="loading"
@@ -84,7 +83,7 @@ const ns = useNamespace('image')
 const rawAttrs = useRawAttrs()
 const attrs = useAttrs()
 
-const imageSrc = ref<string | undefined>()
+const imageSrc = ref('')
 const hasLoadError = ref(false)
 const isLoading = ref(true)
 const showViewer = ref(false)
@@ -135,8 +134,8 @@ const loadImage = () => {
 
   // reset status
   isLoading.value = true
-  hasLoadError.value = !props.src
-  imageSrc.value = props.src || undefined
+  hasLoadError.value = false
+  imageSrc.value = props.src === null ? '' : props.src
 }
 
 function handleLoad(event: Event) {
--- a/packages/components/image/src/image.vue
+++ b/packages/components/image/src/image.vue
@@ -1,6 +1,6 @@
 <template>
   <div ref="container" :class="[ns.b(), $attrs.class]" :style="containerStyle">
-    <slot v-if="hasLoadError" name="error">
+    <slot v-if="hasLoadError || isEmpty" name="error">
       <div :class="ns.e('error')">{{ t('el.image.error') }}</div>
     </slot>
     <template v-else>
@@ -85,6 +85,7 @@ const rawAttrs = useRawAttrs()
 const attrs = useAttrs()
 
 const imageSrc = ref<string | undefined>()
+const isEmpty = ref(false)
 const hasLoadError = ref(false)
 const isLoading = ref(true)
 const showViewer = ref(false)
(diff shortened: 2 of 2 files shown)


chenxch:
I think in this case, you should consider using an `empty` state `slot` that triggers the error event when there is `null`. This way, it will prevent rendering the data from the request and directly trigger the error operation set by the user.

warmthsea:
> I think in this case, you should consider using an `empty` state `slot` that triggers the error event when there is `null`. This way, it will prevent rendering the data from the request and directly trigger the error operation set by the user.

Without rendering, I can't get the `Event ` parameter (`emit('error', event)`).

chenxch:
> > I think in this case, you should consider using an `empty` state `slot` that triggers the error event when there is `null`. This way, it will prevent rendering the data from the request and directly trigger the error operation set by the user.
> 
> Without rendering, I can't get the `Event ` parameter (`emit('error', event)`).

Yes, when the value is null or '', it will not be rendered, so the current behavior is correct.

chenxch (approved):
LGTM.

btea:
Seems similar to the `placeholder` slot in some ways. 🤔

btea:
IMO, this is not a bug and may not need to be fixed. Directly binding a null value to the src of the img tag in Vue will not trigger a request.

warmthsea:
> IMO, this is not a bug and may not need to be fixed. Directly binding a null value to the src of the img tag in Vue will not trigger a request.

I saw the vue/core PR.

btea:
@warmthsea So, we should keep it like this and let users handle it on their own. And, as the comments over there said, this seems to be a breaking change for previous users.

warmthsea:
> @warmthsea So, we should keep it like this and let users handle it on their own. And, as the comments over there said, this seems to be a breaking change for previous users.

Agreed, that's it for the time being.

PR #17067, opened 2024-06-03
docs(components): [image] change `@error` parameter types

warmthsea:
https://github.com/element-plus/element-plus/blob/03cd828969e4273931f3fee33a362a83a715d291/packages/components/image/src/image.vue#L148C1-L152C2

image


btea (approved):
Thanks!

PR #17079, opened 2024-06-04
feat(components):  [autocomplete] `popper-class` support object

Heising:


btea:
Hi, thanks for your contribution. It seems that supporting strings is sufficient for now, could you explain the reason for your change?

Heising:
> Hi, thanks for your contribution. It seems that supporting strings is sufficient for now, could you explain the reason for your change?

need to dynamically change the CSS of the dropdown menu, this is a feature that Vue already has, and you should not disable it just by using strings

PR #17099, opened 2024-06-06
docs: change cdn online demo link to correct show result

warmthsea:
### Before:

image

### After:

image


btea (approved):
Thank you ! 👍

PR #17106, opened 2024-06-06
fix(components): [pagination] assertValidUsage warning

sujianqingfeng:
fix pagination assertValidUsage warning


tolking:

```
WARNING

We'll detect some deprecated usages, if your pagination don't appeared or worked as expected, please check rules below:

You have to define one of total and page-count, otherwise we can't determine count of total pages.When both defined, page-count taken as priority.
If current-page is defined, you have to listen current-page change, by also define @update:current-page, otherwise pagination didn't work.
If page-size is defined while page size selector displayed(sizes included in layout), you have to listen page-size change as well, by define @update:page-size, otherwise change of page size didn't work.
```

Based on the document's description, the current assessment is correct; I don't understand why this modification is necessary.

PR #17115, opened 2024-06-07
add installation

Euraxluo:


chenxch:
https://element-plus.org/en-US/guide/installation.html#using-package-manager

PR #17150, opened 2024-06-11
feat(components): improve event and supplement docs

warmthsea:
### ❓ Why:

- When I use `el-date-pick`, there is no `@clear` event when clearing the value. I can only listen through `watch`, which is not very convenient.
- I also found that some things are not documented and marked.

#### ` el-autocomplete `

 - [supplement docs](https://github.com/element-plus/element-plus/blob/8bd642f567a027058b543f36cc688e01785591b5/packages/components/autocomplete/src/autocomplete.ts#L141C1-L149C2): 
   - input
   - blur
   - focus
   - clear

#### `el-cascader`, `el-date-picker`, `el-datetime-picker`,  `el-time-picker`, `el-time-select`
- add event:
  - clear


packages/components/color-picker/src/color-picker.vue
```diff
@@ -320,10 +321,12 @@ function clear() {
   debounceSetShowPicker(false)
   emit(UPDATE_MODEL_EVENT, null)
   emit('change', null)
+  emit('clear')
   if (props.modelValue !== null && props.validateEvent) {
     formItem?.validate('change').catch((err) => debugWarn(err))
   }
   resetColor()
+  emit('clear')
```
btea, line 329: Why is it executed twice here?

packages/components/color-picker/src/color-picker.vue
```diff
@@ -298,6 +298,7 @@ function handleConfirm() {
 function confirmValue() {
   const value = color.value
   emit(UPDATE_MODEL_EVENT, value)
+  emit('confirm', value)
```
btea, line 301: Seems to be duplicated with `confirm`.

btea (approved):
LGTM! Thanks!

kooriookami (approved):

PR #17170, opened 2024-06-12
style(components): prevent Dialog width from over the screen

aiokr:

---

When the screen width is smaller than the width of the Dialog (common in mobile phones), the Dialog will exceed the screen, causing the close button in the upper right corner and the action button in the lower right corner to be not directly visible.
I added the max-width: 100vw; property to prevent the Dialog width from exceeding the viewport width.


tolking:

The default width of the Dialog component is 50%, which theoretically should not exceed the screen width. If users configure the width themselves, it is better to prioritize their own configuration, as there may be situations where scrolling is required due to excessive content.

PR #17172, opened 2024-06-13
chore(vscode): set `md` defaultFormatter to prettier

warmthsea:
about: https://github.com/element-plus/element-plus/pull/16364

After setting, other contents do not need to be changed, because the original style is to use `prettier`.

screenshots


btea (approved):
Thanks!

PR #17181, opened 2024-06-13
docs(components): [autocomplete] fetch-suggestions revise type

Heising:
type: definePropType<AutocompleteFetchSuggestions>([Function, Array])
source code array type


btea (approved):
Thanks!

PR #17245, opened 2024-06-17
fix(docs): [image] fix `placeholder` demo

makedopamine:
In lower version of `Vue Macros`, there is a bug in `hoistStatic` feature that causes the issue.

Before: 
image

After:
image

### DIFF
--- a/docs/vite.config.ts
+++ b/docs/vite.config.ts
@@ -72,6 +72,9 @@ export default defineConfig(async ({ mode }) => {
       VueMacros({
         setupComponent: false,
         setupSFC: false,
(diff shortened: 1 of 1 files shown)


warmthsea:
Using `ref` can also solve this problem of docs.

4e6a139392d34f80c720a00660d15d72

makedopamine:
True, it works as `ref()`  is a `CallExpression`  AST node that won't be hoisted.  Even though it's sort of a workaround in this case, we'd better not fix the issue in this way, because the issue will remain in the code base as a potential problem.

tolking (approved):

PR #17283, opened 2024-06-20
fix(components): [upload] allow httpRequest return Thenable

toyobayashi:

`instanceof Promise` is unsafe when user passing a native `async function` as callback and global Promise is overwritten by some third party packages.

### DIFF
--- a/packages/components/menu/src/menu.ts
+++ b/packages/components/menu/src/menu.ts
@@ -24,6 +24,7 @@ import {
   iconPropType,
   isObject,
   isString,
+  isThenable,
   mutable,
 } from '@element-plus/utils'
 import { useNamespace } from '@element-plus/hooks'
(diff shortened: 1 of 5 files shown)


btea:
Thanks for your contribution. 

Can you explain the rationale for this change? It seems to me that `isFunction(val.then)` is easier to modify.

toyobayashi:
The `then` property of a `Thenable` may be getter, access `then` may throw error

btea:
Yeah, the situation you mentioned may also exist. However, this is too special. In this case, I think it is reasonable to throw an error as expected.

toyobayashi:
Sorry for late response. Catching error in `then` getter can make `onError` callback work fine. I'd like to align the behavior like `Promise.resolve` or `await` keyword.

Example 1

```js
const promise = Promise.resolve({
  get then () {
    throw new Error('error')
  }
})

// not throw here

promise.catch(err => {
  console.log(err) // will go here
})
```

Example 2

```js
const promise = (async function () {
  // throw new Error('The error in this line will throw immediately (sync)')

  await {
    get then () {
      throw new Error('error')
    }
  }
})()

// not throw here

promise.catch(err => {
  console.log(err) // will go here
})
```

PR #17292, opened 2024-06-21
refactor(components): [slider]optimize ts types

tyj-321:

### DIFF
--- a/packages/components/slider/src/composables/use-slide.ts
+++ b/packages/components/slider/src/composables/use-slide.ts
@@ -19,9 +19,9 @@ export const useSlide = (
 
   const slider = shallowRef<HTMLElement>()
 
-  const firstButton = ref<SliderButtonInstance>()
+  const firstButton = ref()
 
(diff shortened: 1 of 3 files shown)


tyj-321:
I think using any type is not good enough, so I change the tooltip type

packages/components/slider/src/composables/use-slide.ts
```diff
@@ -19,9 +19,9 @@ export const useSlide = (
 
   const slider = shallowRef<HTMLElement>()
 
-  const firstButton = ref<SliderButtonInstance>()
+  const firstButton = ref()
```
tolking, line 22: Why do you do this?

tyj-321: sorry, if I don't do this. when run `pnpm build`, it will show an error like this
image

today, I pushed again, I found another way to solve this error, you can review it, thanks

tolking (approved):
LGTM

PR #17327, opened 2024-06-26
refactor(components): [statistics] Allows props.value to be a string.

kindaFunnyz:
refactor(components): [statistics] Allows props.value to be a string. Formatter will format only number. Docs now cover string and DayJs as valid props.
Actually, even before this changes, component worked with strings pretty fine, but Vue shows annoying warnings.
Also in docs allowed only number, but in code "DayJs" also allowed, so i made some changes here too.

### DIFF
--- a/docs/en-US/component/statistic.md
+++ b/docs/en-US/component/statistic.md
@@ -40,8 +40,8 @@ statistic/card
 ### Statistic Attributes
 
 | Attribute         | Description                    | Type                                                                | Default |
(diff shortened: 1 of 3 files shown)


btea:
The value of `statistic` component only accepts `number`, and the value of `countdown` component accepts `number` and `Dayjs`.

The string doesn't seem to make sense for the statistic component, so probably doesn't need to be modified.

tolking:
Thank for your contribution

IMO, The string type of value and formatter are prone to conflict

Issue #17083, opened 2024-06-05
[Style] [table] el-table 表格样式排版小bug


mpmp22:
Bug Type: **`Style`**

## Environment

- Vue Version: `3.4.27`
- Element Plus Version: `2.7.4`
- Browser / OS: `Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:126.0) Gecko/20100101 Firefox/126.0`
- Build Tool: `CDN`

### Related Component

- `el-table`

### Reproduction Link

[CodePen](https://codepen.io/mpmp22/pen/bGyWZoV)

### Steps to reproduce

直接通过浏览器的 HTML 标签导入 Element Plus 使用 el-table 表格 样式排版是一种列表展示，它不是表格展示！

### What is Expected?

样式正常

### What is actually happening?

样式出错

### Additional comments

_(empty)_

warmthsea:
It doesn't look like a simple style error, it may be a data rendering anomaly.

warmthsea:
### Use CDN:

❌❌❌❌❌❌

``` vue
<el-table :data="tableData">
    <el-table-column prop="address" label="Address" />
    <el-table-column prop="name" label="Name" />
    <el-table-column prop="date" label="Date" /> 
</el-table>
```
😱😱😱
![image](https://github.com/element-plus/element-plus/assets/45450994/f6abe83a-4780-4a78-9fe9-e1d147d389e4)

✅✅✅✅✅✅

``` vue
<el-table :data="tableData">
    <el-table-column prop="address" label="Address"></el-table-column>
    <el-table-column prop="name" label="Name"></el-table-column>
    <el-table-column prop="date" label="Date"></el-table-column>
</el-table>
```

``` vue
<el-table :data="tableData">
    <el-table-column prop="address" label="Address">
        <template #default="scope">
            {{scope.row.address}}
        </template>
    </el-table-column>
    <el-table-column prop="name" label="Name">
        <template #default="scope">
            {{scope.row.name}}
        </template>
    </el-table-column>
    <el-table-column prop="date" label="Date">
        <template #default="scope">
            {{scope.row.date}}
        </template>
    </el-table-column>
</el-table>
```

![image](https://github.com/element-plus/element-plus/assets/45450994/61e5fb5e-da1e-4140-af1a-b8068f25257e)

warmthsea:
``` vue
<el-input v-model="title" style="width: 240px" placeholder="Please input" />
[...]

PR #17404, opened 2024-07-04
docs: add single closed tag tip for using CDN in `HTML`

warmthsea:
https://github.com/element-plus/element-plus/issues/17083
https://github.com/element-plus/element-plus/issues?q=is%3Aissue+cdn+table

image

Although `el-table` has a record (new issues still appear dozens of times), other components using single closing tags also have this problem, so I think it is better to add a reminder here.

### Add

image


btea (approved):
Thanks!

Issue #7155, opened 2022-04-14
[Bug Report] [TypeScript] [All] pnpm 安装后 全局组件类型支持失效


huzedong2015:
Bug Type: **`TypeScript`**

## Environment

- Vue Version: `3.2.32`
- Element Plus Version: `2.1.9`
- Browser / OS: `chrome 99`
- Build Tool: `Vite`

### Related Component

- `All`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cCBsYW5nPVwidHNcIj5cbmltcG9ydCB7IHJlZiB9IGZyb20gJ3Z1ZSdcbmltcG9ydCB7IFVzZXIgfSBmcm9tICdAZWxlbWVudC1wbHVzL2ljb25zLXZ1ZSdcblxuY29uc3QgbXNnID0gcmVmKCdIZWxsbyBXb3JsZCEnKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPCEtLSBFbGVtZW50IFBsdXMgaWNvbnMgLS0+XG4gIDxlbC1pY29uPjxVc2VyIC8+PC9lbC1pY29uPlxuXG4gIDxoMT57eyBtc2cgfX08L2gxPlxuICA8ZWwtaW5wdXQgdi1tb2RlbD1cIm1zZ1wiIC8+XG48L3RlbXBsYXRlPiIsImltcG9ydC1tYXAuanNvbiI6IntcImltcG9ydHNcIjp7fX0ifQ==)

### Steps to reproduce

这个附件链接是无效，实际要在vs code复现
如果用npm安装是生效的，pnpm安装就不行
// tsconfig.json
{
  "compilerOptions": {
    // ...
    "types": ["element-plus/global"]
  }
}

### What is Expected?

全局组件vs code能正确提示类型

### What is actually happening?

没有提示

### Additional comments

如果在.d.ts文件
/// <reference types="element-plus/global" /> 引入就可以了。

sxzz:
请参考 https://github.com/sxzz/element-plus-best-practices 仓库

PR #17430, opened 2024-07-05
chore(build): update declare module name for volar

jardenliu:
- [x] Update declare module name for volar,  '@vue/runtime-core' => 'vue'

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -1,5 +1,5 @@
 // GlobalComponents for Volar
-declare module '@vue/runtime-core' {
+declare module 'vue' {
   export interface GlobalComponents {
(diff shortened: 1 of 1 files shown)


abitwhy:
According to [Web Components and TypeScript](https://vuejs.org/guide/extras/web-components.html#web-components-and-typescript), this would really close #7155. It's very important for [Full Import](https://element-plus.org/en-US/guide/quickstart.html#full-import) & TypeScript users!
By the way, VueRouter has accept [this change](https://github.com/vuejs/router/pull/2295).

tolking (approved):

btea (approved):
Thanks!

PR #17444, opened 2024-07-05
refactor(docs): upgrade vitepress

makedopamine:


xxholly32:
well done ~, hope core teammember can review

> https://github.com/element-plus/element-plus/discussions/17262

docs/.vitepress/vitepress/components/vp-demo.vue
```diff
@@ -126,25 +106,6 @@ const copyCode = async () => {
...
-            class="op-btn"
-            tabindex="0"
-            role="button"
-            @click="copyCode"
-            @keydown.prevent.enter="copyCode"
-            @keydown.prevent.space="copyCode"
-          >
-            <i-ri-file-copy-line />
-          </ElIcon>
-        </ElTooltip>
```
tolking, line 147: IMO, It might be better to keep this copy button so that users don't have to expand the code first before copying it.

tolking (commented):
LGTM

internal/build-utils/package.json
```diff
@@ -21,6 +28,6 @@
     "consola": "^2.15.3"
   },
   "devDependencies": {
-    "unbuild": "^0.7.4"
```
YunYouJun, line 24: `unbuild@2.0` has some incompatible updates in my mind, which I recommend not mixing into vitepress PR.

YunYouJun (changes requested):

docs/.vitepress/vitepress/components/vp-demo.vue
YunYouJun: I think there's no harm in keeping them both(ep copy & vitepress copy in code), partly to be compatible with past user habits.

YunYouJun:
The rest of the content seems to be working well, thank you for your work.

internal/build-utils/package.json
makedopamine: If I were not to upgrade `unbuild`, `npm run docs:dev` would fail.
Here is why: After upgrading, VitePress uses `vite@5.x` under the hood, which only supports ESM. And `unbuild@0.7.4 --stub` outputs a non-standard `.mjs` file that cannot be consumed correctly by ESM.

docs/.vitepress/vitepress/components/vp-demo.vue
makedopamine: Well, I will revert the code tomorrow.

internal/build-utils/package.json
YunYouJun: Ok, then let's move forward.
@jw-foss is working on pnpm monorepo about `utils` and `constants`.
Do you have any other suggestions?

makedopamine: I have no idea.🤔  Is there any connection between them? Upgrading `unbuild` doesn't seem to affect the work you mentioned.

YunYouJun (approved):

PR #17551, opened 2024-07-16
fix(components): SSR hydration error caused by disabled prop of teleport

tolking:

1. Vue's built-in Teleport component will cause hydration errors after set the disabled prop. This change avoids using the Teleport component after set the disabled prop, thereby avoiding hydration errors.

- [the hydration errors demo](https://preview-17472-element-plus.surge.sh/en-US/component/dialog.html), open Dialog

2. The previous ElTeleport component has not been used, so I refactor it to implement this function.


packages/components/drawer/src/drawer.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <teleport to="body" :disabled="!appendToBody">
+  <el-teleport to="body" :disabled="!appendToBody">
```
btea, line 2: Maybe we can make `to` here configurable later?

tolking: Agree, It should be consistent with the ElDialog

btea (approved):

PR #17576, opened 2024-07-17
 fix(components):[loading] loading directive should dispose old instance before create a new instance

akamerar:
In some extreme cases, the loading component can add multiple instances of loading to a single element at the same time, overwriting the previous instances and making them inaccessible. loading directive should dispose old instance before create a new instance


tolking:
Can you provide an example of issues?

akamerar:
> Can you provide an example of issues?

Sorry, it seems like a historical issue, I can only reproduce this issue in 2.1.9, not in 2.7.7, it should be fixed, thank you

PR #17051, opened 2024-06-02
docs(components): [radio & checkbox] add using `value` tip and perf examples

warmthsea:
Closed: https://github.com/element-plus/element-plus/issues/16892

Edit components:  `radio`、`checkbox`

- add using `value` tip
  - change title to top
- perf examples
  - delete `Basic usage`  disabled examples
  - delete useless css

### Before:

https://element-plus.org/en-US/component/radio.html
https://element-plus.org/en-US/component/checkbox.html

### After:

捕获1
捕获

PR #17588, opened 2024-07-18
docs: perf highlight display and simplify code

warmthsea:
about: https://github.com/element-plus/element-plus/pull/17444, https://github.com/element-plus/element-plus/pull/17051

# Fixed the issue of abnormal highlight color after `vitepress` version upgrade.

## VitePress Update Before:
捕获1

## Before:
image

## After:
image

# Simplify code

## Before:
image

## After:
image


btea (approved):
Thanks!

PR #17594, opened 2024-07-19
docs(components): [menu] fix invalid HTML code in the example

oliver139:

`<div>` is not supposed to be the child of `<ul>`
This PR removed the gap created by `div.flex-grow` and instead using `margin-right: auto` to create the gap

### DIFF
--- a/docs/examples/menu/left-and-right.vue
+++ b/docs/examples/menu/left-and-right.vue
@@ -6,14 +6,13 @@
     :ellipsis="false"
     @select="handleSelect"
   >
-    <el-menu-item index="0">
+    <el-menu-item index="0" class="img-logo">
       <img
         style="width: 100px"
         src="/images/element-plus-logo.svg"
         alt="Element logo"
       />
     </el-menu-item>
-    <div class="flex-grow" />
     <el-menu-item index="1">Processing Center</el-menu-item>
     <el-sub-menu index="2">
       <template #title>Workspace</template>
@@ -40,7 +39,7 @@ const handleSelect = (key: string, keyPath: string[]) => {
 </script>
 
 <style>
-.flex-grow {
-  flex-grow: 1;
+.el-menu--horizontal > .el-menu-item.img-logo {
+  margin-right: auto;
 }
 </style>
--- a/docs/examples/menu/left-and-right.vue
+++ b/docs/examples/menu/left-and-right.vue
@@ -6,7 +6,7 @@
     :ellipsis="false"
     @select="handleSelect"
   >
-    <el-menu-item index="0" class="img-logo">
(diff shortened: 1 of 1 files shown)


chenxch:
Thank you for your contribution, but this one does not seem to be as robust as the original demo, for example in rtl mode.

oliver139:
> Thank you for your contribution, but this one does not seem to be as robust as the original demo, for example in rtl mode.

Oh yes, I really missed RTL. Will fix it.

oliver139:
> Thank you for your contribution, but this one does not seem to be as robust as the original demo, for example in rtl mode.

But isn't that RTL mode is not well support in Element Plus yet?
At least the same example is currently also failed for having a wrong padding on the right to the logo on RTL mode.
The docs website is also not providing RTL language for now (and I believe that there is no plan for that?).

It would be hard and meaningless to handle RTL if the framework is not well supported or having no plan to support it (e.g. providing prop `:rtl="true"` for the components and adding `dir="rtl"` attribute on the components etc.).

docs/examples/menu/left-and-right.vue
```diff
@@ -40,7 +39,7 @@ const handleSelect = (key: string, keyPath: string[]) => {
 </script>
 
 <style>
-.flex-grow {
-  flex-grow: 1;
+.el-menu--horizontal > .el-menu-item.img-logo {
```
btea, line 42: No need to add additional classes.
```suggestion
.el-menu--horizontal > .el-menu-item:nth-child(1) {
```

tolking:
Why not just change the `div` to `li`?

oliver139:
> Why not just change the `div` to `li`?

Then actually there will be an accessibility problem.

But if we talking about accessibility, I believe that the current HTML is also being a bit fail.
Where when a screen reader reading it, it would announce the number of item in the `<ul>` list.
If we include the logo into the list, it will announce there are 3 items in the list, where we should expect it counts the number of items in the menu only.

So actually I think the best approach would be:
```html
<div class="topbar">
  <div class="topbar__left">
    <!-- Logo -->
  </div>
  <div class="topbar__right">
    <!-- Menu -->
  </div>
</div>
```

```css
.topbar {
  display: flex;
  flex-flow: row nowrap;
  justify-content: space-between;
  align-items: center;
}
```
But yes, it might already a bit too complicated for just a example

docs/examples/menu/left-and-right.vue
oliver139: Yep, I did think about this too.

btea (approved):

tolking (approved):

tolking:
Thank for your contribution

PR #17599, opened 2024-07-19
chore: improve intellisense in codebase

makedopamine:
- Compatible with the latest Vue language plugin.  It no longer adds the `GlobalComponents` declaration from @vue/runtime-core to the global scope. See [fda6ad3](https://github.com/vuejs/language-tools/commit/fda6ad360490bb2822ae5309bdd9530ed75cc7a5#diff-1d432278152c845126fa9e323391375cf8fa8806a2cad5c1da6282dae4019947L15)
- Remove a few lines of code I think are redundant.


tolking:
I noticed that the `components.d.ts` file generated by `unplugin-vue-components` in the docs directory also has a similar issue. Perhaps we should update the version of `unplugin-vue-components`

.npmrc
```diff
@@ -1,4 +1,5 @@
 shamefully-hoist=true
 strict-peer-dependencies=false
 shell-emulator=true
-package-manager-strict=false
\ No newline at end of file
+package-manager-strict=false
+https-proxy=socks://127.0.0.1:1080
```
tolking, line 5: ??
Perhaps this file should not be modified

makedopamine: ...my bad

tolking (approved):

docs/components.d.ts
```diff
@@ -1,9 +1,11 @@
...
-// We suggest you to commit this file into source control
+/* eslint-disable */
+// @ts-nocheck
+// Generated by unplugin-vue-components
 // Read more: https://github.com/vuejs/core/pull/3399
-import '@vue/runtime-core'
+export {}
 
-declare module '@vue/runtime-core' {
+/* prettier-ignore */
```
btea, line 7: This file is included in both the `.prettierignore` and the `.eslintignore` files, so there should be no need to add relevant comments here.

makedopamine: True, but the file is auto-generated, we shouldn't modify it manually.

btea: Okay.

btea (approved):

Issue #17439, opened 2024-07-05
[Component] [table, table-column] el-table-column组件设置了sort-orders为 "['descending', null]" 通过点击升序按钮  依旧可以升序


liuhoi:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.3.4`
- Element Plus Version: `2.7.5`
- Browser / OS: `Chrome 126.0.6478.127 / win11`
- Build Tool: `Vite`

### Related Component

- `el-table`
- `el-table-column`

### Reproduction Link

[...]

PR #17608, opened 2024-07-21
fix(components): [table] fix sorting issue in table-column

xixiIBN5100:

issues https://github.com/element-plus/element-plus/issues/17439

### DIFF
--- a/packages/components/table/src/table-header/event-helper.ts
+++ b/packages/components/table/src/table-header/event-helper.ts
@@ -160,7 +160,6 @@ function useEvent<T>(props: TableHeaderProps<T>, emit) {
     event.stopPropagation()
     const order =
       column.order === givenOrder ? null : givenOrder || toggleOrder(column)
-
     const target = (event.target as HTMLElement)?.closest('th')
 
     if (target) {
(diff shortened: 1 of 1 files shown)


packages/components/table/src/table-header/event-helper.ts
```diff
@@ -160,8 +160,22 @@ function useEvent<T>(props: TableHeaderProps<T>, emit) {
...
+    ) {
+      return
+    }
+
+    if (
+      hasClass(clickTarget, 'descending') &&
+      !column.sortOrders.includes('descending')
+    ) {
+      return
+    }
```
tolking, line 178: Wouldn't it be more appropriate to put this part of the code after `if (!column.sortable) return`?

xixiIBN5100: Thanks for the correction, changes have been made

tolking (approved):

packages/components/table/src/table-header/event-helper.ts
```diff
@@ -172,6 +171,21 @@ function useEvent<T>(props: TableHeaderProps<T>, emit) {
...
```
btea, line 187: This could probably be adjusted as follows, as there seem to be too many if statements at the moment.
```ts
const clickTarget = event.currentTarget
if (['ascending', 'descending'].some(str =>  hasClass(clickTarget, str) && !column.sortOrders.includes(str))) {
	return
}
```

xixiIBN5100: It works successfully, thanks for the guidance

btea (approved):
Thanks!

PR #17644, opened 2024-07-23
chore(vscode): fix `unocss` highlighting on `windows` caused by node path diff

warmthsea:
about: https://github.com/unocss/unocss/issues/3972

- windows

Before: 

image

After:

image


btea (approved):
Thanks!

PR #17650, opened 2024-07-24
feat(components): [carousel] export `activeIndex`

warmthsea:
about: https://github.com/element-plus/element-plus/discussions/17648


tolking (approved):

btea (approved):
Thanks!

PR #17654, opened 2024-07-25
fix(build-utils): ignore the effects of directories when exclude files

tolking:

Build error when `['node_modules', 'test', 'mock', 'gulpfile', 'dist']` is included in the project directory.

### DIFF
--- a/internal/build-utils/src/pkg.ts
+++ b/internal/build-utils/src/pkg.ts
@@ -31,7 +31,8 @@ export const getPackageDependencies = (
 
 export const excludeFiles = (files: string[]) => {
   const excludes = ['node_modules', 'test', 'mock', 'gulpfile', 'dist']
-  return files.filter(
-    (path) => !excludes.some((exclude) => path.includes(exclude))
-  )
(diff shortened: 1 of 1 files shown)


btea (commented):
I don't quite understand the error you mentioned, could you explain it in more detail?

tolking:
> I don't quite understand the error you mentioned, could you explain it in more detail?

- directories

```
test
  |- element-plus
```

- pnpm build

in pkg.ts

``` ts
const excludeFiles = (files) => { // The files same like ['/.../test/element-plus/packages/constants/aria.ts', ....]
  const excludes = ['node_modules', 'test', 'mock', 'gulpfile', 'dist'];
  return files.filter(
    (path) => !excludes.some((exclude) => path.includes(exclude)) // path.includes(exclude) === true
  ); // files === []
}
```

btea:
I seem to understand what you mean. There will be a problem when the path where the source code is located contains the same value as in `excludes`.

btea (approved):

PR #17684, opened 2024-07-26
fix(element-plus): Fixed an error with @types/lodash when building

lucky980421:
image
将类型文件放在生产依赖，在构建的时候会出现意料之外的错误


btea:
Thanks for your contribution, but it looks like a duplicate of #17677.

PR #17699, opened 2024-07-29
feat(hooks): use-draggable support resetPostion

donga-gao:

### DIFF
--- a/docs/en-US/component/dialog.md
+++ b/docs/en-US/component/dialog.md
@@ -175,6 +175,12 @@ When using `modal` = false, please make sure that `append-to-body` was set to **
 | open-auto-focus  | triggers after Dialog opens and content focused  | ^[Function]`() => void` |
 | close-auto-focus | triggers after Dialog closed and content focused | ^[Function]`() => void` |
 
+### Exposes
+
+| Name         | Description   | Type                    |
+| ------------ | ------------- | ----------------------- |
+| resetPostion | reset postion | ^[Function]`() => void` |
+
 ## FAQ
 
 #### Using dialog in SFC, the scope style does not take effect.
--- a/docs/en-US/component/dialog.md
+++ b/docs/en-US/component/dialog.md
@@ -179,7 +179,7 @@ When using `modal` = false, please make sure that `append-to-body` was set to **
 
 | Name         | Description   | Type                    |
 | ------------ | ------------- | ----------------------- |
-| resetPostion | reset postion | ^[Function]`() => void` |
+| resetPostion ^(2.8.0) | reset postion | ^[Function]`() => void` |
 
 ## FAQ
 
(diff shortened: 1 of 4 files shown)


tolking:
Thank for your contribution.

Could you briefly explain why this feature is needed?

wzc520pyfm:
The draggabled Dialog can be completely moved out of the window after setting overflow, and it will still be outside the window even if it is opened again. Providing a reset position function would solve this, or could we design a better API?

donga-gao:
> The draggabled Dialog can be completely moved out of the window after setting overflow, and it will still be outside the window even if it is opened again. Providing a reset position function would solve this, or could we design a better API?

you know the scene！

donga-gao:
> The draggabled Dialog can be completely moved out of the window after setting overflow, and it will still be outside the window even if it is opened again. Providing a reset position function would solve this, or could we design a better API?

you know the scene！

donga-gao:
> The draggabled Dialog can be completely moved out of the window after setting overflow, and it will still be outside the window even if it is opened again. Providing a reset position function would solve this, or could we design a better API?

you know the scene！

tolking:
I understand.

You can first improve the Exposes in the document, [refer to](https://github.com/element-plus/element-plus/blob/dev/docs/en-US/component/popover.md#exposes)

tolking (approved):

docs/en-US/component/dialog.md
```diff
@@ -175,6 +175,12 @@ When using `modal` = false, please make sure that `append-to-body` was set to **
 | open-auto-focus  | triggers after Dialog opens and content focused  | ^[Function]`() => void` |
 | close-auto-focus | triggers after Dialog closed and content focused | ^[Function]`() => void` |
 
+### Exposes
+
+| Name         | Description   | Type                    |
+| ------------ | ------------- | ----------------------- |
+| resetPostion | reset postion | ^[Function]`() => void` |
```
btea, line 182:
```suggestion
| resetPostion ^(2.8.0) | reset postion | ^[Function]`() => void` |
```

docs/en-US/component/dialog.md
```diff
@@ -175,6 +175,12 @@ When using `modal` = false, please make sure that `append-to-body` was set to **
 | open-auto-focus  | triggers after Dialog opens and content focused  | ^[Function]`() => void` |
 | close-auto-focus | triggers after Dialog closed and content focused | ^[Function]`() => void` |
 
+### Exposes
+
+| Name         | Description   | Type                    |
+| ------------ | ------------- | ----------------------- |
+| resetPostion ^(2.8.0) | reset postion | ^[Function]`() => void` |
```
btea, line 182: The version number is added here, which causes the width to change. It should be formatted to be consistent with the header.

btea (approved):
Thanks!

PR #17711, opened 2024-07-30
docs(docs): [icon] add loops to display icon

lxKylin:

image


tolking:
Thank for your contribution. 

I do not understand, why do we need to add it?

lxKylin:
> Thank for your contribution.
> 
> I do not understand, why do we need to add it?

Maybe some users don't know how to loop an icon

tolking:
This doesn't seem necessary.

btea:
Thank you for your contribution. However, I think this should be the usage of dynamic rendering components in Vue, and we don't need to add a separate example.
Moreover, it seems that no one has any questions about this, so maybe we should close it first.

PR #17723, opened 2024-07-31
feat(components): teleport to `Element` judgement disable

warmthsea:
First of all, I don't think this is a bug in `element-plus`. I just added content to this PR to solve the following problems (because this problem will directly cause the program to fail to run):

image

Fixed: 
https://github.com/element-plus/element-plus/issues/17710
https://github.com/element-plus/element-plus/issues/16058
https://github.com/element-plus/element-plus/issues/15245
https://github.com/element-plus/element-plus/issues/7061

https://github.com/Tencent/wujie/issues/883
https://github.com/Tencent/wujie/issues/679
https://github.com/Tencent/wujie/issues/564
https://github.com/umijs/qiankun/issues/2408

Refer to:
https://github.com/vuejs/core/issues/2015

I released a package in this PR.
https://www.npmjs.com/package/@warmthsea/element-plus-fix-select

```
"dependencies": {
   "element-plus": "npm:@warmthsea/element-plus-fix-select",
}
```

Repo:
https://github.com/warmthsea/element-plus/tree/play/teleport
https://github.com/Lisheri/wujie_v3_elm_demo   (use `^2.7.8`)  ( [change to play host link](https://github.com/Lisheri/wujie_v3_elm_demo/blob/f0c6a1a8eb1035779f405f177982fe22a964fd93/wujie_demo_main/src/views/AboutView.vue#L40) )

Fix after demo repo (use `npm:@warmthsea/element-plus-fix-select`): ✅
https://github.com/warmthsea/wujie_v3_elm_demo

PS:  https://github.com/element-plus/element-plus/tree/dev/packages/components/teleport
The code does not seem to be used.


warmthsea:
@Lisheri If you have time, take the test.

Lisheri:
> @Lisheri If you have time, take the test.

nb!
I tested a few pages, it looks ok.

packages/hooks/use-load-teleport/index.ts
```diff
@@ -0,0 +1,27 @@
...
+export const useLoadTeleport = (
+  appendTo: AppendTo | Ref<AppendTo> | ComputedRef<AppendTo>
+) => {
+  const isLoadTeleport = ref(false)
+
+  if (!isClient) {
+    isLoadTeleport.value = true
+  } else if (unref(appendTo)) {
+    const el =
+      typeof unref(appendTo) === 'string'
```
btea, line 16: You can use the built-in `isString` method.

btea:
There are code conflicts that need to be resolved.

warmthsea:
> There are code conflicts that need to be resolved.

I saw that, I will make some more code changes.

I think it is necessary to advance this pr to avoid additional operations of users.

tolking:
Thanks for your contribution.

IMO, Disabling Teleport by force may not be a suitable solution. In some cases, it is necessary to use Teleport. 

eg.

- Nesting Dialog components within Dialog components.
- Using Image Viewer in Table Component.

If possible, it would be better to solve why the Dom of the `Teleport to` does not exist.

Or add a document like SSR to show [how to avoid this issues](https://github.com/element-plus/element-plus/issues/17710#issuecomment-2266347462),

tolking:
Fixed by #18121

Issue #17751, opened 2024-08-01
[TypeScript] [tooltip, tooltip-v2] ts errors when a custom effect is set to a tooltip


AxelConceicao:
Bug Type: **`TypeScript`**

## Environment

- Vue Version: `3.4.35`
- Element Plus Version: `2.7.8`
- Browser / OS: `Chrome lastest`
- Build Tool: `Vite`

### Related Component

- `el-tooltip`
- `el-tooltip-v2`

### Reproduction Link

[...]

PR #17860, opened 2024-08-12
chore(components): improve type for popper effect

wzc520pyfm:
Abort #17758 #17751 #17040 #17030

At #17751 , we are forced to set the props type related to `popperEffect` to `PopperEffect | string`, causing `popperEffect ` to be deduced as `string`.

But now, here is a better solution to improve the ts type of `popperEffect`! It preserves the type hints for `light` and `drak` and accepts any string. The only downside is that it cannot pass the eslint check.

image

image


wzc520pyfm:
`{}` may be replaced by  `NonNullable<unknown>`, i will try it, tonight

packages/components/menu/src/menu.ts
```diff
@@ -139,7 +139,9 @@ export const menuProps = buildProps({
    * @description Tooltip theme, built-in theme: `dark` / `light` when menu is collapsed
    */
   popperEffect: {
-    type: definePropType<PopperEffect | string>(String),
+    type: definePropType<PopperEffect | (string & NonNullable<unknown>)>(
+      String
+    ),
```
tolking, line 144:
```suggestion
    type: definePropType<PopperEffect>(String),
```

```ts
export type PopperEffect = typeof effects[number] | (string & {})
```

tolking (commented):
I found that `(string & NonNullable<unknown>)` and `(string & {})` are the same, maybe we can use `(string & {})` directly

wzc520pyfm:
> I found that `(string & NonNullable<unknown>)` and `(string & {})` are the same, maybe we can use `(string & {})` directly

`{}` cannot pass the eslint check.

tolking:
> > I found that `(string & NonNullable<unknown>)` and `(string & {})` are the same, maybe we can use `(string & {})` directly
> 
> `{}` cannot pass the eslint check.

I see

btea:
Is it possible to extract a common type from so many identical types?

packages/components/menu/src/menu.ts
wzc520pyfm: Agree, I have updated.

wzc520pyfm:
> Is it possible to extract a common type from so many identical types?

Agree, I have updated.

btea (approved):

tolking (approved):

PR #17941, opened 2024-08-19
feat(components): [input] remove extra size classname on the textarea

LoTwT:

## description

The prop `size` will work when the `type` is not `textarea`.

image

However, when using textarea like `<el-input type="textarea" size="large" />`, currently it will add an extra class name `el-input--large`.

image

This PR is to conditionally add the size classname when the type of the input is not `textarea`.

### DIFF
--- a/packages/components/input/__tests__/input.test.tsx
+++ b/packages/components/input/__tests__/input.test.tsx
@@ -141,6 +141,11 @@ describe('Input.vue', () => {
     expect(wrapper.classes('el-input--large')).toBe(true)
   })
 
+  test('textarea size', () => {
+    const wrapper = mount(() => <Input type="textarea" size="large" />)
+    expect(wrapper.classes('el-input--large')).not.toBe(true)
+  })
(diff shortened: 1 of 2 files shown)


btea:
Thank you for your contribution, but I think we can keep it. Some users may customize the style based on different size class names.

LoTwT:
> Thank you for your contribution, but I think we can keep it. Some users may customize the style based on different size class names.

Do you mean element-plus won't set specific styles with `el-input--large` and `el-input--small`, but some users may have done? Keeping the class names is a historical issue?

btea:
Yes, I think this can be convenient for users to customize styles. WDYT? @tolking

tolking:
> Yes, I think this can be convenient for users to customize styles. WDYT?

agree

It seems to have been this way from the beginning.

btea:
@LoTwT  Based on the above reasons, I will close this PR. Looking forward to your contribution again.

PR #17974, opened 2024-08-21
feat(components): [dialog] support entire dialog `vertical-center`

warmthsea:
#### Demo

image


warmthsea:
😥 I have never seen the `align-center` attribute. 😂

PR #18048, opened 2024-08-28
refactor: import shared utilities from `@element-plus/utils`

makedopamine:
Import shared utilities from `@element-plus/utils`  instead of `@vue/shared` for consistency.


btea (approved):
LGTM

tolking (approved):

PR #18051, opened 2024-08-28
chore: optimize dts imports

DDDDD12138:

I noticed that the `typings/components.d.ts` file is almost identical to the `global.d.ts` file. The `global.d.ts` file is only copied to the `dist/element-plus` folder after the build process and has no other use. 
Therefore, I propose merging these two files into one. This will also prevent issues where type references might be missing in `global.d.ts`, causing automatic imports to fail.

#### Before
before
#### After
after2


tolking (approved):
LGTM

typings/global.d.ts
```diff
@@ -101,6 +99,8 @@ declare module 'vue' {
     ElWatermark: typeof import('element-plus')['ElWatermark']
     ElTour: typeof import('element-plus')['ElTour']
     ElTourStep: typeof import('element-plus')['ElTourStep']
+    ElAnchor: typeof import('element-plus')['ElAnchor']
+    ElAnchorLink: typeof import('element-plus')['ElAnchorLink']
```
btea, line 103: Why have these two lines of code been moved here?

DDDDD12138: I used git mv to rename the `typings/components.d.ts` file to `typings/global.d.ts`, but for some reason, github is comparing it with the `global.d.ts` file in the parent directory.

tolking: Perhaps we can adjust the order of the ElAnchor and ElAnchor (L102,L103 -> L9,L10)

DDDDD12138: ok, I’ve updated it.

btea (approved):

DDDDD12138:
It seems like it hasn’t been merged yet. What else do I need to do?

warmthsea (approved):
LGTM.

PR #18114, opened 2024-09-01
fix(components): [select-v2] cannot find name `handleBlur`

makedopamine:
In #16746, the CI report indicates an error: `Cannot find name 'handleBlur'`. The error seems to be unintentionally introduced due to `// @tsnocheck`

### DIFF
--- a/packages/components/select-v2/src/useSelect.ts
+++ b/packages/components/select-v2/src/useSelect.ts
@@ -105,7 +105,7 @@ const useSelect = (props: ISelectV2Props, emit) => {
     afterComposition: (e) => onInput(e),
   })
 
-  const { wrapperRef, isFocused } = useFocusController(inputRef, {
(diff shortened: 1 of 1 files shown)


packages/components/select-v2/src/useSelect.ts
```diff
@@ -105,7 +105,7 @@ const useSelect = (props: ISelectV2Props, emit) => {
     afterComposition: (e) => onInput(e),
   })
 
-  const { wrapperRef, isFocused } = useFocusController(inputRef, {
+  const { wrapperRef, isFocused, handleBlur } = useFocusController(inputRef, {
```
btea, line 108: Both `handleClickOutside` in this file and `v-click-outside:[popperRef]="handleClickOutside"` in the `select.vue` file can be removed.

tolking: link #17691

makedopamine:
I closed this PR because there is already a PR that fixes the issue.

PR #18208, opened 2024-09-06
docs(components): [table] revise `getSelectionRows` type

makedopamine:
Update the type `() => void` to `() => any[]`


btea (approved):
Thanks!

PR #18219, opened 2024-09-08
ci: release pr commit is made on the original repository dev

warmthsea:
### 1、push release

image

### 2、pr release

image

The address of the initial comment always points to the first executed code in my local environment.
But in some environments, it points to the latest one, such as vercel.

```sh
pnpm add https://pkg.pr.new/warmthsea/element-plus@21
```

The specified commit points to a specific code. Every time the pr is updated, the commit address of the office comment will change to point to the latest

image

image

Why not wait until the document is packaged before executing?
- Robots don't comment


.github/workflows/publish-commit-pkg.yml
```diff
@@ -1,14 +1,16 @@
...
+name: Publish Commit Pkg
 
 on:
-  pull_request_review:
-    types: [submitted]
+  push:
+    branches:
+      - dev
+    tags:
+      - '!**'
```
btea, line 8: It could be adjusted to the initial state, executed after the pr is created and the document is built so the creator can test it himself. 🤔

warmthsea: After building the document, execute ci. The pr.new.pkg bot will not comment, so I write it directly as pr execution.

warmthsea: > It could be adjusted to the initial state, executed after the pr is created and the document is built so the creator can test it himself. 🤔

I tested two PRs and asked a warehouse to install and download them. I found that the execution results of the robot's initial comment address were different in different environments. I don't know where the problem is.
So now I think it's best not to let the robot comment, and do it according to what you said above

https://github.com/warmthsea/elp-pkg-pr-release-test

https://github.com/warmthsea/element-plus/pull/20
https://github.com/warmthsea/element-plus/pull/21

btea (approved):

.github/workflows/publish-pr-commit-pkg.yml
```diff
@@ -22,7 +27,7 @@ jobs:
       - name: Setup pnpm
         uses: pnpm/action-setup@v4
         with:
-          version: 8.14.1
+          version: 9.5.0
```
tolking, line 30:
```diff
- with:
-   version: 9.5.0
```

Perhaps it can be consistent with other CI

tolking (approved):

PR #18224, opened 2024-09-08
chore(build): remove redundant code

makedopamine:
`components.ts` outputs a components.json file that seems to be unused.


btea (approved):
It really doesn't seem to be of any use. 😄

warmthsea (approved):

tolking (approved):

PR #18237, opened 2024-09-10
feat(components): [autocomplete] add getData to exposes

dormadekhin:


btea:
Hi, could you describe the usage scenario in which you need this function?

dormadekhin:
> Hi, could you describe the usage scenario in which you need this function?

Hi! I need to programmatically call a data request when clicking on a button outside the component

btea:
If I understand correctly, you can store the value you want in a variable, and then fetch the value from the variable in the `fetch-suggestions` function.

dormadekhin:
> If I understand correctly, you can store the value you want in a variable, and then fetch the value from the variable in the `fetch-suggestions` function.

I meant something else. I need to initiate the loading of the list of suggestions when clicking on another element, or rather on the icon next to it.

btea:
OK, it seems that exposing this method will help developers to play more freely. However, you need to update the documentation as well.

dormadekhin:
> OK, it seems that exposing this method will help developers to play more freely. However, you need to update the documentation as well.

Done

packages/components/autocomplete/src/autocomplete.vue
```diff
@@ -402,5 +402,7 @@ defineExpose({
   close,
   /** @description highlight an item in a suggestion */
   highlight,
+  /** @description run load results */
```
btea, line 405: The description here could perhaps be tweaked. `loading suggestion list`

docs/en-US/component/autocomplete.md
```diff
@@ -110,3 +110,4 @@ autocomplete/custom-loading
 | loading          | remote search loading indicator             | ^[object]`Ref<boolean>`                   |
 | popperRef        | el-tooltip component instance               | ^[object]`Ref<ElTooltipInstance>`         |
 | suggestions      | fetch suggestions result                    | ^[object]`Ref<record<string, any>>`       |
+| getData          | run load data                               | ^[Function]`(queryString: string) => void`|
```
btea, line 113: It would be a good idea to mark in which version it will be available.
```suggestion
| getData ^(2.8.3) | run load data                               | ^[Function]`(queryString: string) => void`|
```

packages/components/autocomplete/src/autocomplete.vue
dormadekhin: Done

docs/en-US/component/autocomplete.md
dormadekhin: Done

btea (approved):

dormadekhin:
Hi @btea, when will this feature merge to master branch?

docs/en-US/component/autocomplete.md
```diff
@@ -110,3 +110,4 @@ autocomplete/custom-loading
 | loading          | remote search loading indicator             | ^[object]`Ref<boolean>`                   |
 | popperRef        | el-tooltip component instance               | ^[object]`Ref<ElTooltipInstance>`         |
 | suggestions      | fetch suggestions result                    | ^[object]`Ref<record<string, any>>`       |
+| getData ^(2.8.3) | loading suggestion list                     | ^[Function]`(queryString: string) => void`|
```
btea, line 113:
```suggestion
| getData ^(2.8.4) | loading suggestion list                     | ^[Function]`(queryString: string) => void`|
```

btea (approved):
Thanks for your contribution!

PR #18253, opened 2024-09-12
fix: 内存中的游离节点, 滚动条dom节点未释放

zmcode:
image
轮动条销毁 的时候, 存在游离节点, 内存未释放, 如果一个项目使用过多, 会导致大量内存泄露

### DIFF
--- a/packages/components/scrollbar/src/scrollbar.vue
+++ b/packages/components/scrollbar/src/scrollbar.vue
@@ -32,6 +32,7 @@ import {
   onActivated,
   onMounted,
   onUpdated,
+  onBeforeUnmount,
   provide,
   reactive,
   ref,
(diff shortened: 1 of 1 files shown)


warmthsea:
Hi @zmcode , thank you very much for your contribution, could you provide a before and after example in playground?

LZQCN:
This PR does not make sense.

If there is a memory leak, it is more likely happening through the following way:
First, the `el-scrollbar` component provides the `scrollbarRef` to child components within its slot.
```js
provide(
  scrollbarContextKey,
  reactive({
    scrollbarElement: scrollbarRef,
    wrapElement: wrapRef,
  })
)
```
Then, the component that uses it references the `scrollbarRef` but may cause a memory leak due to improper code.
The most common reason is that some events are subscribed to through an event bus and in the event handler `scrollbarRef.value` is referenced. However, if the event is not unsubscribed during the component's destruction, the value of `scrollbarRef.value` will be continuously referenced by the event handler's closure and not released.

btea:
@LZQCN Thank you for the detailed analysis.

@zmcode Can you show the comparison results before and after the modification? By the way, there are many errors in the code that need to be adjusted.

zmcode:
这是来自QQ邮箱的假期自动回复邮件。
 
你好，我最近正在休假中，无法亲自回复你的邮件。我将在假期结束后，尽快给你回复。

PR #18255, opened 2024-09-12
chore: upgrade to pnpm 9

makedopamine:


btea (approved):

btea:
@warmthsea  You can create a PR, update the pnpm version, and then merge it with this PR.
https://github.com/element-plus/element-plus/blob/647af2cd225df449467096d6df3d07c02329ee95/.github/workflows/pkg-pr-new.yml#L25

makedopamine:
@warmthsea 

> Version is optional when there is a packageManager field in the package.json

According to the [documentation](https://github.com/pnpm/action-setup), it doesn't seem we need to set pnpm version manually.

warmthsea:
I think can merge this PR directly,  and merge https://github.com/element-plus/element-plus/pull/18219

warmthsea (approved):

PR #18283, opened 2024-09-14
refactor(eslint-config): upgrade to eslint 9

makedopamine:
This PR changes a lot of files. I've attempted to break it down into individual commits logically meaningful, so reviewing commit by commit might be easier.


tolking:
Sorry, I just noticed this commit. #21949 has already completed the same upgrade. I'll close this commit for now, you can choose to restore it depending on your situation.

PR #18427, opened 2024-09-30
chore: fix docs/play dev sass deprecation warning

warmthsea:
- About: https://github.com/element-plus/element-plus/pull/18376

     _Impact of sass version updates._

- Before:

image
image

- After:

image
image


btea (approved):
LGTM

PR #18428, opened 2024-09-30
docs: add sass compatible version description

warmthsea:
🌀 Merge https://github.com/element-plus/element-plus/pull/18427 After
About: https://github.com/element-plus/element-plus/pull/18376#issuecomment-2381360685

image

### DIFF
--- a/docs/.vitepress/config/vite.ts
+++ b/docs/.vitepress/config/vite.ts
@@ -58,6 +58,13 @@ const alias: AliasOptions = [
 export const getViteConfig = ({ mode }: { mode: string }): ViteConfig => {
   const env = loadEnv(mode, process.cwd(), '')
   return {
+    css: {
+      preprocessorOptions: {
+        scss: {
+          silenceDeprecations: ['legacy-js-api'],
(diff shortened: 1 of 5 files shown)


lvzhenbo:
The modern-compiler option only applies to the sass-embedded package, the sass package is the modern option

btea (approved):

warmthsea:
> ~~The modern-compiler option applies to the sass-embedded package, the sass package is the modern option~~
> 
> 
> 
> Oh, my mistake was that it was recommended to use the sass-embedded package together with the modern-compiler option

I tested it yesterday, and there seems to be no problem with this configuration.

warmthsea:
Regarding the deprecated API warning of sass, we may have two solutions in the future:

1. Pay attention to whether the PR of `gulp-sass` is merged
https://github.com/dlmanning/gulp-sass/pull/846

2. Use other solutions instead of `gulp-sass`



```
Deprecation Warning: The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0.
```

PR #18439, opened 2024-10-03
feat(components): [transfer] render costum panel body

LoTwT:

## description

This PR is to enhance the ElTransfer so that it can render the custom panel body. Based on this new feature, ElTransfer could work with other components such as ElTree.

Related discussion: #15762.

The TransferPanel exposes a new slot `body`, its default value is the original checkbox group. Users could use the slot `left-panel-body` and `right-panel-body` exposed on the Transfer with the props `fileredData`, `checkedKey`, and a method `toggleChecked` that are bind to the slot to render data. To get more flexibility with the original behavior of `no data` and `no match data`, there are two new props `remainLeftPanelBody` and `remainRightPanelBody`.

### DIFF
--- a/packages/components/transfer/src/transfer.ts
+++ b/packages/components/transfer/src/transfer.ts
@@ -138,6 +138,14 @@ export const transferProps = buildProps({
     type: Boolean,
     default: true,
   },
+  /**
+   * @description whether to remain the left panel body when having no data or match data
+   */
+  remainLeftPanelBody: Boolean,
+  /**
+   * @description whether to remain the right panel body when having no data or match data
+   */
+  remainRightPanelBody: Boolean,
 } as const)
 export type TransferProps = ExtractPropTypes<typeof transferProps>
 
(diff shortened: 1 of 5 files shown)


LoTwT:
Here is the example https://stackblitz.com/edit/vitejs-vite-hfpoyz?file=src%2FApp.vue using the `pkg-pr-new`.

After reviewing and this PR can be merged, I will add a description to the docs.

packages/components/transfer/src/transfer.ts
```diff
@@ -143,9 +138,16 @@ export const transferProps = buildProps({
...
     default: true,
   },
+  /**
+   * @description whether to remain the left panel body when having no data or match data
+   */
+  remainLeftPanelBody: Boolean,
+  /**
+   * @description whether to remain the right panel body when having no data or match data
+   */
+  remainRightPanelBody: Boolean,
```
tolking, line 148: Thanks you for your contribution.

If users can already customize the main content, then there seems to be no need to add these two attributes.

PR #18479, opened 2024-10-09
feat(components): [date-picker] add active class for shortcuts

Dchih:
add active class to active shortcut
remove active class when user picks date
changes includes date-picker month-picker year-picker & their range picker

reasons:
Usually, we cannot tell at a glance what the time we choose on the timetable means. But when we add an 'active' to a shortcut, it becomes very easy to see.


Dchih:
element-plus offered an 'active' class for shortcuts active style. We can use this class to implement our effects

PR #18482, opened 2024-10-09
fix(hooks): rendering multiple popper container DOM in SSR/SSG

tolking:

By viewing the [document](https://element-plus.org/zh-CN/component/overview.html), it can be found that there are two `<div id="el-popper-container-1024">` appearing in the DOM.

### DIFF
--- a/packages/hooks/use-popper-container/index.ts
+++ b/packages/hooks/use-popper-container/index.ts
@@ -3,8 +3,6 @@ import { isClient } from '@element-plus/utils'
 import { useGetDerivedNamespace } from '../use-namespace'
 import { useIdInjection } from '../use-id'
 
-let cachedContainer: HTMLElement
-
 export const usePopperContainerId = () => {
   const namespace = useGetDerivedNamespace()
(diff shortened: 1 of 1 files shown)


packages/hooks/use-popper-container/index.ts
```diff
@@ -37,10 +35,9 @@ export const usePopperContainer = () => {
     // for this we need to disable the caching since it's not really needed
     if (
       process.env.NODE_ENV === 'test' ||
-      !cachedContainer ||
       !document.body.querySelector(selector.value)
     ) {
-      cachedContainer = createContainer(id.value)
+      createContainer(id.value)
```
btea, line 40: It's strange, why did we use a variable to save it in the first place?

tolking: It should be to avoid duplicate creation of DOM. Same as `document.body.querySelector(selector.value)`

btea (approved):

makedopamine (approved):

PR #18553, opened 2024-10-14
chore(docs): switch Vue to the version used by VitePress

makedopamine:
`@vitejs/plugin-vue` resolves Vue from project root by default, which means it used 3.2.x before, so I manually specified the version as 3.4.x that is used by Vitepress. 
Since `defineOptions` has been integrated into the Vue core as of version 3.3.x, we can remove `vue-macros`.

### DIFF
--- a/docs/.vitepress/config/index.mts
+++ b/docs/.vitepress/config/index.mts
@@ -8,8 +8,9 @@ import { nav } from './nav'
 import { mdPlugin } from './plugins'
 import { sidebars } from './sidebars'
 import { getViteConfig } from './vite'
+import { vueCompiler } from './vue-compiler'
 
-import type { UserConfigFn } from 'vitepress'
+import type { UserConfig } from 'vitepress'
 
 const buildTransformers = () => {
   const transformer = () => {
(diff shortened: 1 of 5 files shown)


btea (approved):
LGTM 👍

warmthsea (commented):
-

warmthsea (approved):

warmthsea:
I found that this PR seems to cause some problems.


## [#18553 Docs Preview](https://preview-18553-element-plus.surge.sh/en-US/) nprogress load unusual
https://staging.element-plus.org/zh-CN/

image


## [#18554 Docs Preview](https://preview-18554-element-plus.surge.sh/en-US/) normal

image

screenshots

btea:
When I start the document switching page locally, it looks normal.

HyramTang:
Hi @makedopamine 
first, thank you for your work;

I found a problem. After this PR adjustment, the function of the Dropdown component will be invalid when previewing the document website.
You can check the PR document preview [WEBSITE PREVIEW IS READY](https://preview-18553-element-plus.surge.sh/) link, find the Dropdown component in it, or directly check that there is no Dropdown content in the language switch dropdown component in the upper right corner.

This problem only occurs in WEBSITE PREVIEW documents.

> [](https://preview-18553-element-plus.surge.sh)

tolking:
Indeed, there was an issue with the ElDropdown component of the preview link starting from this PR. @btea 

img

PR #18586, opened 2024-10-18
fix: progress bar get stuck when navigating between routes

makedopamine:
Vue 3.4 improved the reactivity system in https://github.com/vuejs/core/pull/5912.

> If computed new value does not change, computed, effect, watch, watchEffect, render dependencies will not be triggered

In `vp-content.vue`, running `nprogress.done()` depends on the computed value `isNotFound` which no longer triggers a re-render after the route changes.


btea (approved):

warmthsea (approved):
👍👍

PR #18657, opened 2024-10-24
feat(components): add `showProgressBar`, `actions`, and `timerControls` props 

geforseN:
This PR introduces several enhancements to the `Notification` component in Element Plus, inspired by features from [Nuxt UI Notification](https://ui.nuxt.com/components/notification). The main goal is to provide more flexible interaction options and improve the user experience by offering the following features:

1. **Progress Bar**:  
   A progress bar is displayed at the bottom of the notification, providing users with a visual indication of how much time remains before the notification closes. By default, the progress bar is not shown, consistent with the previous implementation.

2. **Actions**:  
   Allows the addition of customizable action buttons, providing more interactive notifications. Each action includes:
   - `label`: (Required) Button text. Must be unique and non-empty.
   - `execute` (Required): Handler for the click event.
   - `keepOpen` (Optional): Determines whether to keep the notification open after execution. The default is `false`, but it can be set to `true` to keep the notification open or to `'until-resolved'` to await the return of `execute`, after which the notification will close. 
   - `disableAfterExecute` (Optional): Disables the button and prevents further calls to `execute` after the first execution. Defaults to `true` unless `keepOpen` is set to `true`. It is not recommended to change this value, as it prevents multiple execute calls.
   - `button` (Optional): [ElButton](https://element-plus.org/en-US/component/button) component props. Ignores any `onclick` properties, as `execute` handles the click logic.

3. **Timer Controls**:  
	Determines whether the timer should pause and resume on hover or reset and restart. The default option is to reset and restart, consistent with the previous implementation.

#### Additional Improvements
- **Documentation**:  
   Updated documentation covers the new `showProgressBar`, `actions`, and `timerControls` props, explaining how each feature works, along with clarifications on keyboard event handling (delete, backspace, escape).

- **Refactoring**:  
   Updated existing tests and added new ones to ensure the correct behavior of the `showProgressBar`, `actions`, and `timerControls` features.

#### Important
- Please ensure that you update the version numbers (X.Y.Z) in the component documentation to reflect the correct version.

#### Related Issue
- Closes #18542


warmthsea:
@geforseN 

Thank you very much for your contribution. We will review this PR carefully when we have time, which may take some time.

btea:
Can you resolve the code conflict?

packages/components/focus-trap/src/utils.ts
```diff
@@ -82,7 +82,7 @@ export const tryFocus = (
 ) => {
   if (element && element.focus) {
     const prevFocusedElement = document.activeElement
-    let cleanup: boolean = false
+    let cleanup = false
```
btea, line 85: Maybe we can keep the type declaration here.

packages/components/notification/src/composables/use-actions.ts
```diff
@@ -0,0 +1,93 @@
+import { type Ref, computed, ref } from 'vue'
+import { type MaybeComputedRef, resolveUnref as toValue } from '@vueuse/core'
+import { type Mutable, debugWarn, keysOf } from '@element-plus/utils'
+import type { ButtonProps } from '@element-plus/components/button'
+import type { NotificationAction, NotificationProps } from '../notification'
```
btea, line 5: To keep the code style consistent, we'd better separate the variable introduction and type introduction and put the type introduction at the bottom.

packages/components/notification/src/composables/use-actions.ts
```diff
@@ -0,0 +1,93 @@
...
+) {
+  const actions_ = computed(() => {
+    const actionsValue = toValue(actions)
+    if (!actionsValue) {
+      return []
+    }
+    const preparedActions = actionsValue
+      .filter(
+        (action) =>
+          typeof action.execute === 'function' &&
```
btea, line 26: Here you can use the built-in function `isFunction`/`isString`

packages/components/notification/src/composables/use-actions.ts
```diff
@@ -0,0 +1,93 @@
...
+  button: Partial<ButtonProps>
+  disabled: Ref<boolean>
+  onClick: (event: MouseEvent) => Promise<void>
+}
+
+export function useActions(
+  actions: MaybeComputedRef<NotificationProps['actions']>,
+  closeNotification: () => void
+) {
+  const actions_ = computed(() => {
```
btea, line 18:
```suggestion
  const _actions = computed(() => {
```
Internal variables may have a leading underscore.

packages/components/notification/src/notify.ts
```diff
@@ -34,12 +35,17 @@ const notifications: Record<
 const GAP_SIZE = 16
 let seed = 1
 
+function parseOptions(options: NotificationParams) {
+  if (typeof options === 'string' || isVNode(options)) {
```
btea, line 39:
```suggestion
  if (isString(options) || isVNode(options)) {
```

packages/components/package.json
```diff
@@ -9,6 +9,9 @@
   "peerDependencies": {
     "vue": "^3.2.0"
   },
+  "devDependencies": {
+    "jsdom-testing-mocks": "^1.13.1"
+  },
```
btea, line 14: I think we not need introduce new dep, the vitest already support mock.

https://vitest.dev/api/vi.html#vi-mock

geforseN: I added `jsdom-testing-mocks` only to mock the [Web Animations API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API) because after replacing [`useTimeoutFn`](https://vueuse.org/shared/useTimeoutFn/) with [`Animation`](https://developer.mozilla.org/en-US/docs/Web/API/Animation), the tests for `ElNotification` started failing.

I couldn't find a way to mock Animation properly without using this library. If there's an alternative solution, I'm open to suggestions.

patienry (commented):
Thank you for your contribution! The progress bar functionality was released in **v2.14.4**. If you'd like, you can create a new PR for the remaining functionality.
https://element-plus.org/en-US/component/notification#with-progress-bar

Issue #18642, opened 2024-10-23
[Component] [date-picker] date组件week选择问题


055frostFairies:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.12`
- Element Plus Version: `2.8.6`
- Browser / OS: `Google Chrome版本 129.0.6668.101（正式版本） （64 位）`
- Build Tool: `Vite`

### Related Component

- `el-date-picker`

### Reproduction Link

[...]

PR #18662, opened 2024-10-25
fix(components): [date-picker] fix dayjs parse week exception. #18642

CherishTheYouth:

- [X] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).
- [X] Make sure you are merging your commits to `dev` branch.
- [X] Add some descriptions and refer to relative issues for your PR.

### Issue
- #18642 

### Cause of the problem
- /time-picker/src/utils.ts

```ts
export const parseDate = function (
  date: string | number | Date,
  format: string | undefined,
  lang: string
) {
  const day =
    isEmpty(format) || format === 'x'
      ? dayjs(date).locale(lang)
      : dayjs(date, format).locale(lang)
  return day.isValid() ? day : undefined
}

```
1. Dayjs version 2.11.3 has an exception when parsing date data in YYYY ww format. It cannot parse correctly.
2. The plug-ins required to parse YYYY ww format are not extended when dayjs is used in utils

### My Solution

- Update dayjs version to 2.11.13
- Extened  weekOfYear and customParseFormat plugin
```ts
dayjs.extend(weekOfYear)
dayjs.extend(customParseFormat)
```


packages/components/time-picker/src/utils.ts
```diff
@@ -1,7 +1,14 @@
 import dayjs from 'dayjs'
+import customParseFormat from 'dayjs/plugin/customParseFormat'
+import weekOfYear from 'dayjs/plugin/weekOfYear'
+
 import { isArray, isDate, isEmpty } from '@element-plus/utils'
 
 import type { Dayjs } from 'dayjs'
+
+dayjs.extend(weekOfYear)
+dayjs.extend(customParseFormat)
```
btea, line 10: Maybe we should not add additional plugins into the code, but let users import them when needed.

btea (commented):
This PR should only require upgrading the dayjs version. There seem to be a lot of changes after you upgraded the dayjs version. 

In my opinion, the code content changed after the update should be similar to #18738. Maybe the pnpm version you use is different from the one in the source code. Please check it.

btea:
The changes in #18738 make more sense, so we have merged it. He has added you as `Co-authored-by` in #18738, so I will close this PR.

PR #18734, opened 2024-11-01
fix(components): [pagination] fix total undefined trigger current-change

xcy960815:


xcy960815:
### background

Fixed an issue where `current-change` is triggered when `total` is `undefined` in `el-pagination` component.

### Change

- Modified the processing logic of `total` to ensure that the event will not be triggered when it is `undefined`.

### Test steps

1. When using `el-pagination`, set `total` to `undefined`.
2. Confirm that the `current-change` event is no longer triggered.

warmthsea:
I don't quite understand the meaning of your PR, can you explain it in detail and provide examples?

xcy960815:
After you reply, you have thought for a long time. After you have responded for a long time, please note that the current page is assigned to be uncertain, although the user has a guarantee that the total number must be a number of numbers.

xcy960815:
This is the bug demo address
https://github.com/xcy960815/el-pagination-bug，If there is something wrong, please give me some advice.

tolking:

This modification seems pointless; when `v-model:current-page` is bound to a value other than 1, the `current-change` event will still be triggered.

`pageCount` and `total` cannot both be undefined at the same time; your usage is not reasonable.

PR #18737, opened 2024-11-01
fix: downgrade vue-tsc to 2.0.28

makedopamine:
Downgrade `vue-tsc` to version 2.0.28 due to https://github.com/vuejs/language-tools/pull/4644 increases the size of `.d.ts` files by almost 1MB.

Thank @btea for pointing out the problem.


btea (approved):
👍

tolking (approved):

PR #18768, opened 2024-11-04
docs: use native fetch for changelogs

Jungzl:

- drop axios, use native fetch instead;
- drop unused dep octokit;


tolking (approved):

btea (approved):
LGTM.

I'm not sure why we introduced `axios` initially, and the documentation does not mention the need for ssr.

thinkasany:
> LGTM.
> 
> I'm not sure why we introduced `axios` initially, and the documentation does not mention the need for ssr.

My guess is that the fetch module was only built into Node.js 18, and the fetch API can be called globally, whereas previously you needed to import the node-fetch package.
image

btea:
Oh, your guess should be right.

PR #18805, opened 2024-11-06
docs: update promotion for feature

thinkasany:
更新promotion，方便提示用户了解新的feature功能

image


btea:
Thank you for your contribution, but `promotion` indicates whether the component was newly added to the version. It is enough to add the corresponding version tag in the table for the newly added' prop'.

PR #18824, opened 2024-11-07
docs(components): use `""` replace null for no icon setting

thinkasany:
修复icon的type

image

image

### DIFF
--- a/docs/en-US/component/page-header.md
+++ b/docs/en-US/component/page-header.md
@@ -39,7 +39,7 @@ page-header/custom-icon
 ## No icon
 
 Sometimes the page is just full of elements, and you might not want the icon to show up on the page,
-you can set the `icon` attribute to `null` to get rid of it.
+you can set the `icon` attribute to `""` to get rid of it.
 
 :::demo
 
--- a/docs/examples/page-header/additional-sections.vue
+++ b/docs/examples/page-header/additional-sections.vue
@@ -1,5 +1,5 @@
 <template>
-  <el-page-header :icon="null">
+  <el-page-header icon="">
     <template #content>
       <div class="flex items-center">
         <el-avatar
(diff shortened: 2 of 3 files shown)


packages/components/page-header/src/page-header.ts
```diff
@@ -1,14 +1,14 @@
...
+import type { Component, ExtractPropTypes } from 'vue'
 import type PageHeader from './page-header.vue'
 
 export const pageHeaderProps = buildProps({
   /**
    * @description icon component of page header
    */
   icon: {
-    type: iconPropType,
+    type: definePropType<string | Component | null>([String, Object, Function]),
```
btea, line 11: Maybe you should add a `withNullIconPropType` in the same file as the `iconPropType` type.

btea:
This is a legacy issue. Setting the icon to an empty string should also achieve the desired effect. Perhaps we should change the documentation.

packages/components/page-header/src/page-header.ts
thinkasany: done.

thinkasany:
> This is a legacy issue. Setting the icon to an empty string should also achieve the desired effect. Perhaps we should change the documentation.

what I need to do, could you please elaborate a little bit?

warmthsea (commented):
Here we also need.

https://github.com/element-plus/element-plus/blob/ead10025339937e584f1814d19418332b48d897a/docs/examples/page-header/additional-sections.vue#L2

thinkasany:
> Here we also need.
> 
> https://github.com/element-plus/element-plus/blob/ead10025339937e584f1814d19418332b48d897a/docs/examples/page-header/additional-sections.vue#L2

done.

warmthsea (approved):

btea (approved):
Thanks!

PR #18828, opened 2024-11-07
fix(components): [select] adjust width of select-dropdown

DDDDD12138:

### Description
The border of the select is implemented using `box-shadow`, which does not occupy width, whereas the border of the popper is implemented using `border`. This results in the select-dropdown being 2px wider than the select.

image

### DIFF
--- a/packages/components/select-v2/src/useSelect.ts
+++ b/packages/components/select-v2/src/useSelect.ts
@@ -266,7 +266,8 @@ const useSelect = (props: ISelectV2Props, emit: SelectEmitFn) => {
   )
 
   const calculatePopperSize = () => {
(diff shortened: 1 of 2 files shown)


warmthsea:
Maybe it was reasonable, after the change, the borders of other selection boxes cannot be covered, which looks weird.

screenshots

----

Refer to:
https://ui.shadcn.com/docs/components/select

image

DDDDD12138:
> Maybe it was reasonable, after the change, the borders of other selection boxes cannot be covered, which looks weird.

My UI aims for the dropdown of the Select component to be arrow-free. However, after implementing it locally, I noticed that the dropdown seems misaligned with the Select. This is why I submitted this PR. Perhaps this issue won't be as noticeable when there are other elements on the page.

image

PR #18872, opened 2024-11-12
fix(components): [color-picker] optimize the flickering issue

momei-LJM:
Continuously clicking on the input box panel will cause it to blink.
I think he should be allowed to click again to close it(maybe).
before:
screenshots

after: 
screenshots


packages/components/color-picker/src/color-picker.vue
```diff
@@ -244,8 +247,8 @@ function setShowPicker(value: boolean) {
   showPicker.value = value
 }
 
+// const debounceSetShowPicker = debounce(setShowPicker, 100, { leading: true })
```
btea, line 250: Why write the same code comment here?

momei-LJM: sorry,I've removed it

tolking (approved):

btea:
Can you provide a detailed explanation? Based on the example you provided, I don't seem to understand the difference between before and after the modification.

momei-LJM:
> Can you provide a detailed explanation? Based on the example you provided, I don't seem to understand the difference between before and after the modification.

When I click the second time, I can't fold the drop-down selector; But the flicker turned on again

btea (approved):

PR #18885, opened 2024-11-12
feat(components): [input-tag] new component

tolking:

The Component styles reference [Select component](https://element-plus.org/en-US/component/select.html#create-new-items)

close #14111, close #13689


docs/.vitepress/crowdin/en-US/pages/component.json
```diff
@@ -106,6 +106,10 @@
         "link": "/input-number",
         "text": "Input Number"
       },
+      {
+        "link": "/input-tag",
+        "text": "Input Tag"
```
btea, line 111: Here we can add the tag `"promotion": "2.9.0"`.

packages/components/input-tag/src/input-tag.ts
```diff
@@ -0,0 +1,126 @@
...
+import { tagProps } from '@element-plus/components/tag/src/tag'
+
+import type { ExtractPropTypes } from 'vue'
+
+export const inputTagProps = buildProps({
+  /**
+   * @description binding value
+   */
+  modelValue: {
+    type: definePropType<string[]>([Array]),
```
btea, line 24:
```suggestion
    type: definePropType<string[]>(Array),
```

warmthsea:
Add a new attribute that disabled input but can delete tag, and add `index` parameters for `@tag-remove` and `default slot`. Will this make the scalability higher?

It has different functionality from [multiple-select](https://preview-18885-element-plus.surge.sh/en-US/component/select.html#basic-multiple-select), so I suggest adding the above parameters.

docs/examples/input-tag/tag.vue
```diff
@@ -0,0 +1,55 @@
...
+  'info',
+  'warning',
+  'danger',
+]
+const effect: Array<TagProps['effect']> = ['light', 'dark', 'plain']
+
+const tagType = ref<TagProps['type']>('primary')
+const tagEffect = ref<TagProps['effect']>('plain')
+const input = ref<string[]>(['tag1', 'tag2', 'tag3'])
+</script>
```
thinkasany, line 55:
```suggestion
<template>
  <div>
    <el-segmented v-model="tagType" :options="type" />
    <el-segmented v-model="tagEffect" :options="effect" style="margin-left: 1rem" />
  </div>
  <br />
  <el-input-tag
    v-model="input"
    :tag-type="tagType"
    :tag-effect="tagEffect"
    placeholder="Please input"
  >
    <template #tag="{ value }">
      <div class="flex items-center">
        <el-icon class="mr-1">
          <ElementPlus />
        </el-icon>
        <span>{{ value }}</span>
      </div>
    </template>
  </el-input-tag>
</template>

<script lang="ts" setup>
import { ref } from 'vue'
import { ElementPlus } from '@element-plus/icons-vue'
import type { TagProps } from 'element-plus'

const type: Array<TagProps['type']> = [
  'primary',
  'success',
  'info',
  'warning',
  'danger',
]
const effect: Array<TagProps['effect']> = ['light', 'dark', 'plain']

const tagType = ref<TagProps['type']>('primary')
const tagEffect = ref<TagProps['effect']>('plain')
const input = ref<string[]>(['tag1', 'tag2', 'tag3'])
</script>
```

thinkasany: The same effect `segmented` can use more concise code, which makes the demo easier to use
image

docs/examples/input-tag/tag.vue
```diff
@@ -0,0 +1,55 @@
...
+      />
+    </el-radio-group>
+    <el-radio-group v-model="tagEffect" aria-label="effect control">
+      <el-radio-button
+        v-for="item in effect"
+        :key="item"
+        :label="item"
+        :value="item"
+      />
+    </el-radio-group>
```
thinkasany, line 18:
```suggestion
    <el-segmented v-model="tagType" :options="type" />
    <el-segmented v-model="tagEffect" :options="effect" style="margin-left: 1rem" />
```

thinkasany: The same effect `segmented` can use more concise code, which makes the demo easier to use.

image

docs/en-US/component/input-tag.md
```diff
@@ -0,0 +1,139 @@
...
+
+## API
+
+### Attributes
+
+| Name                  | Description                                                            | Type                                                        | Default |
+| --------------------- | ---------------------------------------------------------------------- | ----------------------------------------------------------- | ------- |
+| model-value / v-model | binding value                                                          | ^[array]`string[]`                                          | —       |
+| max                   | max number tags that can be enter                                      | ^[number]                                                   | —       |
+| trigger               | the key to trigger input tag                                           | ^[enum]`'Enter' \| 'Space'`                                 | Enter   |
```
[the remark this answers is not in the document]
tolking: For users, the Enter key and numpadEnter key should be the same.

I have already done compatibility processing, so there is no need to add numpadEnter enum here

tolking:
> Add a new attribute that disabled input but can delete tag

If the InputTag has already been config disabled attribute, and the tag can be delete, it will change bind value, and the disabled attribute will have no effect.

> and add `index` parameters for `@tag-remove` and ~`default slot`~. Will this make the scalability higher?

After delete the tag, the corresponding index is no longer the same as before, which is more likely to cause errors.

> It has different functionality from [multiple-select](https://preview-18885-element-plus.surge.sh/en-US/component/select.html#basic-multiple-select), so I suggest adding the above parameters.

Is it about attributes related to `collapse-tags`? I think this component is operated by users through a keyboard. After collapse the tags, users need to switch to a mouse to know the previous input value or the value that is about to be deleted. This operation affects the user experience.

btea (approved):

packages/components/input-tag/src/input-tag.ts
```diff
@@ -0,0 +1,126 @@
...
+  /**
+   * @description placeholder
+   */
+  placeholder: String,
+  /**
+   * @description native input readonly
+   */
+  readonly: Boolean,
+  /**
+   * @description native input readonly
```
thinkasany, line 83:
```suggestion
   * @description whether to show clear button
```

xingyixiang:
When I input long text into the input box, there may be issues with the display

packages/components/input-tag/src/input-tag.ts
thinkasany: image

xingyixiang:
Drag should not be allowed when disabled.

xingyixiang:
Drag should not be allowed when disabled.

thinkasany:
目前好像操作只能往前拖拽，前面的tag往后拖拽就不好使了。

tolking:
是只有拖到tag上才有效

thinkasany:
> 是只有拖到tag上才有效

acro的这个细节很舒服

image

xingyixiang:
感觉可以增加一个tagColor: (value: string, index: number) => color；来自定义标签颜色

warmthsea (approved):

docs/en-US/component/input-tag.md
```diff
@@ -0,0 +1,150 @@
...
+
+You can set a limit on the number of tags that can be added.
+
+:::demo
+
+input-tag/max
+
+:::
+
+## Disabled
```
thinkasany, line 40: 也许这个demo可以优化成 `status`, 更好的展示 states: disabled, readonly 
新组件能让用户在使用之前直接看到效果感觉更好

image

tolking: 这样应该足够了，我觉得和 Input 保持一致会更好

btea (approved):

btea:
@tolking I think it can be merged first.

PR #18006, opened 2024-08-26
fix(components): [tag] leave animations not working

makedopamine:
I filed an [issue](https://github.com/vuejs/core/issues/11703) to Vue and was aware that the problem is actually caused by a breaking change in Vue3.  I think we can either fix #17255 with monkey-patch like this PR, or remove leave animations (possibly even entire animations for consistency) to follow Transition design.

PR #18928, opened 2024-11-17
fix(components): [tag] make stubbed Transition work in tests

purepear:
By default Vue Test Utils is using stubbed Transition/TransitionGroup components. In this [recent ElementPlus PR](https://github.com/element-plus/element-plus/pull/18006) we start relying on some internal logic of the Transition component which breaks existing unit tests. A potential workaround pointer by the author of the PR is [to disable the default behaviour of Vue Test Utils](https://github.com/element-plus/element-plus/pull/18006#issuecomment-2348607574) but it seems like we can avoid workarounds by protecting the code in case the Transition internals are undefined.

### DIFF
--- a/packages/components/tag/src/tag.vue
+++ b/packages/components/tag/src/tag.vue
@@ -75,6 +75,9 @@ const handleClick = (event: MouseEvent) => {
 
 const handleVNodeMounted = (vnode: VNode) => {
   // @ts-ignore
-  vnode.component.subTree.component.bum = null
+  if (vnode?.component?.subTree?.component?.bum) {
+    // @ts-ignore
+    vnode.component.subTree.component.bum = null
+  }
 }
 </script>


packages/components/tag/src/tag.vue
```diff
@@ -75,6 +75,9 @@ const handleClick = (event: MouseEvent) => {
 
 const handleVNodeMounted = (vnode: VNode) => {
   // @ts-ignore
-  vnode.component.subTree.component.bum = null
+  if (vnode?.component?.subTree?.component?.bum) {
```
makedopamine, line 78:
```suggestion
  if(vnode.component!.subTree.type === 'transition-stub') return
```
I think this will make the intent of the code clearer

makedopamine (approved):
LGTM. I just left one minor comment.

packages/components/tag/src/tag.vue
purepear: I don't think we should be test specific but rather protect the code in general (not just for Vue Test Utils but in case some Transition internals change, etc.)

purepear:
@btea Let me know if you'd like me to change something 🙏

btea (approved):
Thanks!

PR #18931, opened 2024-11-18
fix(components): [select] modify the logic of update the watch option

### LINKED ISSUE #18871: [Component] [select] el-select数据回填时触发请求opt列表，如果鼠标在输入框内，会导致数据格式化失败
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.12`
- Element Plus Version: `2.8.7`
- Browser / OS: `chrome125`
- Build Tool: `Vue CLI`

### Related Component

- `el-select`

### Reproduction Link

[Github Repo](https://github.com/element-plus/element-plus/blob/dev/packages/components/select/src/useSelect.ts#L318)

### Steps to reproduce

使用场景是这样的：
我们结构类似 {label: 'xx', value: 1}
1、为el-select组件（远程api）回填数据code 1。（我们这边，会马上以当前数据触发组件请求，作为接口入参）
2、鼠标马上放到el-select输入框内
这时，会弹出下拉框，但是组件已经赋值了并显示code 1，接口在请求中，鼠标不动保持下拉框状态，当接口请求完成的时候，下拉框的列表会更新，但是el-select显示的值不会更新成label “xx”，这个是非常严重的困扰。

看 useSelect.ts 314行

if (
  (!props.filterable &&
    !props.defaultFirstOption &&
    !isUndefined(props.modelValue)) ||
  !Array.from(inputs).includes(document.activeElement as HTMLInputElement)
) {
  setSelected()
}

看这个源代码，当用户焦点在输入框内的时候，永远不会触发 setSelected，意味着数据回填时，鼠标如果在el-select的输入框内时，数据无法格式化显示。

### What is Expected?

用户鼠标在哪里，都不影响显示格式化

鼠标如果在el-select的输入框内时，数据无法格式化显示

### Additional comments

_(empty)_

YiMo1:
fix #18871 

related issue: #13346 

相关Issue: #13346

### DIFF
--- a/packages/components/select/src/useSelect.ts
+++ b/packages/components/select/src/useSelect.ts
@@ -312,15 +312,7 @@ export const useSelect = (props: ISelectProps, emit) => {
     () => {
       if (!isClient) return
       // tooltipRef.value?.updatePopper?.()
-      const inputs = selectRef.value?.querySelectorAll('input') || []
-      if (
-        (!props.filterable &&
-          !props.defaultFirstOption &&
-          !isUndefined(props.modelValue)) ||
-        !Array.from(inputs).includes(document.activeElement as HTMLInputElement)
-      ) {
-        setSelected()
-      }
+      setSelected()
       if (
         props.defaultFirstOption &&
         (props.filterable || props.remote) &&


warmthsea:
Yes, the original code looks strange. 🤔

YiMo1:
> Yes, the original code looks strange. 🤔

是的，我不知道为什么会有这些复杂的条件，也许是性能上的问题？但我认为只要选项发生了更改就直接重新设置。

Yes, I don't know why there are these complex conditions, maybe it's a performance issue? But I think as long as the options are changed, they can be reset directly.

406087475:
想问下这个修复能及时修复吗？这都审核多久了

makedopamine:
I think this change looks good to merge. I can't think of a case where we need the if condition. @tolking @btea Please chime in if you have any thoughts on this change.

tolking:
It looks like it can be deleted

tolking (approved):

btea:
I haven't found why it was written this way, and the current modifications seem fine.

btea (approved):

makedopamine:
I noticed right now that SelectV2 has very similar code. For consistency, the if condition should be removed from SelectV2 as well, but it can be done in a separate PR.
https://github.com/element-plus/element-plus/blob/e17ec9a8a841aaedb400bd8afbe770d1c0d7ad51/packages/components/select-v2/src/useSelect.ts#L839-L843

@msidolphin  The git history shows you authored [this commit](https://github.com/element-plus/element-plus/pull/3092/files). If you are still around, could we have your feedback in case the change breaks anything?

makedopamine (approved):

PR #18932, opened 2024-11-18
feat(i18n): add Malay(ms) translation

hcw2175:
add locale: ms

### DIFF
--- a/packages/locale/lang/ms.ts
+++ b/packages/locale/lang/ms.ts
@@ -2,7 +2,7 @@ export default {
   name: 'ms',
   el: {
     breadcrumb: {
-      label: 'Breadcrumb', // to be translated
+      label: 'Breadcrumb',
     },
     colorpicker: {
       confirm: 'OK',
@@ -50,7 +50,7 @@ export default {
       month12: 'Disember',
       week: 'minggu',
       weeks: {
-        sun: 'Sun', // to be translated
+        sun: 'Sun',
         mon: 'Mon',
         tue: 'Tue',
         wed: 'Wed',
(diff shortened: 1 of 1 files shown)


packages/locale/lang/ms.ts
```diff
@@ -0,0 +1,184 @@
...
+      titles: ['Senarai 1', 'Senarai 2'],
+      filterPlaceholder: 'Masukkan kata kunci',
+      noCheckedFormat: '{total} item', // to be translated
+      hasCheckedFormat: '{checked}/{total} diperiksa',
+    },
+    image: {
+      error: 'FAILED',
+    },
+    pageHeader: {
+      title: 'Belakang', // to be translated
```
plainheart, line 172: Please remove the comment `// to be translated` if one item has been translated.

packages/locale/lang/ms.ts
```diff
@@ -0,0 +1,184 @@
...
+    },
+    tree: {
+      emptyText: 'Tiada Data',
+    },
+    transfer: {
+      noMatch: 'Tiada data yang sepadan',
+      noData: 'Tiada Data',
+      titles: ['Senarai 1', 'Senarai 2'],
+      filterPlaceholder: 'Masukkan kata kunci',
+      noCheckedFormat: '{total} item', // to be translated
```
plainheart, line 165: Can you also translate the rest of the items?

plainheart (changes requested):

packages/locale/lang/ms.ts
hcw2175: Have been removed

packages/locale/lang/ms.ts
hcw2175: Have been modified

plainheart (approved):

warmthsea:
https://github.com/element-plus/element-plus/blob/dev/docs/en-US/guide/i18n.md
Can you add this new language to this file?

PR #18933, opened 2024-11-18
docs: replaced '/' with '|'  in type

zzjiaxiang:
In the docs, '/' or '|' separated types, and I think using '|' is reasonable,So I replaced '/' with '|' to unify the style.
Partial examples:
before:
image
after:
image


btea:
Thanks for your contribution, but I think it seems unnecessary.  WDYT？@tolking

tolking:
I think this modification does more harm than good.

The `\|` is easy to conflict with the `|` of the table, and it is not easy to find the mistake.

btea:
@zzjiaxiang Thank you for your contribution. 

However, based on the above considerations, we believe that the modification of this PR is unnecessary, so I will close this PR.

PR #18937, opened 2024-11-19
docs(components): [table-v2] unified display style of parameters

warmthsea:
#### Before

image

#### After

image


docs/en-US/component/table-v2.md
```diff
@@ -295,70 +295,70 @@ table-v2/manual-scroll
...
+| ------------------------- | -------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------ | --------- |
+| cache                     | Number of rows rendered in advance to boost the performance                                                                | `number`                                               | 2         |
+| estimated-row-height      | The estimated row height for rendering dynamic height rows                                                                 | `number`                                               | —         |
+| header-class              | Customized class name passed to header wrapper                                                                             | `string` / Function<[HeaderClassGetter](#typings)>     | —         |
+| header-props              | Customized props name passed to header component                                                                           | `object` / Function< [HeaderPropsGetter](#typings)>    | —         |
```
btea, line 303:
```suggestion
| header-props              | Customized props name passed to header component                                                                           | `object` / Function<[HeaderPropsGetter](#typings)>    | —         |
```

btea (commented):
Is there a better way to write this function type? It seems a bit strange at the moment.

QQ_1732025244044

warmthsea:
> Is there a better way to write this function type? It seems a bit strange at the moment.

~~Function<[HeaderClassGetter]>~~ ——>  HeaderClassGetter
or like this（but bit long）
image

btea:
It looks perfect, and it doesn’t matter if it’s a little longer when placed in a pop-up window. 👍

warmthsea:
> It looks perfect, and it doesn’t matter if it’s a little longer when placed in a pop-up window. 👍


`TableV2` many types is  long, I think it's better not to change to  pop-up.

image

<hr />

image

btea (approved):
The current modification works fine for me.

PR #19001, opened 2024-11-23
chore: optimize docs style

zzjiaxiang:
1. Remove this blank area to make the content appear more compact 

before:
image
after:
image

2. Make the directory area scrollable, allowing developers to quickly find the APIs they want

before:
image

after:
image

### DIFF
--- a/docs/.vitepress/vitepress/components/doc-content/vp-table-of-content.vue
+++ b/docs/.vitepress/vitepress/components/doc-content/vp-table-of-content.vue
@@ -19,8 +19,8 @@ const removeTag = (str: string) => str.replace(/<span.*<\/span>/g, '')
 <template>
   <aside ref="container" class="toc-wrapper">
     <nav class="toc-content">
-      <h3 class="toc-content__heading">Contents</h3>
-      <el-scrollbar max-height="calc(100vh - 110px)">
+      <el-scrollbar max-height="calc(100vh - var(--header-height) - 55px)">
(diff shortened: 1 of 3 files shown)


btea (approved):
LGTM

warmthsea:
I think the display on the right is fine now, it looks weird after the change.

#### Before

image

#### After

image

btea:
@warmthsea I compared the official websites of vitepress and vite, and the sidebars are both in this style. Maybe we can keep it consistent.

warmthsea:
> @warmthsea I compared the official websites of vitepress and vite, and the sidebars are both in this style. Maybe we can keep it consistent.

IMO(Right column):

1、[vue.org](https://vuejs.org/api/application.html#app-use) also using vitepress,  the right sidebar is very similar to [element-plus.org](https://element-plus.org/en-US/component/container.html#container-api).

2、The biggest difference between [element-plus.org](https://element-plus.org) and other vitepress websites is the width of the main body. 

image


After the change：
  -  it is asymmetrical. 
  - The lower right corner looks very crowded

image


<hr />

In addition, I don't think it's necessary to keep the style exactly the same as other websites. The above is my personal opinion. If I can't convince you, maybe more people need to discuss it.

btea:
Oh, I thought you were talking about the vertical line on the right. After the modification, the margin on the right is indeed too small.

zzjiaxiang:
I adjusted the right margin to keep it consistent with the previous one.

PR #14240, opened 2023-09-09
refactor(hooks): rewrite composition as a composable function

tolking:
Please make sure these boxes are checked before submitting your PR, thank you!

- [ ] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).
- [ ] Make sure you are merging your commits to `dev` branch.
- [ ] Add some descriptions and refer to relative issues for your PR.

### <samp>🤖 Generated by Copilot at 63e5cd3</samp>

Refactored several components to use the `useComposition` hook for handling input element's composition events. The `useComposition` hook is a new addition to the `@element-plus/hooks` package that simplifies the logic and reduces the code duplication. Deleted the unused `useInput` hook and its related file.

## Related Issue

Fixes #\_\_\_.

### <samp>🤖 Generated by Copilot at 63e5cd3</samp>

*  Add `useComposition` hook to `@element-plus/hooks` package to handle composition events of input elements ([link](https://github.com/element-plus/element-plus/pull/14240/files?diff=unified&w=0#diff-48dbc0893378838f4dc797551501452a45559f7bd19098dbbd8e97cb06ae2c8eR1-R50), [link](https://github.com/element-plus/element-plus/pull/14240/files?diff=unified&w=0#diff-a2275f5046e45dab2261fa65b8b67eecf411212ce5ba7b26c72ad8e7115c0d87R30))
[...]

PR #19010, opened 2024-11-25
fix(components): [select] change states.filtertedOptionsCount to filteredOptionsCount.value

makedopamine:
Looks like this was accidentally left out from [#15325](https://github.com/element-plus/element-plus/pull/15352/files#diff-ae55c3efc549b49d40b3f5b535ac2fc1572374d9309b6439f8d529b06edb1368L165).

### DIFF
--- a/packages/components/select/src/useSelect.ts
+++ b/packages/components/select/src/useSelect.ts
@@ -738,7 +738,7 @@ export const useSelect = (props: ISelectProps, emit) => {
     }
     if (
       states.options.size === 0 ||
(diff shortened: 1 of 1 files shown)


warmthsea (approved):

btea:
It looks like it was modified by #14240 refactoring.

btea:
It looks like it was modified by #14240 refactoring.

btea (approved):

tolking (approved):

PR #19024, opened 2024-11-27
chore(deps): update gulp-sass use new Sass JS API

warmthsea:
https://github.com/dlmanning/gulp-sass/pull/846
https://github.com/element-plus/element-plus/pull/18428#issuecomment-2448868980

Before:

image

After:

image

### DIFF
--- a/packages/theme-chalk/package.json
+++ b/packages/theme-chalk/package.json
@@ -35,7 +35,7 @@
     "cssnano": "^6.0.5",
     "gulp-autoprefixer": "^8.0.0",
     "gulp-rename": "^2.0.0",
-    "gulp-sass": "^5.1.0",
+    "gulp-sass": "^6.0.0",
     "postcss": "^8.4.35"
   },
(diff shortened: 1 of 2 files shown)


btea (approved):

warmthsea:
It's strange that although there is no warning when building, there will still be a warning when running. 🤔

btea:
> It's strange that although there is no warning when building, there will still be a warning when running. 🤔

What conditions do you mean by `running`?

warmthsea:
> What conditions do you mean by `running`?

/play dev

image

image

btea:
For vite related issues, we can upgrade the version or modify the related configuration.
https://github.com/vitejs/vite/pull/17937

warmthsea:
> For vite related issues, we can upgrade the version or modify the related configuration. [vitejs/vite#17937](https://github.com/vitejs/vite/pull/17937)

It seems that users must manually configure or upgrade Vite to 6.0 to avoid warnings.

PR #19081, opened 2024-12-02
docs:[message-box] update `inputType` type

zhangenming:

as https://github.com/element-plus/element-plus/blob/dev/packages/components/message-box/src/message-box.type.ts#L173-L174


docs/en-US/component/message-box.md
```diff
@@ -186,7 +186,7 @@ The corresponding methods are: `ElMessageBox`, `ElMessageBox.alert`, `ElMessageB
...
 | inputPlaceholder                  | placeholder of input                                                                                                                     | ^[string]                                                                          | ''                                               |
-| inputType                         | type of input                                                                                                                            | ^[string]                                                                          | text                                             |
+| inputType                         | type of input                                                                                                                            | ^[string]`'text' \| 'textArea' \| 'password' \| 'number'`                          | text                                             |
```
btea, line 189:
```suggestion
| inputType                         | type of input                                                                                                                            | ^[enum]`'text' \| 'textarea' \| 'password' \| 'number'`                          | text                                             |
```

btea (approved):
It seems that only these values ​​are marked in the comments, but there is no complete restriction in the code logic. I am not sure whether it is necessary to make synchronous changes in the code.

The current changes are also acceptable to me.

cc @tolking @makedopamine

tolking:
The `type` attribute also supports other values,  which can easily lead to misunderstandings. At the same time, the relevant annotation information is quite outdated, and I think it needs to be updated with reference to the [MDN input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input).

I think we can refer to #17860 to improve the `type` attribute of the ElInput, then use it here. 

eg.

`export type InputType = 'text' | 'textarea' | 'number' | 'password' | 'email' | 'search' | 'tel' | 'url' | (string & NonNullable<unknown>)`

In docs ^[enum]`'text' | 'textarea' | 'number' | 'password' | 'email' | 'search' | 'tel' | 'url' | string`

WDYT

btea:
Mainly I'm not sure if setting other types here makes sense.

tolking:
I think it is meaningful, such as `'email '|' search '|' tel '|' url'`, which can control the display of the keyboard interface on the mobile phone.

PR #19098, opened 2024-12-03
chore: [select] remove @ts-nocheck directive

makedopamine:


packages/components/select/src/option-group.vue
```diff
@@ -57,18 +60,22 @@ export default defineComponent({
       children.value.some((option) => option.visible === true)
     )
 
-    const isOption = (node) =>
-      node.type?.name === 'ElOption' && !!node.component?.proxy
+    const isOption = (
+      node: VNode
+    ): node is VNode & { component: OptionInternalInstance } =>
+      (node.type as Component).name === 'ElOption' && !!node.component?.proxy
```
btea, line 66: Can the optional chaining be removed here? Can we be sure that type exists?

makedopamine: I think it's ok because the type of `node.type` is `string | VNode | Component | typeof Text_2 | typeof Static | typeof Comment_2 | typeof Fragment | typeof TeleportImpl | typeof SuspenseImpl`, not including nullish.

btea (approved):

packages/components/select/src/useSelect.ts
```diff
@@ -48,23 +47,29 @@ import {
...
 
-export const useSelect = (props: ISelectProps, emit) => {
+export const useSelect = (props: SelectProps, emit: SelectEmitFn) => {
   const { t } = useLocale()
   const contentId = useId()
   const nsSelect = useNamespace('select')
   const nsInput = useNamespace('input')
 
-  const states = reactive({
+  const states: SelectStates = reactive({
```
tolking, line 67:
```suggestion
  const states = reactive<SelectStates>({
```

packages/components/select/src/useOption.ts
```diff
@@ -63,7 +72,7 @@ export function useOption(props, states) {
 
   const updateOption = (query: string) => {
     const regexp = new RegExp(escapeStringRegexp(query), 'i')
-    states.visible = regexp.test(currentLabel.value) || props.created
+    states.visible = regexp.test(currentLabel.value as string) || props.created
```
tolking, line 75: `currentLabel: ComputedRef<string | number | boolean>`

How about using `String()` functions for conversion?

packages/hooks/use-namespace/index.ts
```diff
@@ -70,9 +70,9 @@ export const useNamespace = (
       ? _bem(namespace.value, block, blockSuffix, element, modifier)
       : ''
   const is: {
-    (name: string, state: boolean | undefined): string
+    (name: string, state: boolean | string | undefined): string
     (name: string): string
-  } = (name: string, ...args: [boolean | undefined] | []) => {
+  } = (name: string, ...args: [boolean | string | undefined] | []) => {
```
tolking, line 75: The `state` support string looks quite strange, perhaps we can avoid modifying it here

makedopamine:
Sure. But I’d like to merge #20109 first, since #19352 didn't import the used types due to `@ts-nocheck`, or we could add the missing types in a separate PR. Which one do you perfer?

warmthsea:
> #20109 f

Can wait until #20109 is resolved before dealing with conflicts and other issues in this PR.

packages/components/select/src/select.ts
```diff
@@ -1,4 +1,6 @@
+import { type ExtractPropTypes } from 'vue'
```
btea, line 1: Type imports can be placed below.

packages/components/select/src/useSelect.ts
```diff
@@ -15,6 +15,7 @@ import {
   debounce as lodashDebounce,
 } from 'lodash-unified'
 import { useResizeObserver } from '@vueuse/core'
+import { ScrollbarInstance } from 'element-plus'
```
btea, line 18:
```suggestion
import type { ScrollbarInstance } from '@element-plus/components/scrollbar'
```

btea (commented):
I left two comments, you can take a look. 🏓

packages/components/select/src/select.ts
```diff
@@ -8,13 +10,16 @@ import { buildProps, definePropType, iconPropType } from '@element-plus/utils'
...
+
 import type {
   Options,
   Placement,
   PopperEffect,
 } from '@element-plus/components/popper'
+import type { OptionValue } from './type'
 
-export const SelectProps = buildProps({
+export const selectProps = buildProps({
```
btea, line 22: In order not to affect the original users, perhaps we can export both `dada` and `dada` at the same time, and remove the capitalized one in a future version.

packages/components/select/src/select.ts
```diff
@@ -252,3 +263,17 @@ export const SelectProps = buildProps({
   ...useEmptyValuesProps,
   ...useAriaProps(['ariaLabel']),
 })
+/* eslint-disable @typescript-eslint/no-unused-vars */
+export const selectEmits = {
```
btea, line 267: The variable does not appear to be referenced.

btea: ok.

btea (approved):

warmthsea (approved):

PR #19099, opened 2024-12-03
feat(components): [dialog/drawer] add custom sub classes

Dsaquel:
# I would like to configure/access to styles of dialog/drawer slots.
## At the current version we can access in two ways:

1. ### With container classes
```vue
<template>
   <el-drawer v-model="drawer">
       <template #header>
          <div class="header-container">content</div>
      </template>
      <div class="body-container">content</div>
      
      <template #footer>
         <div class="footer-container">content</div>
      </template>
  
    </el-drawer>
</template>

<style>
.header-container {}
.body-container {}
.footer-container {}
</style>
```
**inconvenient**: cant access to the real slots style

2. ### Target wanted classes with `:deep`

```css
:deep(.el-drawer__body) {}
```
**inconvenient**: not elegant + uncomfortable

## Proposal's

1. ### Having 3 mores props `header-class-name` , `body-class-name`, and `footer-class-name`
```vue
<template>
 <el-drawer v-model="drawer" header-class-name="header-container" body-class-name="body-container" footer-class-name="footer-container">
     <template #header>content</template>

    content
    
    <template #footer>content</template>
  </el-drawer>
</template>

<style>
.header-container {}
.body-container {}
.footer-container {}
</style>
```

**advantage**: simple + no breaking change, this will respond to 98% of the usage

2. Keep `1` and having a sub component <el-drawer-content> like [naive drawer](https://www.naiveui.com/en-US/os-theme/components/drawer#DrawerContent-Props)

```vue
<template>
 <el-drawer v-model="drawer" >
    <el-drawer-content header-class-name="header-container" body-class-name="body-container" footer-class-name="footer-container">
         <template #header>content</template>
    
        content
        
        <template #footer>content</template>
    </el-drawer-content>
  </el-drawer>
</template>

<style>
.header-container {}
.body-container {}
.footer-container {}
</style>
```

**advantage**: more scalable for customisation + splitting logic about slots content in intern but its gonna be a breaking change

The real issue is from this [discussion](https://github.com/element-plus/element-plus/discussions/19060).

If you have others solutions im open for it.

(Im preparing for `1`)


btea:
If I understand correctly, if you want the class name added through `header/body/footer-class-name` to take effect in the scoped style tag, you still need to add `:deep`, which seems no different from the original.

Dsaquel:
@btea Actually yes, i think the main usage for these classes names will be to add global css where you set your custom `header/body/footer` class names and depends of your need you'll set directly your class in the component props.
Example:
```css
/* global.css */

.custom-header {
color: red
}
```

```vue
//App.vue
import ./global.css
```

```vue
// random file
<template>
 <el-drawer v-model="drawer" header-class-name="custom-header">
    <template #header>
      this text is red
    </template>
  </el-drawer>
</template>
```

I honestly don't think this will be used much and i understand if we dont merge this PR.

btea:
If you want to customize class safety, you can add it yourself through the provided slot. The newly added prop seems redundant.

Anyway, thank you very much for your contribution! ❤️

Dsaquel:
@btea completely understandable.
And i just realize that we can find a way to do this like:
```css
/* global.css */

.my-custom-drawer-style {
  :deep(.el-drawer__header) {
    color: red
  }
  :deep(.el-drawer__body) {
    color: red
  }
}
```
and use it with `class="my-custom-drawer-style"` of `drawer` or `dialog`.

Thanks for your feedback ❤️

btea:
Yeah, this is the solution I often use in my projects. 👍

btea:
@Dsaquel Referring to the comments [here](https://github.com/element-plus/element-plus/discussions/19060#discussioncomment-11504073), I think this PR seems to be reopened. The newly added prop is really convenient when used in combination with other style libraries.

Dsaquel:
@btea Yeah maybe this will be usefull with unocss/tailwind, completely forgot about that.

Do we need test case ?

btea:
> Do we need test case ?

Yeah, I think it's better to add it.

I found that the corresponding prop name has been changed in the current document, removing `-name`, but it is not in the code, so perhaps it needs to be synchronized.

Dsaquel:
ready to review 👍

btea (approved):
Thanks!

warmthsea (approved):
LGTM

docs/en-US/component/dialog.md
```diff
@@ -154,6 +154,9 @@ dialog/events
 | top                        | value for `margin-top` of Dialog CSS, default is 15vh                                                | ^[string]                           | ''      |
 | modal                      | whether a mask is displayed                                                                          | ^[boolean]                          | true    |
 | modal-class                | custom class names for mask                                                                          | ^[string]                           | —       |
+| header-class ^(2.9.2)      | custom class names for header slot                                                                   | ^[string]                           | —       |
+| body-class ^(2.9.2)        | custom class names for default slot                                                                  | ^[string]                           | —       |
+| footer-class ^(2.9.2)      | custom class names for footer slot                                                                   | ^[string]                           | —       |
```
tolking, line 159: I feel that the description here may lead to misunderstandings, as these attributes actually apply to the wrapper of the slot

packages/components/dialog/src/dialog.vue
```diff
@@ -60,7 +63,9 @@
                   :title-id="titleId"
                   :title-class="ns.e('title')"
                 />
-                <slot v-else name="title" />
+                <slot v-else name="title">
+                  <!-- DEPRECATED SLOT -->
+                </slot>
```
tolking, line 68: Why do we need to modify this place? I do not think it is necessary. We have already reminded the user through `useDepreciated`.

docs/en-US/component/dialog.md
```diff
@@ -154,6 +154,9 @@ dialog/events
 | top                        | value for `margin-top` of Dialog CSS, default is 15vh                                                | ^[string]                           | ''      |
 | modal                      | whether a mask is displayed                                                                          | ^[boolean]                          | true    |
 | modal-class                | custom class names for mask                                                                          | ^[string]                           | —       |
+| header-class ^(2.9.2)      | custom class names for header slot                                                                   | ^[string]                           | —       |
```
tolking, line 157: ^(2.9.2) -> ^(2.9.3)

packages/components/dialog/src/dialog.vue
Dsaquel: I saw in `drawer.vue` we do this, wanted to be consistent. I agree we shouldnt affect other things that isnt related to the PR :+1:

packages/components/dialog/src/dialog-content.ts
```diff
@@ -27,6 +27,18 @@ export const dialogContentProps = buildProps({
    * @description whether the Dialog takes up full screen
    */
   fullscreen: Boolean,
+  /**
+   * @description custom class names for header slot
```
tolking, line 31: The description here is different from the document

tolking (approved):

PR #19255, opened 2024-12-12
chore: [tree] remove unused arg

zzjiaxiang:
ev.target.checked  seems to be always equal to value


btea (approved):
Thanks!

warmthsea (approved):
LGTM, Thanks!

PR #19277, opened 2024-12-14
chore: [message-box] remove @ts-nocheck directive

zzjiaxiang:


packages/components/message-box/src/index.vue
```diff
@@ -286,16 +285,16 @@ export default defineComponent({
...
       inputErrorMessage: '',
-      message: null,
+      message: '',
       modalFade: true,
       modalClass: '',
       showCancelButton: false,
       showConfirmButton: true,
       type: '',
-      title: undefined,
+      title: '',
```
tolking, line 297: > L39 `v-if="title !== null && title !== undefined"`

This modification will affect the previous judgment. I think it may be less error-prone to modify the type than to modify the initial value.

packages/components/message-box/src/index.vue
```diff
@@ -458,8 +458,8 @@ export default defineComponent({
     }
 
     const getInputElement = () => {
-      const inputRefs = inputRef.value.$refs
-      return (inputRefs.input || inputRefs.textarea) as HTMLElement
+      const inputRefs = inputRef?.value?.$refs
```
tolking, line 461:
```suggestion
      const inputRefs = inputRef.value?.$refs
```

tolking (approved):

btea (approved):
Thanks!

PR #19278, opened 2024-12-14
docs: add Playground link in navbar

Dsaquel:
I think it would be useful to have an Playground link in the navbar.

image

Edit: I dont know how can i test the chinese doc version and if someone can translate "Playground" into chinese it would be great ! Thanks 😀


btea:
> Edit: I dont know how can i test the chinese doc version and if someone can translate "Playground" into chinese it would be great ! Thanks 😀

Then we can just translate it simultaneously on Crowdin.

btea (approved):
LGTM

PR #19280, opened 2024-12-15
chore: [table] optimization type check

zzjiaxiang:


packages/components/table/src/filter-panel.vue
```diff
@@ -50,8 +50,7 @@
           :class="[
             ns.e('list-item'),
             {
-              [ns.is('active')]:
-                filterValue === undefined || filterValue === null,
+              [ns.is('active')]: isNil(filterValue),
```
tolking, line 53: I recommend using isPropAbsent instead of isNil

tolking (approved):

btea (approved):
Thanks!

PR #19454, opened 2024-12-30
docs: `Select` default placeholder is "Select"

aghArdeshir:

No relative issues (and I couldn't create a new issue through GitHub). This MR fixes a misleading part of the Docs: https://element-plus.org/en-US/component/select.html In the "Select" guides, the default value of the Select is mentioned to be "-" but it is actually "Select"


btea:
Thank you for your contribution, but if a default value is given, no matter what language the locale is set to, the default placeholder of select will be `Select`.
https://github.com/element-plus/element-plus/blob/7d76dea667f0cefd5608192b2b9a58f30aeaf33a/packages/components/select/src/useSelect.ts#L345

aghArdeshir:
@btea thanks. You are totally right. This was a thoughtless change, and I meant to only update the docs, shouldn't have touched the default props.

And I didn't pay attention to the fact that the default placeholder will be different per language, which makes it a bit hard to put in API docs.

I'll close this MR, as I can't think of any good way to put this in the Docs 
image

PR #19539, opened 2025-01-07
docs: [collapse] change event parameter type error

zzjiaxiang:
image

### DIFF
--- a/docs/examples/collapse/custom-icon.vue
+++ b/docs/examples/collapse/custom-icon.vue
@@ -58,6 +58,7 @@
 import { ref } from 'vue'
 import { CaretRight } from '@element-plus/icons-vue'
 import type { CollapseModelValue } from 'element-plus'
+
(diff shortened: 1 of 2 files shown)


docs/examples/collapse/custom-icon.vue
```diff
@@ -57,9 +57,9 @@
 <script lang="ts" setup>
 import { ref } from 'vue'
 import { CaretRight } from '@element-plus/icons-vue'
-
+import type { CollapseModelValue } from 'element-plus'
 const activeNames = ref(['1'])
```
Dsaquel, line 61: Maybe should add an escape line here
```suggestion
import type { CollapseModelValue } from 'element-plus'

const activeNames = ref(['1'])
```

btea (approved):
Thanks!

btea (approved):
Thanks!

PR #19544, opened 2025-01-07
ci: run pkg.new when push

warmthsea:
https://github.com/element-plus/element-plus/commits/test-pr.new/

I tested it on other branches and it worked fine.
I think it might be related to my account being banned.


btea (approved):

btea:
It doesn't seem to be the cause, the problem still exists.

PR #19638, opened 2025-01-14
docs: add code file-title and icon

warmthsea:
### [Before](https://element-plus.org/en-US/guide/quickstart.html)

image

### [After](https://preview-19638-element-plus.surge.sh/en-US/guide/quickstart.html)

image


btea (approved):
Looks great.

PR #19671, opened 2025-01-16
fix(ci): move all bundle to docs dist

Dsaquel:
cc @btea

Was only move files but not folders with `mv dist/element-plus/*.*`


btea:
Thanks a lot, I hadn't noticed the syntax difference before.

btea (approved):

PR #19687, opened 2025-01-18
docs: [tooltip] fix type error

zzjiaxiang:
image

[DOMRect.fromRect()](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect/fromRect_static#rectangle) All properties default to `0`

### DIFF
--- a/docs/examples/tooltip/virtual-trigger.vue
+++ b/docs/examples/tooltip/virtual-trigger.vue
@@ -16,19 +16,11 @@ import { onMounted, onUnmounted, ref } from 'vue'
 
 const visible = ref(false)
 const triggerRef = ref({
-  getBoundingClientRect() {
-    return position.value
-  },
+  getBoundingClientRect: () => position.value,
 })
+const position = ref(DOMRect.fromRect())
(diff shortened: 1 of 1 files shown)


warmthsea (approved):
LGTM, Thanks!

warmthsea:
After the change, it seems that Nuxt will report an error. 🤔

docs/examples/tooltip/virtual-trigger.vue
```diff
@@ -16,26 +16,17 @@ import { onMounted, onUnmounted, ref } from 'vue'
 
 const visible = ref(false)
 const triggerRef = ref({
-  getBoundingClientRect() {
-    return position.value
-  },
+  getBoundingClientRect: () => position.value,
 })
+const position = ref(DOMRect.fromRect())
```
tolking, line 21: > rendering pages...ReferenceError: DOMRect is not defined

This causes SSR, SSG errors because `DOMRect` does not exist at server-side rendering.

tolking (approved):

PR #19694, opened 2025-01-20
docs: [dropdown] add `button-props` prop

smallbonelu:
add the missed button-props attribute for the dropdown component


btea (approved):
Thanks!

PR #19697, opened 2025-01-20
docs: [scrollbar] fix type error

zzjiaxiang:
image


docs/examples/scrollbar/manual-scroll.vue
```diff
@@ -17,26 +17,27 @@
 
 <script lang="ts" setup>
 import { onMounted, ref } from 'vue'
-import { ElScrollbar } from 'element-plus'
+import { isNumber } from 'lodash-unified'
```
btea, line 20: To be compatible with playground, we should not import lodash-unifiend, we can use typeof directly.

docs/examples/scrollbar/manual-scroll.vue
```diff
@@ -17,26 +17,27 @@
...
+
+import type { ScrollbarInstance } from 'element-plus'
+import type { Arrayable } from '@element-plus/utils'
```
btea, line 23:
```suggestion
type Arrayable<T> = T | T[]
```

DDDDD12138:
I think the event type should be `((value: number) => void) | ((value: number[]) => void)`, rather than `(value: Arrayable<number>) => xxxx`.

btea:
From the implementation point of view, the two are equivalent here. In order to be consistent with the internal type declaration, `(value: Arrayable<number>) => xxxx` is reasonable.

DDDDD12138:
This might not be user-friendly. Even though I know clearly that `isRange` is false, I still have to check if it is an array.

zzjiaxiang:
I use `as` replace type check

btea (approved):
Thanks!

PR #19712, opened 2025-01-21
docs: [input-number] fix type error

zzjiaxiang:
image
change should be `(currentValue: number | undefined, oldValue: number | undefined) => void`


warmthsea (approved):
LGTM, Thanks!

Issue #11930, opened 2023-03-09
[Component] [select] el-select default-first-option 属性在搜索有延迟时默认选中第一个值ui错误


qjinle:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.47`
- Element Plus Version: `2.2.36`
- Browser / OS: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/110.0.0.0 Safari/537.36`
- Build Tool: `Vite`

### Related Component

- `el-select`

### Reproduction Link

[...]

PR #19806, opened 2025-02-10
fix(components): [select] typed value causes `default-first-option` to fail

noki-maker:
Fix: `default-first-option` when `optionsArray` changed (#11930)

### DIFF
--- a/packages/components/select/__tests__/select.test.ts
+++ b/packages/components/select/__tests__/select.test.ts
@@ -2975,6 +2975,69 @@ describe('Select', () => {
 
       vi.useRealTimers()
     })
+
+    it.only('should work when options changed', async () => {
+      vi.useFakeTimers()
+
+      const wrapper = _mount(
+        `
+        <el-select v-model="value" filterable default-first-option :filterMethod="filterMethod">
+          <el-option
+            v-for="option in options"
+            :key="option.value"
+            :value="option.value"
+            :label="option.label"
+          />
+        </el-select>
+      `,
+        () => ({
+          value: '',
+          options: [
+            {
+              value: 'a',
+              label: 'a',
+            },
+          ],
+        }),
+        {
+          methods: {
+            filterMethod() {
+              setTimeout(() => {
+                this.options = [
+                  {
+                    value: 0,
+                    label: 0,
+                  },
+                ]
+              }, 200)
+            },
+          },
+        }
+      )
+
+      const select = wrapper.findComponent({ name: 'ElSelect' })
+      const selectVm = select.vm as any
+      const input = wrapper.find('input')
+      input.element.focus()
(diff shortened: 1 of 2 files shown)


packages/components/select/src/useSelect.ts
```diff
@@ -406,19 +406,17 @@ export const useSelect: useSelectType = (props: ISelectProps, emit) => {
...
-    (val) => {
-      if (isNumber(val) && val > -1) {
-        hoverOption.value = optionsArray.value[val] || {}
-      } else {
-        hoverOption.value = {}
-      }
-      optionsArray.value.forEach((option) => {
-        option.hover = hoverOption.value === option
-      })
+  watch([() => states.hoveringIndex, () => optionsArray.value], (val) => {
```
btea, line 409:
```suggestion
  watch([() => states.hoveringIndex, optionsArray], ([val]) => {
```

btea:
Could you add a test case?

noki-maker:
> Could you add a test case?

Yes.

packages/components/select/__tests__/select.test.ts
```diff
@@ -2975,6 +2975,69 @@ describe('Select', () => {
 
       vi.useRealTimers()
     })
+
+    it('should work when options changed', async () => {
+      vi.useFakeTimers()
+
+      const wrapper = _mount(
+        `
+        <el-select v-model="value" filterable default-first-option :filterMethod="filterMethod">
```
btea, line 2984: Perhaps it would be better to use `remote-method` as in the #11930 example.

noki-maker: That is a good suggestion.

btea (approved):
Thanks!

packages/components/select/src/useSelect.ts
```diff
@@ -406,19 +406,16 @@ export const useSelect: useSelectType = (props: ISelectProps, emit) => {
...
+  watch([() => states.hoveringIndex, optionsArray], ([val]) => {
+    if (isNumber(val) && val > -1) {
+      hoverOption.value = optionsArray.value[val] || {}
+    } else {
+      hoverOption.value = {}
     }
-  )
+    optionsArray.value.forEach((option) => {
+      option.hover = hoverOption.value === option
+    })
```
warmthsea, line 417:
```suggestion
    props.defaultFirstOption &&
      optionsArray.value.forEach((option) => {
        option.hover = hoverOption.value === option
      })
```

warmthsea: 😂 
Looks need revert change
~~props.defaultFirstOption &&~~

warmthsea (approved):
LGTM, Thanks!

PR #19816, opened 2025-02-10
refactor(components): [menu] change to `script setup` writing  and optimizing code

zzjiaxiang:


btea (approved):
Thanks!

packages/components/menu/src/menu-collapse-transition.vue
```diff
@@ -3,60 +3,52 @@
     <slot />
   </transition>
 </template>
-<script lang="ts">
-import { defineComponent } from 'vue'
+<script lang="ts" setup>
 import { useNamespace } from '@element-plus/hooks'
 import { addClass, hasClass, removeClass } from '@element-plus/utils'
```
warmthsea, line 9:
```suggestion
```

warmthsea (approved):
LGTM, Thanks!

warmthsea (commented):
https://github.com/element-plus/element-plus/blob/dev/packages/components/menu/src/menu-item.vue

This file can also be modified together.

packages/components/menu/src/menu-item-group.vue
warmthsea, line 1: 1、this function need `defineExpose`.

```
return {
  parentMenu,
  rootMenu,
  active,
  nsMenu,
  nsMenuItem,
  handleClick,
}
```

2、@ts-nocheck fix would be better.

packages/components/menu/src/menu-item.vue
warmthsea, line 1: 1、this function need `defineExpose`.

```
return {
  parentMenu,
  rootMenu,
  active,
  nsMenu,
  nsMenuItem,
  handleClick,
}
```

2、@ts-nocheck fix would be better.

warmthsea (commented):
1、this function need `defineExpose`.

```
return {
  parentMenu,
  rootMenu,
  active,
  nsMenu,
  nsMenuItem,
  handleClick,
}
```

2、@ts-nocheck fix would be better.

packages/components/menu/src/menu.ts
warmthsea, line 1:
```ts
if (i !== -1) {
  openedMenus.value.splice(i, 1)
}
```
 🔽

```ts
if (i !== -1) openedMenus.value.splice(i, 1)
```

...

I think this kind of style doesn’t need to be changed, the previous readability is very good.

zzjiaxiang:
1. `
return {
      parentMenu,
      rootMenu,
      active,
      nsMenu,
      nsMenuItem,
      handleClick,
    }`
Is it unnecessary to expose methods that have not been called through ref and have not been declared in the document?

2.   `export interface MenuItemRegistered {
  index: string
  indexPath: string[]
  active: boolean
}
` 
`  index: {
    type: definePropType<string | null>([String, null]),
    default: null,
  },`
These two type declarations are inconsistent. We should determine the `index` type and fix it in the next PR

warmthsea:
> 1. ` return { parentMenu, rootMenu, active, nsMenu, nsMenuItem, handleClick, }`
>    Is it unnecessary to expose methods that have not been called through ref and have not been declared in the document?

Yes, some users will use it even if the documentation does not record it.

packages/components/menu/src/menu-item.vue
```diff
@@ -55,63 +54,50 @@ import { menuItemEmits, menuItemProps } from './menu-item'
...
-      nsMenuItem,
-      handleClick,
-    }
-  },
+onBeforeUnmount(() => {
+  subMenu.removeSubMenu(item)
+  rootMenu.removeMenuItem(item)
 })
+
+defineExpose({ parentMenu, rootMenu, active, nsMenu, nsMenuItem, handleClick })
```
warmthsea, line 102:
```suggestion
defineExpose({
  parentMenu,
  rootMenu,
  active,
  nsMenu,
  nsMenuItem,
  handleClick,
})
```

warmthsea (approved):

Issue #18433, opened 2024-10-01
[Component] [table] component does not properly clean up side effects after unmounting


diyan-slavov-infoleven:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.10`
- Element Plus Version: `2.8.4`
- Browser / OS: `any`
- Build Tool: `Vite`

### Related Component

- `el-table`

### Reproduction Link

[CodeSandbox](https://codesandbox.io/p/devbox/trusting-resonance-9rz3y9)

### Steps to reproduce

Run `npm run test`

The test completes successfully, but the `doLayout` gets called after the test has finished and it causes an uncaught exception that is reported by Vitest.

### What is Expected?

When the toggle button is clicked, the ElTable component will be unmounted from the document.

It should be expected that when the component unmounts, it properly clears all of the side effects that it has registered during its lifecycle.

### What is actually happening?

The `doLayout` function inside ElTable is debounced with a 50ms delay as can be seen here https://github.com/element-plus/element-plus/blob/18a77d5829df0b7db064d634cd45645befe77435/packages/components/table/src/table.vue#L281

When this function is called, it is going to run later, even if the <ElTable /> component has been unmounted from the document. 

This causes issues in test environments, because if the test completes before the debounced doLayout function is called, it will be invoked after the environment (jsdom or alike) has been torn down, which results in uncaught Vitest errors after the test suit has been completed.

### Additional comments

The function that debounce returns has a `cancel` method, which should be called from the `onUnmounted` hook inside <ElTable /> component.

It should either be cancelled, or the implementation of the `doLayout` should check if the component has not unmounted, before performing the side effects.

This is how I imagine a fix would look like

```js
const debouncedUpdateLayout = debounce(doLayout, 50)
onUnmounted(() => debouncedUpdateLayout.cancel())
```

The `useLockscreen` hook, which is used inside the Dialog component also has the same problem

https://github.com/element-plus/element-plus/blob/18a77d5829df0b7db064d634cd45645befe77435/packages/hooks/use-lockscreen/index.ts#L50

The `setTimeout` call can also run after the browser environment has been torn down, leading to the same problem. The code uses nullish coalescent operator for references ( document?.body and subsequently && document ), however, in non-browser environment simply referring to a `document` variable is going to result in a reference error.

For a possible solution, we can just check if the document variable is undefined

```js
  const cleanup = () => {
    setTimeout(() => {
      if (typeof document === 'undefined') {
        return
      }

[...]

PR #19837, opened 2025-02-11
fix(components): [table] compatible with window non-existence

purepear:
ElTable's `doLayout()` is debounced and scheduled (relying on `setTimeout()`) which causes problems in testing(Vitest).

Sometimes the testing environment(jsdom) is destroyed before doLayout() executes which causes error due to requestAnimationFrame not existing. 

To mitigate this issue we can check if `window` is defined which is a good indication that jsdom has not been thorn down yet.

This PR is similar in spirit to https://github.com/element-plus/element-plus/pull/18445 and fixes very similar issue to the one described in https://github.com/element-plus/element-plus/issues/18433

image

### DIFF
--- a/packages/components/table/src/table/style-helper.ts
+++ b/packages/components/table/src/table/style-helper.ts
@@ -113,6 +113,10 @@ function useStyle<T>(
       layout.updateElsHeight()
     }
     layout.updateColumnsWidth()
+
+    // When the test case is running, the context environment simulated by jsdom may have been destroyed,
+    // and window.requestAnimationFrame does not exist at this time.
+    if (typeof window === 'undefined') return
     requestAnimationFrame(syncPosition)
   }
   onMounted(async () => {


purepear:
Thanks for the suggestion, @Dsaquel. I'm not super familiar with how `doLayout()` is scheduled but it seems like it's triggered on events (resize, data watchers, others?). `useRafFn` on the other hand seem more akin to continuous render loop like in games (even has fps param). Not sure if it's best fit for purpose - events controlling a loop with pause/resume 🤔 and it'll be a good chunk of reworking quite fiddly code.

Dsaquel:
the error occurs because `requestAnimationFrame` is being access while window not exist.
Its like writing this:
 ```ts
    window.requestAnimationFrame(syncPosition)
```
For this use case i wanna use `useRafFn` because it wrap `requestAnimationFrame` and run only in an browser (`window`) context. My thought behind is to let vueuse check the compatibility for window properties being access in this context of error.

**Comment by purepear** [comment]
Hi @Dsaquel, I understand that useRafFn checks for `window` internally but it does more than that which I don't think fits our use case. It's not just a `requestAnimationFrame` wrapper that checks for `window`, it immediately starts an indefinite loop that needs to be explicitly paused and resumed in order to achieve what we want. This just doesn't seem the right way to fix the issue but I'm happy to be proven wrong :)

**Comment by purepear** [comment]
Just to clarify what is happening:
1. ElTable onBeforeUnmounted clears any scheduled calls to `doLayout` (by cancelling the debounce `doLayoutDebounced.cancel()`) 
2. ElTableColumn onBeforeUnmounted runs after that and schedules new calls to `doLayout` (by calling scheduleLayout -> which calls `doLayoutDebounced`)
3. The `doLayout` calls are delayed 50ms which sometimes falls in the interval where the jsdom environment is destroyed but the vitest has not finished yet.

We could mitigate this by doing `doLayoutDebounced.cancel()` on ElTable onUnmounted. That will be executed after the ElTableColumn onBeforeUnmounted and will clear the scheduled calls. I'm wish we could just cancel once on onUnmounted but then there will be risk of scheduled call when the component is unmounted but onUnmounted has not been executed yet? 🤔  Therefore we might have to have this:

```ts
    onBeforeUnmount(() => {
      debouncedUpdateLayout.cancel();
    });
    onUnmounted(() => {
      debouncedUpdateLayout.cancel();
    })
```

Please let me know if you prefer this over the `if (typeof window === undefined) return` solution (which seems ok to me)

Dsaquel:
I really don't know what else to say, i thought i was clear in my explication.
Lets see what others people think about it.

purepear:
@Dsaquel I created a very simple representation of doLayout & requestAnimationFrame. Would you be able to help me understand how we could take advantage of `useRafFn` by adjusting the code and sharing the link to your implementation? 🙏 

[Link to vueuse playground](https://play.vueuse.org/#N4IgDghgxg1hDmBTAziAXAbVAOwgW0XRADcBXRAWgBNE8BLEAGhGQHtSAnKQtEU7MDHhMQNZFA50wAFzqtsRAGrkABABFadFQAoAFhAA2AMxV1sKgGIdE2KLoCUp5CogqaxRAdZgz8FaVkDOmkATxcDLwB3ZxD2FWlWFUjJaUQVAFVsOg8OZEMVZTSAGToAIw4ISRQVI1YOAtUAJhUAMhUAZhFOAyJdaWkwZDQAemH+QXgAOihWPGGyShp6YaDS4bMaAA9JvAArVABfRhx8HhAAAQXSZERh5H1rKhE2Tm4icaERMQkpWXkiADKD0QVAaiHSN38gWCdBQky6HB6vD6AyGow+Uxmc0u5Gut3ulRB5wAjAAGSbEyak9bYLY7fYgI4nAhEHGIPHDGbWZ7sLhnDFfFA-GRyBS8ADCrAiiCgf3MrBMKBu2Fk+UKKkleDArGQMPkGWhshQCKRIBRgxGYwEQmms2GbI5XMQJPJlOpG0Q2z2hwAuswjHQDMa0KAAIJgMCTBboUCpLUGCCpIgAHio2QAOuYVPBJFQALzpkAAP0LmZUKjACe4BZAMxVNmkFG49fqwVoyCbDcQHFLWd0NfE1hsvfLtRVNbw8lYvYAfGWVKmM1ny6V4DWgvA+hQAKyk0luSowNBUQ87vcj8sVmuNC8ruo0Dg1jjsWkgi9z5cL5CQcypTbSGsABZNgMQsZ2AYAkhUA4jhUCCVF0aCDmTO4fw-csULTYgP0w7IZ2eYVpCIOgtTqaQ4JUawTAOGpnzwFQAHIFgYzMSO1DhyMgvEACUICMCxsBoow6MYh0bk5OpEBY7BMzrZByMiFQ80oxAjG0Ul7Fk+R5IQpSVLUjTM0zIx+FlUU3FYIoIFiAJtEcYB52sABHch5NDLI8ETUUrFObQ7KUmc4PnctIijQxVGUyINlYUKzGwbsAHU6CoaRdGChCwoMCKkmi2LsHijgAAlEDoTdpHnA5NMEozcCoKgAFEPBVEp5JsbttELaxdQAL0QQtGAsqybOkexGRgoA)

packages/components/table/src/table/style-helper.ts
```diff
@@ -109,6 +109,10 @@ function useStyle<T>(
   })
 
   const doLayout = () => {
+    // When the test case is running, the context environment simulated by jsdom may have been destroyed,
+    // and the window does not exist at this time.
+    if (typeof window === 'undefined') return
+
```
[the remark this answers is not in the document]
purepear: Hi @btea, thanks for looking into it!

Unfortunately [useRafFn](https://vueuse.org/core/useRafFn/) works differently than [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame)
* `requestAnimationFrame(callback)` calls the callback **only once** on the next frame
* `useRafFn(callback)` calls the callback **on every** frame unless paused

Internally `useRafFn` is doing something like this:
```js
function useRafFn(callback) {
  function loop() {
    callback()
    requestAnimationFrame(loop)
  }
  
  loop()
}
```

To demonstrate the difference, here is a demo of a counter:
* [using requestAnimationFrame](https://play.vueuse.org/#N4IgDghgxg1hDmBTAziAXAbVAOwgW0XRADcBXRAWgBNE8BLEAGhGQHtSAnKQtEU7MDHhMQNZFA50wAFzqtsRAGrkABABFadFQAoAFhAA2AMxV1sKgGIdE2KLoCUp5CogqaxRAdZgz8FaVkDOmkATxcDLwB3ZxD2FWlWFUjJaUQVAFVsOg8OZEMVZTSAGToAIw4ISRQVI1YOAtUAJhUAMhUAZhFOAyJdaWkwZDQAemH+QXgAOihWPGGyShp6YaDS4bMaAA9JvAArVABfRhx8HhAAAQXSZERh5H1rKhE2Tm4icaERMQkpWXkiADKD0QVAaiHSN38gWCdBQky6HB6vD6AyGow+Uxmc0u5Gut3ulRB5wAjAAGSbEyak9bYLY7fYgI4nAhEHGIPHDGbWZ7sLhnDFfFA-GRyBS8ADCrAiiCgf3MrBMKBu2Fk+UKKkleDArGQMPkGWhshQCKRIBRgxGYwEQmms2GbI5XMQJPJlOpG0Q2z2hwAuswjHQDMa0KAAIJgMCTBboUCpLUGCCpIgAHio2RUYAT3AAvAAdEAzFU2aQUbhF+rBWjIUvFxAcfMqXR5lgSRA2Bu1FXNvDyVgN1KbaTNgAsmwM+YAfLnzCpU9kJ8BgCoZvxpCoDgdk8M08Qp9gtzuJ89hdIiHQtXU10vrCYDjUOLMVAByBZP6fn7UcK-+G4AJQgRgWNgd5GA+eDPg6NycnUiBvtg06Fsga4riqKjZioN7aKS9jTtORj8LKooqMgIS2AACjqerYNojjANOKjLuwKpRoY5AANRsdOBy4dg+G2HKbisEUECxAENEqHRM7WAAjuQSGhlkeCJqKVinNoJHkZRco4cBPFUEJInsNINGMhuQA)
* [using useRefFn](https://play.vueuse.org/#N4IgDghgxg1hDmBTAziAXAbVAOwgW0XRADcBXRAWgBNE8BLEAGhGQHtSAnKQtEU7MDHhMQNZFA50wAFzqtsRAGrkABABFadFQAoAFhAA2AMxV1sKgGIdE2KLoCUp5CogqaxRAdZgz8FaVkDOmkATxcDLwB3ZxD2FWlWFUjJaUQVAFVsOg8OZEMVZTSAGToAIw4ISRQVI1YOAtUAJhUAMhUAZhFOAyJdaWkwZDQAemH+QXgAOihWPGGyShp6YaDS4bMaAA9JvAArVABfRhx8HhAAAQXSZERh5H1rKhE2Tm4icaERMQkpWXkiADKD0QVAaiHSN38gWCdBQky6HB6vD6AyGow+Uxmc0u5Gut3ulRB5wAjAAGSbEyak9bYLY7fYgI4nAhEHGIPHDGbWZ7sLhnDFfFA-GRyBS8ADCrAiiCgf3MrBMKBu2Fk+UKKkleDArGQMPkGWhshQCKRIBRgxGYwEQmms2GbI5XMQJPJlOpG0Q2z2hwAuswjHQDMa0KAAIJgMCTBboUCpLUGCCpIgAHio2RUYAT3AAvAAdEAzFU2aQUbhF+rBWjIUvFxAcfMqXR5lgSRA2Bu1FXNvDyVgN1KbaTNgAsmwM+YAfLnzCpU9kJ8BgCoZvxpCoDgdk8M08Qp9gtzuJ89hdIiHQtXU10vrCYDjUOLMVAByBZP6fn7UcK-+G4AJQgRgWNgd5GA+eDPg6NycnUiBvtg06Fsga4riqKjZioN7aKS9jTtORj8LKooqMgIS2AACjqerYNojjANOKjLuwKpRoY5AANRsdOBy4dg+G2HKbisEUECxAENEqHRM54v+gHUSR5GUXKOHATxVBCSJ7DSDRjIbkAA)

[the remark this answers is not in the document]
btea: Yeah, so I think your judgment about `window` can be placed directly before the requestAnimationFrame call.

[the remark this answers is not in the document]
Dsaquel: @purepear Hi, im sorry i didn't check deeply what `useRafFn` was doing. Sorry for bothering.

packages/components/table/src/table/style-helper.ts
```diff
@@ -113,7 +113,12 @@ function useStyle<T>(
       layout.updateElsHeight()
     }
     layout.updateColumnsWidth()
-    requestAnimationFrame(syncPosition)
+
+    // When the test case is running, the context environment simulated by jsdom may have been destroyed,
+    // and window.requestAnimationFrame does not exist at this time.
+    if (typeof window !== 'undefined') {
+      requestAnimationFrame(syncPosition)
+    }
```
btea, line 121: Maybe line 127 should also be added.

purepear: Line 127 is within onMounted() which shouldn't cause a debounce/delayed call after (after test env destroy).

btea: OK, you can install element-plus from this link to verify whether your local test can pass.

https://github.com/element-plus/element-plus/pull/19837#issuecomment-2651810474

purepear: Thanks for the hint @btea! 

I've already tested the change locally by patching node_modules directly:
* increased the vitest delay between jsdom destroyed and vitest exits
* applied this change to the element-plus dist file :)

Local tests are passing 🥳

btea (approved):

btea:
@Dsaquel Do you have any other suggestions? If not, let's merge it.

Dsaquel:
@btea Thanks for consideration, i honestly prefer have consistency with https://github.com/element-plus/element-plus/pull/18445 and do a one line check.
Like you said @purepear here https://github.com/element-plus/element-plus/pull/19837#discussion_r1956115211
Sorry again @purepear for my unappropriated attitude.

btea:
Thank you for your reminder. It’s also good to keep the format consistent with #18445. What do you think? @purepear

purepear:
Thank you both!
Don't sweat it @Dsaquel 🙂 ❤️ 

I updated to 1-line early return

btea:
@purepear @Dsaquel  Thanks!

PR #19846, opened 2025-02-12
chore: [tree] export types

zzjiaxiang:


warmthsea (commented):
This will cause old users to report errors when using the examples. 🤔 

image

btea (approved):
Thanks!

warmthsea (approved):

PR #19865, opened 2025-02-14
chore: change create issue use

warmthsea:
### Before:

image

### After:

https://github.com/warmthsea/create-issue-test/issues

image

image


btea (approved):
👍

PR #19887, opened 2025-02-17
refactor(components): [time-picker] refactor code

zzjiaxiang:


btea (commented):

But I think this kind of refactoring is not necessary, because there is no obvious benefit and too much of this kind of refactoring information will overwrite the original commit record, making it difficult to clearly know the original record of fixing the problem or adding functions.

PR #19889, opened 2025-02-17
feat(components): [slider] add `persistent` prop

zzjiaxiang:


packages/components/slider/src/slider.vue
```diff
@@ -50,6 +51,7 @@
         :vertical="vertical"
         :tooltip-class="tooltipClass"
         :placement="placement"
+        :persistent="persistent"
```
btea, line 54: It can be obtained through `inject`, just like `showTooltip`, instead of passing it through prop.

packages/components/slider/src/button.vue
```diff
@@ -59,11 +59,16 @@ const initData = reactive<SliderButtonInitData>({
   oldValue: props.modelValue,
 })
 
+const btnPersistent = computed(() =>
```
btea, line 62:
```suggestion
const tooltipPersistent = computed(() =>
```

docs/en-US/component/slider.md
```diff
@@ -85,33 +85,34 @@ slider/show-marks
...
+| marks                       | marks, type of key must be `number` and must in closed interval `[min, max]`, each mark can custom style                                        | ^[object]`SliderMarks`                                                                                                                                                      | —       |
+| validate-event              | whether to trigger form validation                                                                                                              | ^[boolean]                                                                                                                                                                  | true    |
+| persistent ^(2.9.5)         | when slider inactive and `persistent` is `false` , popconfirm will be destroyed. `persistent` always be `false` when `show-tooltip ` is `false` | ^[boolean]                                                                                                                                                                  | true    |
```
btea, line 114: Perhaps the description here should be changed.
```suggestion
| persistent ^(2.9.5)         | when slider tooltip inactive and `persistent` is `false` , popconfirm will be destroyed. `persistent` always be `false` when `show-tooltip ` is `false` | ^[boolean]                                                                                                                                                                  | true    |
```

packages/components/slider/src/slider.ts
```diff
@@ -168,6 +168,13 @@ export const sliderProps = buildProps({
     type: Boolean,
     default: true,
   },
+  /**
+   * @description when slider inactive and `persistent` is `false` , popconfirm will be destroyed. `persistent` always be `false` when `show-tooltip ` is `false`
```
btea, line 172: This should be synchronized, and the table format in the above document may need to be formatted.

btea (approved):
Thanks!

warmthsea (approved):

PR #19912, opened 2025-02-20
chore: [dropdown] remove @ts-nocheck & fix type error

warmthsea:
This function is not used in the project.

https://github.com/element-plus/element-plus/blob/37f3447eca7430488725573c76a20ee597cc4f54/packages/components/dropdown/src/useDropdown.ts#L19C1-L23C4


packages/components/dropdown/src/dropdown.ts
```diff
@@ -130,6 +129,7 @@ export const dropdownProps = buildProps({
    */
   role: {
     type: String,
+    values: roleTypes,
```
btea, line 132: Maybe we shouldn't restrict the list to specific values.

warmthsea: https://github.com/element-plus/element-plus/blob/dev/packages/components/popper/src/popper.ts#L32C1-L38C12

The code here affects the type error.
This is the only way to fix it.

btea (approved):

PR #19924, opened 2025-02-21
docs(components): [message] add a "Won't close automatically" example. 

zhuft-moon:
1. docs(components): [config-provider] add default values
2. docs(components): [message] add a "Won't close automatically" example. Because the "duration" attribute was mentioned above, no example was provided.

### DIFF
--- a/docs/en-US/component/config-provider.md
+++ b/docs/en-US/component/config-provider.md
@@ -90,11 +90,11 @@ In this section, you can learn how to use Config Provider to provide experimenta
 
 | Attribute          | Description                                                                    | Type       | Default |
 | ------------------ | ------------------------------------------------------------------------------ | ---------- | ------- |
-| max                | the maximum number of messages that can be displayed at the same time          | ^[number]  | —       |
-| grouping ^(2.8.2)  | merge messages with the same content, type of VNode message is not supported   | ^[boolean] | —       |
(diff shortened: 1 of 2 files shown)


docs/en-US/component/config-provider.md
```diff
@@ -90,11 +90,11 @@ In this section, you can learn how to use Config Provider to provide experimenta
 
 | Attribute          | Description                                                                    | Type       | Default |
 | ------------------ | ------------------------------------------------------------------------------ | ---------- | ------- |
-| max                | the maximum number of messages that can be displayed at the same time          | ^[number]  | —       |
-| grouping ^(2.8.2)  | merge messages with the same content, type of VNode message is not supported   | ^[boolean] | —       |
-| duration ^(2.8.2)  | display duration, millisecond. If set to 0, it will not turn off automatically | ^[number]  | —       |
-| showClose ^(2.8.2) | whether to show a close button                                                 | ^[boolean] | —       |
-| offset ^(2.8.2)    | set the distance to the top of viewport                                        | ^[number]  | —       |
+| max                | the maximum number of messages that can be displayed at the same time          | ^[number]  | 3       |
```
btea, line 93: Thank you for your contribution, but there is no default value set here. The default values ​​of these parameters are set internally by the message component.

zhuft-moon: Ok, I have cancelled the default value change and just submit another one

warmthsea (approved):
LGTM, Thanks!

PR #19941, opened 2025-02-22
fix(components): [carousel] Resolve issue with height: 100% not working

SpanManX:
屏幕截图 2025-02-22 195540


btea:

But this change seems unnecessary. You can add style styles directly to the carousel component, which will be inherited to the component's root element by default.

PR #19943, opened 2025-02-23
docs(components): [el-tooltip] add example using append-to API

SpanManX:
docs: Add an example of the tooltip component using the append-to API .

为Tooltip 组件添加 append-to API使用的示例。

issues #19905


btea (approved):
Thanks!

warmthsea (approved):
LGTM, Thanks!

PR #19898, opened 2025-02-18
ci: update lint pr title

warmthsea:
### ❓ Why

- 1、CI has a problem, [`CommitMessage::Unqualified`](https://github.com/element-plus/element-plus/pulls?q=is%3Apr+label%3ACommitMessage%3A%3AUnqualified+is%3Aclosed) not work.
- 2、[`CommitMessage::Qualified`](https://github.com/element-plus/element-plus/pulls) seems redundant, too many tags looks messy.

PR #19464, opened 2024-12-31
fix(component): resolve 'cursor' style loss caused by 'pointer-events: none'

DDDDD12138:
Please make sure these boxes are checked before submitting your PR, thank you!

- [x] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).
- [x] Make sure you are merging your commits to `dev` branch.
- [x] Add some descriptions and refer to relative issues for your PR.

### Description

Due to pr #19320, using `pointer-events: none` to block mouse events caused the `cursor: not-allowed` style to fail. 
This PR resolves the issue by retaining the `cursor: not-allowed` style and removing the `tabindex` attribute when the component is `disabled`, achieving the same effect as `pointer-events: none`.

Related Components: input, select, mention, picker, color-picker, input-tag

### Preview
[...]

PR #19947, opened 2025-02-24
ci: fix lint-pr-title comment

warmthsea:
Rel: 
- https://github.com/element-plus/element-plus/pull/19898

😅 But lint ci `header-max-length` is `72`, some PR titles will not pass https://github.com/element-plus/element-plus/pull/19464#issuecomment-2677556553

Can't seem to set it via script https://commitlint.js.org/reference/cli.html

I'm not sure if I should modify this file. 
https://github.com/element-plus/element-plus/blob/ef8a6bb7dfc089dde8ff0cbfd0721cf042fa96ad/commitlint.config.ts#L69

### Before:

image

### After

image


btea (approved):
No modification is required. This rule should appear when the code is submitted, unless the `--no-verify` parameter is added directly to bypass it. I think it can be kept as is.

PR #20008, opened 2025-03-01
fix(components): [tree-select] use up/down/enter keys select error

### LINKED ISSUE #19990: [Component] [tree-select] el-tree-select 当persistent:false 时 ，v-model 初始化值，el-tree-select无法正确获得selectedLabel值，selectlabel 为value值,以及键盘上下控制hover
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.13`
- Element Plus Version: `2.9.5`
- Browser / OS: `Chrome 133.0.6943.127/macOS 14.7`
- Build Tool: `Vite`

### Related Component

- `el-tree-select`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtdHJlZS1zZWxlY3RcbiAgICB2LW1vZGVsPVwidmFsdWVcIlxuICAgIDpkYXRhPVwiZGF0YVwiXG4gICAgOnJlbmRlci1hZnRlci1leHBhbmQ9XCJmYWxzZVwiXG4gICAgOnBlcnNpc3RlbnQ9XCJmYWxzZVwiXG4gICAgc3R5bGU9XCJ3aWR0aDogMjQwcHhcIlxuICAvPlxuICA8ZWwtZGl2aWRlciAvPlxuICBzaG93IGNoZWNrYm94OlxuICA8ZWwtdHJlZS1zZWxlY3RcbiAgICB2LW1vZGVsPVwidmFsdWVcIlxuICAgIDpkYXRhPVwiZGF0YVwiXG4gICAgOnJlbmRlci1hZnRlci1leHBhbmQ9XCJmYWxzZVwiXG4gICAgc2hvdy1jaGVja2JveFxuICAgIHN0eWxlPVwid2lkdGg6IDI0MHB4XCJcbiAgLz5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5pbXBvcnQgeyByZWYgfSBmcm9tICd2dWUnXG5cbmNvbnN0IHZhbHVlID0gcmVmKCczLTEnKVxuXG5jb25zdCBkYXRhID0gW1xuICB7XG4gICAgdmFsdWU6ICcxJyxcbiAgICBsYWJlbDogJ0xldmVsIG9uZSAxJyxcbiAgICBjaGlsZHJlbjogW1xuICAgICAge1xuICAgICAgICB2YWx1ZTogJzEtMScsXG4gICAgICAgIGxhYmVsOiAnTGV2ZWwgdHdvIDEtMScsXG4gICAgICAgIGNoaWxkcmVuOiBbXG4gICAgICAgICAge1xuICAgICAgICAgICAgdmFsdWU6ICcxLTEtMScsXG4gICAgICAgICAgICBsYWJlbDogJ0xldmVsIHRocmVlIDEtMS0xJyxcbiAgICAgICAgICB9LFxuICAgICAgICBdLFxuICAgICAgfSxcbiAgICBdLFxuICB9LFxuICB7XG4gICAgdmFsdWU6ICcyJyxcbiAgICBsYWJlbDogJ0xldmVsIG9uZSAyJyxcbiAgICBjaGlsZHJlbjogW1xuICAgICAge1xuICAgICAgICB2YWx1ZTogJzItMScsXG4gICAgICAgIGxhYmVsOiAnTGV2ZWwgdHdvIDItMScsXG4gICAgICAgIGNoaWxkcmVuOiBbXG4gICAgICAgICAge1xuICAgICAgICAgICAgdmFsdWU6ICcyLTEtMScsXG4gICAgICAgICAgICBsYWJlbDogJ0xldmVsIHRocmVlIDItMS0xJyxcbiAgICAgICAgICB9LFxuICAgICAgICBdLFxuICAgICAgfSxcbiAgICAgIHtcbiAgICAgICAgdmFsdWU6ICcyLTInLFxuICAgICAgICBsYWJlbDogJ0xldmVsIHR3byAyLTInLFxuICAgICAgICBjaGlsZHJlbjogW1xuICAgICAgICAgIHtcbiAgICAgICAgICAgIHZhbHVlOiAnMi0yLTEnLFxuICAgICAgICAgICAgbGFiZWw6ICdMZXZlbCB0aHJlZSAyLTItMScsXG4gICAgICAgICAgfSxcbiAgICAgICAgXSxcbiAgICAgIH0sXG4gICAgXSxcbiAgfSxcbiAge1xuICAgIHZhbHVlOiAnMycsXG4gICAgbGFiZWw6ICdMZXZlbCBvbmUgMycsXG4gICAgY2hpbGRyZW46IFtcbiAgICAgIHtcbiAgICAgICAgdmFsdWU6ICczLTEnLFxuICAgICAgICBsYWJlbDogJ0xldmVsIHR3byAzLTEnLFxuICAgICAgICBjaGlsZHJlbjogW1xuICAgICAgICAgIHtcbiAgICAgICAgICAgIHZhbHVlOiAnMy0xLTEnLFxuICAgICAgICAgICAgbGFiZWw6ICdMZXZlbCB0aHJlZSAzLTEtMScsXG4gICAgICAgICAgfSxcbiAgICAgICAgXSxcbiAgICAgIH0sXG4gICAgICB7XG4gICAgICAgIHZhbHVlOiAnMy0yJyxcbiAgICAgICAgbGFiZWw6ICdMZXZlbCB0d28gMy0yJyxcbiAgICAgICAgY2hpbGRyZW46IFtcbiAgICAgICAgICB7XG4gICAgICAgICAgICB2YWx1ZTogJzMtMi0xJyxcbiAgICAgICAgICAgIGxhYmVsOiAnTGV2ZWwgdGhyZWUgMy0yLTEnLFxuICAgICAgICAgIH0sXG4gICAgICAgIF0sXG4gICAgICB9LFxuICAgIF0sXG4gIH0sXG5dXG48L3NjcmlwdD5cbiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQGxhdGVzdC9kaXN0L2luZGV4LmNzcycsICdodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvdGhlbWUtY2hhbGsvZGFyay9jc3MtdmFycy5jc3MnXS5tYXAoKHN0eWxlKSA9PiB7XG4gICAgcmV0dXJuIG5ldyBQcm9taXNlKChyZXNvbHZlLCByZWplY3QpID0+IHtcbiAgICAgIGNvbnN0IGxpbmsgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCdsaW5rJylcbiAgICAgIGxpbmsucmVsID0gJ3N0eWxlc2hlZXQnXG4gICAgICBsaW5rLmhyZWYgPSBzdHlsZVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdsb2FkJywgcmVzb2x2ZSlcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignZXJyb3InLCByZWplY3QpXG4gICAgICBkb2N1bWVudC5ib2R5LmFwcGVuZChsaW5rKVxuICAgIH0pXG4gIH0pXG4gIHJldHVybiBQcm9taXNlLmFsbFNldHRsZWQoc3R5bGVzKVxufVxuIiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IFwiRVNOZXh0XCIsXG4gICAgXCJqc3hcIjogXCJwcmVzZXJ2ZVwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwidHlwZXNcIjogW1wiZWxlbWVudC1wbHVzL2dsb2JhbC5kLnRzXCJdLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcImFsbG93SnNcIjogdHJ1ZSxcbiAgICBcImNoZWNrSnNcIjogdHJ1ZVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9ydW50aW1lLWRvbUBsYXRlc3QvZGlzdC9ydW50aW1lLWRvbS5lc20tYnJvd3Nlci5qc1wiLFxuICAgIFwiQHZ1ZS9zaGFyZWRcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvc2hhcmVkQGxhdGVzdC9kaXN0L3NoYXJlZC5lc20tYnVuZGxlci5qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L1wiLFxuICAgIFwiQGVsZW1lbnQtcGx1cy9pY29ucy12dWVcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlQDIvZGlzdC9pbmRleC5taW4uanNcIlxuICB9LFxuICBcInNjb3Blc1wiOiB7fVxufSIsIl9vIjp7fX0=)

### Steps to reproduce

1. 使用select的props，persistent设置为false
2. 初始化v-model 

键盘上下控制选项hover 时，没有悬浮显示，但是回车却能选中值
1. 使用键盘keydown 和keyup
2. 进行回车

### What is Expected?

1. 获得当前value的正确的selectlabel
2. 有明确hover显示

1. selectlabel 为value值
2. hover样式没有明确显示，但是回车却能选中值，造成视觉误差
### Additional comments

_(empty)_

Image

SpanManX:
close #19990
fix(components): [tree-select] fix background color to up/down keys, fix Enter selection bug.

**The bugs mentioned below can all be reproduced on the official website.**

**以下所提 BUG 都可在官网复现**

https://element-plus.org/zh-CN/component/tree-select.html

**Fix tree-select component bugs:**
1. No background color on element when pressing "Up/Down" keys.
2. Pressing Enter key does not select the data.

**This PR fixes issue #19990.**

**修改 tree-select 组件 BUG：**
1. 键盘按“上下键” 元素没有背景色。
2. 按 Enter 键没有选中数据。

**该 PR 修复了 issue #19990。**

### Original code test GIF images / 原始代码测试GIF图
error

Clicking the triangle iconto expand QQ20250302-021446 the tree and then pressing the "Up/Down" keys or Enter will trigger useKeydown.ts -> handleKeydown() instead of useSelect.ts -> navigateOptions().

The two different "console" logs printed in the console are to more clearly display and distinguish which function from which file was executed in the GIF.

点击三角形图标 QQ20250302-021446 展开树形后再按键盘“上下键或Enter键”，会执行 useKeydown.ts ->handleKeydown() ，而不是执行 useSelect.ts -> navigateOptions() 。

控制台打印的两个不同的 “console” 是为更直观的在 GIF 图上展示和区分执行了哪个文件下的函数。

### The test after changing the code / 修改代码后的测试
success


SpanManX:
After modifying the code, I tested all the functionalities related to the select component, and no other functionalities were affected. If any issues are found that affect other functionalities, I can make further modifications.

SpanManX:
@yujinpan 你好，请帮忙review一下这个PR。

packages/components/tree/src/tree-node.vue
```diff
@@ -132,6 +133,10 @@ export default defineComponent({
       type: Boolean,
       default: false,
     },
+    selectFocus: {
+      type: Function,
+      default: null,
+    },
```
yujinpan, line 139: The new prop doesn't seem to belong in the tree, try using the `node-expand/node-collapse` events in the TreeSelect component instead.

packages/theme-chalk/src/tree.scss
```diff
@@ -75,6 +75,9 @@
     &:hover {
       background-color: getCssVar('tree-node-hover-bg-color');
     }
+    &:has(.#{$state-prefix}hovering) {
+      background-color: getCssVar('tree-node-hover-bg-color');
+    }
```
yujinpan, line 80: Same as above.

yujinpan (dismissed):
Thanks for your contribution.

But there are still some issues to be fixed:

- Up/Down does not work in collapsed tree nodes (may have other effects when using `renderAfterExpand`)
- Left/Right does not expand/collapse tree nodes

Finally we need some tests to check the code.

SpanManX:
> Thanks for your contribution.
> 
> But there are still some issues to be fixed:
> 
> * Up/Down does not work in collapsed tree nodes (may have other effects when using `renderAfterExpand`)
> * Left/Right does not expand/collapse tree nodes
> 
> Finally we need some tests to check the code.

我已经全部重写完毕，并且发现并修改了另外一个BUG。明天会更新上代码，并加上代码描述

packages/components/tree/src/tree-node.vue
SpanManX: 已删除 selectFocus，改为在 tree.ts onNodeExpand() 和 onNodeCollapse() 内调用 select.value?.focus()

packages/theme-chalk/src/tree.scss
SpanManX: 样式移动到了 tree-select.scss

SpanManX:
> Thanks for your contribution.
> 
> But there are still some issues to be fixed:
> 
> * Up/Down does not work in collapsed tree nodes (may have other effects when using `renderAfterExpand`)
> * Left/Right does not expand/collapse tree nodes
> 
> Finally we need some tests to check the code.

@yujinpan 你好，已按照要求修改了代码，请帮忙review一下
Hello, I have modified the code as requested. Please help review it.

I modified the code implementation in a simpler way:

1. Switched to using the tree component's keyboard events.
2. Used the left and right arrow keys to expand and collapse the tree.
3. Pressing Enter checks whether a value can be selected. By default, pressing Enter on a parent node expands it, considering cases where checkStrictly and lazy properties are set to true.
4. Resolved the impact of lazy loading and renderAfterExpand on the background color display order when navigating with the up and down arrow keys.

我用更简单的方式修改了该代码实现的功能。

1. 改用 tree 组件的键盘事件。
2. 左右按键展开折叠 tree。
3. 回车判断是否可选中值。默认属性父级按下回车展开节点，并判断 checkStrictly 和 lazy 属性为 true 的情况。
4. 解决了 lazy 和 renderAfterExpand 加载数据对上下按键背景色显示顺序造成的影响。

SpanManX:
@yujinpan 你好，已按照要求修改了代码，请帮忙review一下

SpanManX:
@yujinpan 你好，是否还需要修改？

packages/components/tree-select/src/cache-options.ts
```diff
@@ -48,6 +52,8 @@ export default defineComponent({
       { flush: 'post', immediate: true }
     )
 
+    select.states.isUseTreeKeydown = props.isUseTreeKeydown
```
Dsaquel, line 55: because it will always be true we can set to `true`  i think

Dsaquel (commented):
I left some comment, but over all i think it's good :+1: .
As @yujinpan said, can you add a test case ?

SpanManX:
> I left some comment, but over all i think it's good 👍 . As @yujinpan said, can you add a test case ?

I’ve added test cases, but the trigger firing order in the test cases is different from that in the browser. Is there a demo I can refer to?

Dsaquel:
I also didn't find a way to test it. Input somehow can't yield active element. I'll see tomorrow how to do it.

Dsaquel:
With we can `input.trigger('keydown.down') `active element but somehow `.find` or `.querySelector` can't find `.hovering`. Even with [pr](https://github.com/element-plus/element-plus/pull/20499). Can't say if it's an issue or we are doing in the wrong way.
For now we can use `navigateOptions` to mock the keydown just like in [select.test.ts](https://github.com/element-plus/element-plus/blob/dev/packages/components/select/__tests__/select.test.ts#L784-L786).

packages/components/select/src/useSelect.ts
```diff
@@ -761,6 +778,8 @@ export const useSelect = (props: ISelectProps, emit: SelectEmits) => {
         navigateOptions(direction)
       }
       nextTick(() => scrollToOption(hoverOption.value))
+    } else if (states.isUseTreeKeydown) {
+      useTreeKeydown()
```
Dsaquel, line 782: Should handle the update of `hoveringIndex` i think.

SpanManX: In the case of using renderAfterExpand and lazy, the order of hoveringIndex becomes incorrect.

在使用 `renderAfterExpand`  和 `lazy` 的情况下，`hoveringIndex`  顺序是错乱的

Dsaquel:
Lgtm

packages/components/select/src/useSelect.ts
SpanManX: Okay, I'll look into whether there's a way to have the best of both worlds — adding test cases without changing the original logic.

Dsaquel (approved):

warmthsea:
There seems to be a problem 👀 

1. focus select
2. down active level 1 _(2 normal)_
3. enter _(popper close & icon error)_
4. enter _(data error & icon normal)_

screenshots

SpanManX:
> There seems to be a problem 👀
> 
> 1. focus select
> 2. down active level 1 _(2 normal)_
> 3. enter _(popper close & icon error)_
> 4. enter _(data error & icon normal)_
> 

The modification has been completed

SpanManX:
I found another method. Wait for me to update the code

SpanManX:
> change the keyboard methods

I've already changed the keyboard methods   @Dsaquel  @warmthsea

Dsaquel:
After reading the new solution i think the previous solution was a bit more reasonable, the reason of this is because you don't want to involve tree/tree-select logic in the select component to keep the code agnostic as much as possible among other reasons like preserving tree shaking.

SpanManX:
> After reading the new solution i think the previous solution was a bit more reasonable, the reason of this is because you don't want to involve tree/tree-select logic in the select component to keep the code agnostic as much as possible among other reasons like preserving tree shaking.

I've changed the code back to the previous version

Dsaquel:
@SpanManX After looking back, the change looks quite hacky so i checked when the navigate keyboard start not working.
You can see in pr https://github.com/element-plus/element-plus/pull/15768/files there are two added lines that break the navigation.
With this context maybe we can find another solution :thinking: .

Thanks for your patience :heart: .

SpanManX:
> @SpanManX After looking back, the change looks quite hacky so i checked when the navigate keyboard start not working. You can see in pr https://github.com/element-plus/element-plus/pull/15768/files there are two added lines that break the navigation. With this context maybe we can find another solution 🤔 .
> 
> Thanks for your patience ❤️ .

Changed to Only when the `filterable` field is true can `select.value?.focus()` be called.

Dsaquel:
Sorry, I just re-check [#15768](https://github.com/element-plus/element-plus/pull/15768) is just out of context in this pr i mixed up with something else.

My only blocking thing is the changes in the select folder.
I will checkout later if we can have a all changes just in tree-select folder.

packages/components/tree/src/model/useKeydown.ts
```diff
@@ -26,6 +26,32 @@ export function useKeydown({ el$ }: UseKeydownOption, store: Ref<TreeStore>) {
...
+      (node.store.checkStrictly &&
+        ![EVENT_CODE.left, EVENT_CODE.right].includes(code))
+    ) {
+      currentItem.click()
+    } else if (!node.expanded) {
+      node.expand()
+    } else if (node.expanded) {
+      node.collapse()
+    }
+  }
```
Dsaquel: I didn't find the difference with a regular click event. Can you explain what does this ?

packages/components/tree-select/src/select.ts
```diff
@@ -37,6 +38,28 @@ export const useSelect = (
...
+            // el-select-dropdown__item => el-tree-node__content => el-tree-node__content
+            hoverOption?.$el?.parentNode?.parentNode?.focus()
+          })
+        }
+      },
+      {
+        capture: true,
+      }
+    )
+  })
```
Dsaquel, line 86: @SpanManX I took some code of your previous solution and setted here.
From now this solution will fix the down event but can't resolve the up event. 
The best we can have is [handleKeydown](https://github.com/element-plus/element-plus/blob/dev/packages/components/tree/src/model/useKeydown.ts#L29-L109) for a good solution.
But until we refact it, i think we can merge this.

SpanManX:
`tree.ts` `onNodeClick -> select.value?.focus()` Using this method or not makes no difference

Dsaquel:
@SpanManX I see you have got the `mousedown.up` working— so cool!

I will review a bit later.

Dsaquel:
Sometimes the keydown up/down is blocked after expanding some nodes:
Peek 2025-07-28 00-41

SpanManX:
> Sometimes the keydown up/down is blocked after expanding some nodes

I‘ve already fixed the issue, but PR #21526 needs to be awaiting review

Dsaquel:
Yes i saw the pr, i'll review it just a bit later.

packages/components/tree-select/src/select.ts
```diff
@@ -37,6 +38,49 @@ export const useSelect = (
     { flush: 'post' }
   )
 
+  const focusLastNode = (listNode) => {
+    if (listNode.at(-1).expanded && listNode.at(-1).childNodes.at(-1)) {
+      focusLastNode([listNode.at(-1).childNodes.at(-1)])
```
Dsaquel: **nit**:

I think we can simplify this.

packages/components/tree-select/src/select.ts
```diff
@@ -37,6 +38,49 @@ export const useSelect = (
...
+    } else {
+      const el = tree.value.el$?.querySelector(
+        `[data-key="${listNode.at(-1).key}"]`
+      )
+      el.focus()
```
Dsaquel:
```suggestion
      el?.focus()
```

packages/components/tree-select/src/tree.ts
```diff
@@ -179,7 +179,8 @@ export const useTree = (
       } else if (props.expandOnClickNode) {
         e.proxy.handleExpandIconClick()
       }
-      select.value?.focus()
+      // TODO: Using this method or not makes no difference
+      // select.value?.focus()
```
Dsaquel: Let's keep it for now.

Dsaquel (commented):
I forgot we merged the fixed old pr, added just some nits.

Dsaquel:
Now we should be able to add a test as well.

packages/components/tree-select/src/tree.ts
SpanManX: Using this method may result in situations like this [ref](https://github.com/element-plus/element-plus/pull/20008#issuecomment-3051965051)

464114002-7980d6ad-3075-4b55-b931-95dbfe42eb96

Dsaquel: Okay didn't notice that, we can remove both line then 👍 .

Dsaquel (approved):
Thanks!

Dsaquel (approved):

packages/components/tree-select/__tests__/tree-select.test.tsx
```diff
@@ -485,6 +492,70 @@ describe('TreeSelect.vue', () => {
...
+    const treeWrapper = wrapper.findComponent(Tree)
+    const allNodes = treeWrapper.findAll('.el-tree-node')
+    const len = allNodes.length
+    for (let i = 0; i < len; i++) {
+      if (allNodes[i + 1]) {
+        defineGetter(allNodes[i + 1].element, 'focus', handleFocus)
+        // 模拟按下下箭头键
+        allNodes[i].element.dispatchEvent(
+          new KeyboardEvent('keydown', {
+            code: 'ArrowDown',
```
tolking:
```suggestion
            code: EVENT_CODE.down,
```

packages/components/tree-select/src/select.ts
```diff
@@ -38,6 +39,50 @@ export const useSelect = (
...
 
+  const focusLastNode = (listNode) => {
+    const lastNode = listNode.at(-1)
+    if (lastNode.expanded && lastNode.childNodes.at(-1)) {
+      focusLastNode([lastNode.childNodes.at(-1)])
+    } else {
+      const el = tree.value.el$?.querySelector(
+        `[data-key="${listNode.at(-1).key}"]`
+      )
+      el?.focus()
```
tolking:
```suggestion
      el?.focus({ preventScroll: true })
```

packages/components/tree-select/src/select.ts
```diff
@@ -38,6 +39,50 @@ export const useSelect = (
...
+  }
+
+  onMounted(() => {
+    useEventListener(
+      () => select.value?.$el,
+      'keydown',
+      async (evt) => {
+        const { dropdownMenuVisible } = select.value!
+        if (
+          (EVENT_CODE.down === evt.key || EVENT_CODE.up === evt.key) &&
```
tolking:
```suggestion
        const code = getEventCode(evt)
        const { dropdownMenuVisible } = select.value!
        if (
          ([EVENT_CODE.down, EVENT_CODE.up].includes(code)) &&
```

packages/components/tree-select/__tests__/tree-select.test.tsx
```diff
@@ -485,6 +492,70 @@ describe('TreeSelect.vue', () => {
...
+    function handleFocus() {
+      return () => (flag = true)
+    }
+    vi.useFakeTimers()
+    const selectWrapper = wrapper.find('.el-select__wrapper')
+    await selectWrapper.trigger('click')
+    await selectWrapper.trigger('focus')
+    const focused = wrapper.find('.is-focused')
+    expect(focused.exists()).toBe(true)
+    await selectWrapper.find('input').trigger('keydown.ArrowDown')
```
tolking:
```suggestion
    await selectWrapper.find('input').trigger('keydown', { code: EVENT_CODE.down })
```

It might be best to save it consistently with other test files

tolking (commented):
LGTM.

It would be best if some details could be further optimized

packages/components/tree-select/__tests__/tree-select.test.tsx
```diff
@@ -485,6 +493,71 @@ describe('TreeSelect.vue', () => {
...
+    }
+    vi.useFakeTimers()
+    const selectWrapper = wrapper.find('.el-select__wrapper')
+    await selectWrapper.trigger('click')
+    await selectWrapper.trigger('focus')
+    const focused = wrapper.find('.is-focused')
+    expect(focused.exists()).toBe(true)
+    await selectWrapper
+      .find('input')
+      .trigger('keydown', { key: EVENT_CODE.down })
```
tolking, line 517:
```suggestion
      .trigger('keydown', { code: EVENT_CODE.down })
```

packages/components/tree-select/__tests__/tree-select.test.tsx
```diff
@@ -485,6 +493,71 @@ describe('TreeSelect.vue', () => {
...
+    })
+
+    await nextTick()
+    vi.useFakeTimers()
+    const selectWrapper = wrapper.find('.el-select__wrapper')
+    await selectWrapper.trigger('click')
+    await selectWrapper.trigger('focus')
+    const focused = wrapper.find('.is-focused')
+    expect(focused.exists()).toBe(true)
+    await selectWrapper.find('input').trigger('keydown', { key: EVENT_CODE.up })
```
tolking, line 556:
```suggestion
    await selectWrapper.find('input').trigger('keydown', { code: EVENT_CODE.up })
```

packages/components/tree-select/src/select.ts
```diff
@@ -38,6 +40,51 @@ export const useSelect = (
...
+        const code = getEventCode(evt)
+        const { dropdownMenuVisible } = select.value!
+        if (
+          [EVENT_CODE.down, EVENT_CODE.up].includes(code) &&
+          dropdownMenuVisible
+        ) {
+          await nextTick()
+          // wait navigateOption to finish
+          setTimeout(() => {
+            if (EVENT_CODE.up === evt.key) {
```
tolking:
```suggestion
            if (EVENT_CODE.up === code) {
```

packages/components/tree-select/src/select.ts
```diff
@@ -38,6 +40,51 @@ export const useSelect = (
...
+              const listNode = tree.value.store.root.childNodes
+              focusLastNode(listNode)
+              return
+            }
+            // el-select-dropdown__item => el-tree-node__content => el-tree-node__content
+            select.value.optionsArray[
+              select.value.states.hoveringIndex
+            ].$el?.parentNode?.parentNode?.focus()
```
tolking:
```suggestion
            ].$el?.parentNode?.parentNode?.focus({ preventScroll: true })
```

packages/components/tree-select/__tests__/tree-select.test.tsx
SpanManX: using `code` does not trigger `@keydown.down.stop.prevent="navigateOptions('next')"`

packages/components/tree-select/__tests__/tree-select.test.tsx
SpanManX: The same problem

tolking (approved):

tolking:
@yujinpan cc

SpanManX:
@yujinpan  你好，请帮忙review一下这个PR。

PR #20017, opened 2025-03-03
fix: component [radio] disabled radio label should not be colored as placeholder

qq1037305420:
Thank you for element plus!
when radio is disabled, the label color should be disabled instead of placeholder.

### DIFF
--- a/packages/theme-chalk/src/radio.scss
+++ b/packages/theme-chalk/src/radio.scss
@@ -114,7 +114,7 @@ $radio-font-size: map.merge(
         }
       }
       & + span.#{$namespace}-radio__label {
-        color: getCssVar('text-color', 'placeholder');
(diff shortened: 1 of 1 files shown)


btea:
But the `--el-disabled-text-color` variable reference is also the `--el-text-color-placeholder` variable, and the value is the same. So, I don't think it is necessary to modify it.

image

qq1037305420:
Thank you for element plus!
However when i overwrite el-text-color-placeholder to transparent in a disabled form, radio's label is also transparent.

btea:
Yes, because `--el-disabled-text-color` and `--el-text-color-placeholder` are the same value. If you only want to override `--el-text-color-placeholder`, you can add a `--el-disabled-text-color: #a8abb2` to your form to override the label color alone.

PR #20035, opened 2025-03-06
fix(components): [dropdown] `button-props` as a `Partial` type

Dsaquel:
`button-props` should be partial in this situation because it have default value from required props value's.
image

[ref discussion](https://github.com/element-plus/element-plus/discussions/19883)


warmthsea (approved):
LGTM

btea (approved):
Thanks!

PR #20050, opened 2025-03-07
refactor: [select, select-v2]: unified name & supplement type

warmthsea:
- unified name
- export Instance

### DIFF
--- a/packages/components/select-v2/src/defaults.ts
+++ b/packages/components/select-v2/src/defaults.ts
@@ -16,6 +16,7 @@ import { useTooltipContentProps } from '@element-plus/components/tooltip'
 import { ArrowDown, CircleClose } from '@element-plus/icons-vue'
 import { tagProps } from '../../tag'
 import { defaultProps } from './useProps'
+import SelectV2 from './select.vue'
 
 import type { Option, OptionType } from './select.types'
 import type { Props } from './useProps'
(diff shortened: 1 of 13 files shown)


packages/components/select-v2/src/defaults.ts
```diff
@@ -27,7 +28,7 @@ import type {
   PopperEffect,
 } from '@element-plus/components/popper'
 
-export const SelectProps = buildProps({
+export const selectV2Props = buildProps({
```
btea, line 31: These variables are exported and may be used by users, so we should not modify the variable names directly.

kooriookami:
This can be merged now.

kooriookami (approved):

Dsaquel (approved):

btea:
Would it be too radical to just delete it? Maybe we can keep both the old and new names. And we can add a todo comment and leave it for deletion in 3.0. 🤔

warmthsea:
> Would it be too radical to just delete it? And we can add a todo comment and leave it for deletion in 3.0. 🤔

I think it's best to deal with it now, and add it to the update notice.

>  Maybe we can keep both the old and new names. 

There is a conflict, this won't work.

_old:_     **SelectProps (props value)**
_new:_   **SelectProps (type)**

btea (approved):
Makes sense.

We can note the relevant information in advance and add it to the relevant log information released next time.

PR #20051, opened 2025-03-07
chore(eslint-config): add `line-between-blocks` rules

warmthsea:
- unified rules


internal/eslint-config/index.js
warmthsea, line 1: Only added the rule here, other changes executed `npm run lint:fix`

btea (approved):
LGTM

PR #20087, opened 2025-03-11
style(components): [checkbox] modify the icon position when selected

dengguangmiaoyyds:
before:
checkbox-before

after:
checkbox-after

### DIFF
--- a/packages/theme-chalk/src/checkbox.scss
+++ b/packages/theme-chalk/src/checkbox.scss
@@ -169,7 +169,7 @@ $checkbox-bordered-input-width: map.merge(
         border-color: getCssVar('checkbox-checked-input-border-color');
 
         &::after {
-          transform: translate(-50%, -50%) rotate(45deg) scaleY(1);
+          transform: translate(-45%, -60%) rotate(45deg) scaleY(1);
           border-color: getCssVar('checkbox-checked-icon-color');
         }
       }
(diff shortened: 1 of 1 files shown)


warmthsea (commented):
## It seems that `translate(-45%, -60%)` is more suitable

## Screen x 125%

- ### [Before](https://element-plus.org/en-US/component/checkbox.html#basic-usage)

image

- ### [After](https://preview-20087-element-plus.surge.sh/en-US/component/checkbox.html#basic-usage) `transform: translate(-50%, -50%)`

image

- ### `translate(-45%, -60%)`

image



## Screen x 500%

image

image


image

dengguangmiaoyyds:
@warmthsea you're right, thank you. :smile:

warmthsea (approved):

btea:
I scaled the styles 5x and they looked pretty much the same, so I'm not sure if the changes were necessary. 🤔 

|before|after|
|-----|------|
|image|image|

warmthsea:
> I scaled the styles 5x and they looked pretty much the same, so I'm not sure if the changes were necessary. 🤔


Even if the effect is not obvious, I think the change is the best solution.

I am going to submit a **large-related PR** and found that the best result is after the change. Manually setting px always results in a little deviation.



### Before
*100%
image
*500%
image


### After
*100%
image
*500%
image

Dsaquel (approved):
LGTM, it's feels more natural on x125-500 as it is more center.

PR #20066, opened 2025-03-09
fix: borken import paths in declaration files

makedopamine:
fix #20065 

I didn't look into which specific commit caused the issue, but it's likely related to upgrading TypeScript version.

Note: this PR leads to a minor [performance regression](https://github.com/user-attachments/assets/7d586f6b-9a74-4a83-942a-b953e35a77d7) for intellisense, though, given the changes, no surprises there.

PR #20109, opened 2025-03-12
chore: emits smaller declaration files

makedopamine:
#19419 and #19352 reduced DTS size with explicit type annotation. This PR reverts those changes and introduces a workaround further reduces DTS size without explicit type annotation.

[Performance comparsion](https://github.com/user-attachments/assets/e1d4a2a9-73e8-40f5-a7a7-8051805d56af) of this PR vs #20066


warmthsea:
I compared dist before and after, I found that it did get smaller, mainly because:

- Use `ref<xxInstance>` instead of `ref<InstanceType<typeof ElXxx>>`
- Delete the display type of `defineExpose`

Can you explain in detail why you added `& unknown`? I don't know your motivation.


| file-size | before | after|
|-------|-------|-------|
| dist   |   41.8MB |40.8MB   |
| element-plus  | 38.6MB |  37.8MB |
| types|   2.89MB |2.50MB   |
| tsconfig.web |  359k |501k | 

image
image

image

btea (approved):
Looks great! 👍

warmthsea:
> It's to prevent component types from being inlined. You can try removing `& unknown` from TooltipInstance and regenerate the declaration files, then you'll notice `element-plus/es/components/table/src/filter-panel.vue.d.ts` inlines `element-plus/es/components/tooltip/src/tooltip.vue.d.ts` at line 56.

Indeed.

image
image

warmthsea (approved):

tsconfig.base.json
warmthsea, line 1: After the change, the quick jump jumps to `node_modules`.

### Before
screenshots

### After
screenshots

PR #20124, opened 2025-03-13
style(components): [popconfirm] modify icon style

zhuft-moon:
before:
before1

after:

after1


zhuft-moon:
Well, I think it's better, ant design also puts Icon at the top~
image

btea:

This change is breaking change for users, and we should try to avoid such operations.

This kind of style modification is mainly a customization requirement, and it is better for users to modify the style by themselves.

After all, some users may think that the current layout is better, which cannot be concluded. I think it is better to keep the current situation.

PR #20126, opened 2025-03-13
docs: add placement demo for popover/popconfirm

thinkasany:
增加 popover popconfirm 的 placement demo

方便呈现 placement 不同位置的效果，利于功能的展示 和 debug 调试。

image
image

### DIFF
--- a/docs/en-US/component/popconfirm.md
+++ b/docs/en-US/component/popconfirm.md
@@ -7,6 +7,16 @@ lang: en-US
 
 A simple confirmation dialog of an element click action.
 
+## Placement
+
+popconfirm has 9 placements.
+
+:::demo Use attribute `content` to set the display content when hover. The attribute `placement` determines the position of the popconfirm. Its value is `[orientation]-[alignment]` with four orientations `top`, `left`, `right`, `bottom` and three alignments `start`, `end`, `null`, and the default alignment is null. Take `placement="left-end"` for example, popconfirm will display on the left of the element which you are hovering and the bottom of the popconfirm aligns with the bottom of the element.
+
+popconfirm/placement
+
+:::
+
 ## Basic usage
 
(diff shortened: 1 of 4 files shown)


thinkasany:
hi ~ @warmthsea, 这个 PR 考虑采纳吗

btea (commented):
It looks good and can quickly preview the placement effects in different positions, providing a reference for users to configure.

The description needs some tinkering.

docs/en-US/component/popconfirm.md
```diff
@@ -7,6 +7,16 @@ lang: en-US
 
 A simple confirmation dialog of an element click action.
 
+## Placement
+
+popconfirm has 9 placements.
+
+:::demo Use attribute `content` to set the display content when click the reference element. The attribute `placement` determines the position of the popconfirm. Its value is `[orientation]-[alignment]` with four orientations `top`, `left`, `right`, `bottom` and three alignments `start`, `end`, `null`, and the default alignment is null. Take `placement="left-end"` for example, popconfirm will display on the left of the element which you are hovering and the bottom of the popconfirm aligns with the bottom of the element.
```
btea, line 14: It seems that the `content` attribute is useless here, only the `title` will take effect.

thinkasany, line 14:
```suggestion
:::demo Use attribute `title` to set the display content when click the reference element. The attribute `placement` determines the position of the popconfirm. Its value is `[orientation]-[alignment]` with four orientations `top`, `left`, `right`, `bottom` and three alignments `start`, `end`, `null`, and the default alignment is null. Take `placement="left-end"` for example, popconfirm will display on the left of the element which you are hovering and the bottom of the popconfirm aligns with the bottom of the element.
```

thinkasany: done.

btea (approved):
Thanks!

thinkasany:
> Thanks!

Thanks for your review too!

PR #20144, opened 2025-03-15
refactor: [table] remove ts-nocheck directive and use script-setup syntax

zzjiaxiang:
先移除这一点吧, table里面有太多的 `ts-nocheck` 了


Dsaquel:
I was also trying removing some `ts-nocheck` for `el-table` but it seems like there is a pr here.
About the pr changes i think we should keep `null` to make sure to break anything even though there are some tests.

After that i think this pr should be considered to help developer debugging/adding new feature for `el-table` in the future.

zzjiaxiang:
> I was also trying removing some `ts-nocheck` for `el-table` but it seems like there is a pr here. About the pr changes i think we should keep `null` to make sure to break anything even though there are some tests.
> 
> After that i think this pr should be considered to help developer debugging/adding new feature for `el-table` in the future.



> I was also trying removing some `ts-nocheck` for `el-table` but it seems like there is a pr here. About the pr changes i think we should keep `null` to make sure to break anything even though there are some tests.
> 
> After that i think this pr should be considered to help developer debugging/adding new feature for `el-table` in the future.

I see that you have submitted multiple PR to remove 'ts nocheck' and have merged them So I closed this

PR #20171, opened 2025-03-18
chore(eslint-config): changed `no-unused-vars` to `warn`

warmthsea:
### Before

image

### After

image


btea:
Perhaps it would be better to make the errors more obvious. 🤔

PR #20173, opened 2025-03-18
fix(components): [menu] Fix the macro task in `onMouseEnter` is triggered after submenu click on mobile

### LINKED ISSUE #20154: [Component] [menu] el-menu Collapse 折叠面板 在折叠状态下移动端为什么要点击两次才能关闭子菜单
Bug Type: **`Component`**

## Environment

- Vue Version: `3.4.31`
- Element Plus Version: `2.9.1`
- Browser / OS: `Chrome`
- Build Tool: `Vite`

### Related Component

- `el-menu`

### Reproduction Link

[Docs](https://element-plus.org/zh-CN/component/menu.html)

### Steps to reproduce

无

### What is Expected?

在手机端点击一次就关闭子菜单

在手机端点击两次才关闭子菜单

### Additional comments

_(empty)_

comp-squirrel:
fix https://github.com/element-plus/element-plus/issues/20154

The sequence of triggering a click event on the desktop is mouseenter -> click -> mouseleave.
image

However, on mobile devices, the mouseenter and click events are triggered simultaneously. As a result, the timeout in the mouseenter event enters the next event loop.
image

This causes the `openMenu` method to be executed after the click, which reactivates the parent menu. The solution is to avoid using the timeout on mobile devices, ensuring that the mouseenter event is executed before the click event.

After the modifications, the interaction behavior was tested and confirmed to meet expectations on the following platforms:
- [x] Windows 11 / Chrome Developer Tools (simulating mobile devices)
- [x] iPadOS 18.3 / Safari
- [x] Android 15 / Chromium
- [x] HarmonyOS 4.2 / Chromium


comp-squirrel:
@btea does we need to solve this problem?

packages/components/menu/src/sub-menu.ts
```diff
@@ -213,6 +213,11 @@ export default defineComponent({
     const doDestroy = () =>
       vPopper.value?.popperRef?.popperInstanceRef?.destroy()
 
+    const isMobile = () =>
+      /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
+        window.navigator.userAgent
+      )
```
btea, line 219: This method can be placed in the following file, and the isClient check can be added.

https://github.com/element-plus/element-plus/blob/dev/packages/utils/browser.ts

comp-squirrel: Done, thanks!

PR #20176, opened 2025-03-18
feat(components): [date-picker]  Support hh:mm:ss string format

SpanManX:
Add the string type time format hh:mm:ss to the default-time attribute


btea:

But I think this is unnecessary. It is better to limit the qualified date of delivery, otherwise it will only cause more uncertainty and add a lot of additional judgment logic.

PR #20179, opened 2025-03-19
docs: optimize example import

warmthsea:
- Remove unnecessary component imports
- Replace `InstanceType<typeof` with an existing `XxInstance`
- Remove import type blank Line

### DIFF
--- a/docs/examples/checkbox/intermediate.vue
+++ b/docs/examples/checkbox/intermediate.vue
@@ -18,7 +18,6 @@
 
 <script lang="ts" setup>
 import { ref } from 'vue'
-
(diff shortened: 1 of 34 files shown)


btea:
> * Remove import type blank Line

We keep a line of space between variable imports and type imports in our code, maybe we can keep this habit in our examples as well.

warmthsea:
> > * Remove import type blank Line
> 
> We keep a line of space between variable imports and type imports in our code, maybe we can keep this habit in our examples as well.

It is not necessary. I think remove blank lines is more suitable for users. Most of the existing examples do not have blank lines.

btea (approved):

PR #20202, opened 2025-03-20
chore: [table-v2] remove ts-nocheck directive

zzjiaxiang:


packages/components/table-v2/src/composables/use-columns.ts
```diff
@@ -67,15 +66,12 @@ function useColumns(
...
-      },
-      {}
-    )
-  })
+  const columnsStyles = computed(() =>
+    unref(_columns).reduce<Record<KeyType, CSSProperties>>((style, column) => {
+      style[column.key] = calcColumnStyle(column, unref(fixed), props.fixed)
+      return style
+    }, {})
+  )
```
warmthsea, line 74: `return` does not need to be deleted.

warmthsea (approved):
LGTM, Thanks!

tolking (approved):

PR #20211, opened 2025-03-20
refactor: buildProps

makedopamine:
This PR brings the following:
- `as const` is no longer needed when calling buildProps
- Stricter type checking for the `props` parameter of buildProps.  Have a look at [this test case](https://github.com/element-plus/element-plus/blob/6cea66f24c51da2993355956b53c8418266a6321/packages/utils/__tests__/vue/props.test.ts#L64) for details.
- Simplify the return type of buildProps, which improves type hints in IDEs.  Check out `tooltip.vue.d.ts` [before](https://preview-20181-element-plus.surge.sh/bundle/es/components/tooltip/src/tooltip.vue.d.ts) and [after](https://preview-20211-element-plus.surge.sh/bundle/es/components/tooltip/src/tooltip.vue.d.ts) to get a rough sense of what’s changed.
- Upgrade esbuild to support const modifier for type parameters.


makedopamine:
Basically because esbuild changed how it preserves line breaks and comments when minification isn’t enabled. Please have a look at [the diff](https://www.diffchecker.com/6SMLYVQ7/) of `index.full.js`.

btea (approved):
Looks great! 👍

btea:
There is a code conflict.

packages/utils/vue/props/runtime.ts
```diff
@@ -1,122 +1,47 @@
...
+      if (validator) valid ||= validator(val)
+
+      if (!valid && allowValuesText) {
+        warn(
+          `Invalid prop: validation failed${
+            key ? ` for prop "${key}"` : ''
+          }. Expected one of [${allowValuesText}], got value ${JSON.stringify(
+            val
+          )}.`
+        )
```
tolking, line 37: I notice that you moved the conversion of `values` from the `buildProp` function to the `buildProps` function.

I'm worried that someone will use the buildProp function directly in a component.

makedopamine: If calling `buildProp` directly can also create a validator, the console warning might show which prop validation failed incorrectly, because its return value can be assigned to any prop. Since `buildProp` could give the impression that it's for creating validators, how about renaming it to `inferProp` or something?

makedopamine:
Closing in favor of using type-only `defineProps` instead of `buildProps`

PR #20216, opened 2025-03-21
chore(style): [table] remove duplicate defined style code

Putia3:
Above to define the same box-shadow Style code, again defined here


btea (approved):
Thanks!

warmthsea (approved):
LGTM, Thanks!

PR #20309, opened 2025-03-31
chore(eslint-config): add `newline-after-import` rules

warmthsea:
- Most of the **existing codes** follow this rule.


btea (approved):
LGTM

PR #20325, opened 2025-04-02
fix(utils): nuxt build error

warmthsea:
fix: https://github.com/element-plus/element-plus/discussions/20317

image


packages/utils/strings.ts
```diff
@@ -1,10 +1,6 @@
 import { capitalize as toCapitalize } from '@vue/shared'
 
-export {
-  camelize,
-  hyphenate,
-  hyphenate as kebabCase, // alias
-} from '@vue/shared'
+export { camelize, hyphenate } from '@vue/shared'
```
btea, line 3: Maybe we could add `const kebabCase = hyphenate`

If someone has introduced `hyphenate` in element-plus, deleting it directly may result in an error.

btea:
Having said that, I think our writing seems to be correct, it seems that there is a problem with the logic of the plugin `esbuild-transpile`. 🤔

btea (approved):

pei-zheng-yi:
什么时候能发个版本出来 @btea

warmthsea:
> 什么时候能发个版本出来 @btea

image

PR #20335, opened 2025-04-03
docs: resolve code copy button alignment issue

DDDDD12138:
Fix the alignment issue of the code block copy button's `copied` element. Changed the `::before` element's positioning from `relative` to `absolute`, resolving the 1px vertical gap caused by the browser's user agent stylesheet (padding-block: 1px).

|before|after|
|---|---|
|image | image|


btea (approved):
Thanks!

PR #20355, opened 2025-04-07
refactor(components): [message] use message type constants

w2xi:
- Update tests to use `messageTypes` for dynamic type checking
- Refactor `Message` type definition to use mapped type for better scalability


warmthsea (approved):

btea (approved):
Thanks!

PR #20408, opened 2025-04-12
feat(components): [card] add `header-class` and `footer-class` for slots

Dsaquel:
resolve https://github.com/element-plus/element-plus/discussions/20382


btea (approved):
Thanks!

PR #20413, opened 2025-04-14
docs: add right tab scroll

### LINKED ISSUE #20412: [Docs] [All] 【文档官网】右边导航条bug
Bug Type: **`Docs`**

## Environment

- Vue Version: `3.5.13`
- Browser / OS: `Chrome/135.0.0.0/Windows NT 10.0`
- Build Tool: `Other`

### Related Component

- `All`

### Reproduction Link

[Docs](https://element-plus.org/zh-CN/component/table.html)

### Steps to reproduce

打开链接右侧导航条内容多的情况下，没有滚动条

### What is Expected?

内容显示全支持快捷导航，显示滚动条

内容显示不全，无滚动条

### Additional comments

warmthsea:
close: https://github.com/element-plus/element-plus/issues/20412

refer to: 
- https://vite.dev/guide/features.html
- https://vitest.dev/config/#threads

| [Before](https://element-plus.org/en-US/component/table.html) | [After](https://preview-20413-element-plus.surge.sh/en-US/component/table.html) |
| ------- | ----- |
|     screenshots       |    screenshots      |

### DIFF
--- a/docs/.vitepress/vitepress/components/doc-content/vp-table-of-content.vue
+++ b/docs/.vitepress/vitepress/components/doc-content/vp-table-of-content.vue
@@ -52,6 +52,7 @@ const removeTag = (str: string) => str.replace(/<span.*<\/span>/g, '')
       <sponsor-right-logo-small-list />
       <sponsor-right-text-list />
     </nav>
+    <div class="toc-content-mask" />
   </aside>
 </template>
 
--- a/docs/.vitepress/vitepress/styles/content/table-of-content.scss
+++ b/docs/.vitepress/vitepress/styles/content/table-of-content.scss
@@ -2,6 +2,7 @@
 
 .toc-wrapper {
   display: none;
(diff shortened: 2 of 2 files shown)


btea (approved):
LGTM 👍

Dsaquel:
I was not aware of the issue and your pr. I went to do a pr, done almost pretty much the same diff change and the same reference too. That pretty funny.

That said i have some questions because when i came for the solution only this was sufficient:
```css
    height: calc(100vh - var(--header-height) - 32px);
    overflow-y: auto;
    scrollbar-width: none;
```
Why do we need the extras diff ?

warmthsea:
> Why do we need the extras diff ?

There is a slight diff from the vite website. Our right side is not a position: fixed.

### 1. mask

image

### 2. padding

image

Dsaquel:
I see now, thank 👍

PR #20414, opened 2025-04-14
chore(deps): drop esbuild-kit/cjs-loader

Dsaquel:
This pr does two things:
1: Dropping `esbuild-kit/cjs-loader` because it is deprecated we use tsx instead.

2: While Dropping `esbuild-kit/cjs-loader` in `commitlint.config.js`.
>// commitlint uses `ts-node` to load typescript config, it's too slow. So we replace it with `esbuild`.
>
This comment is no longer accurate. [ref](https://github.com/conventional-changelog/commitlint/pull/3722)
And i think it was too overkill to install `esbuild-kit/cjs-loader` to load only one file (`commitlint.config.ts`). We can do the same as before but with `tsx` but again i think its too much for any noticable performance (since commitlint dont use ts-node anymore). You can try commit to see.


commitlint.config.mjs
```diff
@@ -40,7 +40,9 @@ const subjectComplete = gitStatus
   ?.replace(/\//g, '%%')
   ?.match(/packages%%components%%((\w|-)*)/)?.[1]
 
+/** @type { import('cz-git').UserConfig } */
 export default {
+  extends: "@commitlint/config-conventional",
```
Dsaquel, line 45: This extend was not used anywhere so added it.
https://github.com/element-plus/element-plus/blob/dev/package.json#L72

package.json
```diff
@@ -68,12 +68,11 @@
     "normalize-wheel-es": "^1.2.0"
   },
   "devDependencies": {
-    "@commitlint/cli": "^18.6.1",
-    "@commitlint/config-conventional": "^18.6.3",
+    "@commitlint/cli": "^19.0.0",
+    "@commitlint/config-conventional": "^19.0.0",
```
Dsaquel, line 72: the bump feat with the migration of `commitlint.config.js` from commonjs to esm
https://github.com/conventional-changelog/commitlint/pull/3850

package.json
```diff
@@ -92,8 +91,8 @@
     "concurrently": "^7.2.2",
     "consola": "^2.15.3",
     "csstype": "^2.6.20",
-    "cz-git": "^1.3.8",
-    "czg": "^1.3.8",
+    "cz-git": "^1.11.1",
+    "czg": "^1.11.1",
```
Dsaquel, line 95: Same as above, it is a bump to use commitlintjs in esm way.
[cz-git](https://cz-git.qbb.sh/faq/#error-cannot-require-es-module-not-supported)
>Please upgrade cz-git or czg to the latest version.
>

internal/build/package.json
```diff
@@ -7,7 +7,7 @@
   "module": "./dist/index.mjs",
   "types": "./dist/index.d.ts",
   "scripts": {
-    "start": "gulp --require @esbuild-kit/cjs-loader -f gulpfile.ts",
+    "start": "gulp --require tsx/dist/cjs/index.cjs -f gulpfile.ts",
```
Dsaquel, line 10: `tsx/dist/cjs/index.cjs` is a workaround while waiting for tsx to be included in gulp https://github.com/gulpjs/interpret/pull/96
[issue](https://github.com/gulpjs/interpret/issues/95)

btea (approved):

warmthsea (approved):
LGTM 👍

commitlint.config.mjs
warmthsea, line 45:
```suggestion
  extends: '@commitlint/config-conventional',
```

warmthsea:
### I used Beyond Compare to compare the dist directory before and after the change:

- I found that the directory structure was changed, and a new /es/package directory was added.

<hr />

image

<hr />

I think it is best to keep the `dist directory` structure the same as the original one,
to prevent unexpected problems.

Are there any parameters that can be configured?

Dsaquel:
>I found that the directory structure was changed, and a new /es/package directory was added.
>
Yes, that very strange because sometimes it happens with `esbuild-kit/cjs-loader` it happens too.
For example https://github.com/element-plus/element-plus/pull/20413
when installed the zip, click directly on the link if you want to try.
>https://pkg.pr.new/element-plus/element-plus@20413
>
See that `packages` is added and don't respect `dist/element-plus/package.json` exports.
Note that for some other pr/people it respect it like https://github.com/element-plus/element-plus/pull/20211
I don't know what cause this too.

This pr is a good opportunity to see why i think 👍

Dsaquel:
I just realized that @btea pushed the fixed yesterday, thanks <3

I rebased it and its good now, thanks @warmthsea to noticed this 🫡

warmthsea:
Yeah, this is caused by the import path.
https://github.com/element-plus/element-plus/pull/20415/files

btea:
There is a conflict that needs to be resolved.

warmthsea (approved):

warmthsea:
My local build keeps failing. 🤔 Can you all build successfully?

image
image

Dsaquel:
Yes @warmthsea it seems like there is a conflict in pnpm lock, i took my changes exluded from pnpm lock, rebased to dev and everything is good.
I'll try to see how can i preserve the commits order and while exlude the pnpm lock in my previous commits to add it at the end of it.

Dsaquel:
It should be good now :+1:

warmthsea:
> It should be good now 👍

👍 👍

PR #20438, opened 2025-04-15
docs: attach nprogress on router event

Dsaquel:
### Preview
el_1

note the difference with the current algolia search.

cc @makedopamine (https://github.com/element-plus/element-plus/pull/18586)

### DIFF
--- a/docs/.vitepress/vitepress/components/vp-app.vue
+++ b/docs/.vitepress/vitepress/components/vp-app.vue
@@ -1,7 +1,6 @@
 <script setup lang="ts">
 import { onMounted } from 'vue'
 // import { ElMessageBox } from 'element-plus'
-import nprogress from 'nprogress'
 // import dayjs from 'dayjs'
 import { isClient, useEventListener, useToggle } from '@vueuse/core'
 import { EVENT_CODE } from 'element-plus'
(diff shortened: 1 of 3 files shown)


docs/.vitepress/vitepress/components/vp-app.vue
```diff
@@ -47,72 +46,44 @@ useEventListener('keydown', (e) => {
...
-  //   }
-  // }
-  // unregister sw
   navigator?.serviceWorker?.getRegistrations().then((registrations) => {
     for (const registration of registrations) {
       registration.unregister()
     }
   })
 })
+// if (lang.value === 'zh-CN') {
```
makedopamine, line 56: Any reason why you moved these commented out lines?

Dsaquel: It is an github or git diff false positive i think because i deplace `navigator?.serviceWorker` to be closer with onMounted and not the comments.

makedopamine: Nah. These lines are inside onMounted before the changes.

Dsaquel: I didn't saw it

makedopamine:
> note the difference with the current algolia search.

I couldn't spot the difference.🤦‍♂️  Can you spell out the difference for me?

Dsaquel:
>I couldn't spot the difference.🤦‍♂️ Can you spell out the difference for me?

No no it is my fault: the step to reproduce is to not use mouse but the enter key.
Like: 
1. search ctrl + k 
2. search for an item and hover or use arrow keys to select this item
3. click the enter key

makedopamine (approved):

Dsaquel:
Thanks 🙇

warmthsea (approved):
👍

PR #20480, opened 2025-04-18
feat(components): [link] add control for when underlines should appear

### LINKED ISSUE #20471: [Style] [link] el-link组件underline配置无效
Bug Type: **`Style`**

## Environment

- Vue Version: `3.5.13`
- Browser / OS: `Chrome/135 windows10`
- Build Tool: `Vite`

### Related Component

- `el-link`

### Reproduction Link

[Docs](https://element-plus.org/zh-CN/component/link.html#%E4%B8%8B%E5%88%92%E7%BA%BF)

### Steps to reproduce

官网自己的都没有下划线

### What is Expected?

下划线生效

下划线没有生效

### Additional comments

_(empty)_

comp-squirrel:
- feat(components): [link] support controlling when underlines should appear
- docs(components): [link] update doc `Underline` block
- style(components): [link] remove useless styles

---

Change:
underline

---

close: https://github.com/element-plus/element-plus/issues/20471


docs/en-US/component/link.md
```diff
@@ -27,13 +27,23 @@ link/disabled
 
 :::
 
-## Underline
+## Underline Link ^(next)
```
btea, line 30:
```suggestion
## Underline Link ^(2.9.9)
```

packages/theme-chalk/src/link.scss
```diff
@@ -80,9 +94,12 @@
...
 
-      @include when(underline) {
-        &:hover:after {
-          border-color: getCssVar('link', 'text-color');
+      @include when(disabled) {
+        color: getCssVar('link', 'disabled-text-color');
+        cursor: not-allowed;
+
+        &:after {
+          border-color: getCssVar('link', 'disabled-text-color');
```
btea, line 102: This section appears to be repeated with line 60.

kooriookami:
This prop is similar to el-card's shadow which has three available values. Perhaps underline could use these values: `always` | `never` | `hover`. The default value is `hover`. What do you think? @btea

btea:
Sounds good, this way there is no need to add a new prop. 👍 

Just for compatibility reasons, maybe we should support booleans as well?

comp-squirrel:
I initially considered the same approach, but introduced a new prop to avoid potential breaking changes.
What do you think about temporarily supporting both formats - the explicit values `'always'` | `'never'` | `'hover'` alongside the current `boolean` option for backward compatibility? We could then phase out the boolean support in the next major version.

btea:
@comp-squirrel I agree with you. 

WDYT? @kooriookami

kooriookami:
@btea I agree too.

kooriookami:
Please add `useDeprecated` function when underline is using boolean.

kooriookami (approved):
LGTM

btea (approved):
Thanks!

zhouxianjun:
@comp-squirrel @btea After upgrading to 2.9.9, use underline="always" type exception:
(property) underline?: EpPropMergeType<BooleanConstructor, unknown, unknown> | undefined

comp-squirrel:
> @comp-squirrel @btea After upgrading to 2.9.9, use underline="always" type exception: (property) underline?: EpPropMergeType<BooleanConstructor, unknown, unknown> | undefined

The issue you mentioned appears to be caused by using a version prior to `2.9.9`. Could you please confirm the currently used version with the following code?

image

zhouxianjun:
> underline="always"

Sorry, I tested it with a new project and there was no problem with the type. The reason was that my vite cache caused the dts file to be abnormal.

PR #20524, opened 2025-04-22
fix(components): [input-number] reset non numeric value on chromium

### LINKED ISSUE #20522: [Component] [input-number] Why can the numeric input field accept letters?
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.13`
- Element Plus Version: `2.9.8`
- Browser / OS: `Chrome  135.0.7049.96`
- Build Tool: `Vite`

### Related Component

- `el-input-number`

### Reproduction Link

[Docs](https://element-plus.org/en-US/component/input-number.html)

### Steps to reproduce

You can test this with the example on the official website.

### What is Expected?

The numeric input field should not allow entering letters

Image

The numeric input field should not allow entering letters, but I tried it and found that all letters except "e" cannot be entered.

### Additional comments

_(empty)_

Dsaquel:
fixes #20522 #13975

edit: I also encounter this issue on brave (chromium)

### DIFF
--- a/packages/components/input-number/__tests__/input-number.test.tsx
+++ b/packages/components/input-number/__tests__/input-number.test.tsx
@@ -495,6 +495,19 @@ describe('InputNumber.vue', () => {
     expect(num2.value).toBe(8)
   })
 
+  test('clear when non-numeric is set', async () => {
+    const num = ref(0)
+    const handleInput = vi.fn()
+    const wrapper = mount(() => (
+      <InputNumber v-model={num.value} onInput={handleInput} />
+    ))
+    const inputWrapper = wrapper.find('input')
+    inputWrapper.setValue('2e')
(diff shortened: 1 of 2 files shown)


Dsaquel:
I'll add tests case when i have time :+1: (tonight)

packages/components/input-number/__tests__/input-number.test.tsx
```diff
@@ -495,6 +495,19 @@ describe('InputNumber.vue', () => {
     expect(num2.value).toBe(8)
   })
 
+  test('clear when non-numeric is set', async () => {
```
Dsaquel, line 498: the test might be useless because it passes also in dev branch

DDDDD12138:
fix: #19513 fix: https://github.com/element-plus/element-plus/issues/19514#issuecomment-257431739

PR #20530, opened 2025-04-23
style(components): [form-item] Fix label line break style

kooriookami:
Before:
image
After:
image


btea:
I'm not sure if this makes sense, it seems that after this modification the label text wrapping will cause the distance between the form items to increase.

In fact, we should generally avoid this situation and manually set `label-width: auto` or other fixed values.

cc @warmthsea @makedopamine @tolking

kooriookami:
Actually element ui is the same behavior.
image

PR #20532, opened 2025-04-23
feat(components): [input-tag] add delimiter

Dsaquel:
Offer the opportunity to add tags with delimiter. Works with string and regexp.

resolve https://github.com/element-plus/element-plus/discussions/20490

cc @tolking (#18885)

### DIFF
--- a/docs/en-US/component/input-tag.md
+++ b/docs/en-US/component/input-tag.md
@@ -67,6 +67,16 @@ input-tag/draggable
 
 :::
 
+## Delimiter ^(2.9.9)
+
+You can add a tag when a delimiter is matched.
+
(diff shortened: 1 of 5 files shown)


warmthsea (approved):
LGTM

btea (approved):
Thanks!

dormadekhin:
Hello @Dsaquel!
Should delimiter work when inserting text (ctrl-v)? now the behavior is that the characters are simply deleted from the inserted string and all are added with one tag.

Dsaquel:
@dormadekhin Agree, it should work when pasting text with delimiter, thanks for the report i will fix the issue.

PR #20586, opened 2025-04-28
build: rollup dts

makedopamine:
Intellisense performance: [this PR vs 2.9.9](https://github.com/user-attachments/assets/73db311c-2976-4905-81c1-aacb0540532d)

### DIFF
--- a/internal/build/package.json
+++ b/internal/build/package.json
@@ -9,13 +9,15 @@
   "scripts": {
     "start": "gulp --require tsx/dist/loader.cjs -f gulpfile.ts",
     "dev": "pnpm run stub",
-    "stub": "unbuild --stub"
+    "stub": "unbuild --stub",
+    "gen-dts": "gulp --require tsx/dist/loader.cjs -f gulpfile.ts generateTypesDefinitions"
   },
   "peerDependencies": {
     "vue": "^3.2.25"
   },
   "dependencies": {
     "@element-plus/build-constants": "workspace:*",
+    "@microsoft/api-extractor": "^7.52.5",
     "@pnpm/find-workspace-packages": "^4.0.16",
     "@pnpm/logger": "^4.0.0",
     "@rollup/plugin-commonjs": "^22.0.1",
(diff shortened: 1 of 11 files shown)


internal/build/src/tasks/types-definitions.ts
```diff
@@ -1,15 +1,78 @@
...
-import { buildOutput } from '@element-plus/build-utils'
+import { Extractor, ExtractorConfig } from '@microsoft/api-extractor'
+import { buildOutput, epRoot, projRoot } from '@element-plus/build-utils'
 import { pathRewriter, run } from '../utils'
 
 export const generateTypesDefinitions = async () => {
+  const typesDir = path.join(buildOutput, 'types', 'packages')
+  const entryDir = path.join(typesDir, 'element-plus')
+  const entryFilePath = path.join(entryDir, 'index.d.ts')
+  const bundlePath = path.join(typesDir, 'element-plus.d.ts')
```
btea, line 14: Why do we need to change the output file name here? Maybe we can just use the original `index.d.ts`.

btea (approved):

internal/build/package.json
```diff
@@ -9,13 +9,15 @@
   "scripts": {
     "start": "gulp --require tsx/dist/loader.cjs -f gulpfile.ts",
     "dev": "pnpm run stub",
-    "stub": "unbuild --stub"
+    "stub": "unbuild --stub",
+    "gen-dts": "gulp --require tsx/dist/loader.cjs -f gulpfile.ts generateTypesDefinitions"
```
btea, line 13: This command seems to be deletable.

warmthsea (commented):
The import names in these two places can also be changed by the way.

`@vue/runtime-core` -> `vue`

https://github.com/element-plus/element-plus/blob/build/rollup_dts/packages/components/virtual-list/__tests__/scrollbar.test.ts#L2

https://github.com/element-plus/element-plus/blob/build/rollup_dts/typings/env.d.ts#L19

internal/build/package.json
makedopamine: Done

warmthsea (approved):

PR #20616, opened 2025-05-01
refactor(components): standardize injection keys

betavs:


btea (commented):
Thank you for your contribution. Using symbol variables as keys is certainly a good idea.

Of course, using ordinary strings as keys also has another use. When custom components are injected into components through slots, some information can be obtained through the string key with the same name. We are not sure if any user has done this. If modified rashly, it may cause some problems.

betavs:
@btea As you said, it is true, but this standardization is necessary and symbol variables have been exported.

warmthsea (commented):
> Of course, using ordinary strings as keys also has another use. When custom components are injected into components through slots, some information can be obtained through the string key with the same name. We are not sure if any user has done this. If modified rashly, it may cause some problems.

<hr />

I agree.

image

If you want to unify, can temporarily use `string` instead of `Symbols`.

~`export const TREE_NODE_MAP_INJECTION_KEY = Symbol('TreeNodeMap')`~
⬇️
```ts
export const TREE_NODE_MAP_INJECTION_KEY = ‘TreeNodeMap’
```

By the way, it is best to add the suffix `Key` or `KEY` to the variable name (about `provide, inject`).

warmthsea (approved):
LGTM

Dsaquel (approved):
LGTM

PR #20710, opened 2025-05-12
style(theme-chalk): fix form label style

kooriookami:
When there's a global `text-align: right`.
Before:
image
After:
image

When there's a line wrap.
Before:
image
After:
image


btea (approved):
LGTM

PR #20714, opened 2025-05-13
docs: default time example visual display

warmthsea:
| -        |  example display value | 
| ------ | ------- |
| Before         |    image        |  
| After        |     image         |


btea (approved):
LGTM

PR #20750, opened 2025-05-15
feat(components): [table-column] add `expand` slot

ShuaiNingZH:
[discussion](https://github.com/element-plus/element-plus/discussions/14272)

### DIFF
--- a/packages/components/table/src/config.ts
+++ b/packages/components/table/src/config.ts
@@ -127,9 +127,7 @@ export const cellForced = {
       expanded: boolean
     }) {
       const { ns } = store
-      const classes = [
-        column.renderExpand ? ns.e('expand') : ns.e('expand-icon'),
-      ]
+      const classes = [ns.e('expand-icon')]
 
       if (!column.renderExpand && expanded) {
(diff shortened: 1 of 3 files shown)


packages/components/table/src/config.ts
```diff
@@ -116,17 +116,22 @@ export const cellForced = {
...
       expanded: boolean
     }) {
       const { ns } = store
-      const classes = [ns.e('expand-icon')]
-      if (expanded) {
+      const classes = [
+        column.renderExpand ? ns.e('expand') : ns.e('expand-icon'),
+      ]
+
+      if (!column.renderExpand && expanded) {
```
btea, line 134: As far as I understand, even if a custom slot is used, the class name here should not need to be modified.

wjp980108: That is true, I was overthinking it

Dsaquel (commented):
Thanks !

Added just some nit.

btea (approved):
LGTM

Dsaquel (approved):

PR #20757, opened 2025-05-15
feat(components): [config-provider] add global icons configuration

feynmx:
- Global icons added to dialog, dropdown, message, notification, select, and tour components.

#20755 


docs/en-US/component/config-provider.md
```diff
@@ -41,6 +41,14 @@ config-provider/message
 
 :::
 
+## Icons Configurations
```
btea, line 44:
```suggestion
## Icons Configurations ^(2.9.11)
```

btea (commented):
Overall, there is no problem, but the newly added attributes need to add corresponding version tags.

docs/en-US/component/config-provider.md
```diff
@@ -116,6 +125,13 @@ In this section, you can learn how to use Config Provider to provide experimenta
 | offset ^(2.8.2)    | set the distance to the top of viewport                                        | ^[number]  | —       |
 | plain ^(2.9.11)    | whether message is plain                                                       | ^[boolean] | —       |
 
+### Icons Attribute
+
+| Attribute | Description                                               | Type                     | Default |
+| --------- | --------------------------------------------------------- | ------------------------ | ------- |
+| close     | Custom close icon component for dialog, message, etc.     | ^[string] / ^[Component] | —       |
```
Dsaquel, line 132: These attributes should should have a bump version as @btea said.
Also the `dropdown` property is a little bit confusing. Maybe something like `selectSuffix` or `{ select: { suffix: MyIcon }}` could fit well.

Dsaquel (commented):
I think it's a good idea to have a global icon provider :+1: .
Maybe we should list all components that need to inherit from it.

docs/en-US/component/config-provider.md
cn-xufei: Thanks for the suggestion!

I get that dropdown might not be super intuitive at first, but I wanted to avoid suffix since it could be confused with the existing slot or prop. Also, the icon is used in dropdown-like behaviors across multiple components, so dropdown feels like a more general and reusable name.

To help clarify things, I added a mapping table and also updated related components’ descriptions to mention Config Provider `icons`

packages/components/config-provider/src/hooks/use-global-config.ts
```diff
@@ -72,6 +89,21 @@ export function useGlobalComponentSettings(
...
+  success: markRaw(SuccessFilled),
+  info: markRaw(InfoFilled),
+  warning: markRaw(WarningFilled),
+  error: markRaw(CircleCloseFilled),
+}
+
+export function useGlobalIcons() {
+  const globalConfig = useGlobalConfig('icons', defaultIcons)
+  return computed(() => Object.assign(defaultIcons, globalConfig.value))
+}
```
Dsaquel, line 105: If there is no `globalConfig` we will don't have these default icons.
We should keep handling the default value as their place :+1: .

docs/en-US/component/dialog.md
```diff
@@ -145,36 +145,36 @@ dialog/events
...
+| overflow ^(2.5.4)          | draggable Dialog can overflow the viewport                                                                                                       | ^[boolean]                          | false   |
+| center                     | whether to align the header and footer in center                                                                                                 | ^[boolean]                          | false   |
+| align-center ^(2.2.16)     | whether to align the dialog both horizontally and vertically                                                                                     | ^[boolean]                          | false   |
+| destroy-on-close           | destroy elements in Dialog when closed                                                                                                           | ^[boolean]                          | false   |
+| close-icon                 | custom close icon component, default value comes from [Config Provider icons](/en-US/component/config-provider#icons-attribute) `close` property | ^[string] / ^[Component]            | Close   |
```
Dsaquel, line 174: Default values should be handled by their components.
No need these changes in the components doc, it should be clear from his context.

packages/components/message-box/src/index.vue
```diff
@@ -314,15 +320,19 @@ export default defineComponent({
...
 
     const contentId = useId()
     const inputId = useId()
 
     const iconComponent = computed(() => {
       const type = state.type
-      return state.icon || (type && TypeComponentsMap[type]) || ''
+      return (
+        state.icon ||
+        (type && globalIcons.value[type === 'primary' ? 'info' : type]) ||
```
Dsaquel, line 333: `TypeComponentsMap` is seems useful if there is no globalConfig.

docs/en-US/component/dialog.md
cn-xufei: Thanks for your feedback.

My initial thought was that since component `props` have higher priority, it's better to let `ConfigProvider` handle the default values, so that we can keep the components themselves more lightweight and focused on override behavior rather than managing defaults.😊

packages/components/message-box/src/index.vue
cn-xufei: Thanks for the feedback, I replaced TypeComponentsMap with globalIcons to leverage ConfigProvider for consistent icon defaults, Even without ConfigProvider, globalIcons is always available by default, ensuring reliability.  😊

docs/en-US/component/dialog.md
Dsaquel: I understand that, but in order to respect other properties handled by config provider button, link, etc... default values are handled by the end component. Button, link, etc... have no relation with config provider until we create it.

IMO, If you want to handle default values in directly inside config provider then it should be the same for all other component properties handled by config provider to keep consistency. And if so you can open additional PR to suggest this change.

feynmx: Thanks for your input!

I checked the Link component — it also sets default values to undefined, and follows the same priority: component > global (ConfigProvider). So this approach is consistent with existing components.

Let me know if you see any specific inconsistency or potential improvement — happy to discuss!

Dsaquel: >I checked the Link component — it also sets default values to undefined, and follows the same priority: component > global (ConfigProvider). So this approach is consistent with existing components.
>
With the current changes, the default value is controlled in `useGlobalIcons`, which is different from the way Link (and others) components: they let the component control the default value, not the hook.

I understand the idea of putting defaults in `useGlobalIcons` and making it work like hooks such as `useLocale` or `useEmptyValues`. However, those hooks have more complex logic, whereas `globalIcons` is only about forwarding values.

For these reasons, and because there’s no additional logic in the hook, I believe we should follow the same pattern as Link, Button, etc. — letting the component define the default value, and using the hook only to inject the global config if present.

What do you think?

feynmx: Makes sense, I agree. I’ll update the components to set their own default icons, and only use globalIcons from the hook when it’s available.

packages/components/config-provider/src/config-provider-props.ts
```diff
@@ -12,6 +12,16 @@ export type ExperimentalFeatures = {
   // TO BE Defined
 }
 
+export type IconConfigContext = {
+  close?: string | Component
+  arrowDown?: string | Component
```
Dsaquel, line 17: thanks for changing `dropdown` to `arrowDown`, please can you sync the change in the doc.

packages/components/config-provider/src/hooks/use-global-config.ts
```diff
@@ -72,6 +73,10 @@ export function useGlobalComponentSettings(
   }
 }
 
+export function useGlobalIcons() {
+  return useGlobalConfig('icons', {} as IconConfigContext)
```
Dsaquel: I think we can use directly `useGlobalConfig('icons')` into the end components.
I don't think we need `{} as IconConfigContext` as well.

packages/components/select/__tests__/select.test.ts
```diff
@@ -1090,6 +1090,7 @@ describe('Select', () => {
     })
 
     const suffixIcon = wrapper.findComponent(ArrowDown)
+    console.log(wrapper.html())
```
Dsaquel:
```suggestion
```

Dsaquel (commented):
overall it is good.

I'll test a bit deeper later :+1:

docs/en-US/component/config-provider.md
```diff
@@ -85,6 +93,7 @@ In this section, you can learn how to use Config Provider to provide experimenta
...
 | message                 | message related configuration, [see the following table](#message-attribute)                                                                                           | ^[object]`{max?: number}`                                                                                                                                                                                                                                      | see the following table                                                                |
+| icons                   | icons related configuration, [see the following table](#icons-attribute)                                                                                               | ^[object]`{close?: string \| Component, dropdown?: string \| Component}`                                                                                                                                                                                       | —                                                                                      |
```
Dsaquel: We can add the rest icons here.

packages/components/config-provider/src/hooks/use-global-config.ts
```diff
@@ -72,6 +72,10 @@ export function useGlobalComponentSettings(
   }
 }
 
+export function useGlobalIcons() {
+  return useGlobalConfig('icons')
```
Dsaquel: As i said before we can use directly we use `useGlobalConfig('icons')` directly.

packages/components/message-box/src/index.vue
```diff
@@ -283,7 +290,7 @@ export default defineComponent({
       dangerouslyUseHTMLString: false,
       distinguishCancelAndClose: false,
       icon: '',
-      closeIcon: '',
+      closeIcon: props.closeIcon as string | Component,
```
Dsaquel, line 293: why changing this line ?

packages/components/message-box/src/index.vue
```diff
@@ -62,7 +62,7 @@
                 "
               >
                 <el-icon :class="ns.e('close')">
-                  <component :is="closeIcon || 'close'" />
+                  <component :is="closeIconComponent" />
```
Dsaquel, line 65: props have higher priority than iconsConfig.
it should be:

```suggestion
                  <component :is="closeIcon || iconsConfig.close || 'close'" />
```
+ Perhaps no need to declare the props closeIcon anymore.

Please can you handle the same change everywhere you are doing the same logic.

packages/components/message-box/src/index.vue
feynmx: It’s also fine not to include it in props; the original intention was to maintain consistency with the default props behavior across other components.

packages/components/message-box/src/index.vue
feynmx: > props have higher priority than iconsConfig.

Initially, props had higher priority, but to ensure the default values of the component props are preserved, the current logic gives higher precedence to iconsConfig over props, as I understand it.

PR #20758, opened 2025-05-15
docs: add missing button types

Dsaquel:
- add missing button types
- ~~add missing bump version for underline link~~
- `pnpm format`: looks like `tooltip.md` was not formatted


btea (approved):
Thanks!

PR #20764, opened 2025-05-16
docs: remove confusing dialog attributes

Dsaquel:
From the doc:
image

Is more confusing to demonstrate the opposite, the tip helper is sufficient IMO.


btea (approved):
Thanks!

PR #20782, opened 2025-05-19
feat(components): [alert] [result] [message] [message-box] [notification] add primary type

kooriookami:
image
image
image
image
image
image


warmthsea (approved):
LGTM

btea:
I looked at other component libraries of the same type, and this seems to be the design.

Taking alert as an example, success/info/warning/error are enough, and `primary` seems to be a bit unclear.

xingyixiang:
> I looked at other component libraries of the same type, and this seems to be the design.
> 
> Taking alert as an example, success/info/warning/error are enough, and `primary` seems to be a bit unclear.

[ant-design alert](https://ant-design.antgroup.com/components/alert-cn)
The `info` type uses the newly added `primary` type style, and I feel that this new feature is quite good

kooriookami:
@btea Actually, the official design used the primary type. https://www.figma.com/community/file/1305760370797950824
image
image
image

btea:
I forgot about this. Okay, let's merge it.

btea (approved):

PR #20807, opened 2025-05-20
fix(components): [image-viewer] dont persist image preview list

Dsaquel:
Revert to element-ui change https://github.com/ElemeFE/element/blob/c345bb453bf11badb4831a6a3f600c9372b3a336/packages/image/src/image-viewer.vue#L40.

I tested to switch back and forth images and it seems like there is no duplicate call.
edit: browsers can re fetch the image after x time.

resolve https://github.com/element-plus/element-plus/discussions/20805


btea (approved):
LGTM

kooriookami (approved):

PR #20899, opened 2025-05-30
fix(components): [cascader-panel] check if children label are comments

Dsaquel:

close #20898

The issue is that slots.defaults is evaluated as true when there are comments.

### DIFF
--- a/packages/components/cascader-panel/src/node-content.ts
+++ b/packages/components/cascader-panel/src/node-content.ts
@@ -2,14 +2,11 @@
 import { defineComponent, h } from 'vue'
 import { useNamespace } from '@element-plus/hooks'
 
+import type { VNode } from 'vue'
+
 //https://github.com/vuejs/core/issues/4733#issuecomment-933284261
-function isVNodeEmpty(vnodes?: VNode | VNode[] | null) {
-  return (
-    !!vnodes &&
-    (Array.isArray(vnodes)
-      ? vnodes.every((vnode) => vnode.type !== Comment)
-      : vnodes.type !== Comment)
-  )
+function isVNodeEmpty(vnodes?: VNode[]) {
+  return !!vnodes?.every((vnode) => vnode.type !== Comment)
 }
 
 export default defineComponent({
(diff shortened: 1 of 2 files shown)


packages/components/cascader-panel/src/node-content.ts
```diff
@@ -2,6 +2,16 @@
 import { defineComponent, h } from 'vue'
 import { useNamespace } from '@element-plus/hooks'
 
+//https://github.com/vuejs/core/issues/4733#issuecomment-933284261
+function isVNodeEmpty(vnodes?: VNode | VNode[] | null) {
```
makedopamine:
```suggestion
function isVNodeEmpty(vnodes?: VNode[]) {
```
From what I know each slot always returns an array of vnodes.

packages/components/cascader-panel/src/node-content.ts
```diff
@@ -2,6 +2,13 @@
 import { defineComponent, h } from 'vue'
 import { useNamespace } from '@element-plus/hooks'
 
+import type { VNode } from 'vue'
+
+//https://github.com/vuejs/core/issues/4733#issuecomment-933284261
+function isVNodeEmpty(vnodes?: VNode[]) {
+  return !!vnodes?.every((vnode) => vnode.type !== Comment)
```
btea:
```suggestion
  return !!vnodes?.every((vnode) => vnode.type === Comment)
```
Here, as long as the slot content is not all comments, the slot content should be rendered.

`Comment` does not seem to be introduced, and the content judgment method can be called.
https://github.com/element-plus/element-plus/blob/1c3fc40a3c4a8f592306005834995d1eebb7e1d6/packages/utils/vue/vnode.ts#L59-L61

Dsaquel: didn't catch this scenario, thanks

packages/components/cascader-panel/src/node-content.ts
```diff
@@ -2,6 +2,12 @@
 import { defineComponent, h } from 'vue'
 import { useNamespace } from '@element-plus/hooks'
 
+import type { Comment, VNode } from 'vue'
```
btea: According to previous habits, it is best to import types and values ​​separately.

btea (commented):
It would be best if we could add a few more test cases to adapt to various situations.

btea (approved):
LGTM

makedopamine (approved):
Well done👏

PR #20905, opened 2025-06-02
fix(components): [select] display exception when label is `0` or `""`

YiMo1:


Dsaquel (approved):
LGTM,
it will follow the same approach https://github.com/element-plus/element-plus/pull/20608 :+1:

btea (commented):
Can you add a test case?

YiMo1:
> Can you add a test case?

ok

btea (approved):
Thanks!

PR #20915, opened 2025-06-03
fix(components): [tree-select] avoid Enter selecting non-leaf nodes

rzzf:

prevent selecting parent nodes with Enter when only leaf nodes are selectable

fix #20870


packages/components/tree-select/src/tree.ts
```diff
@@ -120,6 +122,30 @@ export const useTree = (
...
+      ) {
+        return
+      }
+
+      const { hoveringIndex } = select.value?.states ?? {}
+      const option = select.value?.optionsArray[hoveringIndex]
+      const { isLeaf } = tree.value?.getNode(option?.value) ?? {}
+
+      if (!isLeaf) {
+        e.stopPropagation()
```
Dsaquel, line 142: Thanks for you contribution,

This modification override the existing enter keydown listener of the select component which I do not think is the most elegant solution.

rzzf: Thank you for your review. I think it might be necessary to add a check for the checkStrictly prop in [selectOption](https://github.com/element-plus/element-plus/blob/6f8e91f771d30683a90d95227330113ef296bb4c/packages/components/select/src/useSelect.ts#L696), for example: `if (checkStrictly) { return } ...`. What do you think? I’d really appreciate your guidance on this!

btea: This situation is caused by the combination of `select` component and `tree` component. I think it is more reasonable to solve it within the tree-select component, and it is best not to modify the logic inside the select.

Dsaquel: Ok i agree, the logic of select should only affect his component if possible.

Note: I feel like this [pr](https://github.com/element-plus/element-plus/pull/20008) can affect only tree-select as well. Maybe i'll try to refact this one later.

packages/components/tree-select/src/tree.ts
```diff
@@ -1,6 +1,8 @@
 // @ts-nocheck
 import { computed, nextTick, toRefs, watch } from 'vue'
 import { isEqual, isNil, pick } from 'lodash-unified'
+import { useEventListener } from '@vueuse/core'
+import { EVENT_CODE } from 'element-plus'
```
Dsaquel: importing directly from 'element-plus' can cause local build errors. [ref](https://github.com/element-plus/element-plus/pull/20415)

```suggestion
import { EVENT_CODE } from '@element-plus/constants'
```

rzzf: done

Dsaquel (approved):

packages/components/tree-select/src/tree.ts
```diff
@@ -120,6 +121,30 @@ export const useTree = (
...
+      const { hoveringIndex } = select.value?.states ?? {}
+      const option = select.value?.optionsArray[hoveringIndex]
+      const { isLeaf } = tree.value?.getNode(option?.value) ?? {}
+
+      if (!isLeaf) {
+        e.stopPropagation()
+      }
+    },
+    { capture: true }
+  )
```
tolking, line 146: Thanks you for your contribution. I think a more complete solution is to refine the logic of the `enter` event based on the `click` event of the current tree-select-option.

PR #20980, opened 2025-06-10
fix(components): input-numbr border has been coverage

jyp114110:
the component of input number border has been covarage when use ElForm  validate

BREAKING CHANGE: n

closed #20979

### DIFF
--- a/packages/theme-chalk/src/input-number.scss
+++ b/packages/theme-chalk/src/input-number.scss
@@ -163,6 +163,7 @@
     }
 
     @include e(decrease) {
+      bottom: 1.2px;
       right: 1px;
       top: auto;
       left: auto;


btea:
This doesn't seem to be the right solution, the decimals will still be rendered differently on different screen devices. Maybe we should adjust the slot area layout.

zhixiaotong:
Now the dom design of `input-number` is problematic. The focus effect of the element is provided by `input`. Since the button is placed on it through absolute positioning, especially through px units, once the browser scales, it will cause the border of the `input` to be blocked. I have tried other methods, but I can't completely solve this problem. Only by redesigning the structure of dom, focus makes the border of the external div highlighted, and the `input` itself removes the border. Since the external div uses `overflow: hidden`, the internal button will not block the external div border, which will be a breaking change.

现在`input-number`的dom设计是有问题的，元素的聚焦效果是有`input`提供，由于按钮是通过绝对定位放在上面，尤其是通过px单位，一旦浏览器缩放，会导致遮挡`input`的边框。我试过使用其他方式，都无法完全解决这个问题，只有重新设计dom的结构，聚焦让外部的div的边框高亮，而`input`本身移除边框，由于外部的div使用`overflow: hidden`, 内部的按钮也不会将外部的div边框遮挡，将是一个breaking 变更。

PR #21006, opened 2025-06-12
fix(components): [upload-dragger] Blink when sliding over sub elements

selicens:
When sliding on a child element, `onDragleave` will be triggered to remove the style, and then `onDragover` will periodically detect that the element is located within the area and add the style, causing flickering

| before | after |  
| ------- | ----- |
| old-drag | new-drag |

### DIFF
--- a/packages/components/upload/src/upload-dragger.vue
+++ b/packages/components/upload/src/upload-dragger.vue
@@ -1,6 +1,5 @@
 <template>
   <div
-    ref="dragRef"
     :class="[ns.b('dragger'), ns.is('dragover', dragover)]"
     @drop.prevent="onDrop"
     @dragover.prevent="onDragover"
@@ -40,7 +39,6 @@ if (!uploaderContext) {
 const ns = useNamespace('upload')
 const dragover = ref(false)
 const disabled = useFormDisabled()
-const dragRef = ref()
 
 const onDrop = (e: DragEvent) => {
   if (disabled.value) return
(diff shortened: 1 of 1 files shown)


Dsaquel (commented):
using only the current event is possible i think:
```ts
  if((e.currentTarget as Element).contains(e.relatedTarget as Element)) return
```
[ref](https://stackoverflow.com/a/54271161)

selicens:
> using only the current event is possible i think:
> 
> ```ts
>   if((e.currentTarget as Element).contains(e.relatedTarget as Element)) return
> ```
> 
> [ref](https://stackoverflow.com/a/54271161)

I had asking GPT a similar question to see if it was feasible. It told me that relatedTarget doesn't work during dragging, but I haven't verified it myself. You can give it a try. If it works, I'd be more than happy to make the modification.

selicens:
image

selicens:
I have tried, and relatedTarget is effective. Please ask the maintainer to decide which method to use @btea @kooriookami

selicens:
```js
const onDragleave = (e: DragEvent) => {
  const rect = dragRef.value.getBoundingClientRect()
  const { clientX: x, clientY: y } = e
  if (x <= rect.left || x >= rect.right || y <= rect.top || y >= rect.bottom) {
      // leave the container
      dragover.value = false
   }
}
```
or
```js
const onDragleave = (e: DragEvent) => {
  if ((e.currentTarget as Element).contains(e.relatedTarget as Element)) return
  dragover.value = false
}
```

packages/components/upload/src/upload-dragger.vue
```diff
@@ -61,4 +63,12 @@ const onDrop = (e: DragEvent) => {
 const onDragover = () => {
   if (!disabled.value) dragover.value = true
 }
+const onDragleave = (e: DragEvent) => {
+  const rect = dragRef.value.getBoundingClientRect()
+  const { clientX: x, clientY: y } = e
+  if (x <= rect.left || x >= rect.right || y <= rect.top || y >= rect.bottom) {
+    // leave the container
+    dragover.value = false
+  }
```
kooriookami: Maybe use `!e.currentTarget.contains(e.relatedTarget)` is more better.

kooriookami (changes requested):

packages/components/upload/src/upload-dragger.vue
selicens: ok

kooriookami (approved):
LGTM

Dsaquel (approved):
Thanks!

PR #21049, opened 2025-06-16
fix(components): [date-picker] pass showWeekNumber to panel

fratzinger:
'showWeekNumber' was not fully implemented. 

It alrady existed in `basic-date-table.vue` but could be accessed. With this PR you can pass prop 'showWeekNumber' to date-picker. It defaults to `false` so the existing behavior is in tact.

While implementing this, I found multiple problems:
- the weekNumber for the weeks won't update, when the month changes. -> fixed in this PR
- the word 'week' in the <th> breaks because it's to wide -> simply removed it
- the <td> for the weekNumber had 'cursor: pointer' css rule -> changed to 'cursor: default'
- color had to much contrast for the <td> -> changed to same contrast as 'next-month'
- column was too wide for my taste -> reduced to 24px

For easier review I recorded a before video:

https://github.com/user-attachments/assets/c404e179-4906-4a88-8e8a-9fd1426dd378

And an after video:

https://github.com/user-attachments/assets/ac512278-67f2-4f2b-9a61-3bac4c0a5056

I added the attribute to the docs in 'date-picker.md' and added the next possible version `v2.10.3` to it.

Let me know if something's left. 💚


packages/components/date-picker/src/date-picker-com/basic-date-table.vue
```diff
@@ -12,7 +12,7 @@
   >
     <tbody ref="tbodyRef">
       <tr>
-        <th v-if="showWeekNumber" scope="col">{{ weekLabel }}</th>
+        <th v-if="showWeekNumber" scope="col" class="week" />
```
warmthsea: Great PR.  

I have two more comments:

1. [`DateTimePicker`](https://github.com/element-plus/element-plus/blob/dev/docs/en-US/component/datetime-picker.md#attributes) component also needs to add documentation.
2. It is best to modify the fixed class name here (and [/date-table.scss](https://github.com/element-plus/element-plus/pull/21049/files#diff-29a0497a8858eb012b67b4afabe041ac9c1b5ffe0d49fc1103ad623578d1b5a0))

fratzinger: Thanks for the review.

I figured 'DateTimePicker' docs missed the additional attribute. Copy/pasted it from 'DatePicker'.

I'm very sorry but I don't understand your second point. I'm happy to make changes if you throw me into the right direction.

warmthsea: Dynamic class name, like this https://github.com/fratzinger/element-plus/blob/fix/components/date-picker-show-week-number/packages/components/date-picker/src/date-picker-com/panel-date-pick.vue#L177

fratzinger: Thanks for the direction. I think I got it. I never worked with that before but I looked closely on other implementations I think now it is right. Please let me know if somethings left.

packages/components/date-picker/src/props/shared.ts
```diff
@@ -61,6 +61,10 @@ export const panelSharedProps = buildProps({
     type: Boolean,
     default: true,
   },
+  showWeekNumber: {
+    type: Boolean,
+    default: false,
+  },
```
Dsaquel: **nit**: `Boolean` will be cast to false by default.
```suggestion
  showWeekNumber: Boolean
```

packages/components/time-picker/src/common/props.ts
```diff
@@ -233,6 +233,10 @@ export const timePickerDefaultProps = buildProps({
     type: Boolean,
     default: true,
   },
+  showWeekNumber: {
+    type: Boolean,
+    default: false,
+  },
```
Dsaquel: same as above.
```suggestion
  showWeekNumber: Boolean,
```

Dsaquel (commented):
Thanks, it can be really useful !
>the word 'week' in the breaks because it's to wide -> simply removed it

I think it confuse can people who using the date picker, maybe we can set the letter **W** like:
KFNLT
This will need to wrap the letter for translation.

/cc @warmthsea @btea

warmthsea:
> I think it confuse can people who using the date picker, maybe we can set the letter 

I think it can be deleted for the following reasons:

1. It looks a bit crowded (even with **w**, but the non-English ones are still there).
2. Most of the translations of [`datepicker.week`](https://github.com/element-plus/element-plus/blob/dev/packages/locale/lang/zh-cn.ts#L47) are commented out now, and we still have to deal with the translation issues, so we might as well delete the commented codes at the same time.



image

fratzinger:
> > I think it confuse can people who using the date picker, maybe we can set the letter
> 
> I think it can be deleted for the following reasons:

I thought about going with `'cw'` which is what we use in our app. But I don't know if this abbreviation exists in every language. E.g. in germany (where I live in) we use the equivalent: 'KW'.  But I don't wanted to start a translation discussion with this PR. But I'm happy to make changes.

If you settle on removing the translation for 'datepicker.week' and want me to remove it from the translation files, please let me know.

Dsaquel:
@warmthsea We can increase the width if show-week-number is present. That said i agree with you, the other translations seems very long.
I don't have opinion on if we should remove week from locale translation tho.

@fratzinger 
can you add a test case ?

fratzinger:
@Dsaquel I added basic tests for normal 'date-picker', `type="daterange"`, `type="datetime"` and `type="datetimerange"`.

It also works with `type="week"` but there are not much tests for `type="week"` so I skipped that.

Dsaquel (approved):
LGTM

warmthsea (approved):
Thanks!

fratzinger:
Thanks for the quick feedback and overall for this great library! 💚 It's awesome to work with it daily.

PR #21066, opened 2025-06-18
fix(components): [tree-v2] treeNode add `expanded` and responds to UI

zhixiaotong:
closed #21023


zhixiaotong:
In order to better reflect the expanded, I modified the example: `Custom node content`.

https://github.com/user-attachments/assets/8f23e5ac-2c0d-4b4c-b1e3-d967933a6da5

btea (commented):
Looks great. Maybe we should also document the new properties that are available after a certain version to reduce user confusion.

zhixiaotong:
> Looks great. Maybe we should also document the new properties that are available after a certain version to reduce user confusion.

Thanks for the review. Do you mean adding a new demo? I just exposed expanded in data.

btea:
Yes, this will mark the version in which you added the expanded value to the node parameter of the default slot.

zhixiaotong:
> Yes, this will mark the version in which you added the expanded value to the node parameter of the default slot.

okay, i'll fix it tomorrow.

zhixiaotong:
https://github.com/user-attachments/assets/c2397c15-0666-4fd7-8be8-bcd24d1c73f3

@btea Please take a look again to see if there are any other additions. I don't think this is a new feature, it can only be regarded as a improvement of the previous function, so there is no test case added

btea (approved):
Thanks!

packages/components/tree-v2/src/tree.vue
```diff
@@ -18,7 +18,7 @@
           :key="data[index].key"
           :style="style"
           :node="data[index]"
-          :expanded="isExpanded(data[index])"
+          :expanded="data[index].expanded"
```
Dsaquel, line 21: Looks like data is not reactive :thinking:.
I'll see later to optimize this.

Dsaquel (approved):
Thanks!

packages/components/tree-v2/src/tree.vue
Dsaquel: Tested and because the data was not proxied, wrapping all nodes inside `reactive` would decrease significantly the performance. So the status quo is good.

PR #21064, opened 2025-06-18
fix(components): [tabs] update tabs order correctly when reordering

w2xi:
Fix #21048

PR #21071, opened 2025-06-18
fix(components): [tabs] orting did not update after changes

snowbitx:
Fixes #21048


btea:
Thanks for your contribution, but it looks like a duplicate of #21064.

snowbitx:
> Thanks for your contribution, but it looks like a duplicate of #21064.

Got it, thanks for the quick review and response!

PR #21090, opened 2025-06-20
refactor(components): [table] remove ts-nocheck

Dsaquel:
This pr prepare will remove `ts-nocheck` for table component, it will facilitate contribution on this component and it has been made to bring generic easily when we'll upgrade vue.

However, it introduce a small breaking change, since the generic across the component has been change:
```diff
- type DefaultRow = any
+ type DefaultRow = Record<PropertyKey, any>

here PropertyKey is a native typescript alias for string | number | symbol
```
This change is necessary to remove `ts-nocheck` properly and add more reliability.

This will break typescript when the user will have:
```diff
- interface SummaryMethodProps<T = Product> {
+ interface SummaryMethodProps<T extends Record<string, any> = Product> {
  columns: TableColumnCtx<T>[]
  data: T[]
}
```
[from this example](https://github.com/element-plus/element-plus/blob/947a88d255b92928e7197285dd8d244579b1c329/docs/examples/table/summary.vue#L40-L43)

It can affect many users.

### DIFF
--- a/packages/components/table/src/store/watcher.ts
+++ b/packages/components/table/src/store/watcher.ts
@@ -20,9 +20,9 @@ import type { StoreFilter } from '.'
 const sortData = <T>(
   data: T[],
   states: {
-    sortingColumn: TableColumnCtx<T>
+    sortingColumn: TableColumnCtx<T> | null
     sortProp: string | null
-    sortOrder: string | null
+    sortOrder: string | number | null
   }
 ) => {
   const sortingColumn = states.sortingColumn
@@ -128,7 +128,7 @@ function useWatcher<T>() {
       (column) => column.type === 'selection'
     )
 
-    let selectColFixLeft
+    let selectColFixLeft: boolean
     if (
       selectColumn &&
       selectColumn.fixed !== 'right' &&
--- a/packages/components/table/src/store/watcher.ts
+++ b/packages/components/table/src/store/watcher.ts
@@ -22,7 +22,7 @@ import type {
 } from '../table/defaults'
 import type { StoreFilter } from '.'
 
-const sortData = <T>(
+const sortData = <T extends DefaultRow>(
   data: T[],
   states: {
     sortingColumn: TableColumnCtx<T> | null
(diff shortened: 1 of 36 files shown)


packages/components/table/src/store/watcher.ts
warmthsea, line 1: All : `Ref<XX> = ref(xx)` changed to `= ref<XX>(xx)`  is better?

Dsaquel: It would be more readable yes,

but when comes using generic:
```diff
-  const data: Ref<T[]> = ref([])
+  const data = ref<T[]>([])
```
[variable ref](https://github.com/element-plus/element-plus/blob/refact/table-remove-ts-nocheck/packages/components/table/src/store/watcher.ts#L65).
 The output type of the variable would be: `Ref<UnwrapRefSimple<T>[]>`. And result to an typecheck error.

For another [variable](https://github.com/element-plus/element-plus/blob/refact/table-remove-ts-nocheck/packages/components/table/src/store/watcher.ts#L66)
 it would throw this error:
>Return type of exported function has or is using name 'ShallowReactiveMarker' from external module "/home/element/Developments/alchemies/element-plus/node_modules/@vue/reactivity/dist/reactivity" but cannot be named.

i don't know the reason of this btw.

packages/components/table/src/table/style-helper.ts
```diff
@@ -320,7 +319,7 @@ function useStyle<T>(
   /**
    * fix layout
    */
-  const handleFixedMousewheel = (event, data) => {
+  const handleFixedMousewheel = (event: WheelEvent, data: any) => {
     const bodyWrapper = table.refs.bodyWrapper
     if (Math.abs(data.spinY) > 0) {
       const currentScrollTop = bodyWrapper.scrollTop
```
warmthsea: These codes seem to be able to delete, or supplement the data type?

https://github.com/ElemeFE/element/blob/c345bb453bf11badb4831a6a3f600c9372b3a336/packages/table/src/table.vue#L396C6-L410C9

Dsaquel: :+1:

Dsaquel:
I will resolve the issues later.

Dsaquel:
https://github.com/element-plus/element-plus/pull/21134#issuecomment-3005109265

PR #21093, opened 2025-06-20
feat(components): [select] expose selectDisabled prop to #tag slot

baozjj:

- [x] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | [中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md).

---

### ✨ Feature Description

In our use case, we needed to fully customize the tag display of <el-select> using the #tag slot. These tags could be individually removed via a close icon.

However, we found that when <el-select> was placed inside a disabled <el-form>, the custom tags could still be removed, which was unexpected. This behavior happened because the default tag rendering handles the disabled state internally, but custom #tag slots do not receive any such information by default.

While it's possible to work around this by manually passing the disabled state from the parent, we believe the <el-select> component should expose its internal disabled state (selectDisabled) to the #tag slot, so users can handle this consistently and naturally within the slot template.

This PR adds that support, enabling developers to implement custom tag behavior that fully respects the disabled state of the component.

---

### ✅ What’s Changed

- Passed `selectDisabled` to the `#tag` slot of `<el-select>`
- Updated slot prop type declarations
- Added unit tests to verify:
  - The presence of `selectDisabled` in the `#tag` slot
  - Correct behavior when used within `<el-form disabled>`

---

### 🧪 Test Plan

- All unit tests pass via `pnpm test`
- New tests confirm the disabled state is reactive and slot-aware
- Ensures no regression to existing slot behaviors

---

### 🔗 Related Issues

_No related issues; this is a proactive enhancement._


btea (commented):
We should add a new example to the documentation, marking the relevant version number, to clearly tell users what version numbers are available.

baozjj:
> We should add a new example to the documentation, marking the relevant version number, to clearly tell users what version numbers are available.

Hi, thanks for the feedback! I’ll add a new example to the documentation showing how to use the selectDisabled prop in the #tag slot, and mark the version number where this feature is introduced.
I’ll submit the update shortly.
Thanks again!

baozjj:
> We should add a new example to the documentation, marking the relevant version number, to clearly tell users what version numbers are available.

Hi,I’ve updated the PR to include a documentation example with the version number 2.10.3 as suggested.
Please let me know if there’s anything else I can improve. Thanks for the review!

btea (approved):
Thanks!

warmthsea:
https://github.com/baozjj/element-plus/blob/feat/select-expose-disabled-to-tag-slot/packages/components/select-v2/src/select.vue#L59-L66

**select-v2** can be changed together.

baozjj:
> https://github.com/baozjj/element-plus/blob/feat/select-expose-disabled-to-tag-slot/packages/components/select-v2/src/select.vue#L59-L66
> 
> **select-v2** can be changed together.

Got it. I’ll apply the same change to select-v2 to keep both versions aligned.

baozjj:
> https://github.com/baozjj/element-plus/blob/feat/select-expose-disabled-to-tag-slot/packages/components/select-v2/src/select.vue#L59-L66
> 
> **select-v2** can be changed together.

I've applied the same change to `select-v2` as suggested to ensure consistency. Please let me know if any further adjustments are needed.

warmthsea:
The examples can be deleted, only the bottom document is updated, refer to https://github.com/element-plus/element-plus/pull/21157/files#diff-1fe792429b649feda5d215381b79aa885bdc2ccba1f3dceb0ae4a454ff58dba5R265

baozjj:
> The examples can be deleted, only the bottom document is updated, refer to https://github.com/element-plus/element-plus/pull/21157/files#diff-1fe792429b649feda5d215381b79aa885bdc2ccba1f3dceb0ae4a454ff58dba5R265

The changes have been pushed. Please help review again when convenient.

Dsaquel:
@baozjj He talked about deleting both `tag-disabled.vue` examples files.

baozjj:
> @baozjj He talked about deleting both `tag-disabled.vue` examples files.

Example files have been removed following the suggestion. Appreciate the guidance.

Dsaquel (approved):
LGTM

warmthsea (approved):

PR #21095, opened 2025-06-21
fix(components): [collapse-item] prevent keyboard navigation when disabled

w2xi:
## Description

Fix issue where disabled collapse items could still be toggled using keyboard navigation (Enter/Space keys).

## Changes

- Added disabled check to `handleEnterClick` function in `use-collapse-item.ts`
- Now keyboard navigation (Enter key) respects the `disabled` prop, matching the behavior of mouse clicks

## Before
- Disabled collapse items would still toggle when pressing Enter key
- Inconsistent behavior between mouse clicks and keyboard navigation

## After  
- Disabled collapse items ignore both mouse clicks and keyboard navigation
- Consistent behavior across all interaction methods

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

Existing tests already cover this functionality:
- `disabled collapse item should not toggle on keyboard enter`
- `disabled collapse item should not toggle on keyboard space`
- All existing tests continue to pass

### DIFF
--- a/packages/components/collapse/__tests__/collapse.test.tsx
+++ b/packages/components/collapse/__tests__/collapse.test.tsx
@@ -428,4 +428,219 @@ describe('Collapse.vue', () => {
       expect(collapseItem.vm.isActive).toBe(false)
     })
   })
+
+  describe('Disabled CollapseItem', () => {
+    test('disabled collapse item should not toggle on click', async () => {
+      const wrapper = mount({
+        data() {
+          return {
+            activeNames: ['1'],
+          }
+        },
+        render() {
+          return (
+            <Collapse v-model={this.activeNames}>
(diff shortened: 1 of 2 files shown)


tolking:
> `:tabindex="disabled ? -1 : 0"`

The reason for the issues is that `tabindex` is -1, but it can still capture focus. Need to be modified to `:tabindex="disabled ? undefined : 0"`.

w2xi:
> > `:tabindex="disabled ? -1 : 0"`
> 
> The reason for the issues is that `tabindex` is -1, but it can still capture focus. Need to be modified to `:tabindex="disabled ? undefined : 0"`.

Oh, I get it. And I see you've submit a PR to solve this problem, thx.

PR #21099, opened 2025-06-22
refactor(docs): partially remove typecript errors

Dsaquel:
I wanted initially to add a `typecheck:docs` command but the element-plus alias path in [docs/tsconfig.json](https://github.com/element-plus/element-plus/blob/dev/docs/tsconfig.json#L15) will cause the typecheck to be effective on `packages` directory too, [more info](https://github.com/vuejs/language-tools/issues/3144).

So I just ran `vue-tsc -p ./docs/tsconfig.json --composite false --noEmit`  and handled errors in docs.
In fact, there were a lot of errors due to the difference of the vue version between vitepress and ours (`3.4.27` **vs** `3.2.37`). I decided to sync vue doc version with the vitepress version, it reduce 94 ts errors.
With the others refact: -138 errors.

That said, it can be a great opportunity to see if components will have issues in the doc, even if it's two minor versions.


docs/.vitepress/config/sponsors.ts
```diff
@@ -1,4 +1,4 @@
-export const rightRichTextSponsors = []
+export const rightRichTextSponsors: never[] = []
```
Dsaquel: `rightRichTextSponsors` and `leftLogoSponsors` made `Sponsor` to be `any`, `never` avoid this.

docs/package.json
```diff
@@ -22,7 +22,6 @@
   },
   "devDependencies": {
     "@crowdin/cli": "^4.5.0",
-    "@docsearch/react": "^3.1.0",
```
Dsaquel, line 25: `@docsearch/js` has the `react` folder too.

docs/components.d.ts
```diff
@@ -160,7 +160,6 @@ declare module 'vue' {
     VpFooter: typeof import('./.vitepress/vitepress/components/globals/vp-footer.vue')['default']
     VpHamburger: typeof import('./.vitepress/vitepress/components/navbar/vp-hamburger.vue')['default']
     VpHeroContent: typeof import('./.vitepress/vitepress/components/vp-hero-content.vue')['default']
-    VpLastUpdatedAt: typeof import('./.vitepress/vitepress/components/doc-content/vp-last-updated-at.vue')['default']
```
Dsaquel, line 163: Not used.

docs/.vitepress/config/sponsors.ts
warmthsea: 1. `name_cn` related codes can be retained.
2. `rightRichTextSponsors` & `leftLogoSponsors` type can refer to the git record.
   - It is best to delete `any` here, the ide display has some problems. 


image

image

warmthsea (approved):

docs/.vitepress/vitepress/components/vp-sponsor-large.vue
```diff
@@ -2,10 +2,12 @@
...
 
-defineProps({
-  itemClass: String,
-  itemStyle: [String, Object, Array],
-})
+import type { CSSProperties } from 'vue'
+
+defineProps<{
+  itemClass?: string
+  itemStyle?: CSSProperties
```
Dsaquel, line 9: @warmthsea I had also this error but after `pnpm install` vue 3.4 will recognize it as  `StyleValue`. Do you still have this error after install?

warmthsea: 👍 It is indeed a problem with my local version. I have submitted another PR.

PR #21123, opened 2025-06-24
docs(components): [table] update `filter-change` event description

dhj-l:
docs(docs): fix filter-change event description in table #21104


warmthsea (approved):
LGTM, Thanks!

PR #21139, opened 2025-06-25
refactor(components): [table] remove `@ts-nocheck`

Dsaquel:
1/3

`@ts-nocheck` removed: **10**.
`@ts-nocheck` remaining: **19**.

It's hard to remove smoothly `ts-nocheck` when trying to handle files that depend on `useStore`, `TableColumnCtx`, `util.ts`. So i decided to copy paste files from https://github.com/element-plus/element-plus/pull/21090.
It's surely this pr who will have the biggest change.

---
#### Summary ([copying](https://github.com/element-plus/element-plus/pull/21090#issue-3161599447)):

This pr prepare will remove `ts-nocheck` for table component, it will facilitate contribution on this component and it has been made to bring generic easily when we'll upgrade vue.

However, it introduce a small breaking change, since the generic across the component has been change:
```diff
- type DefaultRow = any
+ type DefaultRow = Record<PropertyKey, any>

here PropertyKey is a native typescript alias for string | number | symbol
```
This change is necessary to remove `ts-nocheck` properly and add more reliability.

This will break typescript when the user will have:
```diff
- interface SummaryMethodProps<T = Product> {
+ interface SummaryMethodProps<T extends Record<string, any> = Product> {
  columns: TableColumnCtx<T>[]
  data: T[]
}
```
[from this example](https://github.com/element-plus/element-plus/blob/dev/docs/examples/table/summary.vue#L40-L43)

It can affect many users.

### DIFF
--- a/packages/components/table/src/store/index.ts
+++ b/packages/components/table/src/store/index.ts
@@ -118,8 +118,8 @@ function useStore<T extends DefaultRow>() {
       updateColumnOrder: () => void
     ) {
       const array = unref(states._columns) || []
-      if (parent?.children) {
-        parent.children.splice(
+      if (parent) {
+        parent.children?.splice(
           parent.children.findIndex((item) => item.id === column.id),
           1
         )
--- a/packages/components/table/src/store/index.ts
+++ b/packages/components/table/src/store/index.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 import { getCurrentInstance, nextTick, unref } from 'vue'
 import { isNull } from 'lodash-unified'
 import { useNamespace } from '@element-plus/hooks'
@@ -6,14 +5,14 @@ import useWatcher from './watcher'
 
 import type { Ref } from 'vue'
 import type { TableColumnCtx } from '../table-column/defaults'
-import type { Filter, Sort, Table } from '../table/defaults'
+import type { DefaultRow, Filter, Sort, Table } from '../table/defaults'
 
-interface WatcherPropsData<T> {
+interface WatcherPropsData<T extends DefaultRow> {
   data: Ref<T[]>
-  rowKey: Ref<string>
+  rowKey: Ref<string | null>
 }
 
-function replaceColumn<T>(
+function replaceColumn<T extends DefaultRow>(
   array: TableColumnCtx<T>[],
   column: TableColumnCtx<T>
 ) {
@@ -27,7 +26,7 @@ function replaceColumn<T>(
   })
 }
 
-function sortColumn<T>(array: TableColumnCtx<T>[]) {
+function sortColumn<T extends DefaultRow>(array: TableColumnCtx<T>[]) {
   array.forEach((item) => {
     item.no = item.getColumnIndex?.()
     if (item.children?.length) {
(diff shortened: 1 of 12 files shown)


packages/components/table/src/store/index.ts
```diff
@@ -211,14 +210,17 @@ function useStore<T>() {
...
   }
-  const commit = function (name: keyof typeof mutations, ...args) {
+  const commit = function (name: keyof typeof mutations, ...args: any[]) {
     const mutations = instance.store.mutations
     if (mutations[name]) {
-      mutations[name].apply(instance, [instance.store.states].concat(args))
+      ;(mutations[name] as any).apply(
+        instance,
+        [instance.store.states].concat(args)
+      )
```
btea: Here we should be able to use spread syntax instead of apply.

packages/components/table/src/store/tree.ts
```diff
@@ -24,16 +29,18 @@ function useTree<T>(watcherData: WatcherPropsData<T>) {
...
     keys.forEach((key) => {
       if (lazyTreeNodeMap.value[key].length) {
-        const item = { children: [] }
+        const item: typeof res[number] = { children: [] }
         lazyTreeNodeMap.value[key].forEach((row) => {
           const currentRowKey = getRowIdentity(row, rowKey)
-          item.children.push(currentRowKey)
-          if (row[lazyColumnIdentifier.value] && !res[currentRowKey]) {
-            res[currentRowKey] = { children: [] }
+          if (currentRowKey) {
```
btea: If numbers are supported in the future, the value may be 0. Maybe we can make it compatible in advance?

packages/components/table/src/store/index.ts
Dsaquel: d1e3fc494119

packages/components/table/src/store/tree.ts
Dsaquel: I just removed the judgement if that respond to the question. 4bda0d177a4d

packages/components/table/src/util.ts
```diff
@@ -627,9 +646,9 @@ export const getFixedColumnOffset = <T>(
     after = 0,
   } = isFixedColumn(index, fixed, store, realColumns)
   if (!direction) {
-    return
+    return null
```
btea: There seems to be no difference between returning null and undefined here.

Dsaquel: I changed the type arg of `ensurePosition` aswell because `getFixedColumnOffset` is always used along `ensurePosition`, [example](https://github.com/element-plus/element-plus/blob/a174c6e0c3fc0d568683e608c75cef330bf53d80/packages/components/table/src/table-body/styles-helper.ts#L72-L78).

btea (approved):
👍

packages/components/table/src/store/index.ts
```diff
@@ -119,7 +118,7 @@ function useStore<T>() {
       updateColumnOrder: () => void
     ) {
       const array = unref(states._columns) || []
-      if (parent) {
+      if (parent?.children) {
         parent.children.splice(
```
warmthsea: This change may break the original logic, so I think it is best to keep this PR as is.

```suggestion
      if (parent) {
        parent.children?.splice(
```

warmthsea (approved):

PR #21200, opened 2025-07-01
refactor(components): [table] remove `@ts-nocheck`

Dsaquel:
3/3

`@ts-nocheck` removed: **11**.
`@ts-nocheck` remaining: **0**.

### DIFF
--- a/packages/components/table/src/table-column/render-helper.ts
+++ b/packages/components/table/src/table-column/render-helper.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 import {
   Comment,
   computed,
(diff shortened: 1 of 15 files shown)


packages/components/table/src/table-column/render-helper.ts
warmthsea, line 1: I found that most of the values ​​in this file are any. If I don’t add some types now, I’m afraid it will be difficult to change them later. 🤔 

image

Dsaquel: I agree, in these files it's pretty exhausting to handle them correctly without having to do a lot of `keyof typeof`.
I'll see if i can remove some more `any` tonight.

warmthsea (approved):

PR #21222, opened 2025-07-03
feat(types): [components] add props public type

yuhengshen:

组件的`props类型`有两种，一种组件内部使用，一种暴露给外部声明使用
image
image

给所有组件新增public props，供外部使用

由于当前库Vue版本3.2，缺少`ExtractPublicPropTypes`同时为了兼容高版本Vue，临时使用`__ExtractPublicPropTypes`替代。
等Vue依赖升级后可以删除这部分代码，并全局替换`__ExtractPublicPropTypes`->`ExtractPublicPropTypes`。

代码变动：
- 所有组件增加 public props
- packages/utils/vue/props/util.ts 引入__ExtractPublicPropTypes
- 模板调整

### DIFF
--- a/packages/utils/vue/props/util.ts
+++ b/packages/utils/vue/props/util.ts
@@ -9,9 +9,8 @@ export type IfUnknown<T, Y, N> = [unknown] extends [T] ? Y : N
 
 export type UnknownToNever<T> = IfUnknown<T, never, T>
 
-// ExtractPublicPropTypes copy from vue 3.3
 // delete when upgrade to vue 3.3 : start
-
+// __ExtractPublicPropTypes copy from vue 3.3
 // If the type T accepts type "any", output type Y, otherwise output type N.
 // https://stackoverflow.com/questions/49927523/disallow-call-with-any/49928360#49928360
 export type IfAny<T, Y, N> = 0 extends 1 & T ? Y : N
@@ -53,12 +52,12 @@ type PublicOptionalKeys<T> = Exclude<keyof T, PublicRequiredKeys<T>>
  */
 
 declare module 'vue' {
-  export type ExtractPublicPropTypes<O> = {
+  // compatible with higher versions of Vue
+  export type __ExtractPublicPropTypes<O> = {
     [K in keyof Pick<O, PublicRequiredKeys<O>>]: InferPropType<O[K]>
   } & {
     [K in keyof Pick<O, PublicOptionalKeys<O>>]?: InferPropType<O[K]>
   }
 }
 // delete when upgrade to vue 3.3 : end
-
 export {}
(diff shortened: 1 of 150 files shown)


Dsaquel (commented):
Perhaps we can replace ~~`ExtractPublicPropTypes`~~ `ExtractPropTypes` with `__ExtractPublicPropTypes` to avoid migrating the interface name.

yuhengshen:
> Perhaps we can replace `ExtractPublicPropTypes` with `__ExtractPublicPropTypes` to avoid migrating the interface name.

Using __ExtractPublicPropTypes for current resolves type conflicts in Vue 3.3 and above.

Dsaquel:
When upgrading to vue 3.3 or higher we'll make sure to replace with `ExtractPublicPropTypes`.

yuhengshen:
> Perhaps we can replace ~`ExtractPublicPropTypes`~ `ExtractPropTypes` with `__ExtractPublicPropTypes` to avoid migrating the interface name.

- __ExtractPublicPropTypes make some props with default value optional, this will make component internal type error.
- This ensures backward compatibility, and the previously imported ExtractPropTypes will behave consistently in existing code.

No need to migrate the interface name; these two types are useful.

Dsaquel:
Oh, ok i see now. Thanks for explanation.

packages/utils/vue/props/util.ts
```diff
@@ -1,3 +1,5 @@
+import { Prop } from 'vue'
```
Dsaquel:
```suggestion
import type { Prop } from 'vue'
```

Dsaquel (commented):
Looks good.

Dsaquel (commented):
I was worried about what is used for but it can be useful for partial props assignment.

Is there any other useful use case? @yuhengshen

yuhengshen:
> I was worried about what is used for but it can be useful for partial props assignment.
> 
> Is there any other useful use case? @yuhengshen

In most cases, declaring Props types with `Partial` is sufficient, except when certain properties are `required: true`.
image

https://github.com/element-plus/element-plus/blob/cfe8e4e982d529b7c76f296384713b8c656899bb/packages/components/select-v2/src/defaults.ts#L177

I checked and found that there aren't many properties that actually `required: true`. Perhaps we should consider whether introducing this type is really necessary.

But on the other hand, exposing public props to users would be more reasonable.

yuhengshen:
I'm building a dynamic form, so there needs to be clear definitions about which properties are optional.

Dsaquel (approved):

btea (approved):

makedopamine (approved):

packages/utils/vue/props/util.ts
warmthsea, line 1: It seems that this file code causes component **types to be lost** and IDE(Volar) display abnormality.

image


git checkout 58a5852d2dff643e871c26e67a675b426c092094

image

git checkout cfc661c62678901fdc4187262dfaa035369510c0

image

yuhengshen: fix: https://github.com/element-plus/element-plus/pull/21266

PR #21235, opened 2025-07-04
docs: select use minimum base examples & delete input tip

warmthsea:
- select、select-v2 use minimum base examples.
  - _Because it is a commonly used component, it is changed to facilitate user use, [refer to input](https://element-plus.org/en-US/component/input.html#basic-usage)._ 
- delete input warning tip.
  - _This is the most basic usage of vue, I think it can be deleted._ 

| -        | Before | After |
| ------ | ------- | ----- |
| select、select-v2        |     image  |     image     |
|  input tip | image  | image |

### DIFF
--- a/docs/en-US/component/input.md
+++ b/docs/en-US/component/input.md
@@ -7,16 +7,6 @@ lang: en-US
 
 Input data using mouse or keyboard.
 
-:::warning
-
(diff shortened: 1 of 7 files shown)


Dsaquel (approved):
LGTM.

PR #21256, opened 2025-07-04
fix(components): [input-tag] paste multiple delimiters

Dsaquel:

resolve https://github.com/element-plus/element-plus/pull/20532#issuecomment-3034290512

### DIFF
--- a/packages/components/input-tag/__tests__/input-tag.test.tsx
+++ b/packages/components/input-tag/__tests__/input-tag.test.tsx
@@ -237,30 +237,43 @@ describe('InputTag.vue', () => {
   describe('delimiter', () => {
     test('with string', async () => {
       const inputValue = ref<string[]>()
+      const addTag = vi.fn()
       const wrapper = mount(() => (
-        <InputTag v-model={inputValue.value} delimiter="," />
+        <InputTag v-model={inputValue.value} delimiter="," onAdd-tag={addTag} />
       ))
-
       await wrapper.find('input').setValue(`${AXIOM},`)
+
+      expect(addTag).toBeCalledWith(AXIOM)
       expect(wrapper.findAll('.el-tag').length).toBe(1)
       expect(wrapper.find('.el-tag').text()).toBe(AXIOM)
       expect(inputValue.value).toEqual([AXIOM])
     })
     test('with RegExp', async () => {
       const inputValue = ref<string[]>()
+      const addTag = vi.fn()
       const wrapper = mount(() => (
-        <InputTag v-model={inputValue.value} delimiter={/\./} />
+        <InputTag
+          v-model={inputValue.value}
+          delimiter={/\./}
+          onAdd-tag={addTag}
+        />
       ))
-
       await wrapper.find('input').setValue(`${AXIOM}.`)
+
+      expect(addTag).toBeCalledWith(AXIOM)
       expect(wrapper.findAll('.el-tag').length).toBe(1)
       expect(wrapper.find('.el-tag').text()).toBe(AXIOM)
       expect(inputValue.value).toEqual([AXIOM])
     })
     test('paste multiple delimiter', async () => {
       const inputValue = ref<string[]>()
+      const addTag = vi.fn()
       const wrapper = mount(() => (
-        <InputTag v-model={inputValue.value} delimiter={/\./} />
+        <InputTag
+          v-model={inputValue.value}
+          delimiter={/\./}
+          onAdd-tag={addTag}
+        />
       ))
 
       const clipboardData = new DataTransfer()
@@ -269,25 +282,57 @@ describe('InputTag.vue', () => {
       wrapper.find('input').element?.dispatchEvent(evt)
       await wrapper.find('input').trigger('paste')
 
+      const result = [AXIOM, AXIOM, AXIOM, AXIOM]
       expect(wrapper.findAll('.el-tag').length).toBe(4)
+      expect(addTag).toBeCalledWith(result)
       wrapper
         .findAll('.el-tag')
         .forEach((tag) => expect(tag.text()).toBe(AXIOM))
-      expect(inputValue.value).toEqual([AXIOM, AXIOM, AXIOM, AXIOM])
+      expect(inputValue.value).toEqual(result)
     })
     test('should add multiple tags without paste', async () => {
       const inputValue = ref<string[]>()
+      const addTag = vi.fn()
       const wrapper = mount(() => (
-        <InputTag v-model={inputValue.value} delimiter={/\./} />
+        <InputTag
+          v-model={inputValue.value}
+          delimiter={/\./}
+          onAdd-tag={addTag}
+        />
       ))
-
       await wrapper.find('input').setValue(`${AXIOM}.${AXIOM}.${AXIOM}`)
 
+      const result = [AXIOM, AXIOM, AXIOM]
+      expect(addTag).toBeCalledWith(result)
       expect(wrapper.findAll('.el-tag').length).toBe(3)
       wrapper
         .findAll('.el-tag')
         .forEach((tag) => expect(tag.text()).toBe(AXIOM))
-      expect(inputValue.value).toEqual([AXIOM, AXIOM, AXIOM])
+      expect(inputValue.value).toEqual(result)
+    })
+    test('should return single string on paste', async () => {
+      const inputValue = ref<string[]>()
+      const addTag = vi.fn()
+      const wrapper = mount(() => (
+        <InputTag
+          v-model={inputValue.value}
+          delimiter={/\./}
+          onAdd-tag={addTag}
+        />
+      ))
+
+      const clipboardData = new DataTransfer()
+      clipboardData.setData('text', `${AXIOM}.`)
+      const evt = new ClipboardEvent('paste', { clipboardData })
+      wrapper.find('input').element?.dispatchEvent(evt)
+      await wrapper.find('input').trigger('paste')
+
+      expect(addTag).toBeCalledWith(AXIOM)
+      expect(wrapper.findAll('.el-tag').length).toBe(1)
+      wrapper
+        .findAll('.el-tag')
+        .forEach((tag) => expect(tag.text()).toBe(AXIOM))
+      expect(inputValue.value).toEqual([AXIOM])
     })
   })
 
--- a/packages/components/input-tag/__tests__/input-tag.test.tsx
+++ b/packages/components/input-tag/__tests__/input-tag.test.tsx
@@ -1,7 +1,10 @@
+/**
+ * @vitest-environment happy-dom
+ */
+
 import { nextTick, ref } from 'vue'
 import { mount } from '@vue/test-utils'
 import { describe, expect, test, vi } from 'vitest'
-import sleep from '@element-plus/test-utils/sleep'
 import { ComponentSize, EVENT_CODE } from '@element-plus/constants'
 import { InputTagInstance } from '@element-plus/components/input-tag'
 import FormItem from '@element-plus/components/form/src/form-item.vue'
@@ -260,11 +263,12 @@ describe('InputTag.vue', () => {
         <InputTag v-model={inputValue.value} delimiter={/\./} />
       ))
 
-      console.log('val', `${AXIOM}.${AXIOM}.${AXIOM}.${AXIOM}.`)
-      await wrapper
-        .find('input')
-        .setValue(`${AXIOM}.${AXIOM}.${AXIOM}.${AXIOM}.`)
-      await sleep()
+      const clipboardData = new DataTransfer()
+      clipboardData.setData('text', `${AXIOM}.${AXIOM}.${AXIOM}.${AXIOM}.`)
+      const evt = new ClipboardEvent('paste', { clipboardData })
+      wrapper.find('input').element?.dispatchEvent(evt)
+      await wrapper.find('input').trigger('paste')
+
       expect(wrapper.findAll('.el-tag').length).toBe(4)
       wrapper
         .findAll('.el-tag')
--- a/packages/components/input-tag/__tests__/input-tag.test.tsx
+++ b/packages/components/input-tag/__tests__/input-tag.test.tsx
@@ -1,7 +1,3 @@
-/**
- * @vitest-environment happy-dom
- */
-
 import { nextTick, ref } from 'vue'
 import { mount } from '@vue/test-utils'
 import { describe, expect, test, vi } from 'vitest'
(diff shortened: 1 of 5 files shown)


packages/components/input-tag/__tests__/input-tag.test.tsx
```diff
@@ -253,6 +254,23 @@ describe('InputTag.vue', () => {
       expect(wrapper.find('.el-tag').text()).toBe(AXIOM)
       expect(inputValue.value).toEqual([AXIOM])
     })
+    test('paste multiple delimiter', async () => {
+      const inputValue = ref<string[]>()
+      const wrapper = mount(() => (
+        <InputTag v-model={inputValue.value} delimiter={/\./} />
+      ))
+
+      console.log('val', `${AXIOM}.${AXIOM}.${AXIOM}.${AXIOM}.`)
```
btea: This line can be removed.

packages/components/input-tag/__tests__/input-tag.test.tsx
```diff
@@ -253,6 +254,23 @@ describe('InputTag.vue', () => {
...
+      await wrapper
+        .find('input')
+        .setValue(`${AXIOM}.${AXIOM}.${AXIOM}.${AXIOM}.`)
+      await sleep()
+      expect(wrapper.findAll('.el-tag').length).toBe(4)
```
btea, line 280: https://vitest.dev/api/vi.html#vi-waituntil

I think we can use `vi.waitUnti` to replace sleep, which looks more friendly.

packages/components/input-tag/src/composables/use-input-tag.ts
```diff
@@ -41,18 +46,27 @@ export function useInputTag({ props, emit, formItem }: UseInputTagOptions) {
...
+        props.delimiter instanceof RegExp
+          ? props.delimiter
+          : new RegExp(escapeStringRegexp(props.delimiter), 'i')
+      const tags = inputValue?.value?.split(regexp) ?? []
+
+      for (let i = 0; i < tags.length; i++) {
+        const tag = tags[i]
+        if (tag.length !== inputValue.value?.length) {
+          inputValue.value = tag
+          handleAddTag()
```
makedopamine: Calling `handleAddTag` emits `update:modeValue`, `change` and `add-tag` events. Does it make sense to emit them multiple times in one paste action? 🤔

Dsaquel: In fact it make more sense, Im just worried about one thing:

- Should we add an extra event for the `paste` event? If so it's a little bit annoying for the developer to handle two data flows that do the same. Maybe we could emitting the batch update directly via `add-tag` but it will make inconsistent with the name of the event and the payload type (string !== string[]).

I'll try changing this.

packages/components/input-tag/__tests__/input-tag.test.tsx
Dsaquel: Thanks for advice, since no need anymore i just removed it.

packages/components/input-tag/src/composables/use-input-tag.ts
Dsaquel: Let me know if you think it's more valuable to switch more into a `paste` event.

packages/components/input-tag/__tests__/input-tag.test.tsx
```diff
@@ -1,3 +1,7 @@
+/**
+ * @vitest-environment happy-dom
+ */
```
Dsaquel: Since the [clipboard](https://github.com/jsdom/jsdom/issues/1568) api is not currently supported in js-dom, we switch to happy-dom. inspired by [autocomplete test](https://github.com/element-plus/element-plus/blob/dev/packages/components/autocomplete/__tests__/autocomplete.test.tsx#L1-L3).

packages/components/input-tag/src/composables/use-input-tag.ts
```diff
@@ -83,14 +94,29 @@ export function useInputTag({ props, emit, formItem }: UseInputTagOptions) {
...
 
+  const handlePaste = (event: ClipboardEvent) => {
+    const data = event.clipboardData?.getData('text')
+    if (!data) return
+    const regexp =
+      props.delimiter instanceof RegExp
+        ? props.delimiter
+        : new RegExp(escapeStringRegexp(props.delimiter), 'i')
+
+    const tags = data.split(regexp).filter((val) => val && val !== data)
```
btea:
```suggestion
    const tags = data.split(props.delimiter).filter((val) => val && val !== data)
```
If `props.delimiter` is a string, perhaps we should treat it as a normal string, as in line 63.

makedopamine:
If I paste a string without the delimiter into the input, like "foo", the input will be empty. Is that intentional? @Dsaquel

packages/components/input-tag/src/composables/use-input-tag.ts
makedopamine: >Let me know if you think it's more valuable to switch more into a paste event.

Although allowing `add-tag` to accept an array of strings might be confusing, I think it’s better than adding a paste event.

Dsaquel:
> If I paste a string without the delimiter into the input, like "foo", the input will be empty. Is that intentional? @Dsaquel

Nope, there are an another issue when i split the string with the delimiter it should also bulk update as handlePaste do. I'll fix it both.

packages/components/input-tag/src/composables/use-input-tag.ts
```diff
@@ -41,6 +39,29 @@ export function useInputTag({ props, emit, formItem }: UseInputTagOptions) {
...
+    const tags = input
+      .split(props.delimiter)
+      .filter((val) => val && val !== input)
+    if (props.max) {
+      const maxInsert = props.max - (props.modelValue?.length ?? 0)
+      tags.splice(maxInsert)
+    }
+    return {
+      tags,
+      hasDelimited: tags.length && tags[0] !== input,
```
makedopamine: I don’t see why `tags[0] !== input` is needed. At line 54, `val !== input` seems to do the same thing. Could you give me an example?

Dsaquel: >I don’t see why tags[0] !== input is needed

You right, not used anymore.

packages/components/input-tag/src/composables/use-input-tag.ts
Dsaquel: Since i found we can separate the string input with delimiter i think we can reconsider the paste event.

makedopamine:
@Dsaquel 
I found an edge case. Maybe it's not worth addressing but I want to bring it to your attention.

[Reproduction](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtaW5wdXQtdGFnXG4gICAgdi1tb2RlbD1cImlucHV0XCJcbiAgICBkcmFnZ2FibGVcbiAgICBwbGFjZWhvbGRlcj1cIlRyeSB0byBzZXBhcmF0ZSB3b3JkcyB3aXRoICxcIlxuICAgIGRlbGltaXRlcj1cIixcIlxuICAvPlxuPC90ZW1wbGF0ZT5cblxuPHNjcmlwdCBsYW5nPVwidHNcIiBzZXR1cD5cbmltcG9ydCB7IHJlZiB9IGZyb20gJ3Z1ZSdcblxuY29uc3QgaW5wdXQgPSByZWY8c3RyaW5nW10+KFtdKVxuPC9zY3JpcHQ+XG4iLCJlbGVtZW50LXBsdXMuanMiOiJpbXBvcnQgRWxlbWVudFBsdXMgZnJvbSAnZWxlbWVudC1wbHVzJ1xuaW1wb3J0IHsgZ2V0Q3VycmVudEluc3RhbmNlIH0gZnJvbSAndnVlJ1xuXG5sZXQgaW5zdGFsbGVkID0gZmFsc2VcbmF3YWl0IGxvYWRTdHlsZSgpXG5cbmV4cG9ydCBmdW5jdGlvbiBzZXR1cEVsZW1lbnRQbHVzKCkge1xuICBpZiAoaW5zdGFsbGVkKSByZXR1cm5cbiAgY29uc3QgaW5zdGFuY2UgPSBnZXRDdXJyZW50SW5zdGFuY2UoKVxuICBpbnN0YW5jZS5hcHBDb250ZXh0LmFwcC51c2UoRWxlbWVudFBsdXMpXG4gIGluc3RhbGxlZCA9IHRydWVcbn1cblxuZXhwb3J0IGZ1bmN0aW9uIGxvYWRTdHlsZSgpIHtcbiAgY29uc3Qgc3R5bGVzID0gWydodHRwczovL3VucGtnLmNvbS9lbGVtZW50LXBsdXNAbGF0ZXN0L2Rpc3QvaW5kZXguY3NzJywgJ2h0dHBzOi8vdW5wa2cuY29tL2VsZW1lbnQtcGx1c0BsYXRlc3QvdGhlbWUtY2hhbGsvZGFyay9jc3MtdmFycy5jc3MnXS5tYXAoKHN0eWxlKSA9PiB7XG4gICAgcmV0dXJuIG5ldyBQcm9taXNlKChyZXNvbHZlLCByZWplY3QpID0+IHtcbiAgICAgIGNvbnN0IGxpbmsgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCdsaW5rJylcbiAgICAgIGxpbmsucmVsID0gJ3N0eWxlc2hlZXQnXG4gICAgICBsaW5rLmhyZWYgPSBzdHlsZVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdsb2FkJywgcmVzb2x2ZSlcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignZXJyb3InLCByZWplY3QpXG4gICAgICBkb2N1bWVudC5ib2R5LmFwcGVuZChsaW5rKVxuICAgIH0pXG4gIH0pXG4gIHJldHVybiBQcm9taXNlLmFsbFNldHRsZWQoc3R5bGVzKVxufVxuIiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IFwiRVNOZXh0XCIsXG4gICAgXCJqc3hcIjogXCJwcmVzZXJ2ZVwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwidHlwZXNcIjogW1wiZWxlbWVudC1wbHVzL2dsb2JhbC5kLnRzXCJdLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcImFsbG93SnNcIjogdHJ1ZSxcbiAgICBcImNoZWNrSnNcIjogdHJ1ZVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vdW5wa2cuY29tL0B2dWUvcnVudGltZS1kb21AbGF0ZXN0L2Rpc3QvcnVudGltZS1kb20uZXNtLWJyb3dzZXIuanNcIixcbiAgICBcIkB2dWUvc2hhcmVkXCI6IFwiaHR0cHM6Ly91bnBrZy5jb20vQHZ1ZS9zaGFyZWRAbGF0ZXN0L2Rpc3Qvc2hhcmVkLmVzbS1idW5kbGVyLmpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXNcIjogXCJodHRwczovL3VucGtnLmNvbS9lbGVtZW50LXBsdXNAbGF0ZXN0L2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwiaHR0cHM6Ly91bnBrZy5jb20vZWxlbWVudC1wbHVzQGxhdGVzdC9cIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly91bnBrZy5jb20vQGVsZW1lbnQtcGx1cy9pY29ucy12dWVAMi9kaXN0L2luZGV4Lm1pbi5qc1wiXG4gIH0sXG4gIFwic2NvcGVzXCI6IHt9XG59IiwiX28iOnt9fQ==) steps:
1. Copy ",bar" to the clipboard
2. Type "foo" in the input
3. Paste ",bar"

Expected behavior: Two separate tags `foo` and `bar`.
Actual behavior: Only one tag `foobar`.

btea:
@makedopamine  This scene is quite special. 👍

packages/components/input-tag/src/composables/use-input-tag.ts
```diff
@@ -41,6 +39,26 @@ export function useInputTag({ props, emit, formItem }: UseInputTagOptions) {
...
+
+    emit(UPDATE_MODEL_EVENT, list)
+    emit(CHANGE_EVENT, list)
+    emit('add-tag', value)
+    inputValue.value = undefined
+  }
+
+  const getDelimitedTags = (input: string) => {
+    const tags = input
+      .split(props.delimiter)
```
btea, line 53: Maybe we should add a check `if(props.delimiter)`, otherwise it will be cut according to the empty string.

Dsaquel: Sorry i don't understand, do you mean a check on input ?

btea: Sorry, there was something wrong with my comment. I have updated it.

Dsaquel:
@makedopamine thanks for you attention, i will fix it. I didn't know it will take such effort to fix the paste event 😶.

packages/components/input-tag/src/composables/use-input-tag.ts
Dsaquel: With the latest change the judgment is made only in `handleInput`.

Dsaquel:
Ok so i removed `onPaste` because i should've insert the string copied where the cursor index is... Too much work for nothing and there is no reason now to have a paste listener anymore.

That said i didn't made the copy-paste test work. But everything should work now.

makedopamine (approved):
LGTM. Just make sure to update the function signature of `add-tag` in the documentation.

btea (approved):

warmthsea (approved):

PR #21266, opened 2025-07-07
fix(types): [other] cannot find declare module types from installed packages

yuhengshen:

`@microsoft/api-extractor` breaks TypeScript's recognition of declare module 'vue'.
​**fix**:​​ Temporarily resolve this by manually importing it. 
 
Test it with pnpm, it's working now

A better fix may exist.

### DIFF
--- a/packages/utils/vue3.3.polyfill.ts
+++ b/packages/utils/vue3.3.polyfill.ts
@@ -1,0 +1,51 @@
+import type { Prop } from 'vue'
+// delete when upgrade to vue 3.3 : start
+// __ExtractPublicPropTypes copy from vue 3.3
+// If the type T accepts type "any", output type Y, otherwise output type N.
+// https://stackoverflow.com/questions/49927523/disallow-call-with-any/49928360#49928360
+type IfAny<T, Y, N> = 0 extends 1 & T ? Y : N
+
+type InferPropType<T, NullAsAny = true> = [T] extends [null]
+  ? NullAsAny extends true
+    ? any
+    : null
+  : [T] extends [{ type: null | true }]
+  ? any // As TS issue https://github.com/Microsoft/TypeScript/issues/14829 // somehow `ObjectConstructor` when inferred from { (): T } becomes `any` // `BooleanConstructor` when inferred from PropConstructor(with PropMethod) becomes `Boolean`
+  : [T] extends [ObjectConstructor | { type: ObjectConstructor }]
+  ? Record<string, any>
+  : [T] extends [BooleanConstructor | { type: BooleanConstructor }]
+  ? boolean
+  : [T] extends [DateConstructor | { type: DateConstructor }]
+  ? Date
+  : [T] extends [(infer U)[] | { type: (infer U)[] }]
+  ? U extends DateConstructor
+    ? Date | InferPropType<U, false>
+    : InferPropType<U, false>
+  : [T] extends [Prop<infer V, infer D>]
+  ? unknown extends V
+    ? keyof V extends never
+      ? IfAny<V, V, D>
+      : V
+    : V
+  : T
+
+type PublicRequiredKeys<T> = {
+  [K in keyof T]: T[K] extends { required: true } ? K : never
+}[keyof T]
+
+type PublicOptionalKeys<T> = Exclude<keyof T, PublicRequiredKeys<T>>
+
+declare module 'vue' {
+  // compatible with higher versions of Vue
+  /**
+   * Extract prop types from a runtime props options object.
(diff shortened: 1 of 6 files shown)


btea:
Can you describe the problem scenario more specifically? I don't understand why this modification is required.

yuhengshen:
Before:
The original type definition works correctly:
image
However, the output from @microsoft/api-extractor shows a type error:
image
This discrepancy occurs because TypeScript's declare module syntax requires scanning the file first before resolving extended types.

Now:
image


@microsoft/api-extractor loses references to declare module. so the type reference was manually added. It will be removed after upgrading Vue in the future.

btea (approved):
Thanks!

warmthsea:
**/play** still has type problems.

image

yuhengshen:
Global type declarations might need manual import of the utils type; I'll check.

yuhengshen:
@warmthsea I tested it with an empty project using this version and it worked. Have you reloaded Volar?

image

warmthsea:
> @warmthsea I tested it with an empty project using this version and it worked. Have you reloaded Volar?

Yes, new project only import ` "element-plus/global"`, but it did affect the [element-plus/play](https://github.com/element-plus/element-plus/tree/dev/play)

yuhengshen:
> > @warmthsea I tested it with an empty project using this version and it worked. Have you reloaded Volar?
> 
> Yes, new project only import ` "element-plus/global"`, but it did affect the [element-plus/play](https://github.com/element-plus/element-plus/tree/dev/play)

fixed

btea:
It looks like `pnpm typecheck` doesn't pass.

yuhengshen:
The current code includes many temporary fixes. we need upgrade to vue 3.3

packages/utils/vue3.3.polyfill.ts
```diff
@@ -0,0 +1,51 @@
...
+    : V
+  : T
+
+type PublicRequiredKeys<T> = {
+  [K in keyof T]: T[K] extends { required: true } ? K : never
+}[keyof T]
+
+type PublicOptionalKeys<T> = Exclude<keyof T, PublicRequiredKeys<T>>
+
+declare module 'vue' {
```
btea, line 38: Perhaps a safer approach would be to remove the type declaration here and import the type directly using the file path, without having to modify `formattedText`.

yuhengshen: But this approach is not ideal for future upgrades. Currently, upgrading only requires replacing all __ExtractPublicPropTypes and changing some files. Using type imports would necessitate manually modifying numerous files.

btea: I don't think the difference is big. If upgrade in the future, you can just re-export the type here.
`export type { ExtractPublicPropTypes as __ExtractPublicPropTypes } from 'vue'`

I think the advantage of this change is that it is very clear, it will not have the problems mentioned above, and there is no need to change other configuration files.

btea: WDYT？ cc  @warmthsea @Dsaquel

yuhengshen: This will result in two import sources for the same type for a long time.

btea: Yes, this is not a good idea either.

btea:
@yuhengshen It seems to be ok for now.

yuhengshen:
> @yuhengshen It seems to be ok for now.

Yes, I tested it with the following methods

- Using the global type.
- import { xxx } from 'element-plus'.
- import { xxx } from 'element-plus/es/xxx'.

btea:
Ok, let's merge it.

warmthsea:
> > > @warmthsea I tested it with an empty project using this version and it worked. Have you reloaded Volar?
> > 
> > 
> > Yes, new project only import ` "element-plus/global"`, but it did affect the [element-plus/play](https://github.com/element-plus/element-plus/tree/dev/play)
> 
> fixed

Sorry for not replying to you in time.  
Now it seems that it does affect the color parsing of **volar**, and the ide color of the component is lost. I can't think of a good solution.

**/packages**

image

**/packages**

image

**/play**

image

yuhengshen:
@warmthsea 
Sorry for reproducing this error again.

When I added the type field in tsconfig.web.json, Volar started working, but `typecheck` reported type errors.

After I fixed the type error, Volar stopped working again.

I suspect this might be a Volar bug. It seems to load the declare module definitions multiple times, which leads to a TypeScript error: Duplicate identifier '__ExtractPublicPropTypes'.

warmthsea:
I will submit a PR later and have a solution.

PR #21292, opened 2025-07-09
fix(components): [menu] hide arrow right padding incorrect

### LINKED ISSUE #21198: [Style] [menu] Menu 组件在 Container 布局中作为横向菜单栏显示收缩按钮时宽度超出
Bug Type: **`Style`**

## Environment

- Vue Version: `3.5.17`
- Element Plus Version: `2.10.2`
- Browser / OS: `UserAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/121.0.0.0 Safari/537.36 Edg/121.0.0.0`
- Build Tool: `Vite`

### Related Component

- `el-menu`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8RWxDb250YWluZXI+XG4gICAgPEVsQXNpZGU+XG4gICAgICBBc2lkZVxuICAgIDwvRWxBc2lkZT5cblxuICAgIDxFbENvbnRhaW5lcj5cbiAgICAgIDxFbEhlYWRlciBjbGFzcz1cImhlYWRlclwiPlxuICAgICAgICA8ZWwtbWVudSBjbGFzcz1cIm1lbnViYXJcIiBtb2RlPVwiaG9yaXpvbnRhbFwiIGRlZmF1bHQtYWN0aXZlPVwiNWFkXCI+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjFcIj5Qcm9jZXNzaW5nIENlbnRlcjwvZWwtbWVudS1pdGVtPlxuICAgICAgICAgIDxlbC1tZW51LWl0ZW0gaW5kZXg9XCI0XCI+SW5mbzwvZWwtbWVudS1pdGVtPlxuICAgICAgICAgIDxlbC1tZW51LWl0ZW0gaW5kZXg9XCI1MTJcIj5PcmRlcnM8L2VsLW1lbnUtaXRlbT5cbiAgICAgICAgICA8ZWwtbWVudS1pdGVtIGluZGV4PVwiNTIxM1wiPk9yZGVyczwvZWwtbWVudS1pdGVtPlxuICAgICAgICAgIDxlbC1tZW51LWl0ZW0gaW5kZXg9XCI1cXdlXCI+T3JkZXJzPC9lbC1tZW51LWl0ZW0+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjMtMVwiPml0ZW0gb25lPC9lbC1tZW51LWl0ZW0+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjMtMlwiPml0ZW0gdHdvPC9lbC1tZW51LWl0ZW0+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjMtM1wiPml0ZW0gdGhyZWU8L2VsLW1lbnUtaXRlbT5cbiAgICAgICAgICA8ZWwtbWVudS1pdGVtIGluZGV4PVwiNWFkXCI+T3JkZXJzPC9lbC1tZW51LWl0ZW0+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjItMVwiPml0ZW0gb25lPC9lbC1tZW51LWl0ZW0+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjItMlwiPml0ZW0gdHdvPC9lbC1tZW51LWl0ZW0+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjItM1wiPml0ZW0gdGhyZWU8L2VsLW1lbnUtaXRlbT5cbiAgICAgICAgPC9lbC1tZW51PlxuXG4gICAgICAgIDxFbFNwYWNlIGNsYXNzPVwidG9vbGJhclwiPlxuICAgICAgICAgIDxFbElucHV0IC8+XG4gICAgICAgICAgPEVsQnV0dG9uPuaMiemSrjE8L0VsQnV0dG9uPlxuICAgICAgICAgIDxFbEJ1dHRvbj7mjInpkq4xPC9FbEJ1dHRvbj5cbiAgICAgICAgPC9FbFNwYWNlPlxuICAgICAgPC9FbEhlYWRlcj5cbiAgICA8L0VsQ29udGFpbmVyPlxuICA8L0VsQ29udGFpbmVyPlxuPC90ZW1wbGF0ZT5cblxuPHN0eWxlPlxuLmhlYWRlciB7XG4gIGRpc3BsYXk6IGZsZXg7XG4gIGp1c3RpZnktY29udGVudDogc3BhY2UtYmV0d2VlbjtcbiAgYWxpZ24taXRlbXM6IGNlbnRlcjtcbiAgZ2FwOiAxMHB4O1xuICBib3JkZXItYm90dG9tOiAxcHggc29saWQgdmFyKC0tZWwtbWVudS1ib3JkZXItY29sb3IpO1xufVxuXG4ubWVudWJhciB7XG4gIGZsZXg6IGF1dG87XG4gIG1pbi13aWR0aDogMDtcbiAgYm9yZGVyLWJvdHRvbTogbm9uZTtcbn1cblxuLnRvb2xiYXIge1xuICBmbGV4LXNocmluazogMDtcbn1cblxuLm1lbnViYXIgLmVsLXN1Yi1tZW51X19oaWRlLWFycm93ID4gLmVsLXN1Yi1tZW51X190aXRsZSB7XG4gIC8qIHBhZGRpbmctcmlnaHQ6IHZhcigtLWVsLW1lbnUtYmFzZS1sZXZlbC1wYWRkaW5nKSAhaW1wb3J0YW50OyAqL1xufVxuPC9zdHlsZT5cbiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vZmFzdGx5LmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQGxhdGVzdC9kaXN0L2luZGV4LmNzcycsICdodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvdGhlbWUtY2hhbGsvZGFyay9jc3MtdmFycy5jc3MnXS5tYXAoKHN0eWxlKSA9PiB7XG4gICAgcmV0dXJuIG5ldyBQcm9taXNlKChyZXNvbHZlLCByZWplY3QpID0+IHtcbiAgICAgIGNvbnN0IGxpbmsgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCdsaW5rJylcbiAgICAgIGxpbmsucmVsID0gJ3N0eWxlc2hlZXQnXG4gICAgICBsaW5rLmhyZWYgPSBzdHlsZVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdsb2FkJywgcmVzb2x2ZSlcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignZXJyb3InLCByZWplY3QpXG4gICAgICBkb2N1bWVudC5ib2R5LmFwcGVuZChsaW5rKVxuICAgIH0pXG4gIH0pXG4gIHJldHVybiBQcm9taXNlLmFsbFNldHRsZWQoc3R5bGVzKVxufVxuIiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IFwiRVNOZXh0XCIsXG4gICAgXCJqc3hcIjogXCJwcmVzZXJ2ZVwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwidHlwZXNcIjogW1wiZWxlbWVudC1wbHVzL2dsb2JhbC5kLnRzXCJdLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcImFsbG93SnNcIjogdHJ1ZSxcbiAgICBcImNoZWNrSnNcIjogdHJ1ZVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vZmFzdGx5LmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9ydW50aW1lLWRvbUBsYXRlc3QvZGlzdC9ydW50aW1lLWRvbS5lc20tYnJvd3Nlci5qc1wiLFxuICAgIFwiQHZ1ZS9zaGFyZWRcIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvc2hhcmVkQGxhdGVzdC9kaXN0L3NoYXJlZC5lc20tYnVuZGxlci5qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9mYXN0bHkuanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwiaHR0cHM6Ly9mYXN0bHkuanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L1wiLFxuICAgIFwiQGVsZW1lbnQtcGx1cy9pY29ucy12dWVcIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlQDIvZGlzdC9pbmRleC5taW4uanNcIlxuICB9LFxuICBcInNjb3Blc1wiOiB7fVxufSIsIl9vIjp7fX0=)

### Steps to reproduce

打开链接就可以发现，如果把下面注释的样式放开就会正常。

### What is Expected?

期望收缩按钮没有超出宽度。

目前收缩按钮出现后会出现超出宽度的行为。

### Additional comments

_(empty)_

meet-student:

Type

- fix #21198


meet-student:
@Dsaquel @warmthsea  @btea  help review

packages/theme-chalk/src/menu.scss
```diff
@@ -302,6 +302,11 @@
     .#{$namespace}-sub-menu__icon-arrow {
       display: none !important;
     }
+    .#{$namespace}-sub-menu__title {
+      padding-right: calc(
+        #{getCssVar('menu-base-level-padding')}
+      ) !important;
+    }
```
Dsaquel: LGTM, according to https://github.com/element-plus/element-plus/commit/10fb2233f82fd9e1f7799cb729c709e146eae5c2 you can remove `!important` and set this style below line 285.

Meet-student: this ? done

meet-student: this ? done  @Dsaquel  see?

meet-student:
before
image

after
image

btea (approved):

packages/theme-chalk/src/menu.scss
```diff
@@ -298,6 +298,10 @@
     line-height: getCssVar('menu-sub-item-height');
   }
 
+  &.#{$namespace}-sub-menu__hide-arrow .#{$namespace}-sub-menu__title{
+    padding-right: calc(#{getCssVar('menu-base-level-padding')});
```
Dsaquel: **nit**:

```suggestion
  &.#{$namespace}-sub-menu__hide-arrow .#{$namespace}-sub-menu__title {
    padding-right: getCssVar('menu-base-level-padding');
  }
```

packages/theme-chalk/src/menu.scss
Dsaquel: I was not expecting this solution, but it's even better :+1:

packages/theme-chalk/src/menu.scss
meet-student: nice, done @Dsaquel

l246804:
I have a concern regarding the proposed changes. While modifying the CSS styles can address the default width issue, the source code currently hardcodes the width calculation for the collapse button. Wouldn't this still cause inconsistencies when users modify the `--el-menu-base-level-padding` variable?

Have we considered making this calculation dynamic by retrieving the actual button width from the DOM during runtime? This approach might provide a more robust solution that adapts to different padding configurations.

btea:
@l246804 Regarding this point, the component library will only implement the initial functions according to the design to ensure that there will be no problems during normal use.

If the user modifies the variables used during actual use and causes other problems, the developer can only make adjustments based on their actual scenarios. We cannot be compatible with all situations.

l246804:
Thank you for the clarification! I truly appreciate the consideration for maintaining library stability.

That said, since `--el-menu-base-level-padding` is designed as a public CSS variable for theme customization, users following this official approach might still encounter layout issues with the hardcoded collapse button width.

However, I noticed in the source code that the menu items' width is already dynamically calculated. Since the collapse button is part of the same navigation menu component, would it be feasible to apply the same dynamic measurement approach here?

image

btea:
What you are talking about is different from the previous situation. The above variables are only used to configure the padding value, not the fixed width.

The code in the figure above is to calculate the specific item width based on js to determine whether to display hidden items. When js calculates the width, it needs to calculate the actual item width, which includes the set padding.

We cannot set padding based on the width calculated by js, which is putting the cart before the horse.

l246804:
The core issue here is that the `moreItemWidth` value in the JS code is hardcoded as `64`, which fails to align with the CSS implementation due to padding-related discrepancies. Could we instead dynamically calculate moreItemWidth within the JS logic? This would improve error tolerance for users, especially considering that padding values defined via CSS variables are often customized by end-users.

```ts
const moreItemWidth = moreItemRef.value.getBoundingClientRect().width
```

btea:
I seem to understand what you mean. You mean that 64 here should not be a fixed value. If the variable `--el-menu-base-level-padding` changes, we should dynamically get the value of the variable `moreItemWidth`.

But there is still a problem. If we want to get the width of moreItem, we must first make sure it has been rendered. But our current logic is to see if there is enough space left after subtracting the rendered items from the container width. If not, moreItem will be rendered. This seems to be another paradox. 🤔

l246804:
Perhaps we could create a hidden placeholder element with the `el-menu-more-item` (or the actual class used for the collapse button) class and use its computed width for calculations?

This would ensure the measured `moreItemWidth` reflects the actual styles (including padding variables) without requiring the component to be visibly rendered.

It might resolve the paradox by decoupling the width measurement from the rendering logic itself.

btea:
Thank you for the solution. I have considered this before, but it is a bit strange to insert an omitted tag into the project when it is not necessary, which may also cause some confusion for developers.

For example, when running test cases, errors may occur when distinguishing between omitted items and non-omitted items.

l246804:
Understood! To avoid polluting the DOM or interfering with test cases, what if we create a detached DOM element purely for measurement? For example:

```ts
const tempElement = document.createElement('div');
tempElement.className = 'el-menu-more-item'; // Use the actual class
tempElement.style.position = 'absolute';
tempElement.style.left = '-9999px';
tempElement.style.height = '0';
tempElement.style.overflow = 'hidden';
tempElement.style.pointerEvents = 'none';
document.body.appendChild(tempElement);

// Calculate the actual element width
const moreItemWidth = tempElement.offsetWidth;

// When the width calculation is complete, remove the temporary element
document.body.removeChild(tempElement);
```

meet-student:
@l246804 @btea  result ?  or I try next pr ?

Dsaquel:
After reading the discussion i agree that it is reasonable to change the static number `64` as it means to include the icon width + the double padding left and right.

Maybe [here](https://github.com/element-plus/element-plus/blob/dev/packages/components/menu/src/menu.ts#L493) we can bind a ref and get the width dynamically.

meet-student:
@btea @l246804 @Dsaquel ref.value 。 have rendering calculation issues, help review

packages/components/menu/src/menu.ts
```diff
@@ -323,12 +323,57 @@ export default defineComponent({
       return menuItem.offsetWidth + marginLeft + marginRight || 0
     }
 
+    const calcMoreButtonWidth = () => {
```
Dsaquel: @Meet-student Looks like the current change is not a vue-idiomatic solution.

I'll checkout later this pr to see why `ref` is not working.

meet-student: There are issues with timing and rendering conditions .
I can try again later, I don't have time at the beginning .
@Dsaquel

Dsaquel: No problem, we can take our time, there are no reason to rush :+1: .

meet-student: @Dsaquel 
I tried, but this DOM appears after assigning width and cannot be obtained through ref. Before being assigned a value, the menu is fully rendered, so this is how it needs to be handled

The action of obtaining width before rendering more， I think this is similar to write 64

meet-student:
@Dsaquel  help review

Dsaquel:
@Meet-student I should be able to take a look this week.

l246804:
Maybe this would work better:

```html,diff
<script setup lang="ts">
const hiddenMoreButtonRef = ref<HTMLDivElement>()
function getMoreButtonWidth() {
  return hiddenMoreButtonRef.value?.offsetWidth || 64
}
</script>

<template>
  <Teleport to="body">
    <div
      v-if="mode === 'horizontal'"
      ref="hiddenMoreButtonRef"
      :class="nsSubMenu.e('icon-more')"
      style="position: absolute; left: -9999px; height: 0; overflow: hidden; pointer-events: none;"
    />
  </Teleport>
</template>
```

meet-student:
> Maybe this would work better:
> 
> ```
> <script setup lang="ts">
> const hiddenMoreButtonRef = ref<HTMLDivElement>()
> function getMoreButtonWidth() {
>   return hiddenMoreButtonRef.value?.offsetWidth || 64
> }
> </script>
> 
> <template>
>   <Teleport to="body">
>     <div
>       v-if="mode === 'horizontal'"
>       ref="hiddenMoreButtonRef"
>       :class="nsSubMenu.e('icon-more')"
>       style="position: absolute; left: -9999px; height: 0; overflow: hidden; pointer-events: none;"
>     />
>   </Teleport>
> </template>
> ```

Need to remove

l246804:
> > Maybe this would work better:
> > ```
> > <script setup lang="ts">
> > const hiddenMoreButtonRef = ref<HTMLDivElement>()
> > function getMoreButtonWidth() {
> >   return hiddenMoreButtonRef.value?.offsetWidth || 64
> > }
> > </script>
> > 
> > <template>
> >   <Teleport to="body">
> >     <div
> >       v-if="mode === 'horizontal'"
> >       ref="hiddenMoreButtonRef"
> >       :class="nsSubMenu.e('icon-more')"
> >       style="position: absolute; left: -9999px; height: 0; overflow: hidden; pointer-events: none;"
> >     />
> >   </Teleport>
> > </template>
> > ```
> 
> Need to remove

`<Teleport>` should automatically remove the content elements during unmounting.

packages/components/menu/src/menu.ts
```diff
@@ -360,6 +361,8 @@ export default defineComponent({
 
     let isFirstTimeRender = true
     const handleResize = () => {
+      const el = unrefElement(subMenu)
+      if (el) moreItemWidth = calcMenuItemWidth(el) || 64
```
Dsaquel, line 365: Handling the width calculation before `nextTick` l.370 because the `subMenu` will be strangely unavailable.

Dsaquel:
You can try this [demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8RWxDb250YWluZXI+XG4gICAgPEVsQXNpZGU+XG4gICAgICBBc2lkZVxuICAgIDwvRWxBc2lkZT5cblxuICAgIDxFbENvbnRhaW5lcj5cbiAgICAgIDxFbEhlYWRlciBjbGFzcz1cImhlYWRlclwiPlxuICAgICAgICA8ZWwtbWVudSBjbGFzcz1cIm1lbnViYXJcIiBtb2RlPVwiaG9yaXpvbnRhbFwiIGRlZmF1bHQtYWN0aXZlPVwiNWFkXCI+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjFcIj5Qcm9jZXNzaW5nIENlbnRlcjwvZWwtbWVudS1pdGVtPlxuICAgICAgICAgIDxlbC1tZW51LWl0ZW0gaW5kZXg9XCI0XCI+SW5mbzwvZWwtbWVudS1pdGVtPlxuICAgICAgICAgIDxlbC1tZW51LWl0ZW0gaW5kZXg9XCI1MTJcIj5PcmRlcnM8L2VsLW1lbnUtaXRlbT5cbiAgICAgICAgICA8ZWwtbWVudS1pdGVtIGluZGV4PVwiNTIxM1wiPk9yZGVyczwvZWwtbWVudS1pdGVtPlxuICAgICAgICAgIDxlbC1tZW51LWl0ZW0gaW5kZXg9XCI1cXdlXCI+T3JkZXJzPC9lbC1tZW51LWl0ZW0+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjMtMVwiPml0ZW0gb25lPC9lbC1tZW51LWl0ZW0+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjMtMlwiPml0ZW0gdHdvPC9lbC1tZW51LWl0ZW0+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjMtM1wiPml0ZW0gdGhyZWU8L2VsLW1lbnUtaXRlbT5cbiAgICAgICAgICA8ZWwtbWVudS1pdGVtIGluZGV4PVwiNWFkXCI+T3JkZXJzPC9lbC1tZW51LWl0ZW0+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjItMVwiPml0ZW0gb25lPC9lbC1tZW51LWl0ZW0+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjItMlwiPml0ZW0gdHdvPC9lbC1tZW51LWl0ZW0+XG4gICAgICAgICAgPGVsLW1lbnUtaXRlbSBpbmRleD1cIjItM1wiPml0ZW0gdGhyZWU8L2VsLW1lbnUtaXRlbT5cbiAgICAgICAgPC9lbC1tZW51PlxuXG4gICAgICAgIDxFbFNwYWNlIGNsYXNzPVwidG9vbGJhclwiPlxuICAgICAgICAgIDxFbElucHV0IC8+XG4gICAgICAgICAgPEVsQnV0dG9uPuaMiemSrjE8L0VsQnV0dG9uPlxuICAgICAgICAgIDxFbEJ1dHRvbj7mjInpkq4xPC9FbEJ1dHRvbj5cbiAgICAgICAgPC9FbFNwYWNlPlxuICAgICAgPC9FbEhlYWRlcj5cbiAgICA8L0VsQ29udGFpbmVyPlxuICA8L0VsQ29udGFpbmVyPlxuPC90ZW1wbGF0ZT5cblxuPHN0eWxlPlxuQGltcG9ydCAnLi9Db21wLmNzcyc7XG4uaGVhZGVyIHtcbiAgZGlzcGxheTogZmxleDtcbiAganVzdGlmeS1jb250ZW50OiBzcGFjZS1iZXR3ZWVuO1xuICBhbGlnbi1pdGVtczogY2VudGVyO1xuICBnYXA6IDEwcHg7XG4gIGJvcmRlci1ib3R0b206IDFweCBzb2xpZCB2YXIoLS1lbC1tZW51LWJvcmRlci1jb2xvcik7XG59XG5cbi5tZW51YmFyIHtcbiAgZmxleDogYXV0bztcbiAgbWluLXdpZHRoOiAwO1xuICBib3JkZXItYm90dG9tOiBub25lO1xufVxuXG4udG9vbGJhciB7XG4gIGZsZXgtc2hyaW5rOiAwO1xufVxuPC9zdHlsZT5cbiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vcHJldmlldy0yMTI5Mi1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzJywgJ2h0dHBzOi8vcHJldmlldy0yMTI5Mi1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL3RoZW1lLWNoYWxrL2RhcmsvY3NzLXZhcnMuY3NzJ10ubWFwKChzdHlsZSkgPT4ge1xuICAgIHJldHVybiBuZXcgUHJvbWlzZSgocmVzb2x2ZSwgcmVqZWN0KSA9PiB7XG4gICAgICBjb25zdCBsaW5rID0gZG9jdW1lbnQuY3JlYXRlRWxlbWVudCgnbGluaycpXG4gICAgICBsaW5rLnJlbCA9ICdzdHlsZXNoZWV0J1xuICAgICAgbGluay5ocmVmID0gc3R5bGVcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignbG9hZCcsIHJlc29sdmUpXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2Vycm9yJywgcmVqZWN0KVxuICAgICAgZG9jdW1lbnQuYm9keS5hcHBlbmQobGluaylcbiAgICB9KVxuICB9KVxuICByZXR1cm4gUHJvbWlzZS5hbGxTZXR0bGVkKHN0eWxlcylcbn1cbiIsInRzY29uZmlnLmpzb24iOiJ7XG4gIFwiY29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiBcIkVTTmV4dFwiLFxuICAgIFwianN4XCI6IFwicHJlc2VydmVcIixcbiAgICBcIm1vZHVsZVwiOiBcIkVTTmV4dFwiLFxuICAgIFwibW9kdWxlUmVzb2x1dGlvblwiOiBcIkJ1bmRsZXJcIixcbiAgICBcInR5cGVzXCI6IFtcImVsZW1lbnQtcGx1cy9nbG9iYWwuZC50c1wiXSxcbiAgICBcImFsbG93SW1wb3J0aW5nVHNFeHRlbnNpb25zXCI6IHRydWUsXG4gICAgXCJhbGxvd0pzXCI6IHRydWUsXG4gICAgXCJjaGVja0pzXCI6IHRydWVcbiAgfSxcbiAgXCJ2dWVDb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IDMuM1xuICB9XG59XG4iLCJQbGF5Z3JvdW5kTWFpbi52dWUiOiI8c2NyaXB0IHNldHVwPlxuaW1wb3J0IEFwcCBmcm9tICcuL0FwcC52dWUnXG5pbXBvcnQgeyBzZXR1cEVsZW1lbnRQbHVzIH0gZnJvbSAnLi9lbGVtZW50LXBsdXMuanMnXG5zZXR1cEVsZW1lbnRQbHVzKClcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxBcHAgLz5cbjwvdGVtcGxhdGU+XG4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvcnVudGltZS1kb21AbGF0ZXN0L2Rpc3QvcnVudGltZS1kb20uZXNtLWJyb3dzZXIuanNcIixcbiAgICBcIkB2dWUvc2hhcmVkXCI6IFwiaHR0cHM6Ly9mYXN0bHkuanNkZWxpdnIubmV0L25wbS9AdnVlL3NoYXJlZEBsYXRlc3QvZGlzdC9zaGFyZWQuZXNtLWJ1bmRsZXIuanNcIixcbiAgICBcImVsZW1lbnQtcGx1c1wiOiBcImh0dHBzOi8vcHJldmlldy0yMTI5Mi1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwidW5zdXBwb3J0ZWRcIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly9mYXN0bHkuanNkZWxpdnIubmV0L25wbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJDb21wLmNzcyI6Ijpyb290IHtcbiAgLS1lbC1tZW51LWJhc2UtbGV2ZWwtcGFkZGluZzogNDBweDsgLyogYmVmb3JlIGl0IHdhcyAyMHB4ICovXG59IiwiX28iOnsic2hvd0hpZGRlbiI6dHJ1ZSwic3R5bGVTb3VyY2UiOiJodHRwczovL3ByZXZpZXctMjEyOTItZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS9kaXN0L2luZGV4LmNzcyJ9fQ==).

Set this variable manually (i don't know why in repl it doesn't take effect):
```scss
:root {
  --el-menu-base-level-padding: 40px; /* before it was 20px */
}
```
You'll see it does take the padding of the icon item dynamically.

packages/theme-chalk/src/menu.scss
```diff
@@ -298,6 +298,10 @@
     line-height: getCssVar('menu-sub-item-height');
   }
 
+  &.#{$namespace}-sub-menu__hide-arrow .#{$namespace}-sub-menu__title{
```
Dsaquel:
```suggestion
  &.#{$namespace}-sub-menu__hide-arrow .#{$namespace}-sub-menu__title {
```

meet-student:
I'll see your changes.

meet-student:
this ref element maybe always undefind ? @Dsaquel

Dsaquel:
@Meet-student As i said here https://github.com/element-plus/element-plus/pull/21292#discussion_r2217906869 it get undefined only in the `nextTick` call.

meet-student:
> @Meet-student As i said here [#21292 (comment)](https://github.com/element-plus/element-plus/pull/21292#discussion_r2217906869) it get undefined only in the `nextTick` call.

ok

meet-student:
ok ？ @l246804   review over？

l246804:
Thanks for the update! If you guys have no problem with the testing, then I don't have any issues either.

Dsaquel (approved):

btea (approved):
Thanks!

meet-student:
> Thanks for the update! If you guys have no problem with the testing, then I don't have any issues either.

done ! Thanks @l246804 @btea @Dsaquel

PR #21321, opened 2025-07-11
chore: fix dev-volar type resolution

warmthsea:
#### Rel:

- https://github.com/element-plus/element-plus/pull/21266#issuecomment-3061651220


yuhengshen:
Could you explain why this works?

warmthsea:
> Could you explain why this works?

This may be caused by Volar parsing multiple files vue scope conflicts.

about: https://cn.vuejs.org/guide/extras/web-components.html#web-components-and-typescript

btea (approved):

Issue #17997, opened 2024-08-26
[Component] [date-picker] 怎么让日期选择器面板一直显示，能否加一个这样的功能


stupidream:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.4.38`
- Element Plus Version: `2.8.1`
- Browser / OS: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36`
- Build Tool: `CDN`

### Related Component

- `el-date-picker`

### Reproduction Link

[CodePen](https://codepen.io/zhaoyh/pen/qBzKdLG)

### Steps to reproduce

直接行内显示日期面板，不会关闭

### What is Expected?

直接行内显示日期面板，像Calendar组件一样

### What is actually happening?

选择完日期后面板就关闭了

### Additional comments

_(empty)_

![微信截图_20240826095736](https://github.com/user-attachments/assets/655da93e-afaa-499b-962a-d25676be4882)

btea:
It looks like the demand is related to this PR #17168.

We have consulted with the designers before and they think it does not meet the design specifications. Perhaps we can use this opportunity to consult with them for more opinions.

tolking:
Perhaps we can consider separating the panel into a separate component like ElCascader, ElCascaderPanel.

Or export the internal components for users to implement on their own.

btea:
> Perhaps we can consider separating the panel into a separate component like ElCascader, ElCascaderPanel.
> 
> Or export the internal components for users to implement on their own.

Splitting the panels into separate components looks great. 👍

zhixiaotong:
what about color-picker？

qinfeng0214:
> > 那么颜色选择器呢？
> 
> 也许我们可以分离一个`color-panel`组件并重复使用它。

非常需要这个组件，有些场景只需要一个el-date-picker，不需要那个输入框，如果能分离出来就太棒啦，非常感谢你们的付出

a81n9:
确实有很多场景都需要这个功能，期待中...

zhw2590582:
还有下文吗？

a81n9:
Just like [PrimeVue](https://primevue.org/)
[DatePicker inline](https://primevue.org/datepicker/#inline)
[ColorPicker inline](https://primevue.org/colorpicker/#inline)

huanglong6828:
还有下文吗？

[...]

Issue #5269, opened 2022-01-10
[Bug Report] el-dialog uses el-date-picker internally ZIndex Error


MrFu1998:
### Element Plus version
1.3.0-beta.3

### OS/Browsers version
Mac OS/ Chrome 版本 97.0.4692.71（正式版本） (arm64)

### Vue version
3.2.26

### Reproduction Link
https://codesandbox.io/s/element-plus-spa-forked-wqbsw?file=/src/App.vue

### Steps to reproduce
1. open dialog
2. select date-picker
3. close dialog
4. reopen dialog
5. reselect date-picker

### What is Expected?
el-date-picker Hierarchy should be above / 日期选择层级应高于 dialog 层级

### What is actually happening?
Hierarchy should be below / 日期选择层级低于 dialog 层级

PR #21331, opened 2025-07-13
feat(components): [date-picker-panel] externalize `date-picker` core component

Dsaquel:
Requests: https://github.com/element-plus/element-plus/discussions/18011 https://github.com/element-plus/element-plus/discussions/12408 https://github.com/element-plus/element-plus/discussions/16376 https://github.com/element-plus/element-plus/discussions/18298 and maybe others.

---
I forwarded the tests of `date-time-picker` in `date-picker-panel` tests in addition with the specific test case of the component. I didn't forwarded the tests of `date-picker` because there are just too much tests (2000 lines), and i think there will be just small edge case that can be handle individually.


Dsaquel:
WDYT @btea @warmthsea .

Dsaquel:
@Moresl This is currently not possible. I'm currently working to set v-model:visible on this type of component. Where you'll can manually control the open/close of the popper.

btea:
> WDYT @btea @warmthsea .

Looks good, keep it up! 👍

btea:
Maybe we can add an el-date-panel component. refer to https://github.com/element-plus/element-plus/issues/17997

Dsaquel:
Ok, i'm going for it :+1:

zhixiaotong:
@Dsaquel  @btea Thank you all for your efforts on this. Perhaps this is a good opportunity to restart the `el-color-panel`? Both the `color-picker` and `date-picker` require independent display panels, so I hope the team can also push forward work in this area. Please see PR https://github.com/element-plus/element-plus/pull/17168.

Dsaquel:
@zhixiaotong Thanks for pointing this PR, i will in fact work on externalize the template of theses components based on tooltip during the next weeks.

packages/components/date-picker-panel/src/instance.ts
```diff
@@ -0,0 +1,5 @@
+import type { ComponentPublicInstance } from 'vue'
+import type { DatePickerPanelProps } from './props/date-picker-panel'
+
+export type DatePickerPanelInstance =
+  ComponentPublicInstance<DatePickerPanelProps>
```
btea: For component instance types, please refer to #20109.

btea (approved):
Overall it looks great! 👍

warmthsea:
I just saw this PR, hard work. 😃

packages/components/date-picker-panel/src/instance.ts
Dsaquel: done :+1:, There was the same in date-picker i have changed it too.

kooriookami:
image
I think the icon can be the same with date-picker.

kooriookami:
I think to use version 2.11.0 is better, cause it's a new component.

Dsaquel:
I agree, maybe then it should be the next version of 2.10.5 to avoid potentials many conflicts.

warmthsea:
There is a conflict that needs to be resolved.

packages/components/date-picker/style/css.ts
```diff
@@ -1,6 +1,4 @@
 import '@element-plus/components/base/style/css'
-import '@element-plus/theme-chalk/el-date-picker.css'
-import '@element-plus/components/input/style/css'
-import '@element-plus/components/scrollbar/style/css'
+import '@element-plus/theme-chalk/src/date-picker.scss'
```
btea:
```suggestion
import '@element-plus/theme-chalk/el-date-picker.css'
```

packages/components/date-picker-panel/src/date-picker-com/panel-date-range.vue
```diff
@@ -136,8 +141,11 @@
             <button
               v-if="unlinkPanels"
               type="button"
-              :disabled="!enableYearArrow"
-              :class="[ppNs.e('icon-btn'), { 'is-disabled': !enableYearArrow }]"
+              :disabled="!enableYearArrow || disabled"
+              :class="[
+                ppNs.e('icon-btn'),
+                ppNs.is('disabled', !enableYearArrow),
```
btea: Here, `is-disabled` is checked in the same way as above, adding `|| disabled`?

packages/components/date-picker-panel/src/date-picker-com/panel-date-range.vue
```diff
@@ -154,7 +162,7 @@
               :disabled="!enableMonthArrow"
```
btea: Same here.

packages/components/date-picker-panel/src/date-picker-com/panel-date-pick.vue
```diff
@@ -60,7 +63,7 @@
           :class="[
             dpNs.e('header'),
             (currentView === 'year' || currentView === 'month') &&
-              dpNs.e('header--bordered'),
+              dpNs.e('header--border'),
```
btea: Wouldn't it be better to leave it as it is?

warmthsea:
1. Should support `@change` event 🤔 .
2. Add another date-format example. https://element-plus.org/en-US/component/date-picker.html#date-formats

packages/components/date-picker-panel/src/date-picker-com/panel-date-pick.vue
Dsaquel: Yes, and i didn't make the scss follow either :no_mouth:

Dsaquel:
>1. Should support @change event 🤔 .

We can but i don't know if it's worth: The output will be the exact same as the update event and the implementation of the change emit flow is a bit complicated.

packages/components/date-picker-panel/src/date-picker-com/panel-date-pick.vue
```diff
@@ -16,6 +18,7 @@
           v-for="(shortcut, key) in shortcuts"
           :key="key"
           type="button"
+          :disabled="disabled"
```
tolking, line 21: The disable attribute should be applied to both of the following `el-input`

packages/components/date-picker-panel/src/date-picker-com/panel-date-pick.vue
```diff
@@ -263,7 +273,7 @@ const pickerBase = inject(PICKER_BASE_INJECTION_KEY) as any
 const isDefaultFormat = inject(
   ROOT_PICKER_IS_DEFAULT_FORMAT_INJECTION_KEY
 ) as any
-const popper = inject(TOOLTIP_INJECTION_KEY)
+const popper = inject(TOOLTIP_INJECTION_KEY, undefined)
 const { shortcuts, disabledDate, cellClassName, defaultTime } = pickerBase.props
```
tolking:
```diff
- const popper = inject(TOOLTIP_INJECTION_KEY, undefined)
```

IMO, The current component has nothing to do with ElTooltip, and the related logic should be moved to the ElDatePicker component

packages/components/date-picker-panel/src/date-picker-com/panel-date-range.vue
```diff
@@ -363,6 +381,7 @@
         {{ t('el.datepicker.clear') }}
       </el-button>
       <el-button
+        v-if="hasPopper"
```
tolking: IMO, It doesn't feel very friendly to introduce ElTooltip here for judgment. How about providing a slot to implement the relevant logic in the ElDatePick component?

packages/components/date-picker-panel/src/types.ts
```diff
@@ -1,6 +1,6 @@
 import type { Dayjs } from 'dayjs'
 
-export declare type IDatePickerType =
+export type DatePickerType =
```
tolking, line 3:
```suggestion
/** @deprecated use `DatePickerType` instead */
export type IDatePickerType = DatePickerType

export type DatePickerType =
```

For exported type name changes, we can prompt the user to change

packages/components/date-picker-panel/style/css.ts
```diff
@@ -0,0 +1,5 @@
+import '@element-plus/components/base/style/css'
+import '@element-plus/theme-chalk/src/el-date-picker-panel.css'
```
tolking:
```suggestion
import '@element-plus/theme-chalk/el-date-picker-panel.css'
```

packages/components/date-picker-panel/style/css.ts
```diff
@@ -0,0 +1,5 @@
...
+import '@element-plus/components/input/style/css'
+import '@element-plus/components/scrollbar/style/css'
```
tolking, line 4:
```suggestion
```

I found that ElScrollbar component is not used in the component, maybe it can be removed

packages/components/date-picker-panel/style/index.ts
```diff
@@ -0,0 +1,5 @@
+import '@element-plus/components/base/style'
+import '@element-plus/theme-chalk/src/date-picker-panel.scss'
+import '@element-plus/components/input/style'
+import '@element-plus/components/scrollbar/style'
```
tolking, line 4:
```suggestion
```

I found that ElScrollbar component is not used in the component, maybe it can be removed

packages/components/date-picker-panel/src/types.ts
Dsaquel: In fact `IDatePickerType` was never exported in the root path. [play](https://element-plus.run/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cCBsYW5nPVwidHNcIj5cbmltcG9ydCB0eXBlIHsgSURhdGVQaWNrZXJUeXBlIH0gZnJvbSAnZWxlbWVudC1wbHVzJ1xuXG4vLyB0aGlzIHdvcmtzXG4vL2ltcG9ydCB0eXBlIHsgSURhdGVQaWNrZXJUeXBlIH0gZnJvbSAnZWxlbWVudC1wbHVzL2VzL2NvbXBvbmVudHMvZGF0ZS1waWNrZXIvc3JjL2RhdGUtcGlja2VyLnR5cGUnXG48L3NjcmlwdD5cblxuIiwiZWxlbWVudC1wbHVzLmpzIjoiaW1wb3J0IEVsZW1lbnRQbHVzIGZyb20gJ2VsZW1lbnQtcGx1cydcbmltcG9ydCB7IGdldEN1cnJlbnRJbnN0YW5jZSB9IGZyb20gJ3Z1ZSdcblxubGV0IGluc3RhbGxlZCA9IGZhbHNlXG5hd2FpdCBsb2FkU3R5bGUoKVxuXG5leHBvcnQgZnVuY3Rpb24gc2V0dXBFbGVtZW50UGx1cygpIHtcbiAgaWYgKGluc3RhbGxlZCkgcmV0dXJuXG4gIGNvbnN0IGluc3RhbmNlID0gZ2V0Q3VycmVudEluc3RhbmNlKClcbiAgaW5zdGFuY2UuYXBwQ29udGV4dC5hcHAudXNlKEVsZW1lbnRQbHVzKVxuICBpbnN0YWxsZWQgPSB0cnVlXG59XG5cbmV4cG9ydCBmdW5jdGlvbiBsb2FkU3R5bGUoKSB7XG4gIGNvbnN0IHN0eWxlcyA9IFsnaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L2Rpc3QvaW5kZXguY3NzJywgJ2h0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQGxhdGVzdC90aGVtZS1jaGFsay9kYXJrL2Nzcy12YXJzLmNzcyddLm1hcCgoc3R5bGUpID0+IHtcbiAgICByZXR1cm4gbmV3IFByb21pc2UoKHJlc29sdmUsIHJlamVjdCkgPT4ge1xuICAgICAgY29uc3QgbGluayA9IGRvY3VtZW50LmNyZWF0ZUVsZW1lbnQoJ2xpbmsnKVxuICAgICAgbGluay5yZWwgPSAnc3R5bGVzaGVldCdcbiAgICAgIGxpbmsuaHJlZiA9IHN0eWxlXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2xvYWQnLCByZXNvbHZlKVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdlcnJvcicsIHJlamVjdClcbiAgICAgIGRvY3VtZW50LmJvZHkuYXBwZW5kKGxpbmspXG4gICAgfSlcbiAgfSlcbiAgcmV0dXJuIFByb21pc2UuYWxsU2V0dGxlZChzdHlsZXMpXG59XG4iLCJ0c2NvbmZpZy5qc29uIjoie1xuICBcImNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogXCJFU05leHRcIixcbiAgICBcImpzeFwiOiBcInByZXNlcnZlXCIsXG4gICAgXCJtb2R1bGVcIjogXCJFU05leHRcIixcbiAgICBcIm1vZHVsZVJlc29sdXRpb25cIjogXCJCdW5kbGVyXCIsXG4gICAgXCJ0eXBlc1wiOiBbXCJlbGVtZW50LXBsdXMvZ2xvYmFsLmQudHNcIl0sXG4gICAgXCJhbGxvd0ltcG9ydGluZ1RzRXh0ZW5zaW9uc1wiOiB0cnVlLFxuICAgIFwiYWxsb3dKc1wiOiB0cnVlLFxuICAgIFwiY2hlY2tKc1wiOiB0cnVlXG4gIH0sXG4gIFwidnVlQ29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiAzLjNcbiAgfVxufVxuIiwiUGxheWdyb3VuZE1haW4udnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCBBcHAgZnJvbSAnLi9BcHAudnVlJ1xuaW1wb3J0IHsgc2V0dXBFbGVtZW50UGx1cyB9IGZyb20gJy4vZWxlbWVudC1wbHVzLmpzJ1xuc2V0dXBFbGVtZW50UGx1cygpXG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8QXBwIC8+XG48L3RlbXBsYXRlPlxuIiwiaW1wb3J0LW1hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge1xuICAgIFwidnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AdnVlL3J1bnRpbWUtZG9tQGxhdGVzdC9kaXN0L3J1bnRpbWUtZG9tLmVzbS1icm93c2VyLmpzXCIsXG4gICAgXCJAdnVlL3NoYXJlZFwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9zaGFyZWRAbGF0ZXN0L2Rpc3Qvc2hhcmVkLmVzbS1idW5kbGVyLmpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXNcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvZGlzdC9pbmRleC5mdWxsLm1pbi5tanNcIixcbiAgICBcImVsZW1lbnQtcGx1cy9cIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvXCIsXG4gICAgXCJAZWxlbWVudC1wbHVzL2ljb25zLXZ1ZVwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQGVsZW1lbnQtcGx1cy9pY29ucy12dWVAMi9kaXN0L2luZGV4Lm1pbi5qc1wiXG4gIH0sXG4gIFwic2NvcGVzXCI6IHt9XG59IiwiX28iOnt9fQ==)

packages/components/date-picker-panel/style/css.ts
Dsaquel: I tried before but it appear the we have to inject the scrollbar style for time pickers.
image
Perhaps we should instead inject ElTimePicker style instead of ElScrollbar style but the previous ElDatePicker didn't inject it what do you think ?

packages/components/date-picker-panel/src/date-picker-com/panel-date-range.vue
Dsaquel: I agree, i will try externalize this tonight.

packages/components/date-picker-panel/style/css.ts
tolking: Let's not change the code here.

packages/components/date-picker-panel/src/date-picker-com/panel-date-pick.vue
Dsaquel: Done in https://github.com/element-plus/element-plus/pull/21331/commits/85318bef2c095f850a675ab5631944439914b158, investigated and it seems like the watch is not necessary anymore (base [issue](https://github.com/element-plus/element-plus/issues/5269)), [demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZGl2PlxuICAgIDxlbC1idXR0b24gQGNsaWNrPVwib3BlbkRpYWxvZ1wiPm9wZW4gZGlhbG9nPC9lbC1idXR0b24+XG5cbiAgICA8ZWwtZGlhbG9nIHYtbW9kZWw9XCJkaWFsb2dWaXNpYmxlXCIgdGl0bGU9XCJkaWFsb2cgdGl0bGVcIj5cbiAgICAgIDxkaXYgY2xhc3M9XCJibG9ja1wiPlxuICAgICAgICA8c3BhbiBjbGFzcz1cImRlbW9uc3RyYXRpb25cIj5EZWZhdWx0PC9zcGFuPlxuICAgICAgICA8ZWwtZGF0ZS1waWNrZXJcbiAgICAgICAgICB2LW1vZGVsPVwidmFsdWUxXCJcbiAgICAgICAgICB0eXBlPVwiZGF0ZXJhbmdlXCJcbiAgICAgICAgICByYW5nZS1zZXBhcmF0b3I9XCJUb1wiXG4gICAgICAgICAgc3RhcnQtcGxhY2Vob2xkZXI9XCJTdGFydCBkYXRlXCJcbiAgICAgICAgICBlbmQtcGxhY2Vob2xkZXI9XCJFbmQgZGF0ZVwiXG4gICAgICAgID5cbiAgICAgICAgPC9lbC1kYXRlLXBpY2tlcj5cbiAgICAgIDwvZGl2PlxuICAgIDwvZWwtZGlhbG9nPlxuICA8L2Rpdj5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5pbXBvcnQgeyByZWYgfSBmcm9tICd2dWUnXG5cbmNvbnN0IGRpYWxvZ1Zpc2libGUgPSByZWYoZmFsc2UpO1xuZnVuY3Rpb24gb3BlbkRpYWxvZygpIHtcbiAgZGlhbG9nVmlzaWJsZS52YWx1ZSA9IHRydWU7XG59XG5cbmNvbnN0IHZhbHVlMSA9IHJlZihcIlwiKTtcblxuPC9zY3JpcHQ+XG5cbjxzdHlsZT5cbi5lbC1tZW51LXZlcnRpY2FsLWRlbW86bm90KC5lbC1tZW51LS1jb2xsYXBzZSkge1xuICB3aWR0aDogMjAwcHg7XG4gIG1pbi1oZWlnaHQ6IDQwMHB4O1xufVxuPC9zdHlsZT5cbiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vcHJldmlldy0yMTMzMS1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzJywgJ2h0dHBzOi8vcHJldmlldy0yMTMzMS1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL3RoZW1lLWNoYWxrL2RhcmsvY3NzLXZhcnMuY3NzJ10ubWFwKChzdHlsZSkgPT4ge1xuICAgIHJldHVybiBuZXcgUHJvbWlzZSgocmVzb2x2ZSwgcmVqZWN0KSA9PiB7XG4gICAgICBjb25zdCBsaW5rID0gZG9jdW1lbnQuY3JlYXRlRWxlbWVudCgnbGluaycpXG4gICAgICBsaW5rLnJlbCA9ICdzdHlsZXNoZWV0J1xuICAgICAgbGluay5ocmVmID0gc3R5bGVcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignbG9hZCcsIHJlc29sdmUpXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2Vycm9yJywgcmVqZWN0KVxuICAgICAgZG9jdW1lbnQuYm9keS5hcHBlbmQobGluaylcbiAgICB9KVxuICB9KVxuICByZXR1cm4gUHJvbWlzZS5hbGxTZXR0bGVkKHN0eWxlcylcbn1cbiIsInRzY29uZmlnLmpzb24iOiJ7XG4gIFwiY29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiBcIkVTTmV4dFwiLFxuICAgIFwianN4XCI6IFwicHJlc2VydmVcIixcbiAgICBcIm1vZHVsZVwiOiBcIkVTTmV4dFwiLFxuICAgIFwibW9kdWxlUmVzb2x1dGlvblwiOiBcIkJ1bmRsZXJcIixcbiAgICBcInR5cGVzXCI6IFtcImVsZW1lbnQtcGx1cy9nbG9iYWwuZC50c1wiXSxcbiAgICBcImFsbG93SW1wb3J0aW5nVHNFeHRlbnNpb25zXCI6IHRydWUsXG4gICAgXCJhbGxvd0pzXCI6IHRydWUsXG4gICAgXCJjaGVja0pzXCI6IHRydWVcbiAgfSxcbiAgXCJ2dWVDb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IDMuM1xuICB9XG59XG4iLCJQbGF5Z3JvdW5kTWFpbi52dWUiOiI8c2NyaXB0IHNldHVwPlxuaW1wb3J0IEFwcCBmcm9tICcuL0FwcC52dWUnXG5pbXBvcnQgeyBzZXR1cEVsZW1lbnRQbHVzIH0gZnJvbSAnLi9lbGVtZW50LXBsdXMuanMnXG5zZXR1cEVsZW1lbnRQbHVzKClcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxBcHAgLz5cbjwvdGVtcGxhdGU+XG4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvcnVudGltZS1kb21AbGF0ZXN0L2Rpc3QvcnVudGltZS1kb20uZXNtLWJyb3dzZXIuanNcIixcbiAgICBcIkB2dWUvc2hhcmVkXCI6IFwiaHR0cHM6Ly9mYXN0bHkuanNkZWxpdnIubmV0L25wbS9AdnVlL3NoYXJlZEBsYXRlc3QvZGlzdC9zaGFyZWQuZXNtLWJ1bmRsZXIuanNcIixcbiAgICBcImVsZW1lbnQtcGx1c1wiOiBcImh0dHBzOi8vcHJldmlldy0yMTMzMS1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwidW5zdXBwb3J0ZWRcIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly9mYXN0bHkuanNkZWxpdnIubmV0L25wbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJfbyI6eyJzaG93SGlkZGVuIjp0cnVlLCJzdHlsZVNvdXJjZSI6Imh0dHBzOi8vcHJldmlldy0yMTMzMS1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzIn19).

warmthsea:
> 2. Add another date-format example. https://element-plus.org/en-US/component/date-picker.html#date-formats

The formatted example looks a bit crowded in the document, so I'll remove it.

Dsaquel:
>The formatted example looks a bit crowded in the document, so I'll remove it.

Okay 🫡

warmthsea:
Can you take a screenshot to illustrate the `show-confirm` attribute?

Dsaquel:
>Can you take a screenshot to illustrate the show-confirm attribute?

image

By default date-picker-panel doesn't expose his confirm button. Date-picker do want the confirm button when the panel permit it.

warmthsea (approved):
I've added a commit to add the `show-confirm` attribute to the datetime-picker docs.

Other than that, I'm fine, and I've installed and tested it in a real project, and it works fine. 
I think this pr is ready to be merged.

Dsaquel:
>I think this pr is ready to be merged.

Yes indeed, we're good to go.

PR #21333, opened 2025-07-13
docs: fix examples & type consistency

Dsaquel:
For select and select-v2:
- add first arg type for `filter-method`, `remote-method`
- add missing attribute `id`
- make same comments/doc

And just remove `:height="208"` => `:height="200"` for the `tree-v2` examples (no difference observed).


btea (approved):
LGTM

warmthsea (approved):

PR #21371, opened 2025-07-16
fix(components): [pagination] reset icon hover state when page changes

### LINKED ISSUE #21370: [Component] [pagination] pagination分页组件图标显示异常问题
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.1`
- Element Plus Version: `2.10.4`
- Browser / OS: `Chrome 138.0`
- Build Tool: `Other`

### Related Component

- `el-pagination`

### Reproduction Link

[Docs](https://element-plus.org/zh-CN/component/pagination.html)

### Steps to reproduce

选至附加功能
以Total item count为例 
第一步：点击左边的...（第一页右边的icon）  此时icon变成<<（双箭头icon）
第二步：点击第10页   此时左侧的图标还是双箭头<<
第三步（错误现象）：悬浮左侧双箭头icon后移出，图标变成了...

### What is Expected?

期望图标显示有自己的规律

目前这个more的图标没有什么规则，很随机，部分情况下是双箭头，部分情况下是三个点；总之在无意识点击的情况下，会很疑惑

### Additional comments

_(empty)_

rzzf:

When switching pages, the icon's hover state was not properly reset, causing the icon to retain its previous hover effect after pagination. This fix ensures that the hover state is reset correctly whenever the page changes.

fix #21370

### DIFF
--- a/packages/components/pagination/src/components/pager.vue
+++ b/packages/components/pagination/src/components/pager.vue
@@ -147,6 +147,8 @@ watchEffect(() => {
   const halfPagerCount = (props.pagerCount - 1) / 2
   showPrevMore.value = false
   showNextMore.value = false
+  quickPrevHover.value = false
+  quickNextHover.value = false
   if (props.pageCount! > props.pagerCount) {
     if (props.currentPage > props.pagerCount - halfPagerCount) {
       showPrevMore.value = true
--- a/packages/components/pagination/src/components/pager.vue
+++ b/packages/components/pagination/src/components/pager.vue
@@ -70,7 +70,7 @@
 </template>
 
 <script lang="ts" setup>
-import { computed, ref, watchEffect } from 'vue'
+import { computed, ref, watch } from 'vue'
 import { DArrowLeft, DArrowRight, MoreFilled } from '@element-plus/icons-vue'
 import { useLocale, useNamespace } from '@element-plus/hooks'
 import { CHANGE_EVENT } from '@element-plus/constants'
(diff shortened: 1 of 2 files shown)


btea:
Could you provide a before-and-after GIF or video of the modification? I still don't understand what the problem is. 🤔

rzzf:
> Could you provide a before-and-after GIF or video of the modification? I still don't understand what the problem is. 🤔

Please check:

1. Click the icon on the left.
2. Click the fifth page, icon change to `<<`（Expected `...`)

Before:

https://github.com/user-attachments/assets/d908a8b5-bd21-47b0-9af1-353109e77024

After:

https://github.com/user-attachments/assets/f0f48846-5bf3-4553-b1c0-901496f39bcc

packages/components/pagination/src/components/pager.vue
```diff
@@ -143,19 +143,25 @@ const nextMoreKls = computed(() => [
...
+
+    if (pageCount > pagerCount) {
+      showPrev = currentPage > pagerCount - halfPagerCount
+      showNext = currentPage < pageCount - halfPagerCount
     }
-  }
-})
+
+    quickPrevHover.value &&= showPrevMore.value
+    quickNextHover.value &&= showNextMore.value
```
Dsaquel: Any reason to change to a watch?

I feel like just adding the potential reset here is good, there are no re-trigger of the `watchEffect` if that the purpose.

cszhjh: I think there’s a special case to consider: when hovering over the icon and switching pages programmatically, if the icon’s position and visibility don’t change, we might want to preserve the hover state.

So I believe we should rely on the current hover and icon state to decide whether we need to reset it to false. If we use watchEffect, quickPrevHover will be collected as a dependency, which could cause the effect to re-run unnecessarily when mouse events change the hover state — even though it’s unrelated to pagination logic.

What do you think?

rzzf: I think there’s a special case to consider: when hovering over the icon and changing pages programmatically, if the icon's position and visibility don’t change, we might want to preserve the hover state. 

So I think we should rely on the current hover and icon state to decide whether or not we need to reset it to `false`.

You're right about `watchEffect`—I’ll revert it.

What do you think?

rzzf: Regarding `watchEffect`, I think although modifying `quickPrevHover` inside the effect won't cause repeated triggers, changes to `quickPrevHover` from mouse events will still cause unnecessary re-executions, even if those changes are unrelated to the pagination logic.

So I think using `watch` is better—at least it makes the sources we listen to explicit, which is easier to maintain and understand. What do you think?

Dsaquel: Yeah i basically agree with you, i just realized that `&&=` will implicitly read the value and set it as dependency causing unexpect triggers when hover an icon hahaha.
Wanted initially to have a minimal diff but we may need to revert with your previous solution or use `unref` (both are fine for me).

rzzf: hahaha🤣. I have just restored the code, please review it again.

Dsaquel: nice thanks you :+1:

Dsaquel (approved):

btea (approved):
Thanks!

PR #21477, opened 2025-07-23
feat(components): [input-tag] add `collapse-*` prop

eighty-cents:
##What has this PR done?
When used as a search condition, input-tag causes the input box to expand when multiple tag contents are entered, affecting the page layout and making it unsightly. Therefore, using the collapseTag method is more appropriate.

##Screenshot/Screen Recording
image

https://github.com/user-attachments/assets/53c1f318-4a78-4471-a961-6d8d6307b1d9


btea:
This feature looks good. WDYT? @tolking

tolking:
I understand why I need these attributes, we can merge this PR.

but before that

1. need to improve the relevant attribute documents
2. refer to ElSelect to complete the `beforeBlur` configuration of `useFocusController`

eighty-cents:
> I understand why I need these attributes, we can merge this PR.
> 
> but before that
> 
> 1. need to improve the relevant attribute documents
> 2. refer to ElSelect to complete the `beforeBlur` configuration of `useFocusController`

@tolking 
1. Related attribute document has been modified
2. The beforeBlur configuration has been added

packages/components/input-tag/src/input-tag.vue
```diff
@@ -10,25 +10,108 @@
...
+          @dragstart="(event: DragEvent) => handleDragStart(event, index)"
+          @dragover="(event: DragEvent) => handleDragOver(event, index)"
+          @dragend="handleDragEnd"
+          @drop.stop
+        >
+          <slot name="tag" :value="item" :index="index">
+            {{ item }}
+          </slot>
+        </el-tag>
+      </template>
```
tolking: IMO, It would be better to calculate showTagList like ElSelect components

packages/components/input-tag/src/input-tag.vue
```diff
@@ -10,25 +10,108 @@
...
+                disable-transitions
+                @close="handleRemoveTag(index)"
+              >
+                <span :class="ns.e('tags-text')">
+                  <slot name="label" :value="item" :index="index">
+                    {{ item }}
+                  </slot>
+                </span>
+              </el-tag>
+            </div>
```
tolking: IMO, It would be better to calculate collapseTagList like ElSelect components

It seems that unnecessary nested node `input-tag-item`, `tags-text` can be avoided here

docs/en-US/component/input-tag.md
```diff
@@ -37,6 +37,16 @@ input-tag/max
 
 :::
 
+## Collapse Tags
```
tolking:
```suggestion
## Collapse Tags ^(2.11.0)
```

docs/en-US/component/input-tag.md
```diff
@@ -111,29 +121,32 @@ input-tag/prefix-suffix
...
+| draggable             | whether tags can be dragged                                                                                    | ^[boolean]                                                  | false   |
+| delimiter ^(2.9.9)    | add a tag when a delimiter is matched                                                                          | ^[string] / ^[regex]                                        | —       |
+| size                  | input box size                                                                                                 | ^[enum]`'large' \| 'default' \| 'small'`                    | —       |
+| collapse-tags         | whether to collapse tags to a text when multiple selecting                                                     | ^[boolean]                                                  | false   |
+| collapse-tags-tooltip | whether show all selected tags when mouse hover text of collapse-tags. To use this, collapse-tags must be true | ^[boolean]                                                  | false   |
```
tolking:
```suggestion
| collapse-tags ^(2.11.0)         | whether to collapse tags to a text when multiple selecting                                                     | ^[boolean]                                                  | false   |
| collapse-tags-tooltip ^(2.11.0) | whether show all selected tags when mouse hover text of collapse-tags. To use this, collapse-tags must be true | ^[boolean]                                                  | false   |
```

docs/en-US/component/input-tag.md
```diff
@@ -111,29 +121,32 @@ input-tag/prefix-suffix
...
+| readonly              | same as `readonly` in native input                                                                             | ^[boolean]                                                  | false   |
+| autofocus             | same as `autofocus` in native input                                                                            | ^[boolean]                                                  | false   |
+| id                    | same as `id` in native input                                                                                   | ^[string]                                                   | —       |
+| tabindex              | same as `tabindex` in native input                                                                             | ^[string] / ^[number]                                       | —       |
+| max-collapse-tags     | the max tags number to be shown. To use this, collapse-tags must be true                                       | ^[number]                                                   | 1       |
```
tolking:
```suggestion
| max-collapse-tags ^(2.11.0)     | the max tags number to be shown. To use this, collapse-tags must be true                                       | ^[number]                                                   | 1       |
```

packages/components/input-tag/src/input-tag.vue
eighty-cents: > IMO, It would be better to calculate showTagList like ElSelect components

ShowTagList has been added with reference to ElSelect

packages/components/input-tag/src/input-tag.vue
eighty-cents: > collapseTagList
@tolking 
1.CollapseTagList has been added with reference to ElSelect
2.Nested node removed

docs/en-US/component/input-tag.md
tolking: Descriptive information also needs to be optimized

packages/components/input-tag/src/composables/use-input-tag.ts
```diff
@@ -38,6 +40,22 @@ export function useInputTag({ props, emit, formItem }: UseInputTagOptions) {
...
+      : props.modelValue
+  })
+  const collapseTagList = computed(() => {
+    if (!props.collapseTags) {
+      return []
+    }
+    return props.collapseTags
+      ? props.modelValue?.slice(props.maxCollapseTags)
+      : []
+  })
```
tolking:
```suggestion
  const showTagList = computed(() => {
    return props.collapseTags
      ? props.modelValue?.slice(0, props.maxCollapseTags)
      : props.modelValue
  })
  const collapseTagList = computed(() => {
    return props.collapseTags
      ? props.modelValue?.slice(props.maxCollapseTags)
      : []
  })
```

packages/components/input-tag/src/composables/use-input-tag.ts
```diff
@@ -94,7 +112,8 @@ export function useInputTag({ props, emit, formItem }: UseInputTagOptions) {
         if (!inputValue.value && props.modelValue?.length) {
           event.preventDefault()
           event.stopPropagation()
-          handleRemoveTag(props.modelValue.length - 1)
+          const tag = props.modelValue[props.modelValue.length - 1]
+          handleRemoveTag(tag)
```
tolking: I don't understand why we need to change this.

packages/components/input-tag/src/input-tag.vue
```diff
@@ -10,25 +10,83 @@
...
+          @dragstart="(event: DragEvent) => handleDragStart(event, index)"
+          @dragover="(event: DragEvent) => handleDragOver(event, index)"
+          @dragend="handleDragEnd"
+          @drop.stop
+        >
+          <slot name="tag" :value="item" :index="index">
+            {{ item }}
+          </slot>
+        </el-tag>
+      </template>
```
tolking: If the collapseTags is false, doesn't it render the same label?

packages/components/input-tag/src/composables/use-input-tag.ts
eighty-cents: > 

Modified @tolking

packages/components/input-tag/src/input-tag.vue
eighty-cents: > If the collapseTags is false, doesn't it render the same label?

Modified @tolking

packages/components/input-tag/src/composables/use-input-tag.ts
eighty-cents: > I don't understand why we need to change this.

@tolking 
1.Because there are two arrays: showTagList and collapseTagList. If you want to delete according to the index, you need to add some judgment in the deletion method
2.Here I modify it by referring to the delete method of the elesct component, because select does the same thing

tolking: The user can enter the same tags, and using the tag value does not determine which tag the user wants to delete.

showTagList and modelValue have the same index number, collapseTagList and modelValue differ by maxCollapseTags.

packages/components/input-tag/src/input-tag.vue
```diff
@@ -10,25 +10,65 @@
...
```
tolking:
```suggestion
        <el-tag
          v-for="(item, index) in showTagList"
          :key="index"
          :size="tagSize"
          :closable="closable"
          :type="tagType"
          :effect="tagEffect"
          :draggable="closable && draggable"
          disable-transitions
          @close="handleRemoveTag(index)"
          @dragstart="(event: DragEvent) => handleDragStart(event, index)"
          @dragover="(event: DragEvent) => handleDragOver(event, index)"
          @dragend="handleDragEnd"
          @drop.stop
        >
          <slot name="tag" :value="item" :index="index">
            {{ item }}
          </slot>
        </el-tag>
```

packages/components/input-tag/src/input-tag.vue
```diff
@@ -10,25 +10,65 @@
...
+            + {{ modelValue.length - maxCollapseTags }}
+          </el-tag>
+        </template>
+        <template #content>
+          <div :class="ns.e('input-tag-list')">
+            <div
+              v-for="item in collapseTagList"
+              :key="item"
+              :class="ns.e('input-tag-item')"
+            >
```
tolking:
```suggestion
```

I do not think this div node is necessary. right?

packages/components/input-tag/src/input-tag.vue
```diff
@@ -10,25 +10,65 @@
...
+              <el-tag
+                :size="tagSize"
+                :closable="closable"
+                :type="tagType"
+                :effect="tagEffect"
+                disable-transitions
+                @close="handleRemoveTag(item)"
+              >
+                {{ item }}
```
tolking, line 64:
```suggestion
          <slot name="tag" :value="item" :index="index + maxCollapseTags">
            {{ item }}
          </slot>
```

packages/components/input-tag/src/input-tag.vue
eighty-cents: > I do not think this div node is necessary. right?

Removed @tolking

packages/components/input-tag/src/composables/use-input-tag.ts
eighty-cents: > The user can enter the same tags, and using the tag value does not determine which tag the user wants to delete.
> 
> showTagList and modelValue have the same index number, collapseTagList and modelValue differ by maxCollapseTags.

Modified to index

packages/components/input-tag/src/input-tag.vue
eighty-cents: > 

Modified @tolking

packages/components/input-tag/src/input-tag.vue
eighty-cents: > 

Does this refer to the following code

`<div :class="ns.e('input-tag-list')">
    <el-tag
      v-for="(item, index) in collapseTagList"
      :key="index"
      :size="tagSize"
      :closable="closable"
      :type="tagType"
      :effect="tagEffect"
      disable-transitions
      @close="handleRemoveTag(index + maxCollapseTags)"
    >
      {{ item }}
    </el-tag>
</div>`

docs/en-US/component/input-tag.md
eighty-cents: > Descriptive information also needs to be optimized

Modified @tolking

packages/components/input-tag/src/input-tag.vue
tolking: Now this tag slot does not work for tags in ElTooltip content.

eighty-cents: > Now this tag slot does not work for tags in ElTooltip content.

I have modified it. Take a look

tolking (approved):
Overall, there doesn't seem to be any problem.

The rest is that the content of the document needs to be optimized.

In addition, it would be best if you could add some test cases.

btea:
The version number corresponding to the newly added `prop` is not yet added in the documentation.

eighty-cents:
> Overall, there doesn't seem to be any problem.
> 
> The rest is that the content of the document needs to be optimized.
> 
> In addition, it would be best if you could add some test cases.

Unit test added @tolking

eighty-cents:
> The version number corresponding to the newly added `prop` is not yet added in the documentation.

Corresponding version number has been added @btea

btea (approved):

btea:
Thanks!

PR #20453, opened 2025-04-16
refactor(components): [menu-item] set index attribute to required

zzjiaxiang:
现在的 `Menu-Item` 的 `index` 类型为 `string / null ` 与 `MenuItemRegistered` 类型里面 `index` 不一致. 

将 `Menu-Item` 的 `index` 更改为了必填. 这样移除 `ts-nocheck` 指令改动能小一点.

## Summary by CodeRabbit

* **Breaking Changes**
  * Menu items now require an index value; the previous optional/null behavior is removed.

* **Refactor**
  * Tightened typing and initialization across menu components.
  * Simplified event handling and keyboard/focus logic with no intended change to user-facing behavior.

PR #21481, opened 2025-07-23
chore(components): [menu] remove ts-nocheck

zzjiaxiang:
ref #20453

### DIFF
--- a/packages/components/menu/src/menu-item.vue
+++ b/packages/components/menu/src/menu-item.vue
@@ -64,7 +64,7 @@ isPropAbsent(props.index) &&
   debugWarn(COMPONENT_NAME, 'Missing required prop: "index"')
 
 const instance = getCurrentInstance()!
-const rootMenu = inject<MenuProvider>(MENU_INJECTION_KEY)!
+const rootMenu = inject<MenuProvider>(MENU_INJECTION_KEY)
 const nsMenu = useNamespace('menu')
 const nsMenuItem = useNamespace('menu-item')
 if (!rootMenu) throwError(COMPONENT_NAME, 'can not inject root menu')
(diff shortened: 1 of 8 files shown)


packages/components/menu/src/menu-item.vue
```diff
@@ -65,7 +64,7 @@ isPropAbsent(props.index) &&
   debugWarn(COMPONENT_NAME, 'Missing required prop: "index"')
 
 const instance = getCurrentInstance()!
-const rootMenu = inject<MenuProvider>(MENU_INJECTION_KEY)
+const rootMenu = inject<MenuProvider>(MENU_INJECTION_KEY)!
 const nsMenu = useNamespace('menu')
 const nsMenuItem = useNamespace('menu-item')
 if (!rootMenu) throwError(COMPONENT_NAME, 'can not inject root menu')
```
Dsaquel, line 70: `rootMenu` is checkout by the judgment below.

```suggestion
const rootMenu = inject<MenuProvider>(MENU_INJECTION_KEY)
const nsMenu = useNamespace('menu')
const nsMenuItem = useNamespace('menu-item')
if (!rootMenu) throwError(COMPONENT_NAME, 'can not inject root menu')
```

zzjiaxiang: image

Dsaquel: Oh it is used in template, ok we can revert the change.

Dsaquel (approved):
Thanks!

warmthsea (commented):
I think there is no need to add a `null` type,
but to change the `index` in prop to be required.

https://github.com/zzjiaxiang/element-plus/blob/remove-ts-nocheck-menu/packages/components/menu/src/menu-item.ts#L16-L21

PR #21486, opened 2025-07-24
feat(components): [cascader] add header and footer slot

rzzf:

resolve #20967, #21358


rzzf:
[Documentation Preview](https://preview-21486-element-plus.surge.sh/en-US/component/cascader.html#custom-header-footer-of-the-dropdown)

btea:
I think it is enough to keep one or two codes in the example, and focus on the content display of the slot. There is too much data, and it is also troublesome to scroll the page expansion code.

rzzf:
> I think it is enough to keep one or two codes in the example, and focus on the content display of the slot. There is too much data, and it is also troublesome to scroll the page expansion code.

How about reducing the `options`? Because I think the all-select function is quite common, and retaining a simple use case for it might be better.

btea:
We can group two examples together, they belong to the same type of example, one on the left and one on the right.

rzzf:
> We can group two examples together, they belong to the same type of example, one on the left and one on the right.

Please check. Thank you.

docs/examples/cascader/custom-header-footer.vue
```diff
@@ -0,0 +1,117 @@
...
+          >
+            All
+          </el-checkbox>
+        </template>
+      </el-cascader>
+    </div>
+    <div>
+      <p>Custom footer content</p>
+      <el-cascader :options="options">
+        <template #footer> Footer </template>
```
Dsaquel: We can set a button to reset the state instead of a plain string.

rzzf: I'm trying to understand what you mean — are you referring to resetting the `modelValue`? Could you provide an example? Thank you!🙏

Dsaquel: Yeah completely, just like the clear button of the datetimerange for example.
I can't send an example right now but I hope you understand.

rzzf: done，Please check.

Dsaquel (approved):
Thanks!

btea (approved):

PR #21490, opened 2025-07-24
fix(components): [date-picker] Invalid time isn’t fixed on blur

SpanManX:
Fix #21404

# Before

# After


Dsaquel:
Single date selection re-open the date-picker after input and applying the cancel if the date is invalid.

Shouldn't we do the same on range type ? :thinking:

SpanManX:
> Single date selection re-open the date-picker after input and applying the cancel if the date is invalid.
> 
> Shouldn't we do the same on range type ? 🤔

Okay. I have changed the code

packages/components/time-picker/src/common/picker.vue
```diff
@@ -640,8 +641,20 @@ const onUserInput = (e: string) => {
...
+  const parse = parseUserInputToDayjs(targetValue) as Dayjs
+  // If the input is invalid, show the panel
+  if (!parse?.isValid()) {
+    pickerVisible.value = true
+  }
+}
+
 const handleStartInput = (event: Event) => {
   const target = event.target as HTMLInputElement
+  handleTimeInput(target.value)
```
Dsaquel: I think we can reuse `onUserInput`.
https://github.com/SpanManX/element-plus/blob/4820650617acbca042df865f9aa9fc04cdea36da/packages/components/time-picker/src/common/picker.vue#L635-L642

warmthsea (commented):
After the change, each input will trigger the popup box, which does not seem to be the correct behavior.

https://preview-21490-element-plus.surge.sh/en-US/component/date-picker.html#date-range

screenshots

SpanManX:
> After the change, each input will trigger the popup box, which does not seem to be the correct behavior.
> 
> https://preview-21490-element-plus.surge.sh/en-US/component/date-picker.html#date-range

I modified the code according to the requirements. @Dsaquel 

> Single date selection re-open the date-picker after input and applying the cancel if the date is invalid.
> 
> Shouldn't we do the same on range type ? 🤔

SpanManX:
> After the change, each input will trigger the popup box, which does not seem to be the correct behavior.
> 
> https://preview-21490-element-plus.surge.sh/en-US/component/date-picker.html#date-range
> 

Changing the input will also trigger the panel when using a single date picker.

Do I need to make any more changes?

屏幕截图 2025-07-25 191919

Dsaquel:
@warmthsea I indeed recommend this change as a regular single date selection have this behavior.

Dsaquel (approved):
LGTM

tolking:

This issues has been fixed.

PR #21502, opened 2025-07-25
refactor(components): [splitter] format class

thinkasany:
- Clearer: The conditions of each class name are clear at a glance
- Easier to maintain: It is easier to add new conditional class names
- Vue native support: Vue can directly handle class binding in object form
- Better performance: Avoid string concatenation operati

### DIFF
--- a/packages/components/splitter/src/split-bar.vue
+++ b/packages/components/splitter/src/split-bar.vue
@@ -59,7 +59,8 @@ const draggerStyles = computed(() => {
 const draggerPseudoClass = computed(() => {
   const prefix = ns.e('dragger')
   return {
-    [`${prefix}-${props.layout}`]: true,
+    [`${prefix}-horizontal`]: isHorizontal.value,
+    [`${prefix}-vertical`]: !isHorizontal.value,
     [`${prefix}-active`]: startPos.value,
   }
 })
--- a/packages/components/splitter/src/split-bar.vue
+++ b/packages/components/splitter/src/split-bar.vue
@@ -58,11 +58,11 @@ const draggerStyles = computed(() => {
 
 const draggerPseudoClass = computed(() => {
   const prefix = ns.e('dragger')
-  let className = isHorizontal.value
-    ? `${prefix}-horizontal`
-    : `${prefix}-vertical`
-  if (startPos.value) className += ` ${prefix}-active`
-  return className
+  return {
+    [`${prefix}-horizontal`]: isHorizontal.value,
+    [`${prefix}-vertical`]: !isHorizontal.value,
+    [`${prefix}-active`]: startPos.value,
+  }
 })
 
 const startPos = ref<[x: number, y: number] | null>(null)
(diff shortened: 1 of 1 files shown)


packages/components/splitter/src/split-bar.vue
```diff
@@ -58,11 +58,11 @@ const draggerStyles = computed(() => {
...
 const draggerPseudoClass = computed(() => {
   const prefix = ns.e('dragger')
-  let className = isHorizontal.value
-    ? `${prefix}-horizontal`
-    : `${prefix}-vertical`
-  if (startPos.value) className += ` ${prefix}-active`
-  return className
+  return {
+    [`${prefix}-horizontal`]: isHorizontal.value,
+    [`${prefix}-vertical`]: !isHorizontal.value,
```
rzzf, line 63: Would it be better to just use `props.layout` directly here?

```ts
[`${prefix}-${props.layout}`]: true
```

thinkasany: This may not be the case. The style above also uses the cached value of layout.

But the difference may not be big, but from the code conciseness, it seems that the code you provided is better.

Thank you for your review. We will wait for the evaluation of the core members to decide whether to make changes~
```
const draggerStyles = computed(() => {
  return {
    width: isHorizontal.value ? '16px' : '100%',
    height: isHorizontal.value ? '100%' : '16px',
    cursor: isHorizontal.value ? 'col-resize' : 'row-resize',
    touchAction: 'none',
  }
})
```

thinkasany: I thought about it again and decided to refer to yours first~

I will change it back if there is anything inappropriate.

packages/components/splitter/src/split-bar.vue
```diff
@@ -58,11 +58,10 @@ const draggerStyles = computed(() => {
...
   const prefix = ns.e('dragger')
-  let className = isHorizontal.value
-    ? `${prefix}-horizontal`
-    : `${prefix}-vertical`
-  if (startPos.value) className += ` ${prefix}-active`
-  return className
+  return {
+    [`${prefix}-${props.layout}`]: true,
+    [`${prefix}-active`]: startPos.value,
+  }
```
btea, line 65: Thanks for your contribution. Perhaps it's best to leave it as is. Although `props.layout` restricts the value, if the user passes in a wrong value, they will get a useless class name.

thinkasany: Thanks for your reply, I adjusted it and it should be equivalent optimization now.

image

packages/components/splitter/src/split-bar.vue
```diff
@@ -58,11 +58,11 @@ const draggerStyles = computed(() => {
...
   const prefix = ns.e('dragger')
-  let className = isHorizontal.value
-    ? `${prefix}-horizontal`
-    : `${prefix}-vertical`
-  if (startPos.value) className += ` ${prefix}-active`
-  return className
+  return {
+    [`${prefix}-horizontal`]: isHorizontal.value,
+    [`${prefix}-vertical`]: !isHorizontal.value,
+    [`${prefix}-active`]: startPos.value,
```
btea:
```suggestion
    [`${prefix}-active`]: !!startPos.value,
```

btea (approved):
Thanks!

PR #21508, opened 2025-07-27
feat(components): [cascader] add `check-on-click-leaf` attribute

Dsaquel:
- add `check-on-click-leaf` attribute.
- Made the doc a bit more clear with the new added features for `2.10.5`.

### DIFF
--- a/docs/en-US/component/cascader.md
+++ b/docs/en-US/component/cascader.md
@@ -148,33 +148,38 @@ cascader/custom-tag
 
 :::
 
-## Custom Header & Footer of the Dropdown ^(2.10.5)
-
-You can customize both the header and footer of the dropdown using slots.
-
-:::demo Use slot to customize the content.
-
-cascader/custom-header-footer
-
-:::
-
-## Click to Select Node ^(2.10.5)
-
-Allows selecting nodes by clicking, with optional control visibility.
-
-:::demo Select nodes on click via `checkOnClickNode`. You can also hide the selection control with `showPrefix`.
-
-cascader/check-on-click-node
-
-:::
-
-## Show Checked Strategy
+## Show Checked Strategy ^(2.10.5)
 
 Control how selected values are displayed in multiple selection mode.
 
 :::demo In multiple selection mode, you can use `show-checked-strategy` to control how selected values are displayed. The default strategy is `child`, which shows all selected child nodes. The `parent` strategy only shows parent nodes when all their children are selected.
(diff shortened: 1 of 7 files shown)


docs/en-US/component/cascader.md
```diff
@@ -298,6 +303,7 @@ cascader/show-checked-strategy
 | leaf                       | specify which key of node object is used as the node's leaf field                                          | ^[string]                                           | leaf     |
 | hoverThreshold             | hover threshold of expanding options                                                                       | ^[number]                                           | 500      |
 | checkOnClickNode ^(2.10.5) | whether to check or uncheck node when clicking on the node                                                 | ^[boolean]                                          | false    |
+| checkOnClickLeaf ^(2.10.5) | whether to check or uncheck node when clicking on leaf node (last children).                               | ^[boolean]                                          | true     |
```
btea, line 306: Perhaps instead of adding a new attribute, we could extend `checkOnClickNode ` to have a type of `boolean | 'leaf'`? 🤔

Dsaquel: It makes sense to me but since we have `checkOnClickLeaf` for the tree component would it be more consistent to stay as is ?

btea: I just remembered that the tree component also has a property of the same name. To maintain consistency, this is the only way.

packages/components/cascader-panel/src/node-content.tsx
```diff
@@ -18,22 +19,26 @@ export default defineComponent({
...
 
     const label = () => {
       const renderLabel = renderLabelFn?.({ node, data })
       return isVNodeEmpty(renderLabel) ? nodeLabel : renderLabel ?? nodeLabel
     }
     function handleClick() {
-      if (disabled || !checkOnClickNode) return
-      emit('handleSelectCheck', !node.checked)
+      if (
+        (checkOnClickNode || (node.isLeaf && checkOnClickLeaf && multiple)) &&
```
btea: Is it necessary to restrict `multiple` here? 🤔

Dsaquel: There will no have impact on single mode if `checkOnClickLeaf` is true or false if that the question.
But yeah i don't think either `multiple` have impact here.

btea (approved):
LGTM

warmthsea (approved):
LGTM

PR #21532, opened 2025-07-29
docs: format warning tip

warmthsea:
### Description

1. Ensured correct automatic text parsing in [Crowdin](https://crowdin.com/editor/element-plus/883/en-zhcn?view=comfortable&filter=basic&value=0).

2. [Unified](https://github.com/element-plus/element-plus/blob/docs/warn-tip-foramt/docs/en-US/component/button.md?plain=1#L30-L36).

### Now

 | [US](https://element-plus.org/en-US/component/tree.html#selectable) | [ZH](https://element-plus.org/zh-CN/component/tree.html#%E5%8F%AF%E9%80%89%E6%8B%A9) |
| ------- | ----- |
 |      image      |     image  |

<hr />

image

### DIFF
--- a/docs/en-US/component/tree-select.md
+++ b/docs/en-US/component/tree-select.md
@@ -38,8 +38,10 @@ tree-select/check-strictly
 :::
 
 :::warning
+
 When using show-checkbox, since `check-on-click-leaf` is true by default,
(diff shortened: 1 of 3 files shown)


btea (approved):
👍

PR #21543, opened 2025-07-29
feat(components): [radio-group] support options prop rendering

snowbitx:
[](Please make sure these boxes are checked before submitting your PR, thank you!

support options prop rendering    )   [demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtcmFkaW8tZ3JvdXAgdi1tb2RlbD1cInJhZGlvXCIgOm9wdGlvbnM9XCJvcHRpb25zXCIgLz5cbiAgPGRpdj5cbiAgICA8ZWwtcmFkaW8tZ3JvdXAgdi1tb2RlbD1cInJhZGlvXCIgOm9wdGlvbnM9XCJvcHRpb25zMlwiIDpwcm9wcz1cInt2YWx1ZTonaWQnfVwiLz5cbiAgPC9kaXY+XG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IGxhbmc9XCJ0c1wiIHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSAndnVlJ1xuY29uc3QgcmFkaW8gPSByZWYoMylcbmNvbnN0IG9wdGlvbnMgPSBbXG4gIHtcbiAgICB2YWx1ZTogMyxcbiAgICBsYWJlbDogJ09wdGlvbiBBJyxcbiAgfSxcbiAge1xuICAgIHZhbHVlOiA2LFxuICAgIGxhYmVsOiAnT3B0aW9uIEInLFxuICB9LFxuICB7XG4gICAgdmFsdWU6IDksXG4gICAgbGFiZWw6ICdPcHRpb24gQycsXG4gIH0sXG5dXG5jb25zdCBvcHRpb25zMiA9IFtcbiAge1xuICAgIGlkOiAzLFxuICAgIGxhYmVsOiAnT3B0aW9uIEEnLFxuICB9LFxuICB7XG4gICAgaWQ6IDYsXG4gICAgbGFiZWw6ICdPcHRpb24gQicsXG4gIH0sXG4gIHtcbiAgICBpZDogOSxcbiAgICBsYWJlbDogJ09wdGlvbiBDJyxcbiAgfSxcbl1cbjwvc2NyaXB0PiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vcHJldmlldy0yMTU0My1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzJywgJ2h0dHBzOi8vcHJldmlldy0yMTU0My1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL3RoZW1lLWNoYWxrL2RhcmsvY3NzLXZhcnMuY3NzJ10ubWFwKChzdHlsZSkgPT4ge1xuICAgIHJldHVybiBuZXcgUHJvbWlzZSgocmVzb2x2ZSwgcmVqZWN0KSA9PiB7XG4gICAgICBjb25zdCBsaW5rID0gZG9jdW1lbnQuY3JlYXRlRWxlbWVudCgnbGluaycpXG4gICAgICBsaW5rLnJlbCA9ICdzdHlsZXNoZWV0J1xuICAgICAgbGluay5ocmVmID0gc3R5bGVcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignbG9hZCcsIHJlc29sdmUpXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2Vycm9yJywgcmVqZWN0KVxuICAgICAgZG9jdW1lbnQuYm9keS5hcHBlbmQobGluaylcbiAgICB9KVxuICB9KVxuICByZXR1cm4gUHJvbWlzZS5hbGxTZXR0bGVkKHN0eWxlcylcbn1cbiIsInRzY29uZmlnLmpzb24iOiJ7XG4gIFwiY29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiBcIkVTTmV4dFwiLFxuICAgIFwianN4XCI6IFwicHJlc2VydmVcIixcbiAgICBcIm1vZHVsZVwiOiBcIkVTTmV4dFwiLFxuICAgIFwibW9kdWxlUmVzb2x1dGlvblwiOiBcIkJ1bmRsZXJcIixcbiAgICBcInR5cGVzXCI6IFtcImVsZW1lbnQtcGx1cy9nbG9iYWwuZC50c1wiXSxcbiAgICBcImFsbG93SW1wb3J0aW5nVHNFeHRlbnNpb25zXCI6IHRydWUsXG4gICAgXCJhbGxvd0pzXCI6IHRydWUsXG4gICAgXCJjaGVja0pzXCI6IHRydWVcbiAgfSxcbiAgXCJ2dWVDb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IDMuM1xuICB9XG59XG4iLCJQbGF5Z3JvdW5kTWFpbi52dWUiOiI8c2NyaXB0IHNldHVwPlxuaW1wb3J0IEFwcCBmcm9tICcuL0FwcC52dWUnXG5pbXBvcnQgeyBzZXR1cEVsZW1lbnRQbHVzIH0gZnJvbSAnLi9lbGVtZW50LXBsdXMuanMnXG5zZXR1cEVsZW1lbnRQbHVzKClcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxBcHAgLz5cbjwvdGVtcGxhdGU+XG4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvcnVudGltZS1kb21AbGF0ZXN0L2Rpc3QvcnVudGltZS1kb20uZXNtLWJyb3dzZXIuanNcIixcbiAgICBcIkB2dWUvc2hhcmVkXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AdnVlL3NoYXJlZEBsYXRlc3QvZGlzdC9zaGFyZWQuZXNtLWJ1bmRsZXIuanNcIixcbiAgICBcImVsZW1lbnQtcGx1c1wiOiBcImh0dHBzOi8vcHJldmlldy0yMTU0My1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwidW5zdXBwb3J0ZWRcIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJfbyI6eyJzaG93SGlkZGVuIjp0cnVlLCJzdHlsZVNvdXJjZSI6Imh0dHBzOi8vcHJldmlldy0yMTU0My1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzIn19)


packages/components/radio/src/radio-group.vue
```diff
@@ -7,7 +7,16 @@
...
   >
-    <slot />
+    <slot>
+      <Radio
+        v-for="(item, index) in options"
+        :key="index"
+        :value="item[props.props?.value ?? 'value']"
+        :disabled="item[props.props?.disabled ?? 'disabled']"
+      >
+        <span>{{ item[props.props?.label ?? 'label'] }} </span>
```
Dsaquel: I think we can set this under the label attribute.

packages/components/radio/src/radio-group.ts
```diff
@@ -1,6 +1,7 @@
-import { buildProps } from '@element-plus/utils'
+import { buildProps, definePropType } from '@element-plus/utils'
 import { useAriaProps, useSizeProp } from '@element-plus/hooks'
 import { radioEmits } from './radio'
+import { SelectOptionProps } from '@element-plus/components/select/src/select'
 
 import type { ExtractPropTypes, __ExtractPublicPropTypes } from 'vue'
```
Dsaquel:
```suggestion

import type { ExtractPropTypes, __ExtractPublicPropTypes } from 'vue'
import type { SelectOptionProps } from '@element-plus/components/select/src/select'
```

packages/components/radio/src/radio-group.vue
```diff
@@ -28,6 +37,7 @@ import { useId, useNamespace } from '@element-plus/hooks'
 import { debugWarn } from '@element-plus/utils'
 import { radioGroupEmits, radioGroupProps } from './radio-group'
 import { radioGroupKey } from './constants'
+import Radio from './radio.vue'
```
Dsaquel: **nit**:
```suggestion
import ElRadio from './radio.vue'
```

packages/components/radio/src/radio-group.vue
snowbitx: > I think we can set this under the label attribute.

Thanks! Already pushed it to the latest PR 😊

packages/components/radio/src/radio-group.vue
```diff
@@ -7,7 +7,15 @@
     :aria-label="!isLabeledByFormItem ? ariaLabel || 'radio-group' : undefined"
     :aria-labelledby="isLabeledByFormItem ? formItem!.labelId : undefined"
   >
-    <slot />
+    <slot>
+      <ElRadio
```
btea:
```suggestion
      <el-radio
```

packages/components/radio/src/radio-group.vue
```diff
@@ -7,7 +7,15 @@
...
+        v-for="(item, index) in options"
+        :key="index"
+        :value="item[props.props?.value ?? 'value']"
+        :label="item[props.props?.label ?? 'label']"
+        :disabled="item[props.props?.disabled ?? 'disabled']"
```
btea: There is calculation logic in `options`, maybe we should put it into the following ts code.

snowbitx: > There is calculation logic in `options`, maybe we should put it into the following ts code.

Hey, I noticed that el-select and el-checkbox might need similar changes, so I pulled out some common parts (like this [link](https://github.com/element-plus/element-plus/pull/21545/files#diff-045dec09d262543ba6187add4b99a8b3ad58b8ec74cdffdbdaed5343a1b58981R1-R27)) into a shared place. Not sure if this is the best approach for our code style—if it’s not, I’m happy to just make the changes directly in this file. Thanks for taking a look and let me know what you think!

snowbitx: > There is calculation logic in `options`, maybe we should put it into the following ts code.

Or maybe we could keep it consistent with how it's done in el-select? Just a thought—totally open to your suggestions!  [link](https://github.com/element-plus/element-plus/blob/dev/packages/components/select/src/select.vue#L279C13-L287C22)

packages/components/radio/src/radio-group.vue
```diff
@@ -51,6 +60,18 @@ const changeEvent = (value: RadioGroupProps['modelValue']) => {
   nextTick(() => emit(CHANGE_EVENT, value))
 }
 
+const { options = [], props: propConfig } = props
```
warmthsea: [bug](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtcmFkaW8tZ3JvdXAgdi1tb2RlbD1cInJhZGlvXCIgOm9wdGlvbnM9XCJvcHRpb25zXCIgLz5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5pbXBvcnQgeyByZWYgfSBmcm9tICd2dWUnXG5cbmNvbnN0IHJhZGlvID0gcmVmKDMpXG5cbmNvbnN0IG9wdGlvbnMgPSByZWYoW1xuICB7XG4gICAgdmFsdWU6IDMsXG4gICAgbGFiZWw6ICdPcHRpb24gQScsXG4gIH0sXG4gIHtcbiAgICB2YWx1ZTogNixcbiAgICBsYWJlbDogJ09wdGlvbiBCJyxcbiAgfSxcbiAge1xuICAgIHZhbHVlOiA5LFxuICAgIGxhYmVsOiAnT3B0aW9uIEMnLFxuICB9LFxuXSlcblxuc2V0VGltZW91dCgoKT0+e1xuICBvcHRpb25zLnZhbHVlID0gW3tcbiAgICB2YWx1ZTogOSxcbiAgICBsYWJlbDogJ09wdGlvbiBDJyxcbiAgfV1cbn0sMjAwMClcbjwvc2NyaXB0PlxuIiwiZWxlbWVudC1wbHVzLmpzIjoiaW1wb3J0IEVsZW1lbnRQbHVzIGZyb20gJ2VsZW1lbnQtcGx1cydcbmltcG9ydCB7IGdldEN1cnJlbnRJbnN0YW5jZSB9IGZyb20gJ3Z1ZSdcblxubGV0IGluc3RhbGxlZCA9IGZhbHNlXG5hd2FpdCBsb2FkU3R5bGUoKVxuXG5leHBvcnQgZnVuY3Rpb24gc2V0dXBFbGVtZW50UGx1cygpIHtcbiAgaWYgKGluc3RhbGxlZCkgcmV0dXJuXG4gIGNvbnN0IGluc3RhbmNlID0gZ2V0Q3VycmVudEluc3RhbmNlKClcbiAgaW5zdGFuY2UuYXBwQ29udGV4dC5hcHAudXNlKEVsZW1lbnRQbHVzKVxuICBpbnN0YWxsZWQgPSB0cnVlXG59XG5cbmV4cG9ydCBmdW5jdGlvbiBsb2FkU3R5bGUoKSB7XG4gIGNvbnN0IHN0eWxlcyA9IFsnaHR0cHM6Ly9wcmV2aWV3LTIxNTQzLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5jc3MnLCAnaHR0cHM6Ly9wcmV2aWV3LTIxNTQzLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvdGhlbWUtY2hhbGsvZGFyay9jc3MtdmFycy5jc3MnXS5tYXAoKHN0eWxlKSA9PiB7XG4gICAgcmV0dXJuIG5ldyBQcm9taXNlKChyZXNvbHZlLCByZWplY3QpID0+IHtcbiAgICAgIGNvbnN0IGxpbmsgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCdsaW5rJylcbiAgICAgIGxpbmsucmVsID0gJ3N0eWxlc2hlZXQnXG4gICAgICBsaW5rLmhyZWYgPSBzdHlsZVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdsb2FkJywgcmVzb2x2ZSlcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignZXJyb3InLCByZWplY3QpXG4gICAgICBkb2N1bWVudC5ib2R5LmFwcGVuZChsaW5rKVxuICAgIH0pXG4gIH0pXG4gIHJldHVybiBQcm9taXNlLmFsbFNldHRsZWQoc3R5bGVzKVxufVxuIiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IFwiRVNOZXh0XCIsXG4gICAgXCJqc3hcIjogXCJwcmVzZXJ2ZVwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwidHlwZXNcIjogW1wiZWxlbWVudC1wbHVzL2dsb2JhbC5kLnRzXCJdLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcImFsbG93SnNcIjogdHJ1ZSxcbiAgICBcImNoZWNrSnNcIjogdHJ1ZVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vZmFzdGx5LmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9ydW50aW1lLWRvbUBsYXRlc3QvZGlzdC9ydW50aW1lLWRvbS5lc20tYnJvd3Nlci5qc1wiLFxuICAgIFwiQHZ1ZS9zaGFyZWRcIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvc2hhcmVkQGxhdGVzdC9kaXN0L3NoYXJlZC5lc20tYnVuZGxlci5qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9wcmV2aWV3LTIxNTQzLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5mdWxsLm1pbi5tanNcIixcbiAgICBcImVsZW1lbnQtcGx1cy9cIjogXCJ1bnN1cHBvcnRlZFwiLFxuICAgIFwiQGVsZW1lbnQtcGx1cy9pY29ucy12dWVcIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlQDIvZGlzdC9pbmRleC5taW4uanNcIlxuICB9LFxuICBcInNjb3Blc1wiOiB7fVxufSIsIl9vIjp7InNob3dIaWRkZW4iOnRydWUsInN0eWxlU291cmNlIjoiaHR0cHM6Ly9wcmV2aWV3LTIxNTQzLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5jc3MifX0=)

snowbitx: > [bug](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtcmFkaW8tZ3JvdXAgdi1tb2RlbD1cInJhZGlvXCIgOm9wdGlvbnM9XCJvcHRpb25zXCIgLz5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5pbXBvcnQgeyByZWYgfSBmcm9tICd2dWUnXG5cbmNvbnN0IHJhZGlvID0gcmVmKDMpXG5cbmNvbnN0IG9wdGlvbnMgPSByZWYoW1xuICB7XG4gICAgdmFsdWU6IDMsXG4gICAgbGFiZWw6ICdPcHRpb24gQScsXG4gIH0sXG4gIHtcbiAgICB2YWx1ZTogNixcbiAgICBsYWJlbDogJ09wdGlvbiBCJyxcbiAgfSxcbiAge1xuICAgIHZhbHVlOiA5LFxuICAgIGxhYmVsOiAnT3B0aW9uIEMnLFxuICB9LFxuXSlcblxuc2V0VGltZW91dCgoKT0+e1xuICBvcHRpb25zLnZhbHVlID0gW3tcbiAgICB2YWx1ZTogOSxcbiAgICBsYWJlbDogJ09wdGlvbiBDJyxcbiAgfV1cbn0sMjAwMClcbjwvc2NyaXB0PlxuIiwiZWxlbWVudC1wbHVzLmpzIjoiaW1wb3J0IEVsZW1lbnRQbHVzIGZyb20gJ2VsZW1lbnQtcGx1cydcbmltcG9ydCB7IGdldEN1cnJlbnRJbnN0YW5jZSB9IGZyb20gJ3Z1ZSdcblxubGV0IGluc3RhbGxlZCA9IGZhbHNlXG5hd2FpdCBsb2FkU3R5bGUoKVxuXG5leHBvcnQgZnVuY3Rpb24gc2V0dXBFbGVtZW50UGx1cygpIHtcbiAgaWYgKGluc3RhbGxlZCkgcmV0dXJuXG4gIGNvbnN0IGluc3RhbmNlID0gZ2V0Q3VycmVudEluc3RhbmNlKClcbiAgaW5zdGFuY2UuYXBwQ29udGV4dC5hcHAudXNlKEVsZW1lbnRQbHVzKVxuICBpbnN0YWxsZWQgPSB0cnVlXG59XG5cbmV4cG9ydCBmdW5jdGlvbiBsb2FkU3R5bGUoKSB7XG4gIGNvbnN0IHN0eWxlcyA9IFsnaHR0cHM6Ly9wcmV2aWV3LTIxNTQzLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5jc3MnLCAnaHR0cHM6Ly9wcmV2aWV3LTIxNTQzLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvdGhlbWUtY2hhbGsvZGFyay9jc3MtdmFycy5jc3MnXS5tYXAoKHN0eWxlKSA9PiB7XG4gICAgcmV0dXJuIG5ldyBQcm9taXNlKChyZXNvbHZlLCByZWplY3QpID0+IHtcbiAgICAgIGNvbnN0IGxpbmsgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCdsaW5rJylcbiAgICAgIGxpbmsucmVsID0gJ3N0eWxlc2hlZXQnXG4gICAgICBsaW5rLmhyZWYgPSBzdHlsZVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdsb2FkJywgcmVzb2x2ZSlcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignZXJyb3InLCByZWplY3QpXG4gICAgICBkb2N1bWVudC5ib2R5LmFwcGVuZChsaW5rKVxuICAgIH0pXG4gIH0pXG4gIHJldHVybiBQcm9taXNlLmFsbFNldHRsZWQoc3R5bGVzKVxufVxuIiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IFwiRVNOZXh0XCIsXG4gICAgXCJqc3hcIjogXCJwcmVzZXJ2ZVwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwidHlwZXNcIjogW1wiZWxlbWVudC1wbHVzL2dsb2JhbC5kLnRzXCJdLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcImFsbG93SnNcIjogdHJ1ZSxcbiAgICBcImNoZWNrSnNcIjogdHJ1ZVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vZmFzdGx5LmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9ydW50aW1lLWRvbUBsYXRlc3QvZGlzdC9ydW50aW1lLWRvbS5lc20tYnJvd3Nlci5qc1wiLFxuICAgIFwiQHZ1ZS9zaGFyZWRcIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvc2hhcmVkQGxhdGVzdC9kaXN0L3NoYXJlZC5lc20tYnVuZGxlci5qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9wcmV2aWV3LTIxNTQzLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5mdWxsLm1pbi5tanNcIixcbiAgICBcImVsZW1lbnQtcGx1cy9cIjogXCJ1bnN1cHBvcnRlZFwiLFxuICAgIFwiQGVsZW1lbnQtcGx1cy9pY29ucy12dWVcIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlQDIvZGlzdC9pbmRleC5taW4uanNcIlxuICB9LFxuICBcInNjb3Blc1wiOiB7fVxufSIsIl9vIjp7InNob3dIaWRkZW4iOnRydWUsInN0eWxlU291cmNlIjoiaHR0cHM6Ly9wcmV2aWV3LTIxNTQzLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5jc3MifX0=)

Good catch! Fixed it.  [demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtcmFkaW8tZ3JvdXAgdi1tb2RlbD1cInJhZGlvXCIgOm9wdGlvbnM9XCJvcHRpb25zXCIgLz5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5pbXBvcnQgeyByZWYgfSBmcm9tICd2dWUnXG5cbmNvbnN0IHJhZGlvID0gcmVmKDMpXG5cbmNvbnN0IG9wdGlvbnMgPSByZWYoW1xuICB7XG4gICAgdmFsdWU6IDMsXG4gICAgbGFiZWw6ICdPcHRpb24gQScsXG4gIH0sXG4gIHtcbiAgICB2YWx1ZTogNixcbiAgICBsYWJlbDogJ09wdGlvbiBCJyxcbiAgfSxcbiAge1xuICAgIHZhbHVlOiA5LFxuICAgIGxhYmVsOiAnT3B0aW9uIEMnLFxuICB9LFxuXSlcblxuc2V0VGltZW91dCgoKT0+e1xuICBvcHRpb25zLnZhbHVlID0gW3tcbiAgICB2YWx1ZTogOSxcbiAgICBsYWJlbDogJ09wdGlvbiBDJyxcbiAgfV1cbn0sMjAwMClcbjwvc2NyaXB0PlxuIiwiZWxlbWVudC1wbHVzLmpzIjoiaW1wb3J0IEVsZW1lbnRQbHVzIGZyb20gJ2VsZW1lbnQtcGx1cydcbmltcG9ydCB7IGdldEN1cnJlbnRJbnN0YW5jZSB9IGZyb20gJ3Z1ZSdcblxubGV0IGluc3RhbGxlZCA9IGZhbHNlXG5hd2FpdCBsb2FkU3R5bGUoKVxuXG5leHBvcnQgZnVuY3Rpb24gc2V0dXBFbGVtZW50UGx1cygpIHtcbiAgaWYgKGluc3RhbGxlZCkgcmV0dXJuXG4gIGNvbnN0IGluc3RhbmNlID0gZ2V0Q3VycmVudEluc3RhbmNlKClcbiAgaW5zdGFuY2UuYXBwQ29udGV4dC5hcHAudXNlKEVsZW1lbnRQbHVzKVxuICBpbnN0YWxsZWQgPSB0cnVlXG59XG5cbmV4cG9ydCBmdW5jdGlvbiBsb2FkU3R5bGUoKSB7XG4gIGNvbnN0IHN0eWxlcyA9IFsnaHR0cHM6Ly9wcmV2aWV3LTIxNTQzLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5jc3MnLCAnaHR0cHM6Ly9wcmV2aWV3LTIxNTQzLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvdGhlbWUtY2hhbGsvZGFyay9jc3MtdmFycy5jc3MnXS5tYXAoKHN0eWxlKSA9PiB7XG4gICAgcmV0dXJuIG5ldyBQcm9taXNlKChyZXNvbHZlLCByZWplY3QpID0+IHtcbiAgICAgIGNvbnN0IGxpbmsgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCdsaW5rJylcbiAgICAgIGxpbmsucmVsID0gJ3N0eWxlc2hlZXQnXG4gICAgICBsaW5rLmhyZWYgPSBzdHlsZVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdsb2FkJywgcmVzb2x2ZSlcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignZXJyb3InLCByZWplY3QpXG4gICAgICBkb2N1bWVudC5ib2R5LmFwcGVuZChsaW5rKVxuICAgIH0pXG4gIH0pXG4gIHJldHVybiBQcm9taXNlLmFsbFNldHRsZWQoc3R5bGVzKVxufVxuIiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IFwiRVNOZXh0XCIsXG4gICAgXCJqc3hcIjogXCJwcmVzZXJ2ZVwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwidHlwZXNcIjogW1wiZWxlbWVudC1wbHVzL2dsb2JhbC5kLnRzXCJdLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcImFsbG93SnNcIjogdHJ1ZSxcbiAgICBcImNoZWNrSnNcIjogdHJ1ZVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9ydW50aW1lLWRvbUBsYXRlc3QvZGlzdC9ydW50aW1lLWRvbS5lc20tYnJvd3Nlci5qc1wiLFxuICAgIFwiQHZ1ZS9zaGFyZWRcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvc2hhcmVkQGxhdGVzdC9kaXN0L3NoYXJlZC5lc20tYnVuZGxlci5qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9wcmV2aWV3LTIxNTQzLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5mdWxsLm1pbi5tanNcIixcbiAgICBcImVsZW1lbnQtcGx1cy9cIjogXCJ1bnN1cHBvcnRlZFwiLFxuICAgIFwiQGVsZW1lbnQtcGx1cy9pY29ucy12dWVcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlQDIvZGlzdC9pbmRleC5taW4uanNcIlxuICB9LFxuICBcInNjb3Blc1wiOiB7fVxufSIsIl9vIjp7InNob3dIaWRkZW4iOnRydWUsInN0eWxlU291cmNlIjoiaHR0cHM6Ly9wcmV2aWV3LTIxNTQzLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5jc3MifX0=)

packages/components/radio/src/radio-group.vue
```diff
@@ -51,6 +60,17 @@ const changeEvent = (value: RadioGroupProps['modelValue']) => {
   nextTick(() => emit(CHANGE_EVENT, value))
 }
 
+const radioOptions = computed(() => {
```
Dsaquel: We can merge this change to be in the template in order to remove one extra loop.

docs/en-US/component/radio.md
```diff
@@ -61,6 +61,14 @@ radio/radio-button-group
 
 :::
 
+## Options attribute ^(2.10.5)
```
Dsaquel: Versions can be modified to `^2.11.0`.

packages/components/radio/src/radio-group.vue
snowbitx: > We can merge this change to be in the template in order to remove one extra loop.

Good point, updated accordingly.

docs/en-US/component/radio.md
snowbitx: > Versions can be modified to `^2.11.0`.

Updated, thanks!

Dsaquel (commented):
Now LGTM, but maybe the prop `options` should be renamed to `radios`.

Let's see what other people think about this.

btea:
> Now LGTM, but maybe the prop `options` should be renamed to `radios`.
> 
> Let's see what other people think about this.

Using options also looks good and is consistent with other components such as select. 

By the way, we can add some related test cases.

warmthsea (approved):
I think using `options` is good.

packages/components/radio/src/radio-group.ts
```diff
@@ -56,6 +57,12 @@ export const radioGroupProps = buildProps({
     type: Boolean,
     default: true,
   },
+  options: {
+    type: definePropType<Record<string, any>[]>(Array),
+  },
+  props: {
+    type: definePropType<SelectOptionProps>(Object),
+  },
```
jw-foss: I would recommend this to be renamed as `optionProps` or `radioProps` to distinguish from already existing `props` here.

jw-foss (dismissed):

packages/components/radio/src/radio-group.ts
jw-foss: I am also wondering how do you define the attributes for individual `radio` here? It seems like you can only define the `key` for `label/value/disabled`, what happens if I want to define attributes like `data-testid` against each `el-radio` here?

jw-foss: What I would suggest here is that you should be able to combine the props for each `ElRadio` here like the following:

```ts
import type { ComponentProps, ComponentSlots } from 'vue-component-type-helpers'

type Option = ComponentProps<typeof ElRadio> & Record<string, any> // other attributes 
```

Also, just wondering should we also provide a way for the user to render the `radio`'s default slot like `radio-renderer` or something? This is merely a question rather than requirements.

snowbitx: > What I would suggest here is that you should be able to combine the props for each `ElRadio` here like the following:
> 
> ```ts
> import type { ComponentProps, ComponentSlots } from 'vue-component-type-helpers'
> 
> type Option = ComponentProps<typeof ElRadio> & Record<string, any> // other attributes 
> ```
> 
> Also, just wondering should we also provide a way for the user to render the `radio`'s default slot like `radio-renderer` or something? This is merely a question rather than requirements.

Thanks for the great suggestions! 🙏  I've updated the PR and added tests. Please take a look when you have time.Thanks again!

packages/components/radio/src/radio-group.vue
```diff
@@ -7,7 +7,22 @@
...
+      <el-radio
+        v-for="(item, index) in props.options"
+        :key="item.key ?? index"
+        v-bind="{
+          ...item,
+          value: item[props.optionProps?.value ?? 'value'],
+          label: item[props.optionProps?.label ?? 'label'],
+          disabled: item[props.optionProps?.disabled ?? 'disabled'],
+        }"
+      >
```
kooriookami: Please change it refer to #21545

packages/components/radio/src/radio-group.ts
```diff
@@ -1,9 +1,14 @@
-import { buildProps } from '@element-plus/utils'
+import { buildProps, definePropType } from '@element-plus/utils'
 import { useAriaProps, useSizeProp } from '@element-plus/hooks'
 import { radioEmits } from './radio'
+import { ElRadio } from '@element-plus/components'
 
 import type { ExtractPropTypes, __ExtractPublicPropTypes } from 'vue'
+import type { SelectOptionProps } from '@element-plus/components/select/src/select'
 import type RadioGroup from './radio-group.vue'
+import type { ComponentProps } from 'vue-component-type-helpers'
```
kooriookami: Please change it refer to #21545

kooriookami (changes requested):

packages/components/radio/src/radio-group.ts
snowbitx: I've updated it in the latest PR using independent logic.

packages/components/radio/src/radio-group.vue
```diff
@@ -65,6 +83,16 @@ const name = computed(() => {
...
+const aliasProps = computed(() => ({
+  ...radioDefaultProps,
+  ...props.optionProps,
+}))
+const getOptionProps = (option: Record<string, any>) => ({
+  label: option[aliasProps.value.label],
+  value: option[aliasProps.value.value],
+  disabled: option[aliasProps.value.disabled],
+})
+
```
kooriookami: Please keep consistent with #21545

docs/en-US/component/radio.md
```diff
@@ -61,6 +61,14 @@ radio/radio-button-group
 
 :::
 
+## Options attribute ^(2.11.0)
```
kooriookami: Please update the version number.

kooriookami (changes requested):

packages/components/radio/src/radio-group.vue
snowbitx: > Please keep consistent with #21545

Updated, Thanks for the reminder.

packages/components/radio/src/radio-group.ts
```diff
@@ -1,6 +1,6 @@
-import { buildProps } from '@element-plus/utils'
+import { buildProps, definePropType } from '@element-plus/utils'
 import { useAriaProps, useSizeProp } from '@element-plus/hooks'
-import { radioEmits } from './radio'
+import { type RadioPropsPublic, radioEmits } from './radio'
```
kooriookami: Please let import type { xxx } a single line.

packages/components/radio/src/radio-group.vue
```diff
@@ -7,7 +7,20 @@
...
-    <slot />
+    <slot>
+      <el-radio
+        v-for="(item, index) in props.options"
+        :key="item.key ?? index"
+        v-bind="{
+          ...item,
+          ...getOptionProps(item),
+        }"
+      >
```
kooriookami:
```suggestion
        v-for="(item, index) in props.options"
        :key="index"
        v-bind="getOptionProps(item)"
      >
```
Please consistent with #21545

kooriookami (changes requested):

packages/components/radio/src/radio-group.vue
```diff
@@ -7,7 +7,20 @@
...
+        <template v-if="typeof item.render === 'function'">
+          {{ item.render(item, index) }}
+        </template>
```
Dsaquel: I think we can keep the consistency with the other options/props for now:

```suggestion
```

packages/components/radio/src/radio-group.vue
```diff
@@ -65,6 +83,19 @@ const name = computed(() => {
...
+  ...props.props,
+}))
+const getOptionProps = (option: Record<string, any>) => {
+  const base = {
+    label: option[aliasProps.value.label],
+    value: option[aliasProps.value.value],
+    disabled: option[aliasProps.value.disabled],
+  }
+  return { ...option, ...base }
+}
```
Dsaquel, line 90: Same remark as https://github.com/element-plus/element-plus/pull/21545/files#r2309488335.

```suggestion
  return {
    label: option[aliasProps.value.label],
    value: option[aliasProps.value.value],
    disabled: option[aliasProps.value.disabled],
  }
}
```

packages/components/radio/src/radio-group.vue
snowbitx: > I think we can keep the consistency with the other options/props for now:

Got it, thanks!

packages/components/radio/src/radio-group.vue
snowbitx: > Same remark as https://github.com/element-plus/element-plus/pull/21545/files#r2309488335.

This modification is intended to make it more flexible to support additional properties, as shown in the image. The radio component follows the same pattern.
image

packages/components/radio/src/radio-group.ts
snowbitx: > Please let import type { xxx } a single line.

Updated, thanks.

packages/components/radio/src/radio-group.vue
Dsaquel: Ok that's good, we have to mention it in documentation then.

We can be inspired by the cascader component:
image

snowbitx: > Ok that's good, we have to mention it in documentation then.
> 
> We can be inspired by the cascader component:

Thanks for the feedback 🙏. I think it’s sufficient to just update the description of options to indicate support for more attributes. As for props, customizing the three — label, value, and disabled — should be enough. What do you think?

Dsaquel: Ok sounds good to me, it feels a bit hazy but we'll gonna polish the documentation based on the user report.

Dsaquel (approved):

packages/components/radio/src/radio-group.vue
snowbitx: > Ok sounds good to me, it feels a bit hazy but we'll gonna polish the documentation based on the user report.

I agree, we can keep it consistent with the checkbox for now and refine later with user feedback. Thanks!

btea (approved):
LGTM

kooriookami (approved):

PR #21545, opened 2025-07-29
feat(components): [checkbox-group] support options

snowbitx:


warmthsea:
Can refer to this PR to change.

- #21543

jw-foss:
I think [this](https://github.com/element-plus/element-plus/pull/21543#discussion_r2268708052) genuinely applies to this PR as well

packages/components/checkbox/src/checkbox-group.vue
```diff
@@ -9,7 +9,15 @@
...
     :aria-labelledby="isLabeledByFormItem ? formItem?.labelId : undefined"
   >
-    <slot />
+    <slot>
+      <el-checkbox
+        v-for="(item, index) in props.options"
+        :key="index"
+        :value="item[props.props?.value ?? 'value']"
+        :label="item[props.props?.label ?? 'label']"
+        :disabled="item[props.props?.disabled ?? 'disabled']"
```
kooriookami: Could you please use `v-bind` like the usage of el-select?
You can refer to packages/components/select/src/select.vue

packages/components/checkbox/src/checkbox-group.ts
```diff
@@ -1,6 +1,7 @@
 import { UPDATE_MODEL_EVENT } from '@element-plus/constants'
 import { useAriaProps, useSizeProp } from '@element-plus/hooks'
 import { buildProps, definePropType, isArray } from '@element-plus/utils'
+import { SelectOptionProps } from '@element-plus/components/select/src/select'
```
kooriookami: It's weird to import the SelectOptionProps. You can define a CheckboxOptionProps.

kooriookami (changes requested):

packages/components/checkbox/src/checkbox-group.vue
snowbitx: Changes have been made as suggested, thanks! 🙌

packages/components/checkbox/src/checkbox-group.vue
```diff
@@ -20,8 +26,10 @@ import { CHANGE_EVENT, UPDATE_MODEL_EVENT } from '@element-plus/constants'
 import { debugWarn } from '@element-plus/utils'
 import { useNamespace } from '@element-plus/hooks'
 import { useFormItem, useFormItemInputId } from '@element-plus/components/form'
+import { useProps } from '@element-plus/components/select-v2/src/useProps'
```
kooriookami: Please don't use functions by other components like `select-v2`. You can write like this:
```
const aliasProps = computed(() => ({ ...defaultProps, ...props.props }))

const getOptionProps = (option: Record<string, any>) => ({
  label: option[aliasProps.value.label],
  value: option[aliasProps.value.value],
  disabled: option[aliasProps.value.disabled],
})
```

kooriookami (changes requested):

packages/components/checkbox/src/checkbox-group.vue
snowbitx: > Please don't use functions by other components like `select-v2`. You can write like this:
> 
> ```
> const aliasProps = computed(() => ({ ...defaultProps, ...props.props }))
> 
> const getOptionProps = (option: Record<string, any>) => ({
>   label: option[aliasProps.value.label],
>   value: option[aliasProps.value.value],
>   disabled: option[aliasProps.value.disabled],
> })
> ```

Updated as per the latest PR, please have a look.

kooriookami:
Using options doesn't seem quite right. Checkboxes have far more than just these attributes. Unless all options include these attributes
image

snowbitx:
> Using options doesn't seem quite right. Checkboxes have far more than just these attributes. Unless all options include these attributes image

You’re absolutely right. I’ve adjusted the [code ](https://github.com/element-plus/element-plus/pull/21545/files#diff-8727357bbafafab8e7c26783c38260e3dd9d7f688c8c4488304f3fa9bb691c5eR71-R78)to support additional fields. Let me know if you’d like any further changes. Thanks!

docs/en-US/component/checkbox.md
```diff
@@ -61,6 +61,14 @@ checkbox/grouping
 
 :::
 
+## Options attribute ^(2.11.0)
```
kooriookami: Please update the version.

packages/components/checkbox/src/checkbox-group.ts
```diff
@@ -54,6 +54,13 @@ export const checkboxGroupProps = buildProps({
     type: Boolean,
     default: true,
   },
+  options: {
+    type: definePropType<CheckboxOption[]>(Array),
+  },
+  optionProps: {
+    type: definePropType<CheckboxOptionProps>(Object),
+    default: () => checkboxDefaultProps,
+  },
```
kooriookami: @element-plus/backers `props` or `optionProps`, which is better?

kooriookami (changes requested):

packages/components/checkbox/src/checkbox-group.ts
Dsaquel: I vote for `props`, the select component has also this configuration.

snowbitx: I've updated the code to use props as suggested, so that it's consistent with the select component. Thanks.😊

kooriookami (approved):
LGTM

btea (approved):
LGTM

packages/components/checkbox/src/checkbox-group.vue
```diff
@@ -53,6 +64,19 @@ const modelValue = computed({
...
+  ...checkboxDefaultProps,
+  ...props.props,
+}))
+const getOptionProps = (option: Record<string, any>) => {
+  const base = {
+    label: option[aliasProps.value.label],
+    value: option[aliasProps.value.value],
+    disabled: option[aliasProps.value.disabled],
+  }
+  return { ...option, ...base }
```
Dsaquel, line 77: We can now optimize this:
```suggestion
  return {
    label: option[aliasProps.value.label],
    value: option[aliasProps.value.value],
    disabled: option[aliasProps.value.disabled],
  }
```

kooriookami: Other options should be included.

Dsaquel: Thanks, didn't notice it.

Then we should update the documentation, currently it mention only label, value and disabled.

We can refer to cascader to mention the props (or partial props allowed):

Image

snowbitx: > Thanks, didn't notice it.
> 
> Then we should update the documentation, currently it mention only label, value and disabled.
> 
> We can refer to cascader to mention the props (or partial props allowed):
> 
> Image

Thank you for your feedback 🙏. Personally, I think it’s sufficient to just support more configuration options through options here. As for prop alias configuration, keeping the commonly used three — label, value, and disabled — should already cover most needs without making things too bloated. What do you think?

Dsaquel: Ok sounds good to me, it feels a bit hazy but we'll gonna polish the documentation based on the user report.

Dsaquel (approved):

PR #21562, opened 2025-07-31
ci: cleanup community bot messages

Dsaquel:


.github/workflows/pr-welcome.yml
```diff
@@ -24,5 +24,3 @@ jobs:
           labels: 'Contribution::Community'
           token: ${{ github.token }}
           pr-number: ${{ github.event.pull_request.number }}
-          greeting-message: 'Hello %user%, thank you for contributing to %repo%, please see our [guideline](%guideline%) to see how to make contribution'
-          greeting-guideline-address: 'https://github.com/element-plus/element-plus/blob/dev/CONTRIBUTING.md'
```
Dsaquel, line 28: This one happens all the time, IMO it should be one time for first time contribution.

btea (approved):

.github/workflows/pr-welcome.yml
btea: That makes sense. Sometimes format checking limits the content of the expression and the page will be modified.

thinkasany:
@Dsaquel 

When you create it for the first time, you will get ``` Error: Validation Failed: {"resource":"IssueComment","code":"unprocessable","field":"data","message":"Body cannot be blank"} ```

https://github.com/element-plus/element-plus/actions/runs/16712942656/job/47300919110?pr=21605

image

Dsaquel:
Oh thanks i will push a fix :+1: .

PR #21565, opened 2025-07-31
style(components): remove unused icon class & autocomplete tweaks

Dsaquel:
- removed `.el-icon-loading` as it is not used. `.is-loading` is the right one.
- it now enable the class vertical-align of autocomplete (very slight difference):
Peek 2025-07-31 23-36
- cursor to default in the popper when the autocomplete is loading.

### DIFF
--- a/packages/components/autocomplete/__tests__/autocomplete.test.tsx
+++ b/packages/components/autocomplete/__tests__/autocomplete.test.tsx
@@ -279,9 +279,9 @@ describe('Autocomplete.vue', () => {
     vi.runAllTimers()
     await nextTick()
 
-    expect(document.body.querySelector('.el-icon-loading')).toBeDefined()
+    expect(document.body.querySelector('.el-icon.is-loading')).toBeDefined()
     await wrapper.setProps({ hideLoading: true })
-    expect(document.body.querySelector('.el-icon-loading')).toBeNull()
+    expect(document.body.querySelector('.el-icon.is-loading')).toBeNull()
   })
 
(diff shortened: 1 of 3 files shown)


packages/components/autocomplete/__tests__/autocomplete.test.tsx
```diff
@@ -279,9 +279,9 @@ describe('Autocomplete.vue', () => {
     vi.runAllTimers()
     await nextTick()
 
-    expect(document.body.querySelector('.el-icon-loading')).toBeDefined()
+    expect(document.body.querySelector('.el-icon.is-loading')).toBeDefined()
```
Dsaquel, line 282: The test was passing because `querySelector` returned `null`. `null` !== `undefined`.

packages/theme-chalk/src/autocomplete.scss
```diff
@@ -82,8 +83,10 @@
       }
     }
 
-    & .#{$namespace}-icon-loading {
-      vertical-align: middle;
+    @include b(icon) {
+      @include when(loading) {
+        vertical-align: middle;
+      }
     }
```
warmthsea: I think it’s OK to just delete these codes, the original ones feel more natural.

 | Before | Now |
 | ------- | ----- |
 |      image      |   image     |


```suggestion
```

Dsaquel: Yeah it's too far in the middle :+1:

warmthsea (approved):

PR #21579, opened 2025-08-01
refactor(components): [message] use `MessageType` replace `messageType`

thinkasany:
Follow TypeScript type naming conventions, use PascalCase, and provide backwards-compatible deprecated types.


warmthsea (approved):
LGTM, Thanks!

PR #21583, opened 2025-08-01
fix(components): [table] table row-key meaningful or use index

yyismoon:

If the row-key is set, getRowIdentity may return the string undefined or null. If the row-key of the child node is repeated, rendering problems may occur when the data is updated.

https://github.com/user-attachments/assets/1d9e8a84-3b26-4aeb-8748-75120ae7be42

Now it will not return the string undefined or null and repeated keyRow, but will return the index, which can be rendered correctly

https://github.com/user-attachments/assets/1a92332d-c199-43a9-bcec-6a48b35db6aa


yyismoon:
before, `getRowIdentity` can return `’undefined‘ ` `’null‘`, now if `row[rowKey]` is meaningless, return `’‘` for next action.

in rowRender, if can not get `getRowIdentity`, return index.

and check key of row is unique, else return index

packages/components/table/src/util.ts
```diff
@@ -187,7 +187,7 @@ export const getRowIdentity = <T extends DefaultRow>(
   if (!row) throw new Error('Row is required when get row identity')
   if (isString(rowKey)) {
     if (!rowKey.includes('.')) {
-      return `${row[rowKey]}`
+      return row[rowKey] ? `${row[rowKey]}` : ''
```
thinkasany:
```suggestion
      return row[rowKey] ?? ''
```

thinkasany: I'm not sure if this is an optimization. At least it looks simpler and I hope it doesn't bother you.

thinkasany:
Carrying irrelevant commits

image

packages/components/table/src/util.ts
```diff
@@ -196,7 +196,7 @@ export const getRowIdentity = <T extends DefaultRow>(
     }
     //TODO: "current" is now any, we just satisfies typecheck here
     // but this function can actually return a number
-    return `${current}`
+    return current ? `${current}` : ''
```
thinkasany, line 199:
```suggestion
    return current ?? ''
```

yyismoon: Using `? ? `may return a number or Boolean value. I hope getRowIdentity will continue to return a string. I have some new ideas about this issue. I will resubmit the PR. Thank you!

PR #21586, opened 2025-08-01
refactor(components): [tour] improve code style

thinkasany:
1. Simplify code and use insets
2. Increase caching
3. Enhance types

### DIFF
--- a/packages/components/tour/src/mask.vue
+++ b/packages/components/tour/src/mask.vue
@@ -58,7 +58,10 @@ const path = computed(() => {
 
 const maskStyle = computed<CSSProperties>(() => ({
   position: 'fixed',
-  inset: 0,
+  left: 0,
+  right: 0,
+  top: 0,
+  bottom: 0,
   zIndex: props.zIndex,
   pointerEvents: props.pos && props.targetAreaClickable ? 'none' : 'auto',
 }))
--- a/packages/components/tour/src/mask.vue
+++ b/packages/components/tour/src/mask.vue
@@ -1,18 +1,5 @@
 <template>
-  <div
-    v-if="visible"
-    :class="ns.e('mask')"
-    :style="({
-    position: 'fixed',
-    left: 0,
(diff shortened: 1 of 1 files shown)


Dsaquel (approved):
LGTM

packages/components/tour/src/mask.vue
```diff
@@ -69,13 +56,18 @@ const path = computed(() => {
...
-const pathStyle = computed<CSSProperties>(() => {
-  return {
-    fill: props.fill,
-    pointerEvents: 'auto',
-    cursor: 'auto',
-  }
-})
+const maskStyle = computed<CSSProperties>(() => ({
+  position: 'fixed',
+  inset: 0,
```
btea: Image

Image

Due to compatibility issues, it is better to use the original style.

thinkasany: image

image

The splitter mask is also resolved using `inset`.

I believe that if compatibility issues do arise, users should adjust the Vite configuration to make the output compatible, rather than the library using the least compatible syntax as a fallback.

btea: I didn't notice this before. 

Of course, users can use tools to package and convert, but to avoid unnecessary problems, we should ensure that the default code is consistent with the official website.

Could you modify the code elsewhere?

thinkasany: I'll first adjust the tour back to top, bottom, left, and right. If `inset: 0` is also needed elsewhere, you can give me a definite answer and I'll complete it.

btea: `segmented` and `splitter`, you can open a new PR to handle this issue.

btea (approved):
Thanks!

packages/components/tour/src/mask.vue
thinkasany: I have created a new PR.

https://github.com/element-plus/element-plus/pull/21601

PR #21601, opened 2025-08-04
style(components): Replace inset to provide better compatibility

thinkasany:
https://github.com/element-plus/element-plus/pull/21586#discussion_r2250004729


btea (approved):
Thanks!

PR #21605, opened 2025-08-04
refactor(components): [tour] improve TourContext type

thinkasany:


thinkasany:
@btea  Hi, could you please review this PR?

packages/components/tour/src/step.ts
```diff
@@ -37,6 +42,7 @@ export const tourStepProps = buildProps({
    */
   closeIcon: {
     type: iconPropType,
+    default: () => CloseComponents.Close,
```
btea: https://github.com/element-plus/element-plus/blob/382d393020f1d0953d851e85564e26baf025bf2c/packages/components/tour/src/step.vue#L117
There is a fallback here, so there is no need to give a default value.

thinkasany: Thanks.  I've completed the repair.

btea (approved):
Thanks!

warmthsea (approved):

PR #21608, opened 2025-08-04
feat(components): [drawer] add `resizable` prop

rzzf:

This PR implements the solution using Splitter, as previously discussed. I've completed the implementation, but feel free to decide whether to adopt this version, as I'm not entirely sure if using Splitter may introduce any side effects.

[Related Comment](https://github.com/element-plus/element-plus/pull/21595#pullrequestreview-3082404972)


packages/theme-chalk/src/drawer.scss
```diff
@@ -78,39 +79,22 @@ $directions: rtl, ltr, ttb, btt;
...
 }
 
 .#{$namespace}-drawer-fade {
+  &-enter-from,
+  &-leave-from,
+  &-enter-active,
+  &-leave-active {
+    .#{$namespace}-splitter-bar__dragger::before {
+      background-color: transparent;
+    }
```
Dsaquel: This class can be externalized in order to not having the 2px bar before hovering.

Before:
Image

After:
Image

Dsaquel (commented):
I like this solution more, i'll try investigate if this will have unexpected side effects.

packages/theme-chalk/src/drawer.scss
rzzf: done

rzzf:
The currently known issues: #21517, ~~#21607~~

packages/theme-chalk/src/drawer.scss
Dsaquel: I think we can override the base 2px handle bar to 1px.
image

docs/examples/drawer/resizable-drawer.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <el-button plain @click="visible = true">Open a resizable Drawer</el-button>
+  <el-drawer v-model="visible" resizable>
```
thinkasany: It looks like ```cursor``` will be obscured by `mask`.




https://github.com/user-attachments/assets/39f27822-20b6-4af6-9e80-d025eb4f1973

thinkasany: I think we can add direction selection to make it easier to debug the effects in different directions. What do you think?

image

rzzf: > It looks like `cursor` will be obscured by `mask`.


It works fine in both the documentation and the Playground. Since I don’t have a Windows machine, please help me fix it if any issues occur on that platform. Thanks you.

packages/theme-chalk/src/drawer.scss
rzzf: > This class can be externalized in order to not having the 2px bar before hovering.

I think it should be hidden during the transition, because if the mouse happens to be at the same position while opening or closing, the following situation may occur. What’s your take on this?

rzzf: > I think we can override the base 2px handle bar to 1px. 

Done, Thanks.

docs/examples/drawer/resizable-drawer.vue
rzzf: > I think we can add direction selection to make it easier to debug the effects in different directions. What do you think?

Done, Thanks.

thinkasany: Indeed, the performance on my Mac is also correct. I will go back to the company tomorrow to see what the reason is.

docs/examples/drawer/resizable-drawer.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <el-radio-group v-model="direction">
+    <el-radio value="ltr">left to right</el-radio>
```
thinkasany: I think it would be more convenient to bind it directly to Button.Group or Segmented, like naive, so that the drawer can be opened directly.

But don't change it yet, as other members may not agree with it. Let's see what everyone thinks.

docs/examples/drawer/resizable-drawer.vue
thinkasany: https://stackoverflow.com/questions/54562080/is-it-possible-to-force-set-cursor-col-resize-with-pure-css?utm_source=chatgpt.com

https://github.com/user-attachments/assets/2a6ab8d9-6685-41fa-8199-b4aaf1b6a652

I saw a similar question on StackOverflow, and I'm sure it's a Windows problem, but I'm not sure about the solution. Maybe I need to listen to other suggestions.

packages/theme-chalk/src/drawer.scss
Dsaquel: >I think it should be hidden during the transition, because if the mouse happens to be at the same position while opening or closing, the following situation may occur. What’s your take on this?

Yeah definitely we should :+1: .

docs/examples/drawer/resizable-drawer.vue
Dsaquel: I agree with you, i vote for Segmented component 😁.

@cszhjh I see you got inspired by the basic example. perhaps we can change it to segmented too in another PR.

thinkasany: basic is a combination of footer configuration and orientation options. It seems not suitable for Segmented.

But here it should be possible.

rzzf: I’ll implement it later using Segmented. Thanks for your suggestions.

Dsaquel: @thinkasany You're right :+1: .

rzzf: Please check again, Thanks.

Dsaquel (approved):
LGTM, Thanks!

rzzf:
I removed the default direction from the example because, with the current interaction, opening the default rtl drawer requires three steps: selecting another direction, closing it, and then clicking `rtl` again, which feels cumbersome.😄

btea (approved):
Thanks!

warmthsea:
I don't think the drawer needs to rely on a **splitter** for dragging, for the following reasons:

1. It adds several layers of unnecessary splitter elements when not in use.
2. Relying on a splitter impacts performance.
3. It causes many other issues.

What do you think?

rzzf:
I agree, he is currently causing a series of issues.😥

thinkasany:
> I don't think the drawer needs to rely on a **splitter** for dragging, for the following reasons:
> 
> 1. It adds several layers of unnecessary splitter elements when not in use.
> 2. Relying on a splitter impacts performance.
> 3. It causes many other issues.
> 
> What do you think?

There are some issues on Windows.

https://github.com/element-plus/element-plus/pull/21608?notification_referrer_id=NT_kwDOBwSz7LUxNzk5Mzk0MjM5MzoxMTc3NDg3MTY&notifications_query=is%3Aunread#discussion_r2278108191

Dsaquel:
>1. It adds several layers of unnecessary splitter elements when not in use.

I agree with you on this one, we can create a sub component to get around that.

>2. Relying on a splitter impacts performance.
>3. It causes many other issues.

It's clear that the splitter component is still new and under testing, having it under the drawer component increase the issues backlog but I think we would have these issues sooner or later anyway. Fixing them sooner will make the component more stable IMO.

Other point:
Among the fact that we don't duplicate the resize logic, with the splitter component we can easily inherit from the feature of the splitter component like min, max, etc...

warmthsea:
I've created a new PR with the core implementation.

https://github.com/element-plus/element-plus/commit/f5a3a81648bd67e8f47b6d5cfcb540d6c21d1f30
- https://github.com/element-plus/element-plus/pull/21932

screenshots


Note that the direction-related code and tests haven't been added yet.

Alixhan:
现在这个抽屉组件底部在展开和收起会出现滚动条，并且无法通过min-width和max-width动态设置宽度，一堆bug出来，是这个pr导致的吗？
https://github.com/element-plus/element-plus/issues/21891#issuecomment-3227110666

Dsaquel:
>I've created a new PR with the core implementation.

Thanks, taking a step back, it is the most reasonable thing to do right now.

bpuns:
This is really bad. Without setting the resizable property, can't it still be displayed as it was before? This caused all my css selectors to become invalid

rzzf:
> This is really bad. Without setting the resizable property, can't it still be displayed as it was before? This caused all my css selectors to become invalid

this is already addressed in #21932

PR #21626, opened 2025-08-05
feat(components): [el-cascader] add closeOnSelected prop

oooah:
add closeOnSelected prop, whether to close panel when select the node, only valid when multiple=false and checkStrictly=false.
 This PR solve #21351


Dsaquel (commented):
Thanks for your contribution,

But i think a more suitable way and general prop to do this is to follow the v-model:visible implementation: https://github.com/element-plus/element-plus/pull/8762.

PR #21628, opened 2025-08-05
fix(components):[el-input-number] change default 'align' prop to 'left'

wyh369352887:
When the align prop is not supported, the default alignment is left-justified. I believe developers who previously needed center alignment would have already implemented corresponding styles themselves. However, after adding center alignment as the new default, those who don’t need it will now be forced to explicitly configure align="left".

fix [#21291](https://github.com/element-plus/element-plus/pull/21291)


btea:

This configuration was recently added. To ensure that the new configuration does not affect existing users, the default value is set to `center`. Therefore, we will not change this default value at this time.

PR #21634, opened 2025-08-05
refactor(components): [message-box] remove duplicates conditional

thinkasany:
- Remove duplicate conditionals: The original code's `{{ !dangerouslyUseHTMLString ? message : '' }} `is redundant because the outer conditional already has a `v-if="!dangerouslyUseHTMLString"` conditional.

- Use the v-text directive: Change {{ message }} to v-text="message" for a cleaner, more concise look.

- Keep the logic clear: Use v-text for safe rendering and v-html for HTML rendering.


btea (approved):
Thanks!

Dsaquel (approved):

PR #21642, opened 2025-08-05
docs(components): [table] supplement `columns` type

warmthsea:
### Rel:

- https://github.com/element-plus/element-plus/pull/21139

### Description

In fact, `columns` are also often used.


Dsaquel (approved):
👍

Dsaquel:
I just occurs that T will stay as `any` since we need to upgrade to vue `3.3`.
That said it is not really a matter.

PR #21668, opened 2025-08-07
feat(components): [color-picker-panel] externalize `color-picker` core component

Dsaquel:
**Mini breaking change**: Changed the related `dropdown` classes mentioned in color picker panel.
image


btea:
In the page example, perhaps you can add an `alpha` and `predefineColors ` example like the color-picker.

packages/components/color-picker/src/color-picker.vue
```diff
@@ -109,21 +93,12 @@
...
-} from 'vue'
-import { debounce } from 'lodash-unified'
-import { ElButton } from '@element-plus/components/button'
+import { computed, mergeProps, nextTick, provide, ref, watch } from 'vue'
+import { debounce, pick } from 'lodash-unified'
 import { ElIcon } from '@element-plus/components/icon'
 import { ClickOutside as vClickOutside } from '@element-plus/directives'
 import { ElTooltip } from '@element-plus/components/tooltip'
-import { ElInput } from '@element-plus/components/input'
+import ElButton from '@element-plus/components/button'
```
tolking:
```suggestion
import { ElButton } from '@element-plus/components/button'
```

It is recommended to keep consistent with the previous code style

packages/components/color-picker/src/color-picker.vue
```diff
@@ -143,23 +118,25 @@ import {
...
-  colorPickerContextKey,
-  colorPickerEmits,
-  colorPickerProps,
-} from './color-picker'
-
+import { colorPickerEmits, colorPickerProps } from './color-picker'
+import ColorPickerPanel, {
+  ROOT_COMMON_COLOR_INJECTION_KEY,
+  colorPickerPanelProps,
+} from '@element-plus/components/color-picker-panel'
```
tolking:
```suggestion
import {
  ElColorPickerPanel,
  ROOT_COMMON_COLOR_INJECTION_KEY,
  colorPickerPanelProps,
} from '@element-plus/components/color-picker-panel'
```

It feels better to keep consistent with the ElTooltip, ElIcon

packages/components/color-picker/src/color-picker.vue
```diff
@@ -17,50 +17,34 @@
     @hide="setShowPicker(false)"
   >
     <template #content>
-      <div
+      <color-picker-panel
+        ref="pickerPanel"
+        v-bind="mergeProps(passsPanelProps, $props)"
```
tolking: The props of the ElColorPickerPanel have been filtered out (passsPanelProps). Why merge all $props values here?

packages/components/color-picker/src/color-picker.vue
```diff
@@ -168,17 +145,20 @@ const { formItem } = useFormItem()
 const colorSize = useFormSize()
 const colorDisabled = useFormDisabled()
 const { valueOnClear, isEmptyValue } = useEmptyValues(props, null)
-
+const commonColor = useCommonColor(props, emit)
 const { inputId: buttonId, isLabeledByFormItem } = useFormItemInputId(props, {
   formItemContext: formItem,
 })
+const { color } = commonColor
```
tolking: I noticed that the ElColorPickerPanel component has already exported the color, perhaps it can be directly used through `pickerPanel.value?.color` to avoid mutual influence between the two components before

packages/components/color-picker/src/color-picker.vue
```diff
@@ -143,23 +118,25 @@ import {
...
+import { useCommonColor } from '@element-plus/components/color-picker-panel/src/composables/use-common-color'
+
+import type { ColorPickerPanelInstance } from '@element-plus/components/color-picker-panel'
 import type { TooltipInstance } from '@element-plus/components/tooltip'
 
 defineOptions({
   name: 'ElColorPicker',
 })
 const props = defineProps(colorPickerProps)
+mergeProps
```
tolking:
```suggestion
```

It is like meaningless code.

packages/components/color-picker/src/color-picker.vue
```diff
@@ -168,17 +145,20 @@ const { formItem } = useFormItem()
...
 })
+const { color } = commonColor
 
-const hue = ref<InstanceType<typeof HueSlider>>()
-const sv = ref<InstanceType<typeof SvPanel>>()
-const alpha = ref<InstanceType<typeof AlphaSlider>>()
 const popper = ref<TooltipInstance>()
 const triggerRef = ref()
-const inputRef = ref()
+const pickerPanel = ref<ColorPickerPanelInstance>()
```
tolking:
```suggestion
const pickerPanelRef = ref<ColorPickerPanelInstance>()
```

packages/components/color-picker-panel/src/color-picker-panel.vue
```diff
@@ -0,0 +1,130 @@
...
+
+defineExpose({
+  /**
+   * @description current color object
+   */
+  color,
+  /**
+   * @description custom input ref
+   */
+  customInput: inputRef,
```
tolking: I feel that using inputRef directly here is more in line with naming conventions

packages/components/color-picker/src/color-picker.vue
Dsaquel: Okay, i added an computed for a fallback into `common.color`, in `:persistent="false"` `pickerPanelRef` will be undefined.

packages/components/color-picker/src/color-picker.ts
```diff
@@ -105,7 +105,7 @@ export type ColorPickerEmits = typeof colorPickerEmits
 export type ColorPickerInstance = InstanceType<typeof ColorPicker> & unknown
 
 export interface ColorPickerContext {
-  currentColor: ComputedRef<string>
+  color: Color
 }
 
 export const colorPickerContextKey: InjectionKey<ColorPickerContext> = Symbol(
```
warmthsea: Can these codes be deleted? They seem to be of no use.

Dsaquel:
@warmthsea I saw you removed `class="box-content"` but there were added to prevent this in the documentation.

image

Mmh taking a step back it should added in the style file.

warmthsea (commented):
I noticed the spacing looked a bit weird and some code wasn't working.

image

<hr />

How about changing it like this?


https://github.com/element-plus/element-plus/pull/21668/files#diff-2b67c528d0e02e7bfc677ba0517450d8015eae18c79ce8e9066b5f5b7aabf984R139-R150

```scss
  @include e(panel) {
    box-sizing: content-box;
    box-shadow: map.get($dropdown, 'menu-box-shadow');
  }
}

.#{$namespace}-color-picker__panel.#{$namespace}-popper {
  padding: 5px;
}
```

image

kooriookami:
image

Please replace the margin with gap.

warmthsea:
> I noticed the spacing looked a bit weird and some code wasn't working.

or https://github.com/element-plus/element-plus/blob/feat/color-picker-panel/packages/components/color-picker/src/color-picker.vue#L3 use `pure`.

I've found that when using it in production, using automatic imports can cause styling differences.

Full imports will load **popper.scss** , while automatic imports using picker alone will not.

Dsaquel:
Thanks for the attention, didn't notice the padding difference. Using pure will indeed solve the problem.
For the other styles of popper i don't see any weird overriding style.
image

Dsaquel:
>How about changing it like this?
>   ```scss
> @include e(panel) {
>     box-sizing: content-box;
>     box-shadow: map.get($dropdown, 'menu-box-shadow');
>   }
> ```

For this i just let the minimal style that are effective in a build where popper scss is not in.
And i remove `box-sizing` since color-picker-panel.scss handle it.
image

packages/theme-chalk/src/color-picker-panel.scss
```diff
@@ -0,0 +1,283 @@
...
+
+  @include e(color-selector) {
+    width: 20px;
+    height: 20px;
+    border-radius: 4px;
+    cursor: pointer;
+
+    &:nth-child(10n + 1) {
+      margin-left: 0;
+    }
```
kooriookami: Need remove.

kooriookami:
Do you think the padding 12px is better? And I add a border-radius 4px.
image
image

Dsaquel:
>Do you think the padding 12px is better? And I add a border-radius 4px.

Yes we will breath a lot more with extra space, i update it.

kooriookami (approved):

btea (approved):

warmthsea (approved):

PR #21684, opened 2025-08-08
feat(components): [config-provider] support `dialog` config

zhixiaotong:

I personally think some dialog attributes are commonly used, so I committed this pr.

https://github.com/user-attachments/assets/691efa1b-902a-48ed-8e7b-ca0ba5d5db51


packages/components/dialog/src/dialog.vue
```diff
@@ -140,7 +141,17 @@ provide(dialogInjectionKey, {
...
+  () => (globalConfig.value?.draggable || props.draggable) && !props.fullscreen
+)
+
+const overlayAlignCenter = computed(
+  () => globalConfig.value?.alignCenter || props.alignCenter
+)
+
+const overflowViewport = computed(
+  () => globalConfig.value?.overflow || props.overflow
+)
```
rzzf: Perhaps `props` should take precedence over `globalConfig`.

Consider this scenario: `globalConfig.value.overflow = true`, but the current component instance requires `props.overflow = false`.

We might take the Button implementation as a [reference](https://github.com/element-plus/element-plus/blob/2010ed02522805b89103345108cad7961c341fb9/packages/components/button/src/use-button.ts#L48).

zhixiaotong: Yes, indeed. After submitting it, I realized this issue while walking. Logically, props should have higher priority, and what's more, the `boolean` type is different. I'll adjust it when I have time later.

zhixiaotong: thx for the review

zhixiaotong:
@Dsaquel plz take a look, test sometimes failed due to `Avatar`.

packages/components/dialog/src/dialog-content.vue
```diff
@@ -59,8 +61,12 @@ const dialogKls = computed(() => [
...
 const composedDialogRef = composeRefs(focusTrapRef, dialogRef)
 
-const draggable = computed(() => props.draggable)
-const overflow = computed(() => props.overflow)
+const draggable = computed(
+  () => props.draggable ?? globalConfig.value?.draggable ?? false
+)
+const overflow = computed(
+  () => props.overflow ?? globalConfig.value?.overflow ?? false
+)
```
Dsaquel: I feel like the received props have already the calculated logic by `useDialog.ts` → `dialog.vue` → `dialog-content.vue`.

docs/en-US/component/config-provider.md
```diff
@@ -93,6 +101,7 @@ In this section, you can learn how to use Config Provider to provide experimenta
...
 | link                    | link related configuration, [see the following table](#link-attribute)                                                                                                 | ^[object]`{type?: string, underline?: boolean \| string}`                                                                                                                                                                                                      | see the following table                                                                |
+| dialog                  | dialog related configuration, [see the following table](#dialog-attribute)                                                                                             | ^[object]`{alignCenter?: boolean, draggable?: boolean, overflow?: boolean, transition?: DialogTransition}`                                                                                                                                                     | see the following table                                                                |
```
Dsaquel:
```suggestion
| dialog  ^(2.11.0)                | dialog related configuration, [see the following table](#dialog-attribute)                                                                                             | ^[object]`{alignCenter?: boolean, draggable?: boolean, overflow?: boolean, transition?: DialogTransition}`                                                                                                                                                     | see the following table                                                                |
```

docs/en-US/component/config-provider.md
```diff
@@ -120,6 +129,15 @@ In this section, you can learn how to use Config Provider to provide experimenta
 | ---------------- | ------------------------- | --------------------------------- | ------- |
 | shadow ^(2.10.5) | when to show card shadows | ^[enum]`always \| never \| hover` | —       |
 
+### Dialog Attribute
+
+| Attribute    | Description                                                                                                                    | Type                                   | Default |
+| ------------ | ------------------------------------------------------------------------------------------------------------------------------ | -------------------------------------- | ------- |
+| align-center | whether to align the dialog both horizontally and vertically                                                                   | ^[boolean]                             | false   |
```
Dsaquel: These attributes too can be bumped to `^(2.11.0)`

packages/components/dialog/src/dialog-content.vue
zhixiaotong: yeah...i made a mistake~

packages/components/dialog/src/dialog.vue
```diff
@@ -121,6 +121,9 @@ const {
   rendered,
   transitionConfig,
   zIndex,
+  _draggable,
+  _alignCenter,
+  _overflow,
```
rzzf, line 126: Just a small suggestion.
I don’t think the `_` is necessary, as it looks a bit strange. When there’s a variable in `setup` with the same name as a `prop`, Vue will "prioritize" the one defined in `setup` in the template.

Example: [Vue Playground](https://play.vuejs.org/#eNp9UsFOwzAM/RWTS4c0tUJwqsokQJOAA0wMiUsuVet2GWkSJekYqvrvuCkrQ5p2s5+f7ee8dOzOmHjXIktZ5gorjAeHvjULrkRjtPXQgcUKeqisbiAiasRVoZXzYKw2Dm6hxEooXA3ZrOMKoHF1CiEC8N8GU1h7K1Q9HyHi5630KURhQkx0mgnQc9VfcnUYTzANp+Wz6BGl1PChrSwvIqJkyaiVVFLisTEy90gZQLa5WnRdaO77LKEsoGYAp3WhNNyYJUfNbM68o92VqOOt04reJBzBWaEbIyTaV+MFaeNsOo+znKR9PQfM2xZ/b6SeDRafJ/Ct2w8YZyuLDu0OOZtqPrc1+rG8XL/gnuKp2OiylcQ+U3xDp2U7aBxp960qSfYRL6h9Cs6SIe9uufeo3OGoQWhwIvA5I7cfzpz+J/c6vjk4SK849Jz8VOd8K8UuBBQ6qT0kI5yM+H+j+h9N8ud8)

image

Dsaquel: Interesting, nice catch :+1:

packages/components/dialog/src/dialog-content.vue
Dsaquel: No worries 😃.

packages/components/dialog/src/dialog.vue
zhixiaotong: Although this is the case, there will be error messages in some versions of `volar`, and variables with the same name in props may give people an illusion. I personally recommend a semantic and obvious way of writing. The underscore is written with reference to `use-button.ts`, and the previous names are `viewportOverflow`, `dialogDraggable`...

rzzf: Sorry I didn't notice that volar may cause errors

zhixiaotong: @cszhjh Thanks for your review and suggestion. @Dsaquel WDYT? Is it necessary to adjust the variable naming?

zhixiaotong: > Sorry I didn't notice that volar may cause errors

I really didn't pay much attention to the internal handling of vue, thank you for pointing it out. ^^

Dsaquel: We can keep underscores :+1: .

btea (approved):
Overall, it looks good, but the documentation still needs to add the corresponding version tag, as @Dsaquel  said.

zhixiaotong:
> Overall, it looks good, but the documentation still needs to add the corresponding version tag, as @Dsaquel said.

Oh... I forgot to modify it yesterday, now it's done~

Dsaquel (approved):
Thanks!

Dsaquel:
Okay but in the example there are a custom transition ant it seems to work's well with draggable :thinking:, can you provide a custom transition causing the issue ?

zhixiaotong:
> Okay but in the example there are a custom transition ant it seems to work's well with draggable 🤔, can you provide a custom transition causing the issue ?

I think my previous description is not accurate, I think it is a problem, but in fact this problem also exists on other component libraries. Both antd vue and naive ui use js to handle animations, and do not support custom animations, so this problem is reflected. In arco, there is also this problem. When you open a pop-up window that can be dragged and pulled, you can directly close the animation. If you drag and close it, there is no animation. I don't know what caused this for the time being. It may be caused by both `animation` and `draggable` handling `transform`.

### element-plus


https://github.com/user-attachments/assets/cb771729-7344-4cbb-92c1-ccdc54cad47e



### arco design

See https://arco.design/vue/component/modal#draggable

https://github.com/user-attachments/assets/585b0ce5-8310-487f-a4d7-826a5723465c



Another documentation issue, the demo of the transition object simulates the pop-up style similar to antd and naive, but I didn't consider the impact of draggable before, resulting in the inaccurate target of draggable shrinkage after dragging, I need to submit a pr to make up for it (I had already found that I was about to modify the state of pr, but found that pr was merged.)

### Before


https://github.com/user-attachments/assets/91b245ae-a02b-4418-82c3-a82aecf5d7ed


### After


https://github.com/user-attachments/assets/bae5000f-b025-4fa9-8c45-61ac754e07b1

zhixiaotong:
The drag-and-drop transform conflicts with the CSS animated transform.
The transform: scale (0.3) translateY (-50px) of the CSS animation completely overrides the drag-and-drop transform, a similar situation would not occur if transiton did not use the transform, but the second issue, I think I need to fix it, may mislead others.
#21695

Dsaquel:
I see now the issue, if both (transition, draggable) use the same div to apply `transform` it will be indeed hard to make a fix. I'll look forward your pr's.

PR #21701, opened 2025-08-11
fix(components): [input] form email validation

### LINKED ISSUE #20706: [Component] [input] form email 验证不正确
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.13`
- Element Plus Version: `2.9.10`
- Browser / OS: `any`
- Build Tool: `Vite`

### Related Component

- `el-input`

### Reproduction Link

[Docs](https://element-plus.org/zh-CN/component/form.html)

### Steps to reproduce

E-mail：aaa@aaa.com

### What is Expected?

识别非email字符

不识别

### Additional comments

_(empty)_

Image

allenxu09:

This PR is to fix #20706 by patching async-validator regex.

## Summary by CodeRabbit

* **Bug Fixes**
  * Email validation is now case-insensitive and more accurate.
  * Character-length calculation now correctly handles extended Unicode characters (supplementary code points).

* **Chores**
  * Applied an updated async-validator patch across distribution builds to deliver these fixes.

### DIFF
--- a/patches/async-validator@4.2.5.patch
+++ b/patches/async-validator@4.2.5.patch
@@ -1,7 +1,16 @@
 diff --git a/dist-node/index.js b/dist-node/index.js
-index 563b966fe2b4dce9583c6d90ceb996215c1104bc..f81c89359f4566b7a1e7b6228227fc858e026670 100644
+index 563b966fe2b4dce9583c6d90ceb996215c1104bc..15ddac46c35e9c4e644ed894d89ef2578aff4095 100644
 --- a/dist-node/index.js
 +++ b/dist-node/index.js
+@@ -457,7 +457,7 @@ var getUrlRegex = (function () {
+ 
+ var pattern$2 = {
+   // http://emailregex.com/
+-  email: /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF]+\.)+[a-zA-Z\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF]{2,}))$/,
++  email: /^(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])$/i,
+   // url: new RegExp(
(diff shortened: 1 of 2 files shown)


allenxu09:
@rzzf ptal! :D

rzzf:
> @rzzf ptal! :D

Thanks for the ping. I'm not very familiar with this area, so I'll wait for others to review.

patches/async-validator@4.2.5.patch
```diff
@@ -1,7 +1,16 @@
...
+index 563b966fe2b4dce9583c6d90ceb996215c1104bc..15ddac46c35e9c4e644ed894d89ef2578aff4095 100644
 --- a/dist-node/index.js
 +++ b/dist-node/index.js
+@@ -457,7 +457,7 @@ var getUrlRegex = (function () {
+ 
+ var pattern$2 = {
+   // http://emailregex.com/
+-  email: /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF]+\.)+[a-zA-Z\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF]{2,}))$/,
++  email: /^(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])$/i,
+   // url: new RegExp(
```
btea, line 11: ~~Why did you submit the comments as well?~~
It looks like it's from the original repository, but it hasn't been modified. Why is it in the diff results?

allenxu09: ~i think it is modified?~
Oh I get what you mean

allenxu09: I think the patch is correct. Maybe you can compare this file across the 2 commits?

allenxu09: The diff view is kind of confusing.

Before my commit:
image

After my commit:
image

allenxu09: hope this will be clearer

btea: https://emailregex.com/

The modified regular expression is quite complex; we can put its source in the comments. Incidentally, we can add a test case for #20706.

allenxu09: Sure, I will work on that. At the same time, I believe the source is already there in the diff, though the regex from this source doesn't take capital letters into account. Should this be mentioned in comments too?

btea: I don't think this needs special explanation. The URL above has specific standard instructions.

btea: I've noticed that the revised rules don't seem to support Chinese characters in email addresses. For example, `张@gmail.com` was supported before the change, but it's not now.

As the website mentioned earlier stated, there is no 100% match regular expression; one that satisfies most situations is sufficient. Perhaps adding custom validation functions to the project would be more convenient and reasonable for handling special cases.

allenxu09: True indeed. Nevertheless cases such as `张@gmail.com` are very rare, and inclusion of Chinese characters (or more widely, the non-ASCII characters) will eventually lead to some form of trade-off. Hence, I think it would be reasonable to replace the rule for now to exempt special situations?

btea: Yes, I also think it's rare. But we can't assume it doesn't exist. If someone relies on the default validation rules to match these cases, our change would be a breaking change for them. 

So, for now, I still prefer to maintain the status quo.

allenxu09: OK, then let’s keep it as is for now. Thanks!

btea: In any case, thank you very much for your contribution. ❤️

PR #21710, opened 2025-08-11
chore: remove renovate bot

Dsaquel:
Talking about notifications in https://github.com/element-plus/element-plus/pull/21704.

We're not using the bot and he cause many undesirable notifications.

image


btea:
I think we can keep this. In fact, it has been running, but we did not update the dependent version in time. After the #21704 merge, its update will not send email notifications to others. 😄

PR #21725, opened 2025-08-12
chore: update crowdin translate link

warmthsea:
image

### Now

In fact, there is no specific page located. It is very messy and I don’t know where it is.
https://crowdin.com/translate/element-plus/all/en-zhcn

image

### After

https://crowdin.com/project/element-plus/zh-CN

image


btea (approved):
LGTM

PR #21747, opened 2025-08-13
feat(components): [message] add `placement` option & method

zhixiaotong:

`arco` and `naive-ui` have similar support. `naive-ui` also supports `'top-left'\'top-right'\'bottom-left'\'bottom-right'`. 

see discussions: https://github.com/element-plus/element-plus/discussions/21542

https://github.com/user-attachments/assets/8754dcf5-7ae9-4806-97ae-d349b68e95d6


David-Xian66:
Yeah that's what I'm saying

zhixiaotong:
@btea Please take a look, thank you, and help me re-run the test task, which keeps reporting some errors. 😀

btea:
Now that placement is supported, perhaps we should support more values, such as `top-left`/`top-right` , etc.

zhixiaotong:
> Now that placement is supported, perhaps we should support more values, such as `top-left`/`top-right` , etc.

Do you think these four additional location attributes need to be added?

btea:
IMO, since two values are supported, it is inevitable that someone will want to support other values in the future. 😃

zhixiaotong:
Okay, i'll add the rest placement options.

zhixiaotong:
done~

zhixiaotong:
Done, it should be noted that `top-left`, `top-right`, `bottom-left`, `bottom-right` are just different positions, but will not change the effect of the original message up or down animation display, and notification left and right moving animation to distinguish. @btea Plz review again, thx. ^^

packages/components/message/src/message.vue
```diff
@@ -14,6 +14,8 @@
         { [ns.m(type)]: type },
         ns.is('closable', showClose),
         ns.is('plain', plain),
+        ns.is('bottom', placement.startsWith('bottom')),
```
Dsaquel: **nit**:

```suggestion
        ns.is('bottom', verticalProperty === 'bottom'),
```

packages/components/message/src/instance.ts
```diff
@@ -12,9 +12,24 @@ export type MessageContext = {
...
 
-export const getInstance = (id: string) => {
+/**
+ * Get or create message instances array for specified placement on demand
+ * @param placement
+ * @returns placement instances
+ */
+export const getOrCreatePlacementInstances = (placement: MessagePlacement) => {
+  if (!placementInstances[placement]) {
+    placementInstances[placement] = shallowReactive<MessageContext[]>([])
```
Dsaquel:
```suggestion
    placementInstances[placement] = shallowReactive([])
```

packages/components/message/src/instance.ts
```diff
@@ -12,9 +12,24 @@ export type MessageContext = {
...
+export const placementInstances = shallowReactive<
+  Record<MessagePlacement, MessageContext[]>
+>({} as Record<MessagePlacement, MessageContext[]>)
 
-export const getInstance = (id: string) => {
+/**
+ * Get or create message instances array for specified placement on demand
+ * @param placement
+ * @returns placement instances
+ */
```
Dsaquel: I think we can delete the comment here.
```suggestion
```

packages/components/message/src/instance.ts
```diff
@@ -12,9 +12,24 @@ export type MessageContext = {
   props: Mutable<MessageProps>
 }
 
-export const instances: MessageContext[] = shallowReactive([])
+export const placementInstances = shallowReactive<
+  Record<MessagePlacement, MessageContext[]>
+>({} as Record<MessagePlacement, MessageContext[]>)
```
Dsaquel: **nit**:

```suggestion
export const placementInstances = shallowReactive({} as Record<MessagePlacement, MessageContext[]>)
```

packages/theme-chalk/src/message.scss
```diff
@@ -16,15 +16,25 @@
...
-  transition: opacity getCssVar('transition-duration'), transform 0.4s, top 0.4s;
+  transition: opacity getCssVar('transition-duration'), transform 0.4s;
   padding: getCssVar('message', 'padding');
   display: flex;
   align-items: center;
   gap: 8px;
 
+  &.is-left {
+    left: 16px;
+    transform: none;
```
Dsaquel: I inspected the style and it seems like for the horizontal classes we override the 'default center' style.
If you can make the `left` and `transform` opt-in for the center style it would be perfect.

Dsaquel (commented):
Very great feature, added just some nitpicks.

zhixiaotong:
> Very great feature, added just some nitpicks.

I'll make the changes later. If I don't have time, I'll do it tomorrow.

packages/theme-chalk/src/message.scss
zhixiaotong: I'm not entirely sure if I fully understood your meaning, I've already submitted my changes. Please let me know if there's anything incorrect. Also, since I'm unaware of the planned next version number, I kept it as `2.11.0` in the commit.

packages/theme-chalk/src/message.scss
```diff
@@ -15,16 +15,28 @@
...
   align-items: center;
   gap: 8px;
 
+  &.is-left {
+    left: 16px;
+    right: auto;
+  }
+
+  &.is-right {
+    left: auto;
```
Dsaquel: I think the left/right `auto` declaration can be now removed as there are no more properties to override.

packages/theme-chalk/src/message.scss
```diff
@@ -90,5 +102,20 @@
 .#{$namespace}-message-fade-enter-from,
 .#{$namespace}-message-fade-leave-to {
   opacity: 0;
-  transform: translate(-50%, -100%);
-}
+  
+  &:not(.is-center) {
```
Dsaquel: Let's try to be more explicit:

```suggestion
  &:is(.is-left, .is-right) {
```

Dsaquel (commented):
One last thing:
This property really useful, can you set it in config-provider ?

packages/theme-chalk/src/message.scss
Dsaquel: You did more than i expected, thanks you :+1:.

Yes, we can keep `2.11.0` for now.

packages/theme-chalk/src/message.scss
zhixiaotong: sure

zhixiaotong:
> One last thing: This property really useful, can you set it in config-provider ?

Is it added to this pr, or another pr? Should it be merged directly into this pr?

zhixiaotong:
@Dsaquel Done. I'm a bit unsure that the default configuration of other messages has a value, but I set the placement default to `undefined` and explicitly declared the default value `'top'` in the `config-provider`. Whether it meets your expectations is different from the existing merging methods of other message configurations. I think that even if other positions (such as `'bottom'`) are declared in the global configuration, if placement: `'top'` is passed during actual calls, `'top'` should be prioritized over the global `'bottom'`. WDYT @btea 

In addition, Avatar test case failed again. I will try to fix this problem. I really don't want to see this error anymore. 😢

```
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  packages/components/avatar/__tests__/avatar.test.tsx > Avatar.vue > image fallback
AssertionError: expected '' to be 'fallback' // Object.is equality

- Expected
+ Received

- fallback

 ❯ packages/components/avatar/__tests__/avatar.test.tsx:59:28
     57|     wrapper.emitted('error') && expect(wrapper.emitted('error')).toBeD…
     58|     await nextTick()
     59|     expect(wrapper.text()).toBe('fallback')
       |                            ^
     60|     expect(wrapper.find('img').exists()).toBe(false)
     61|   })

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯

 Test Files  1 failed | 157 passed (158)
```

btea:
> @Dsaquel Done. I'm a bit unsure that the default configuration of other messages has a value, but I set the placement default to `undefined` and explicitly declared the default value `'top'` in the `config-provider`. Whether it meets your expectations is different from the existing merging methods of other message configurations. I think that even if other positions (such as `'bottom'`) are declared in the global configuration, if placement: `'top'` is passed during actual calls, `'top'` should be prioritized over the global `'bottom'`. WDYT @btea

Yes, the parameters configured during the call have higher priority than the global configuration.

> In addition, Avatar test case failed again. I will try to fix this problem. I really don't want to see this error anymore. 😢

This test case fails randomly and is indeed a bit annoying.

packages/components/message/src/message.ts
```diff
@@ -23,7 +23,19 @@ export const messageTypes = [
...
+export const messagePlacement = [
+  'top',
+  'top-left',
+  'top-right',
+  'bottom',
+  'bottom-left',
+  'bottom-right',
+] as const
+
+export const MESSAGE_DEFAULT_PLACEMENT = messagePlacement[0]
```
Dsaquel:
```suggestion
export const MESSAGE_DEFAULT_PLACEMENT = 'top'
```

It will result to the same type.

packages/theme-chalk/src/message.scss
Dsaquel: They are still here.

packages/components/message/src/message.ts
```diff
@@ -34,6 +46,7 @@ export interface MessageConfigContext {
   offset?: number
   showClose?: boolean
   plain?: boolean
+  placement?: MessagePlacement
```
Dsaquel: The user will must cast the value, i think it's fine in config provider to set this as string.
```suggestion
  placement?: string
```

zhixiaotong: i forgot it...

packages/theme-chalk/src/message.scss
zhixiaotong: It should be done this time~

packages/components/config-provider/__tests__/config-provider.test.tsx
```diff
@@ -401,6 +402,7 @@ describe('config-provider', () => {
       })
       const overrideConfig = reactive({
         max: 1,
+        placement: 'bottom-left' as MessagePlacement,
```
Dsaquel:
```suggestion
        placement: 'bottom-left',
```

docs/examples/config-provider/message.vue
```diff
@@ -10,11 +10,15 @@
 import { reactive } from 'vue'
 import { ElMessage } from 'element-plus'
 
+import type { MessagePlacement } from 'element-plus'
+
 const config = reactive({
   max: 3,
   plain: true,
+  placement: 'bottom' as MessagePlacement,
```
Dsaquel: Now the casting can be removed.

```suggestion
  placement: 'bottom',
```

Dsaquel (approved):
Thanks!

btea:
|before|after|
|---|---|
|message|message-new|

When multiple messages fade away, the modified message appears to lose its transition when moving up.

zhixiaotong:
> before	after
> message	message-new
> When multiple messages fade away, the modified message appears to lose its transition when moving up.

I'll check it out.

zhixiaotong:
https://github.com/user-attachments/assets/2bd0511a-efa9-494f-8465-7a7047d0fcb3

done @btea Avatar failed again, it seems that my previous pr only reduced the probability of failure but did not completely solve it. 😔

btea (approved):
Thanks!

Issue #21749, opened 2025-08-13
[Component] [date-picker] el-date-picker 选择月份后,正常,浏览器切换到其他标签再切换回来,el-date-picker再一次展示选择页面


Djooxx:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.18`
- Element Plus Version: `2.10.7`
- Browser / OS: `chrome  138.0.7204.184 / windows 11`
- Build Tool: `Vite`

### Related Component

- `el-date-picker`

### Reproduction Link

[...]

PR #21761, opened 2025-08-15
fix(components): [date-picker] window focus causes the panel to pop up

ouyang108:
Fix the time picker to automatically expand when switching browser tabs while it is focused

[#21749]


btea:
This doesn't seem to be a correct fix, as it causes many test cases to fail.

PR #21766, opened 2025-08-15
feat(components): [tour] support keyboard switching steps

thinkasany:
This will help us better experience feature previews, as the focus points of different steps can be far apart, making switching with the mouse somewhat cumbersome.

Being able to use the keyboard would certainly be a significant convenience.

### DIFF
--- a/packages/components/tour/src/step.vue
+++ b/packages/components/tour/src/step.vue
@@ -156,7 +156,7 @@ const onClose = () => {
 
 const handleKeydown = (e: KeyboardEvent) => {
   const target = e.target as HTMLElement | null
-  if (!!target && target.isContentEditable) return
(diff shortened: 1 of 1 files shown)


packages/components/tour/src/step.vue
```diff
@@ -152,4 +153,28 @@ const onClose = () => {
   tourOnClose()
   emit('close')
 }
+
+const handleKeydown = (e: KeyboardEvent) => {
+  const target = e.target as HTMLElement | null
+  if (!!target && target.isContentEditable) return
```
btea:
```suggestion
  if (target?.isContentEditable) return
```

btea (approved):
Looks Great.

Dsaquel (approved):
Nice, thanks!

PR #21777, opened 2025-08-15
feat(components): add global locale

thinkasany:
Some locales are universal and do not need to be encapsulated into components.

image

### DIFF
--- a/packages/locale/lang/en.ts
+++ b/packages/locale/lang/en.ts
@@ -1,6 +1,10 @@
 export default {
   name: 'en',
   el: {
+    global: {
+      placeholder: 'Please select',
+      close: 'Close',
+    },
     breadcrumb: {
       label: 'Breadcrumb', // to be translated
     },
--- a/packages/components/tour/src/step.vue
+++ b/packages/components/tour/src/step.vue
@@ -1,7 +1,7 @@
 <template>
   <button
     v-if="mergedShowClose"
-    aria-label="Close"
+    :aria-label="t('el.global.close')"
     :class="ns.e('closebtn')"
     type="button"
     @click="onClose"
--- a/packages/locale/lang/af.ts
+++ b/packages/locale/lang/af.ts
@@ -1,6 +1,10 @@
 export default {
   name: 'af',
   el: {
+    global: {
+      placeholder: 'Please select',
+      close: 'Close',
+    },
     breadcrumb: {
       label: 'Breadcrumb', // to be translated
     },
--- a/packages/locale/lang/ar-eg.ts
+++ b/packages/locale/lang/ar-eg.ts
@@ -1,6 +1,10 @@
 export default {
   name: 'ar-eg',
   el: {
+    global: {
+      placeholder: 'إختر',
+      close: 'أغلق',
+    },
     breadcrumb: {
       label: 'Breadcrumb', // to be translated
     },
--- a/packages/locale/lang/ar.ts
+++ b/packages/locale/lang/ar.ts
@@ -1,6 +1,10 @@
 export default {
   name: 'ar',
   el: {
+    global: {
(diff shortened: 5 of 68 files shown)


packages/locale/lang/en.ts
```diff
@@ -1,6 +1,10 @@
 export default {
   name: 'en',
   el: {
+    global: {
```
Dsaquel, line 4: This file have only keys as component name, i think we should stay as is and add the close translation in the tour object https://github.com/thinkasany/element-plus/blob/142863925ef195bf01cdf5736962f25ad7718098/packages/locale/lang/en.ts#L154.

thinkasany: You're right, but `global` is actually a common practice. Its advantage is that it allows for reuse across many components. Otherwise, maintaining a copy of each component, coupled with the large number of languages, would be prohibitively expensive.

This approach is also used in other large repositories.

https://github.com/ant-design/ant-design/blob/master/components/locale/en_US.ts#L16

Dsaquel: What you say make sense to me, actually we only have the tour component that need to be applied this translation, do you plan continuing the refactor for the other global properties ?

thinkasany: If we consider supporting this, we could replace several of the existing component usages.

The purpose of creating this pull request is to encourage discussion about whether to use global properties for management.

Dsaquel: Okay let's see what other maintainers think about it.

cc @element-plus/backers

btea: Using global has both advantages and disadvantages.

The good thing is that it is easy to manage, as you said. The bad thing is that one value is used in multiple components. Changing one of them will affect numerous components and may cause unknown effects.

thinkasany: The original purpose of this pull request was to add ```aria-label="Close"```. I'm sure you'll encounter similar locales in the future, as they're incredibly common.

Maintaining such a large number of locales would be quite stressful if I had to rewrite them every time I went into the component.

If the translations aren't complex and don't differ significantly, I think global translations can be used. Since the component is already used, I'll leave it as is for now.

thinkasany: My expectation is:

First, create the global variable. Later, developers can choose whether to use component-level or global language configuration. For example, in a11y, global is sufficient for most languages.

This doesn't directly replace existing components, so there won't be any breaking changes.

This will save some maintenance costs in the future, as global will likely suffice.

btea: Thanks for the detailed explanation, it makes sense to me. Go for it.

thinkasany: Maintaining so many languages doesn't seem like an easy task. 

Even with machine translation, there's always the possibility of errors, so I can only translate a portion of it and use English as a fallback for the rest. 

If there is a better way, please help me.

tolking: If we use global language configuration, users won't be able to configure different values for different components, which doesn't feel good

thinkasany: It should be able to cover most scenarios, such as close Button, or a11y. If it is very special, you can also create a component-level locale separately. They can coexist, which does not mean that all of them need to be replaced with this.

tolking: The description of the close button should be as detailed as possible, otherwise the user will not know what to close

thinkasany: https://github.com/element-plus/element-plus/pull/21934

Thanks for your comments and the synchronous language script. I've resubmitted a tour as you requested, but I still retain the suggestion to provide global, because I find that most of the uses of close are actually the same.


image

PR #21820, opened 2025-08-19
chore: update start docs port

warmthsea:
Port **5173** is hard to remember, and I often forget it, so I changed it to 2000.

### DIFF
--- a/docs/en-US/guide/dev-guide.md
+++ b/docs/en-US/guide/dev-guide.md
@@ -20,7 +20,7 @@ the project will install all dependencies.
 With command
 
 ```shell
-pnpm docs:dev
+pnpm docs
 ```
 
 the project will launch website for you to preview all existing component.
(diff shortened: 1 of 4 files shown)


Dsaquel (approved):
LGTM

docs/en-US/guide/dev-guide.md
```diff
@@ -20,7 +20,7 @@ the project will install all dependencies.
 With command
 
 ```shell
-pnpm docs:dev
+pnpm docs
```
btea: Executing `pnpm docs` will cause problems, opening the page `https://www.npmjs.com/package/undefined` in the browser.

docs/en-US/guide/dev-guide.md
```diff
@@ -32,7 +32,7 @@ See [Local development guide](https://github.com/element-plus/element-plus/blob/
 1. With command
 
 ```shell
-pnpm dev
+pnpm run dev
```
btea: I think `pnpm dev` is more concise, so it's better to keep the original one. `pnpm docs` also conflicts with existing commands, so using `docs:dev` seems more reasonable.

PR #21835, opened 2025-08-21
test: fix undefined equal & ignore anchor slot warning

zhixiaotong:
https://github.com/element-plus/element-plus/pull/21489#issuecomment-3208878652

### DIFF
--- a/packages/components/anchor/__tests__/anchor.test.tsx
+++ b/packages/components/anchor/__tests__/anchor.test.tsx
@@ -1,9 +1,27 @@
 import { nextTick, ref } from 'vue'
 import { mount } from '@vue/test-utils'
-import { describe, expect, test, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest'
 import { CHANGE_EVENT } from '@element-plus/constants'
 import Anchor from '../src/anchor.vue'
 import AnchorLink from '../src/anchor-link.vue'
+
+// ignore the slot warning
+const originalWarn = console.warn
+beforeEach(() => {
+  console.warn = (message: string, ...args: any[]) => {
+    if (
+      typeof message === 'string' &&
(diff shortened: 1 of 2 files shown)


packages/components/image/__tests__/image.test.tsx
```diff
@@ -282,7 +282,7 @@ describe('Image.vue', () => {
       }
       await flushPromises()
       // expect no new error event to be emitted
-      expect(wrapper.emitted('error')?.length).toBe(errorCountBefore)
+      expect(wrapper.emitted('error')?.length || 0).toBe(errorCountBefore)
```
Dsaquel, line 285: It is a test case so i think it's fine to return a potential undefined here.

packages/components/anchor/__tests__/anchor.test.tsx
```diff
@@ -1,10 +1,28 @@
...
 import Anchor from '../src/anchor.vue'
 import AnchorLink from '../src/anchor-link.vue'
 
+// ignore the slot warning
+const originalWarn = console.warn
+beforeEach(() => {
+  console.warn = (message: string, ...args: any[]) => {
+    if (
+      typeof message === 'string' &&
+      message.includes('Slot "default" invoked outside of the render function')
```
Dsaquel, line 14: In fact i saw many tests case have warnings but we don't see them until a test fail in their file. I think it is fine to ignore them for now.

packages/components/image/__tests__/image.test.tsx
zhixiaotong: Cause errorCountBefore was processed before, if it is undefined, it will take 0, resulting in unequality and fail.

packages/components/anchor/__tests__/anchor.test.tsx
zhixiaotong: okay, i got it

PR #21857, opened 2025-08-22
ci: add locale:sync scripts

tolking:

Every time a multilingual field is added, it needs to be manually added to other languages, which is too tedious. Add a script to automatically synchronize the content of `en.ts` files to other languages

```shell
pnpm locale:sync
```

### DIFF
--- a/scripts/sync-locale.ts
+++ b/scripts/sync-locale.ts
@@ -5,21 +5,21 @@ import consola from 'consola'
 import { isArray, isObject, isString } from 'lodash-unified'
 import { localeRoot, normalizePath } from '@element-plus/build-utils'
 
-interface LocaleObject {
-  [key: string]: LocaleObject | string | Array<string>
+type TranslatePair = {
+  [key: string]: string | string[] | TranslatePair
 }
 
 const langRoot = resolve(localeRoot, 'lang')
 const enFile = normalizePath(resolve(langRoot, 'en.ts'))
 const localePath = normalizePath(resolve(`${langRoot}/!(en).ts`))
 const commentInfo = '// to be translated'
+const untranslatedRexExp = new RegExp(
+  `\\w+:(\\s+|\\n\\s+).*,?\\s*${commentInfo}$`,
+  'gm'
+)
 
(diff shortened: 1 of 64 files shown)


Dsaquel (approved):
Looks great 😃 .

Dsaquel:
Looks like more a `feat` than a `ci` commit type?

tolking:
The feat tag will be added to the Change log, and users who normally use components will not care about this PR

scripts/sync-locale.ts
```diff
@@ -0,0 +1,243 @@
...
+}
+
+const langRoot = resolve(localeRoot, 'lang')
+const enFile = normalizePath(resolve(langRoot, 'en.ts'))
+const localePath = normalizePath(resolve(`${langRoot}/!(en).ts`))
+const commentInfo = '// to be translated'
+
+function parseLocaleFile(content: string): LocaleObject {
+  try {
+    const untranslatedRexExp = new RegExp(
```
btea: This regular expression can be put outside, without having to create it every time the function is called.

btea (changes requested):

scripts/sync-locale.ts
```diff
@@ -0,0 +1,243 @@
...
+  consola.info(`Found ${localeFiles.length} locale files to process`)
+
+  localeFiles.forEach((filePath) => {
+    const fileName = basename(filePath)
+
+    consola.start(`Processing ${fileName}...`)
+
+    try {
+      const content = readFileSync(filePath, 'utf-8')
+      const targetObject = parseLocaleFile(content)
```
btea, line 209: Maybe we can load the file content directly through `import(filePath)`.


```suggestion
      const targetObject = (await import(filePath)).default
```

tolking: Because I cleared the field with `commentInfo` inside the `parseLocaleFile` function, otherwise we would lose the comment `// to be translated`.

btea: Indeed, I seem to have missed the logic. You deleted all the fields to be translated in the `parseLocaleFile` function.

btea (approved):

PR #21887, opened 2025-08-24
fix(components): [input-tag] add missing Tooltip style

rzzf:
Since `style/index.ts` and `style/css.ts` did not import the Tooltip styles, the related Tooltip styles were missing when importing InputTag on demand.

image

[Reproduction Repo](https://github.com/cszhjh/ep-input-tag-demo)


Dsaquel (approved):
Nice catch 👍 .

warmthsea (approved):
Thanks!

PR #21906, opened 2025-08-26
refactor(components): [drawer] improve slot template

thinkasany:
- Readability: Reduces repeated conditional checks, resulting in cleaner templates.

- Performance: Avoids repeated evaluation of the same condition.
image

In fact, it should be possible to just delete the ```v-if="!$slots.title"``` inside the span, but it seems clearer to use `template`, and it is also done in other places.

image


tolking (approved):

btea (approved):
Thanks!

PR #21911, opened 2025-08-27
fix(components): [drawer] resolve width issue when min/max is set via external style

rzzf:

fix [Issue comment](https://github.com/element-plus/element-plus/issues/21891#issuecomment-3227110666)

### DIFF
--- a/packages/components/drawer/src/drawer.vue
+++ b/packages/components/drawer/src/drawer.vue
@@ -27,6 +27,7 @@
           @release-requested="onCloseRequested"
         >
           <el-splitter
+            ref="splitterRef"
             :class="ns.b('splitter')"
             :layout="isHorizontal ? 'horizontal' : 'vertical'"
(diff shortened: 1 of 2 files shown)


rzzf:
@Dsaquel Hello, I'm not sure if handling the `min/max` in the component's style internally is appropriate. Maybe we should expose `min` and `max` as props instead? What do you think?

Dsaquel:
>Hello, I'm not sure if handling the min/max in the component's style internally is appropriate. Maybe we should expose min and max as props instead? What do you think?

Yes it's more reasonable :+1:.

rzzf:
> Yes it's more reasonable 👍.

Currently, the splitter doesn't consider min and max when initializing its width; it only applies them during dragging, which could be a potential issue. If we decide to use props, I’ll try to incorporate this logic on top of #21548.

PR #21913, opened 2025-08-27
fix(components): date-picker set persistent to false, display bug

oooah:
When persistent is set to false and modelValue has a value, the initialization display has a bug


FrontEndDog:
Thanks!
The `persistent` attribute is not provided in the document, also, I don't think it's necessary to add it.
If it needs to be added, the document needs to be updated.
cc @element-plus/backers

Dsaquel:
Agreed :+1:

PR #21925, opened 2025-08-28
docs: improve navbar link jump

thinkasany:
Use the <a/> tag to jump to the page with the right mouse button or the middle mouse button, making it easier for developers to open a new window for bilingual comparison.

image

Vue is like this, it is very comfortable to use.

https://cn.vuejs.org/

image

### DIFF
--- a/docs/.vitepress/vitepress/components/navbar/vp-translation.vue
+++ b/docs/.vitepress/vitepress/components/navbar/vp-translation.vue
@@ -23,9 +23,15 @@ const toTranslation = () => {
               v-for="l in langs"
               :key="l"
               :class="{ language: true, selected: l === lang }"
-              @click="switchLang(l)"
             >
-              {{ languageMap[l] }}
+              <a
+                :href="`/${l}${
+                  router.route.path.replace(/^\/[^\/]+/, '') || '/'
+                }`"
+                @click.prevent="switchLang(l)"
+              >
+                {{ languageMap[l] }}
+              </a>
             </ElDropdownItem>
             <ElDropdownItem class="language selected" @click="toTranslation">
               {{ locale.help }}


btea:
> Use the tag to jump to the page with the right mouse button or the middle mouse button, making it easier for developers to open a new window for bilingual comparison.

Thanks for your contribution.

But I think this is subjective. Some people like to open multiple windows for comparison, but many others simply switch to content that's easier for them to understand. In these cases, I think opening a new window is unnecessary. It just leads to a proliferation of windows open simultaneously, which is quite cumbersome.

thinkasany:
> > Use the tag to jump to the page with the right mouse button or the middle mouse button, making it easier for developers to open a new window for bilingual comparison.
> 
> Thanks for your contribution.
> 
> But I think this is subjective. Some people like to open multiple windows for comparison, but many others simply switch to content that's easier for them to understand. In these cases, I think opening a new window is unnecessary. It just leads to a proliferation of windows open simultaneously, which is quite cumbersome.

So I used ```@click.prevent``` to prevent the default redirection of the <a> element (i.e., it won't refresh the page or redirect to the URL specified by the href ).

When you left-click, only the front-end method is used to switch the language without refreshing the page; but right-click/middle-click/ctrl+click can still use the href of <a> to open a new tab.

docs/.vitepress/vitepress/components/navbar/vp-translation.vue
```diff
@@ -23,9 +23,15 @@ const toTranslation = () => {
               v-for="l in langs"
               :key="l"
               :class="{ language: true, selected: l === lang }"
-              @click="switchLang(l)"
             >
-              {{ languageMap[l] }}
+              <a
+                :href="`/${l}${
+                  router.route.path.replace(/^\/[^\/]+/, '') || '/'
```
btea, line 29: This calculation logic is more suitable for integration with the implementation of switchLang.

btea:
> So I used `@click.prevent` to prevent the default redirection of the element (i.e., it won't refresh the page or redirect to the URL specified by the href ).
> 
> When you left-click, only the front-end method is used to switch the language without refreshing the page; but right-click/middle-click/ctrl+click can still use the href of to open a new tab.

This is more reasonable. 👍

docs/.vitepress/vitepress/components/navbar/vp-translation.vue
thinkasany: Any code suggestions? I think the current implementation is more reasonable. If I write a switchLang function, I'll have to add mouse click events to determine whether the middle button is clicked or the left and right buttons are clicked. I feel that the href capability of the a tag is the simplest and most semantically compliant.

tolking:

This issue appears to have already been addressed by other PR changes.

Issue #18155, opened 2024-09-04
[Component] [checkbox] el-checkbox's large size is the same size as default size


RoachLin:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.0`
- Element Plus Version: `2.8.1`
- Browser / OS: `UserAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/115.0`
- Build Tool: `Vite`

### Related Component

- `el-checkbox`

### Reproduction Link

[Docs](https://element-plus.org/zh-CN/component/checkbox.html#%E5%9F%BA%E7%A1%80%E7%94%A8%E6%B3%95)

### Steps to reproduce

https://element-plus.org/zh-CN/component/checkbox.html#%E5%9F%BA%E7%A1%80%E7%94%A8%E6%B3%95
when you check the doc, you can easily find the el-checkbox's large size is the same size as default size

### What is Expected?

el-checkbox's large size is bigger than the default size

### What is actually happening?

el-checkbox's large size is the same size as default size

### Additional comments

_(empty)_

betavs:
<img width="1440" alt="image" src="https://github.com/user-attachments/assets/ec73b6c6-f201-4104-b635-5fe4fbd7ab01">

Maybe `size` only acts on `height` at the beginning of the design.

warmthsea:
参考
https://github.com/element-plus/element-plus/issues/14689

Xaypanya:
+1

Issue #14689, opened 2023-10-30
[Component] [checkbox] el-checkbox size属性设置large无效，宽高还是default的默认值


1zxca:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.3.2`
- Element Plus Version: `2.3.9`
- Browser / OS: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36`
- Build Tool: `Vite`

### Related Component

- `el-checkbox`

### Reproduction Link

[...]

PR #21933, opened 2025-08-28
style(components): [radio] large radio should be larger than the default

### LINKED ISSUE #21924: [Component] [radio] large单选框和常规单选框一样大小
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.20`
- Element Plus Version: `2.11.1`
- Browser / OS: `mac`
- Build Tool: `Vite`

### Related Component

- `el-radio`

### Reproduction Link

[Docs](https://element-plus.org/zh-CN/component/radio.html)

### Steps to reproduce

官方文档错误

### What is Expected?

large的radio应该比常规的大

large的radio和常规的一样的大

### Additional comments

_(empty)_

yosong-github:

fix #21924

### DIFF
--- a/packages/theme-chalk/src/radio.scss
+++ b/packages/theme-chalk/src/radio.scss
@@ -9,7 +9,7 @@
 $radio-font-size: () !default;
 $radio-font-size: map.merge(
   (
-    'large': 14px,
+    'large': 16px,
     'small': 12px,
   ),
   $radio-font-size


Dsaquel:
I think this need more further discussion, i assume it is already designed as is from the beginning.

related issues:
https://github.com/element-plus/element-plus/issues/14689
https://github.com/element-plus/element-plus/issues/18155

yosong-github:
> I think this need more further discussion, i assume it is already designed as is from the beginning.
> 
> related issues: #14689 #18155

Do you have any good ideas?

yosong-github:
> I think this need more further discussion, i assume it is already designed as is from the beginning.
> 
> related issues: #14689 #18155



> I think this need more further discussion, i assume it is already designed as is from the beginning.
> 
> related issues: #14689 #18155

I have looked at similar component libraries, such as Ant Design. It seems that their regular single-select buttons do not support changing the size. Moreover, it is also stated in the documentation. Should we specifically mention this in the documentation as well?

Dsaquel:
Many components have the font-size to 14px in large size (e.g. input related components). So changing the font-size large to 16px for one component will differ with other sizable components.

That said, at the moment, we should not modifying anything or thinking to increase the font-size large to 16px globally as it will break to much components.

btea:
I agree with @Dsaquel, the design is like this, we can't make changes easily. Users can override it with variables if they have special needs. refer to https://element-plus.org/en-US/guide/theming.html#how-to-override-it

warmthsea:
In fact, when comparing it to all other UI libraries, you'll find that element-plus `checkbox` and `radio` components are indeed on the small side.

https://ui.shadcn.com/docs/components/checkbox
https://mui.com/material-ui/react-checkbox
https://vuetifyjs.com/en/components/checkboxes
https://reka-ui.com/docs/components/checkbox
https://ant.design/components/checkbox-cn


I think changing to the following sizes would be a better approach (**the font size doesn't need to change**):

small: ~~12px~~ -> 14px
default: ~~14px~~ -> 16px
large: ~~14px~~ -> 18px

However, this will affect related components, such as `tree` and `table` components...

The current `table checkbox` is indeed too small and should definitely be increased.
For `tree` components, `14px` is better to coordinate with the text...

This may require a lot of changes. If you're interested, feel free to continue revising it. If not, I'll submit a pull request soon.

PR #21036, opened 2025-06-15
style(eslint-config): newline before `import type`

Dsaquel:
We usually have this type of import:
```ts
import {} from ''

import type {} from ''
```

This eslint config will transform import that has no break line between `import` and `import type`.
And if it's not the case, it will regroup split import block's together.

---

#### Before
```ts
import {} from ''
import type {} from ''
```

#### After
```ts
import {} from ''

import type {} from ''
```
---

#### Before
```ts
import {} from ''

import {} from ''
import {} from ''
```

#### After
```ts
import {} from ''
import {} from ''
import {} from ''
```
---

Migrated to [eslint-plugin-import-x](https://github.com/un-ts/eslint-plugin-import-x) because [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import) has not been release since october and the feature [newlines-between-types](https://github.com/import-js/eslint-plugin-import/pull/3127) require a release to `2.32` but no [response](https://github.com/import-js/eslint-plugin-import/issues/3184) about it.
eslint-plugin-import-x is a fork but well maintained.

PR #21949, opened 2025-08-29
refactor(style): Update Eslint to V9 and Prettier to V3

FrontEndDog:
重要变化：
1.升级eslint 到V9，配置文件改写为新配置文件，删除.eslintignore文件 使用配置文件中的ignores字段替代

2.升级prettier到V3，prettier升级到V3后 配置 trailingComma的默认值由"es5"变成了"all", 添加  "trailingComma": "es5", 保持原样[链接](https://prettier.io/docs/options#trailing-commas)

3.升级所有eslint相关依赖到最新版本，移除eslint-plugin-eslint-comments, 太久没更新了 不支持eslintV9。

4.使用eslint-plugin-import替换eslint-plugin-import-x，之前替换为eslint-plugin-import-x是因为 #21036 现在这个问题已经在eslint-plugin-import上得到解决。

5.json文件使用解析器由recommended-with-jsonc替换为recommended-with-json5，为了兼容.github/renovate.json5这个文件

changes：

1. Upgrade ESLint to V9, refactor the configuration file to the new format, remove the .eslintignore file, and replace it with the ignores field in the configuration file

2. Upgrade Prettier to V3. After upgrading to V3, the default value of trailingComma changed from “es5” to “all”. Add “trailingComma”: “es5” to keep the original formatting

3. Upgrade all ESLint-related dependencies to the latest versions, remove eslint-plugin-eslint-comments, which has not been updated for a long time and does not support ESLint V9

4. Replace eslint-plugin-import-x with eslint-plugin-import. Previously eslint-plugin-import-x was used due to #21036, but this issue has now been resolved in eslint-plugin-import

5. For JSON files, replace the parser from recommended-with-jsonc to recommended-with-json5 to support .github/renovate.json5


packages/components/affix/src/affix.vue
```diff
@@ -30,7 +30,7 @@ import type { CSSProperties } from 'vue'
 
 const COMPONENT_NAME = 'ElAffix'
 defineOptions({
-  name: COMPONENT_NAME,
+  name: 'ElAffix',
```
btea: This should be left as is; using a const is preferable.

btea (commented):
`internal/eslint-config/index.old.js` can be deleted.

packages/components/affix/src/affix.vue
FrontEndDog: [This rule](https://eslint.vuejs.org/rules/valid-define-options.html) does not allow the use of variables in definiteOptions
I have submitted a [feature request](https://github.com/vuejs/eslint-plugin-vue/issues/2919) to eslint-plugin-vue
If accepted, I will modify the original code

FrontEndDog:
> `internal/eslint-config/index.old.js` can be deleted.

For the convenience of comparison, I will temporarily keep it and delete it before merging

packages/components/affix/src/affix.vue
btea: Perhaps we can comment the relevant rules here now.

packages/components/time-picker/src/common/picker.vue
```diff
@@ -25,11 +25,11 @@
...
       <el-input
         v-if="!isRangeInput"
-        :id="(id as string | undefined)"
+        :id="id as string"
         ref="inputRef"
         container-role="combobox"
-        :model-value="(displayValue as string)"
-        :name="(name as string | undefined)"
+        :model-value="displayValue as string"
+        :name="name as string"
```
Dsaquel: | Before | After |
|--------|-------|
| Image | Image |

In vscode this change will break the colors:

FrontEndDog: This is a issues known to Volar.
https://github.com/vuejs/language-tools/issues/2104
I have already used `<! --Prettier ignore attribute -->` to prevent prettier from automatically formatting.

FrontEndDog: oh no, Using comments in template tags with only one child node will be judged as two child nodes by Vue, causing errors in automated testing. I will use another method to fix this. I think we should avoid using 'as' in attributes.

packages/components/affix/src/affix.vue
FrontEndDog: > Perhaps we can comment the relevant rules here now.

I think we should temporarily follow this rule to prevent future changes from causing other variables to appear in the definiteOptions.

FrontEndDog:
https://github.com/vuejs/eslint-plugin-vue/pull/2920
Waiting for this PR to be released.

FrontEndDog:
@btea @Dsaquel    `eslint-plugin-vue` version 10.5.0 has been released. This PR is ready to merge, can you take some time to review it.

btea (approved):

internal/eslint-config/index.mjs
```diff
@@ -0,0 +1,553 @@
...
+import globals from 'globals'
+import tseslint, { parser as tsParser } from 'typescript-eslint'
+import eslintPluginUnicorn from 'eslint-plugin-unicorn'
+import importPlugin from 'eslint-plugin-import'
+import eslintPluginJsonc from 'eslint-plugin-jsonc'
+import markdown from 'eslint-plugin-markdown'
+import pluginVue from 'eslint-plugin-vue'
+import eslintPluginPrettierRecommended from 'eslint-plugin-prettier/recommended'
+
+export default tseslint.config(
```
Dsaquel: I saw that `tseslint.config` is deprecated maybe we can pass directly with [defineConfig](https://eslint.org/docs/latest/use/configure/configuration-files), this not block the PR tho.

Dsaquel (commented):
Other things:

1. After doing `pnpm lint` it seems like the folder 'docs/.vitepress/dist' is not getting ignored. Surely you can having the same reproduction by running `pnpm docs:build` and `pnpm lint`.
2. After doing `pnpm format` i have 49 files being formatted.

internal/eslint-config/index.mjs
FrontEndDog: Done!

FrontEndDog: 1.I forgot to run `pnpm format` after upgrading Prettier to V3 and Now it can ignore 'docs/. vitepress/dist'.

2.These changes are caused by the upgrade of Prettier V2 to V3.

internal/eslint-config/package.json
```diff
@@ -1,36 +1,34 @@
...
   "version": "0.0.1",
+  "type": "module",
   "description": "ESLint Config",
   "license": "MIT",
   "files": [
-    "index.js"
+    "index.mjs"
   ],
-  "main": "index.js",
+  "main": "index.mjs",
```
Dsaquel: Because now the package is esm friendly by enabling `type: "module"` we can switch `index.mjs` => `index.js` i think.

```suggestion
    "index.js"
],
  "main": "index.js",
```

FrontEndDog: That is right. Done!

Dsaquel (approved):

Dsaquel:
The commit history contains commits that are already been merged, isn't something weird about this or not really important?

cc @element-plus/backers

btea:
@Dsaquel Thanks for your attention. I hadn't noticed this before. Is this caused by a code merge? @FrontEndDog

FrontEndDog:
> @Dsaquel Thanks for your attention. I hadn't noticed this before. Is this caused by a code merge? @FrontEndDog 

I have noticed this.Due to a conflict, this situation may occur after resolving the conflict on VScode. Using `git rebase`+`git push -f` can solve it.

I have already taken care of it.

FrontEndDog:
> > @Dsaquel Thanks for your attention. I hadn't noticed this before. Is this caused by a code merge? @FrontEndDog
> 
> I have noticed this.Due to a conflict, this situation may occur after resolving the conflict on VScode. Using `git rebase`+`git push -f` can solve it.

I have already taken care of it.

btea:
It seems to contain one more record.
image

FrontEndDog:
> It seems to contain one more record. image

I have already deleted it.

btea (approved):

warmthsea (commented):
I think some rules should remain as they are:

1. Now: Volar(IDE) supports color is warning.

Image

2. `typeof import('element-plus')['---']`

I think similar changes can remain as they are. The original code is more concise.

Image

refer: https://github.com/unplugin/unplugin-auto-import

Image

3. <> `,`  the rules can be canceled.

() also No, I think it’s fine to keep it as it is.

Image

FrontEndDog:
@warmthsea 
1.prettier Version 3.x automatically removes these parentheses as unnecessary. see https://github.com/vuejs/language-tools/issues/2104#issuecomment-3092541527. I think this is a bug in Volar. 
I used this method to temporarily solve the problem.  I think we should avoid using 'as' in attributes.
image

2.This is also caused by the upgrade of Prettier V2 to V3. see https://github.com/prettier/prettier/pull/14042 https://github.com/prettier/prettier/issues/14660

3.This is also caused by the upgrade of Prettier V2 to V3. see https://github.com/prettier/prettier/issues/15142

btea:
Regarding the second point mentioned by @warmthsea, I think we can ignore this kind of code generated directly by the tool when formatting.

I can live with or without the comma in the third point. Either way, the formatting tool will automatically handle it, so there's no confusion.

FrontEndDog:
> Regarding the second point mentioned by @warmthsea, I think we can ignore this kind of code generated directly by the tool when formatting.
> 
> I can live with or without the comma in the third point. Either way, the formatting tool will automatically handle it, so there's no confusion.

`typings/global.d.ts` doesn't seem to be generated by a tool.
I tend to change the config `trailingComma` of prettier to all. what do you think? @element-plus/backers

FrontEndDog:
Use emojis to vote: change the config trailingComma of prettier to all.

FrontEndDog:
Use emojis to vote: change the config trailingComma of prettier to none.

FrontEndDog:
Use emojis to vote: keep the config trailingComma of prettier unchanged. es5.

btea:
> `typings/global.d.ts` doesn't seem to be generated by a tool.

Oh, it looks similar to the one generated by unplugin. But I think it looks more comfortable to keep the original format.

FrontEndDog:
> > `typings/global.d.ts` doesn't seem to be generated by a tool.
> 
> Oh, it looks similar to the one generated by unplugin. But I think it looks more comfortable to keep the original format.

OK! I have added `/* prettier ignore */ `to this file

internal/eslint-config/index.js
```diff
@@ -1,502 +1,555 @@
...
-  plugins: ['@typescript-eslint', 'prettier', 'unicorn'],
-  extends: [
-    'eslint:recommended',
-    'plugin:import-x/recommended',
-    'plugin:eslint-comments/recommended',
-    'plugin:jsonc/recommended-with-jsonc',
-    'plugin:markdown/recommended',
-    'plugin:vue/vue3-recommended',
-    'plugin:@typescript-eslint/recommended',
-    'prettier',
```
tolking, line 18: Do we need to import prettier in the new configuration file?

```ts
import prettier from 'eslint-config-prettier'
```

FrontEndDog: No need！·`eslint-plugin-prettier/recommended` has already added `eslint-config-prettier` for us.
image

tolking (approved):
LGTM.

btea:
Let's merge this PR first. Otherwise, it will easily cause conflicts every time we modify other PRs.

PR #21961, opened 2025-08-29
feat(components): [image-viewer] add `error` slot

zzjiaxiang:
close #21952

### DIFF
--- a/packages/components/image-viewer/src/image-viewer.vue
+++ b/packages/components/image-viewer/src/image-viewer.vue
@@ -84,13 +84,11 @@
           <!-- CANVAS -->
           <div :class="ns.e('canvas')">
             <slot
-              v-if="loadError"
+              v-if="loadError && $slots['viewer-error']"
               name="viewer-error"
               :active-index="activeIndex"
               :src="currentImg"
-            >
-              <div :class="ns.e('error')">{{ t('el.image.error') }}</div>
-            </slot>
+            />
             <img
               v-else
               ref="imgRef"
@@ -298,6 +296,7 @@ function handleImgError(e: Event) {
   loadError.value = true
   loading.value = false
   emit('error', e)
+  ;(e.target as HTMLImageElement).alt = t('el.image.error')
 }
 
 function handleMouseDown(e: MouseEvent) {
--- a/packages/components/image-viewer/src/image-viewer.vue
+++ b/packages/components/image-viewer/src/image-viewer.vue
@@ -94,6 +94,7 @@
             <img
               v-else
               ref="imgRef"
+              :key="currentImg"
               :src="currentImg"
               :style="imgStyle"
               :class="ns.e('img')"
--- a/packages/components/image-viewer/src/image-viewer.vue
+++ b/packages/components/image-viewer/src/image-viewer.vue
@@ -340,7 +340,7 @@ function toggleMode() {
 }
 
 function setActiveItem(index: number) {
-  loadError.value &&= false
+  loadError.value = false
   const len = props.urlList.length
   activeIndex.value = (index + len) % len
 }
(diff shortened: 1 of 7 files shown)


packages/components/image-viewer/src/image-viewer.vue
```diff
@@ -337,6 +340,7 @@ function toggleMode() {
 }
 
 function setActiveItem(index: number) {
+  loadError.value &&= false
```
Dsaquel: I think this can be simplified.

```suggestion
  loadError.value = false
```

Dsaquel (commented):
LGTM,

We can add the newly created slot in the api section:
https://preview-21961-element-plus.surge.sh/en-US/component/image.html#image-slots

packages/components/image-viewer/src/image-viewer.vue
zzjiaxiang: Use `&&=` to avoid meaningless assignments

zzjiaxiang:
`Image error Slots ` 和  `Image Viewer error Slots` 两个组件图像加载错误的插槽名字要不要区分开?
会有这种场景,使用图片预览的时候,图像加载错误和预览图像加载错误要自定义的内容是不一样的.

packages/components/image-viewer/src/image-viewer.vue
Dsaquel: It is a very micro optimization, and vue already check if the value is the same as before.
https://github.com/vuejs/core/blob/main/packages/reactivity/src/ref.ts#L136-L143

Generally speaking, if we accept this change, many of the code can be "optimized" like this.

Dsaquel:
>Image error Slots  和 Image Viewer error Slots 两个组件图像加载错误的插槽名字要不要区分开?
会有这种场景,使用图片预览的时候,图像加载错误和预览图像加载错误要自定义的内容是不一样的.

Didn't see the error slot sorry. I though we didn't have it at all.
That said, we can say that the error slot for image preview is available on `2.11.2` in the section api, like you have done earlier in the doc.

Dsaquel (commented):
I see you renamed `error` to `viewer-error`, well done 👍 .

Now LGTM, can you just add a relevant test case ?

Dsaquel (approved):
LGTM, Thanks!

zzjiaxiang:
Simplified code example

packages/components/image-viewer/src/image-viewer.vue
```diff
@@ -83,19 +83,20 @@
...
+              v-else
+              ref="imgRef"
+              :src="currentImg"
+              :style="imgStyle"
+              :class="ns.e('img')"
+              :crossorigin="crossorigin"
+              @load="handleImgLoad"
+              @error="handleImgError"
+              @mousedown="handleMouseDown"
+            />
```
btea, line 103: I think we can keep the original structure and pass the `url` and `index` to the slot `viewer-error`.

packages/components/image-viewer/src/image-viewer.vue
```diff
@@ -287,8 +295,9 @@ function handleImgLoad() {
 }
 
 function handleImgError(e: Event) {
+  loadError.value = true
   loading.value = false
-  ;(e.target as HTMLImageElement).alt = t('el.image.error')
```
btea, line 291: I think the alt attribute can be retained.

btea (approved):
Thanks!

Dsaquel:
In the documentation the preview image error is not really readable, we can set a lighter background i think.

image

btea:
> In the documentation the preview image error is not really readable, we can set a lighter background i think.

I wanted to say this too, but forgot. 😄

zzjiaxiang:
I have adjusted the wrong font color. If you have any good ideas, you can directly change them here

Dsaquel (approved):

PR #21966, opened 2025-08-30
refactor(components): [cascader] adjust the ts type

thinkasany:
It seems that the automatic deduction can be done normally.

image


btea (approved):
Thanks!

PR #21976, opened 2025-08-30
docs(components): [select] demo for adjusting the options property

lxKylin:

I think that since the description of the options property demo mentions props, it would be better to use the props property in the demo code.

### DIFF
--- a/docs/examples/select/options.vue
+++ b/docs/examples/select/options.vue
@@ -2,12 +2,7 @@
   <el-select
     v-model="value"
     :options="options"
-    :props="{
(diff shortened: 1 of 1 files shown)


docs/examples/select/options.vue
```diff
@@ -2,6 +2,12 @@
   <el-select
     v-model="value"
     :options="options"
+    :props="{
```
btea: It is best to define a variable and put it inside the script tag.

btea (approved):
Thanks!

PR #21978, opened 2025-08-31
fix(components): [select] dropdown should always display when focused

Codachen:

### Description

When the select component's input is focused, the dropdown should always be displayed for better user experience (unified with the other components).

[Ref](https://github.com/element-plus/element-plus/pull/21959)

### Preview

Before:
select-collapse

After:
select-focus-open

### DIFF
--- a/packages/components/select/src/useSelect.ts
+++ b/packages/components/select/src/useSelect.ts
@@ -684,15 +684,12 @@ export const useSelect = (props: SelectProps, emit: SelectEmits) => {
 
   const toggleMenu = (event?: Event) => {
     if (selectDisabled.value) return
-
     // We only set the inputHovering state to true on mouseenter event on iOS devices
     // To keep the state updated we set it here to true
     if (isIOS) states.inputHovering = true
 
-    const clickedSuffix =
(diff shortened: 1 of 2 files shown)


tolking:
@Codachen The test has encountered an error and needs to be fixed.

Codachen:
@tolking Thanks for the reminder. I tried updating the tests a few times, but couldn’t get them all to pass. The component behavior looked correct to me in manual testing, but the test run still failed. I wasn’t fully sure about the test logic, so I had to stop. Sorry for leaving it like this 🙏 Please feel free to adjust as needed or share any suggestions.

packages/components/select/src/useSelect.ts
```diff
@@ -682,18 +682,32 @@ export const useSelect = (props: SelectProps, emit: SelectEmits) => {
     }
   }
 
-  const toggleMenu = () => {
+  const toggleMenu = (event?: Event) => {
```
tolking, line 685: You should write the judgment at the top as much as possible and return directly when the conditions are met to avoid executing unnecessary code.

PR #21980, opened 2025-08-31
feat(components): [mention] support props prop

keeplearning66:

According to #21846, make `Mention` also support `props` prop


packages/components/mention/src/mention.vue
```diff
@@ -225,7 +245,7 @@ const handleSelect = (item: MentionOption) => {
 
   emit(UPDATE_MODEL_EVENT, newValue)
   emit(INPUT_EVENT, newValue)
-  emit('select', item, mentionCtx.value.prefix)
+  emit('select', getOriginalOption(item)!, mentionCtx.value.prefix)
```
Dsaquel, line 251: Do we really need `getOriginalOption` ? Maybe we can pass the configured props by the computed to `mention-dropdown.vue` and return the right `option`.

packages/components/mention/src/types.ts
```diff
@@ -1,5 +1,5 @@
 export type MentionOption = {
-  value: string
+  value?: string
```
Dsaquel, line 2: We should not change this line.

packages/components/mention/src/useProps.ts
```diff
@@ -0,0 +1,37 @@
...
+  const getValue = (option: MentionOption) =>
+    get(option, aliasProps.value.value)
+  const getDisabled = (option: MentionOption) =>
+    get(option, aliasProps.value.disabled)
+
+  return {
+    getLabel,
+    getValue,
+    getDisabled,
+  }
```
Dsaquel: 1. No need to create a file, we can directly referenced this into `mention.vue`.
2. We can make the change accordingly with https://github.com/element-plus/element-plus/pull/21543

packages/components/mention/src/mention.vue
```diff
@@ -108,10 +109,22 @@ const computedFallbackPlacements = computed<Placement[]>(() =>
...
+})
+
 const filteredOptions = computed(() => {
   const { filterOption, options } = props
-  if (!mentionCtx.value || !filterOption) return options
-  return options.filter((option) =>
+  const newOptions = options.map((option) => {
+    const newOption = getNewOption(option)
+    return { ...option, ...newOption }
+  })
```
Dsaquel: I think externalize this as a computed is better.

packages/components/mention/src/mention.vue
patienry: I tried returning `originalOption` in `mention-dropdown.vue` yesterday. There will probably be three scenarios.

1. `option={id: '1', name: 'test'}`  `props={value: 'id', label: 'name'}`  `newOption={id: '1', name: 'test', value: '1', label: 'test'}`
2. `option={value: '1', label: 'test'}`  `props={value: 'value', label: 'label'}`  `newOption={value: '1', label: 'test'}`
3. `option={value: '1', label: 'test'}`  without `props`  `newOption={value: '1', label: 'test'}`

If we use `props` and `newOption` in `mention-dropdown.vue` to return `originalOption` while satisfying these three scenarios, it might be a bit complicated. And handling it in `mention.vue` can appropriately reduce the scope of changes, without needing to modify the code in `mention-dropdown.vue`. What do you think? 👀 👀

packages/components/mention/src/types.ts
patienry: If `props` are allowed, the `option` passed by users may not have `value`, for example: `{id: '1', name: 'test'}`.

packages/components/mention/src/useProps.ts
patienry: Done, thank you.

packages/components/mention/src/mention.vue
patienry: Done, thank you.

packages/components/mention/src/mention.vue
Dsaquel: I understand now why we need these change, thanks you.

packages/components/mention/src/mention.vue
```diff
@@ -108,10 +108,31 @@ const computedFallbackPlacements = computed<Placement[]>(() =>
...
+const getNewOption = (option: MentionOption) => {
+  const base = {
+    label: option[aliasProps.value.label],
+    value: option[aliasProps.value.value],
+    disabled: option[aliasProps.value.disabled],
+  }
+  return { ...option, ...base }
+}
+
+const newOptions = computed(() => {
```
Dsaquel: I think we can rename it `options` and do a one line computed.

packages/components/mention/src/mention.vue
```diff
@@ -108,10 +108,31 @@ const computedFallbackPlacements = computed<Placement[]>(() =>
   props.showArrow ? ['bottom', 'top'] : ['bottom-start', 'top-start']
 )
 
+const aliasProps = computed(() => ({
+  ...mentionDefaultProps,
+  ...props.props,
+}))
+
+const getNewOption = (option: MentionOption) => {
```
Dsaquel: I think we can rename it `mapOption`.

Dsaquel (commented):
The emit return an object instead of a string, resulting to a warning.

Image

[demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtbWVudGlvblxuICAgIHYtbW9kZWw9XCJ2YWx1ZVwiXG4gICAgOm9wdGlvbnM9XCJvcHRpb25zXCJcbiAgICA6cHJvcHM9XCJwcm9wc1wiXG4gICAgc3R5bGU9XCJ3aWR0aDogMzIwcHhcIlxuICAgIHBsYWNlaG9sZGVyPVwiUGxlYXNlIGlucHV0XCJcbiAgICBAc2VsZWN0PVwibG9nXCJcbiAgLz5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgc2V0dXAgbGFuZz1cInRzXCI+XG5pbXBvcnQgeyByZWYgfSBmcm9tICd2dWUnXG5cbmNvbnN0IGxvZyA9IGNvbnNvbGUubG9nXG5jb25zdCB2YWx1ZSA9IHJlZignQCcpXG5cbmNvbnN0IG9wdGlvbnMgPSByZWYoW1xuICB7XG4gICAgbGFiZWw6ICdGdXBob2VuaXhlcycsXG4gICAgbmFtZTogJ0Z1cGhvZW5peGVzJyxcbiAgfSxcbiAge1xuICAgIGxhYmVsOiAna29vcmlvb2thbWknLFxuICAgIG5hbWU6ICdrb29yaW9va2FtaScsXG4gIH0sXG4gIHtcbiAgICBsYWJlbDogJ0plcmVteScsXG4gICAgbmFtZTogJ0plcmVteScsXG4gIH0sXG4gIHtcbiAgICBsYWJlbDogJ2J0ZWEnLFxuICAgIG5hbWU6ICdidGVhJyxcbiAgfSxcbl0pXG5jb25zdCBwcm9wcyA9IHtcbiAgdmFsdWU6ICduYW1lJyxcbn1cbjwvc2NyaXB0PlxuIiwiZWxlbWVudC1wbHVzLmpzIjoiaW1wb3J0IEVsZW1lbnRQbHVzIGZyb20gJ2VsZW1lbnQtcGx1cydcbmltcG9ydCB7IGdldEN1cnJlbnRJbnN0YW5jZSB9IGZyb20gJ3Z1ZSdcblxubGV0IGluc3RhbGxlZCA9IGZhbHNlXG5hd2FpdCBsb2FkU3R5bGUoKVxuXG5leHBvcnQgZnVuY3Rpb24gc2V0dXBFbGVtZW50UGx1cygpIHtcbiAgaWYgKGluc3RhbGxlZCkgcmV0dXJuXG4gIGNvbnN0IGluc3RhbmNlID0gZ2V0Q3VycmVudEluc3RhbmNlKClcbiAgaW5zdGFuY2UuYXBwQ29udGV4dC5hcHAudXNlKEVsZW1lbnRQbHVzKVxuICBpbnN0YWxsZWQgPSB0cnVlXG59XG5cbmV4cG9ydCBmdW5jdGlvbiBsb2FkU3R5bGUoKSB7XG4gIGNvbnN0IHN0eWxlcyA9IFsnaHR0cHM6Ly9wcmV2aWV3LTIxOTgwLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5jc3MnLCAnaHR0cHM6Ly9wcmV2aWV3LTIxOTgwLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvdGhlbWUtY2hhbGsvZGFyay9jc3MtdmFycy5jc3MnXS5tYXAoKHN0eWxlKSA9PiB7XG4gICAgcmV0dXJuIG5ldyBQcm9taXNlKChyZXNvbHZlLCByZWplY3QpID0+IHtcbiAgICAgIGNvbnN0IGxpbmsgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCdsaW5rJylcbiAgICAgIGxpbmsucmVsID0gJ3N0eWxlc2hlZXQnXG4gICAgICBsaW5rLmhyZWYgPSBzdHlsZVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdsb2FkJywgcmVzb2x2ZSlcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignZXJyb3InLCByZWplY3QpXG4gICAgICBkb2N1bWVudC5ib2R5LmFwcGVuZChsaW5rKVxuICAgIH0pXG4gIH0pXG4gIHJldHVybiBQcm9taXNlLmFsbFNldHRsZWQoc3R5bGVzKVxufVxuIiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IFwiRVNOZXh0XCIsXG4gICAgXCJqc3hcIjogXCJwcmVzZXJ2ZVwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwidHlwZXNcIjogW1wiZWxlbWVudC1wbHVzL2dsb2JhbC5kLnRzXCJdLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcImFsbG93SnNcIjogdHJ1ZSxcbiAgICBcImNoZWNrSnNcIjogdHJ1ZVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vZmFzdGx5LmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9ydW50aW1lLWRvbUBsYXRlc3QvZGlzdC9ydW50aW1lLWRvbS5lc20tYnJvd3Nlci5qc1wiLFxuICAgIFwiQHZ1ZS9zaGFyZWRcIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvc2hhcmVkQGxhdGVzdC9kaXN0L3NoYXJlZC5lc20tYnVuZGxlci5qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9wcmV2aWV3LTIxOTgwLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5mdWxsLm1pbi5tanNcIixcbiAgICBcImVsZW1lbnQtcGx1cy9cIjogXCJ1bnN1cHBvcnRlZFwiLFxuICAgIFwiQGVsZW1lbnQtcGx1cy9pY29ucy12dWVcIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlQDIvZGlzdC9pbmRleC5taW4uanNcIlxuICB9LFxuICBcInNjb3Blc1wiOiB7fVxufSIsIl9vIjp7InNob3dIaWRkZW4iOnRydWUsInN0eWxlU291cmNlIjoiaHR0cHM6Ly9wcmV2aWV3LTIxOTgwLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5jc3MifX0=).

patienry:
image

image

I found that the warning was caused by the validation of the option's `value`, so I removed this validation. Since supporting props, the option passed by the user may not have `value`.
Also, I set a default value (`''`) for the `label` in the default `filterOption` to avoid errors when calling `toLowerCase`.

packages/components/mention/src/mention.vue
patienry: Done, thank you.

packages/components/mention/src/mention.vue
patienry: Done, thank you.

packages/components/mention/src/mention.ts
```diff
@@ -125,3 +131,15 @@ export type MentionPropsPublic = __ExtractPublicPropTypes<typeof mentionProps>
 export type MentionInstance = InstanceType<typeof Mention> & unknown
 
 export type { MentionOption } from './types'
+
+export type MentionOptionProps = {
+  value?: string
+  label?: string
+  disabled?: string
+}
```
Dsaquel: Maybe we can reuse `MentionOption`.

Dsaquel (commented):
This line should also be changed:
https://github.com/keeplearning66/element-plus/blob/5076aa4797e43a3f87f0f10bbb7735c4fc12fbdb/packages/components/mention/src/mention.vue#L187

Dsaquel:
>I found that the warning was caused by the validation of the option's value, so I removed this validation. Since supporting props, the option passed by the user may not have value.

Okay, maybe we can set `isObject` then ?

packages/components/mention/src/mention.ts
patienry: Done, thank you.

patienry:
> Okay, maybe we can set `isObject` then ?

Okay, this would be better.

packages/components/mention/src/mention.ts
Dsaquel: My fault, i thought `MentionOptionProps` was the same type as `MentionOption`. We can revert to this change.

Also it can be written like:
```ts
export type MentionOptionProps = {
  value?: string
  label?: string
  disabled?: string
  [key]: string
}
```
to accept other properties.

patienry: It’s okay, I’ve made the changes. Thank you!

packages/components/mention/src/mention.vue
```diff
@@ -108,10 +108,26 @@ const computedFallbackPlacements = computed<Placement[]>(() =>
...
+const mapOption = (option: MentionOption) => {
+  const base = {
+    label: option[aliasProps.value.label],
+    value: option[aliasProps.value.value],
+    disabled: option[aliasProps.value.disabled],
+  }
+  return { ...option, ...base }
+}
+
+const options = computed(() => props.options.map((option) => mapOption(option)))
```
Dsaquel: **nit**:

```suggestion
const options = computed(() => props.options.map(mapOption))
```

docs/en-US/component/mention.md
```diff
@@ -137,11 +138,18 @@ and please go to the original component to view the documentation.
...
 ```ts
 type MentionOption = {
-  value: string
+  value?: string
   label?: string
   disabled?: boolean
   [key: string]: any
 }
+
+export type MentionOptionProps = {
```
Dsaquel:
```suggestion
type MentionOptionProps = {
```

docs/en-US/component/mention.md
```diff
@@ -86,22 +86,23 @@ and please go to the original component to view the documentation.
...
+| Name                                 | Description                                                                            | Type                                                                         | Default                                                |
+| ------------------------------------ | -------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------- | ------------------------------------------------------ |
+| options                              | mention options list                                                                   | ^[array]`MentionOption[]`                                                    | `[]`                                                   |
+| props ^(2.11.2)                      | configuration options                                                                  | ^[object]`MentionOptionProps`                                                | `{value: 'value',label: 'label',disabled: 'disabled'}` |
```
Dsaquel:
```suggestion
| props ^(2.11.2)                      | configuration options                                                                  | ^[object]`MentionOptionProps`                                                | `{value: 'value', label: 'label', disabled: 'disabled'}` |
```

Dsaquel (approved):
LGTM

tolking (approved):

btea (approved):

PR #21988, opened 2025-09-01
style(components): [splitter] use ew-ns cursor

warmthsea:
As @thinkasany said, it seems more reasonable to use ew-ns cursor.

- https://github.com/element-plus/element-plus/pull/21595#discussion_r2275672484

screenshots

#### After

screenshots


Dsaquel (approved):
LGTM

PR #21997, opened 2025-09-01
feat(components): [tree-select] add `show-parent-tag` prop

Serect727:
In multiple selection mode with checkboxes, determines whether to display the parent tag itself instead of all its child tags when the parent node is selected.

BREAKING CHANGE :
docs/en-US/component/tree-select.md packages/components/tree-select/src/tree-select.vue  tree.ts


eighty-cents:
1. Here, like ant vue, if you pass the showCheckedStrategy field, and then receive three values, the branches are: 'all', 'parent', 'child', will it be better
2. You still need to improve the documentation. It would be better to put this function point in the example

PR #21999, opened 2025-09-02
refactor(components): improve tooltip type

thinkasany:


btea (approved):

packages/components/tooltip/src/content.vue
```diff
@@ -28,7 +28,7 @@
         :enterable="enterable"
         :pure="pure"
         :popper-class="popperClass"
-        :popper-style="[popperStyle, contentStyle]"
+        :popper-style="[popperStyle!, contentStyle]"
```
warmthsea, line 31: Although I think it is a small change, I think it is not appropriate to use `!` because its type is indeed `undfined`. It is better to use `?? {}`

Also https://github.com/element-plus/element-plus/blob/dev/packages/components/time-picker/src/common/picker.vue#L12

thinkasany: https://github.com/vuejs/core/blob/4b6cb1f52a0f6c5af2a0114b70000e23028eed15/packages/runtime-dom/src/modules/style.ts#L78

I think it's actually ok, ultimately both set it to `''` and then remove it, but the code for the non-null assertion is cleaner.

warmthsea: Indeed, it's okay to be concise.

warmthsea (approved):

PR #22000, opened 2025-09-02
feat(components): [date-picker] add props.unlinkTimePicker to cancel timePicker min/max value linkage

mansonchor:
In `<el-date-picker type="datetimerange">`, the max and min values ​​of the timePicker will be automatically linked and corrected, that can be confusing in some scenarios

Hope to add a props to control this linkage, `unlinkTimePicker` default is `false` so that it is a compatible change

refs issue: https://github.com/element-plus/element-plus/issues/21797


mansonchor:
Please take a look and see if it is suitable. If the idea is okay, I will add unit tests and related documents.

mansonchor:
@cszhjh @btea Could you please take a look at this PR?

Dsaquel:
There are some duplicate issues about this issue and one suggestion i have kept in mind and should be strongly considered is to disable times that go beyond each other.

Dsaquel:
I personally think the current behavior should be removed and adding a new prop to handle both situations bring a lot of complexity for no particular reason/scenario.

mansonchor:
> I personally think the current behavior should be removed and adding a new prop to handle both situations bring a lot of complexity for no particular reason/scenario.我个人认为当前的行为应该被移除，而添加一个新的属性来处理这两种情况毫无理由地带来了不必要的复杂性。

Actually, now if the start time is greater than the end time, the confirm button will be disabled
I think this interaction already ensures the legitimacy of the time value

image

Dsaquel:
From a user perspective we can't know what really cause the disable state of the confirm button.

mansonchor:
> From a user perspective we can't know what really cause the disable state of the confirm button.从用户的角度来看，我们无法知道确认按钮禁用状态的真实原因。

so should we removed the current  behavior? We can add a tooltip to the disabled button

mansonchor:
What additional information do we need to decide whether to remove the current behavior?

pls let me know, I'm glad to promote this.

Dsaquel:
We can get some other opinion before changing it. I checked again and the current behavior is really fine since it's only on the same day.

cc @element-plus/backers 
Shortcut discussion info: 
Currently on a same date when a start time is going above the end time, the end time is synchronously changed with the same value as the start time, same for the inverse scenario
(*when a end time is going below the start time, the start time is synchronously changed with the same value as the end time*).

solution: disable the beyond time for each other:
image

mansonchor:
> solution: disable the beyond time for each other:

that's a good idea

tolking:
@Dsaquel Your suggestion seems more reasonable.

PR #22001, opened 2025-09-02
refactor(components): [splitter] disabled class and style

thinkasany:
1. `!important` is not good for controlling styles

2. It seems that using `.is-disabled` is more intuitive and in line with project style.
image

### DIFF
--- a/packages/components/splitter/src/split-bar.vue
+++ b/packages/components/splitter/src/split-bar.vue
@@ -143,7 +143,7 @@ const EndIcon = computed(() => (isHorizontal.value ? ArrowRight : ArrowDown))
       :class="[
         ns.e('dragger'),
         draggerPseudoClass,
-        resizable ? '' : ns.is('disabled'),
+       ns.is('disabled', !resizable),
         ns.is('lazy', resizable && lazy),
       ]"
       :style="draggerStyles"
--- a/packages/components/splitter/src/split-bar.vue
+++ b/packages/components/splitter/src/split-bar.vue
@@ -48,7 +48,11 @@ const draggerStyles = computed(() => {
   return {
     width: isHorizontal.value ? '16px' : '100%',
     height: isHorizontal.value ? '100%' : '16px',
-    cursor: isHorizontal.value ? 'ew-resize' : 'ns-resize',
+    cursor: !props.resizable
+      ? 'auto'
+      : isHorizontal.value
+      ? 'ew-resize'
+      : 'ns-resize',
     touchAction: 'none',
   }
 })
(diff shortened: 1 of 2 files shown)


packages/theme-chalk/src/splitter.scss
```diff
@@ -69,7 +70,7 @@
     }
 
     &:hover {
-      &::before {
+      &::before:not(.is-disabled) {
```
rzzf: The `is-disabled` class is applied to the `splitter-bar`.

```suggestion
      &:not(.is-disabled)::before {
```

packages/theme-chalk/src/splitter.scss
```diff
@@ -37,12 +37,13 @@
   flex: none;
   position: relative;
   user-select: none;
-  @include e(disable) {
-    cursor: auto !important;
+  .is-disabled {
+    cursor: auto;
     &::before {
-      background-color: getCssVar('border-color-light') !important;
+      background-color: getCssVar('border-color-light');
```
rzzf: Can this be removed? Since the `:not(.is-disabled)` selector is already applied for the hover state.

thinkasany: Thx～ you are right.

thinkasany: I have finished the adjustment. Can you help me see if there are any other problems? Thanks again！

packages/components/splitter/src/split-bar.vue
```diff
@@ -139,7 +143,7 @@ const EndIcon = computed(() => (isHorizontal.value ? ArrowRight : ArrowDown))
       :class="[
         ns.e('dragger'),
         draggerPseudoClass,
-        resizable ? '' : ns.e('disable'),
+        resizable ? '' : ns.is('disabled'),
```
rzzf: I think the condition can be placed inside `ns.is` to make it more consistent.

```suggestion
       ns.is('disabled', !resizable),
```

packages/theme-chalk/src/splitter.scss
rzzf: Looks good👍

tolking (approved):

packages/theme-chalk/src/splitter.scss
```diff
@@ -37,12 +37,7 @@
   flex: none;
   position: relative;
   user-select: none;
-  @include e(disable) {
-    cursor: auto !important;
-    &::before {
-      background-color: getCssVar('border-color-light') !important;
-    }
```
btea, line 44: Is this style lost after modification? 🤔

rzzf: The background-color has already been set here
https://github.com/thinkasany/element-plus/blob/6608758629200fb4ad9f82674276dccc5f05dd9a/packages/theme-chalk/src/splitter.scss#L55

And the cursor has also been reset to auto
https://github.com/thinkasany/element-plus/blob/6608758629200fb4ad9f82674276dccc5f05dd9a/packages/components/splitter/src/split-bar.vue#L52

I think this part can be safely removed.

btea (approved):

PR #22008, opened 2025-09-02
docs: the ElDropdown component in the preview webpage cannot be use

tolking:

link #18553

The Vue version was 3.2.37 during development, but after build the Vue version was 3.4.31 . This is not in line with expectations.

I have tried to unify the Vue version, but it seems that there is no perfect solution

### DIFF
--- a/docs/.vitepress/config/index.mts
+++ b/docs/.vitepress/config/index.mts
@@ -8,6 +8,7 @@ import { nav } from './nav'
 import { mdPlugin } from './plugins'
 import { sidebars } from './sidebars'
 import { getViteConfig } from './vite'
+import { vueCompiler } from './vue-compiler'
 
 import type { UserConfig } from 'vitepress'
 
@@ -78,6 +79,7 @@ const setupConfig = (configEnv) => {
       config: (md) => mdPlugin(md),
     },
     vue: {
+      compiler: vueCompiler,
       template: {
         compilerOptions: {
           hoistStatic: false,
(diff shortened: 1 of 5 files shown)


btea:
Does this mean that there is a problem with our code in the new version of Vue?

tolking:
The official website can be used normally, indicating that there are no issues with the code after building.

The issues occurred when building the preview document, using an alias to point `element-plus` to the `packages` directory, which relied on two different versions of Vue and caused server rendering errors.

makedopamine:
I reverted [fd2be7](https://github.com/element-plus/element-plus/pull/22008/commits/fd2be708e4ceadba9fed846d82214af01583f03f) and added a new commit to fix the issue.


For context:  When running `pnpm run build`,  `@vietjs/plugin-vue` uses `@vue/compiler-sfc` v3.2.37 which doens't support `defineOptins` to compile SFCs, . So like [components/collection/src/collection.vue](https://github.com/element-plus/element-plus/blob/5459cd5e9d3683e7dbd28ec73f4a81905e5f2249/packages/components/collection/src/collection.vue) must first be  processed by `vue-macros` and then by `@vietjs/plugin-vue`. The output looks like this:
```
{
 render() { ... }
 setup() { ... }
}
```
When running `pnpm run docs:build`, since `@vue/compiler-sfc` v3.4.31 is used, `vue-macros` is no longer needed. However, the output changes.
```
{
   setup() { return render }
}
```
Then in [components/collection/src/collection.ts](https://github.com/element-plus/element-plus/blob/5459cd5e9d3683e7dbd28ec73f4a81905e5f2249/packages/components/collection/src/collection.ts#L23), the `setup()` property  of `collection.vue` component object is overridden with a void function, causing the issue.

Note: The difference in output is due to vue-macros, not to different versions of @vitejs/plugin-vue or @vue/compiler-sfc.

btea (approved):

PR #22020, opened 2025-09-03
refactor(components): [menu] optimize condition judgment

lxKylin:

I think the original condition judgment was overly complicated. We can break down the complex condition judgment into clearer logic to enhance the readability of the code.


btea:
I think the original logic looks very clear. If the ternary operation is difficult to understand, you can use `if...else...`. There is no need to create an extra file to implement a function.

lxKylin:
> I think the original logic looks very clear. If the ternary operation is difficult to understand, you can use `if...else...`. There is no need to create an extra file to implement a function.

ok

Dsaquel (approved):

btea (approved):
Thanks!

Issue #21402, opened 2025-07-18
[Component] [calendar] 级联选项数据变更不会重新渲染选项


WangJincheng4869:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.17`
- Element Plus Version: `2.10.4`
- Browser / OS: `UserAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:137.0) Gecko/20100101 Firefox/137.0`
- Build Tool: `Vite`

### Related Component

- `el-calendar`

### Reproduction Link

[...]

PR #22030, opened 2025-09-04
fix(components): [cascader] avoid init on identical options ref changes

### LINKED ISSUE #20269: [Component] [popover, cascader-panel] cascader panel bug in popover
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.13`
- Browser / OS: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36`
- Build Tool: `Vite`

### Related Component

- `el-popover`
- `el-cascader-panel`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtcG9wb3ZlciA6dGVsZXBvcnRlZD1cImZhbHNlXCIgd2lkdGg9XCJhdXRvXCI+XG4gICAgPHRlbXBsYXRlICNyZWZlcmVuY2U+XG4gICAgICA8ZWwtYnV0dG9uPnRlc3Q8L2VsLWJ1dHRvbj5cbiAgICA8L3RlbXBsYXRlPlxuICAgIDxlbC1jYXNjYWRlci1wYW5lbCBzdHlsZT1cIndpZHRoOiBmaXQtY29udGVudFwiIDpvcHRpb25zPVwib3B0aW9uczFcIiAvPlxuICA8L2VsLXBvcG92ZXI+XG5cbiAgPGVsLWNhc2NhZGVyLXBhbmVsIHN0eWxlPVwid2lkdGg6IGZpdC1jb250ZW50XCIgOm9wdGlvbnM9XCJvcHRpb25zMVwiIC8+XG5cbiAgPGVsLXBvcG92ZXIgOnRlbGVwb3J0ZWQ9XCJmYWxzZVwiICB3aWR0aD1cImF1dG9cIj5cbiAgICA8dGVtcGxhdGUgI3JlZmVyZW5jZT5cbiAgICAgIDxlbC1idXR0b24+dGVzdDwvZWwtYnV0dG9uPlxuICAgIDwvdGVtcGxhdGU+XG4gICAgPGVsLWNhc2NhZGVyLXBhbmVsIHN0eWxlPVwid2lkdGg6IGZpdC1jb250ZW50XCIgOm9wdGlvbnM9XCJbb3B0aW9uczJdXCIgLz5cbiAgPC9lbC1wb3BvdmVyPlxuXG4gIDxlbC1jYXNjYWRlci1wYW5lbCBzdHlsZT1cIndpZHRoOiBmaXQtY29udGVudFwiIDpvcHRpb25zPVwiW29wdGlvbnMyXVwiIC8+XG5cbiAgPGVsLXBvcG92ZXIgOnRlbGVwb3J0ZWQ9XCJmYWxzZVwiICB3aWR0aD1cImF1dG9cIj5cbiAgICA8dGVtcGxhdGUgI3JlZmVyZW5jZT5cbiAgICAgIDxlbC1idXR0b24+dGVzdDwvZWwtYnV0dG9uPlxuICAgIDwvdGVtcGxhdGU+XG4gICAgPGVsLWNhc2NhZGVyLXBhbmVsIHN0eWxlPVwid2lkdGg6IGZpdC1jb250ZW50XCIgOm9wdGlvbnM9XCJvcHRpb25zMygpXCIgLz5cbiAgPC9lbC1wb3BvdmVyPlxuICA8ZWwtY2FzY2FkZXItcGFuZWwgc3R5bGU9XCJ3aWR0aDogZml0LWNvbnRlbnRcIiA6b3B0aW9ucz1cIm9wdGlvbnMzKClcIiAvPlxuPC90ZW1wbGF0ZT5cblxuPHNjcmlwdCBsYW5nPVwidHNcIiBzZXR1cD5cblxuY29uc3Qgb3B0aW9uczEgPSBbe1xuICB2YWx1ZTogJ2d1aWRlJyxcbiAgbGFiZWw6ICdHdWlkZScsXG4gIGNoaWxkcmVuOiBbXG4gICAge1xuICAgICAgdmFsdWU6ICdkaXNjaXBsaW5lcycsXG4gICAgICBsYWJlbDogJ0Rpc2NpcGxpbmVzJyxcbiAgICAgIGNoaWxkcmVuOiBbXG4gICAgICAgIHtcbiAgICAgICAgICB2YWx1ZTogJ2NvbnNpc3RlbmN5JyxcbiAgICAgICAgICBsYWJlbDogJ0NvbnNpc3RlbmN5JyxcbiAgICAgICAgfSxcbiAgICAgICAge1xuICAgICAgICAgIHZhbHVlOiAnZmVlZGJhY2snLFxuICAgICAgICAgIGxhYmVsOiAnRmVlZGJhY2snLFxuICAgICAgICB9LFxuICAgICAgICB7XG4gICAgICAgICAgdmFsdWU6ICdlZmZpY2llbmN5JyxcbiAgICAgICAgICBsYWJlbDogJ0VmZmljaWVuY3knLFxuICAgICAgICB9LFxuICAgICAgICB7XG4gICAgICAgICAgdmFsdWU6ICdjb250cm9sbGFiaWxpdHknLFxuICAgICAgICAgIGxhYmVsOiAnQ29udHJvbGxhYmlsaXR5JyxcbiAgICAgICAgfSxcbiAgICAgIF0sXG4gICAgfSxcbiAgICB7XG4gICAgICB2YWx1ZTogJ25hdmlnYXRpb24nLFxuICAgICAgbGFiZWw6ICdOYXZpZ2F0aW9uJyxcbiAgICAgIGNoaWxkcmVuOiBbXG4gICAgICAgIHtcbiAgICAgICAgICB2YWx1ZTogJ3NpZGUgbmF2JyxcbiAgICAgICAgICBsYWJlbDogJ1NpZGUgTmF2aWdhdGlvbicsXG4gICAgICAgIH0sXG4gICAgICAgIHtcbiAgICAgICAgICB2YWx1ZTogJ3RvcCBuYXYnLFxuICAgICAgICAgIGxhYmVsOiAnVG9wIE5hdmlnYXRpb24nLFxuICAgICAgICB9LFxuICAgICAgXSxcbiAgICB9LFxuICBdLFxufV1cblxuY29uc3Qgb3B0aW9uczIgPSB7XG4gIHZhbHVlOiAnZ3VpZGUnLFxuICBsYWJlbDogJ0d1aWRlJyxcbiAgY2hpbGRyZW46IFtcbiAgICB7XG4gICAgICB2YWx1ZTogJ2Rpc2NpcGxpbmVzJyxcbiAgICAgIGxhYmVsOiAnRGlzY2lwbGluZXMnLFxuICAgICAgY2hpbGRyZW46IFtcbiAgICAgICAge1xuICAgICAgICAgIHZhbHVlOiAnY29uc2lzdGVuY3knLFxuICAgICAgICAgIGxhYmVsOiAnQ29uc2lzdGVuY3knLFxuICAgICAgICB9LFxuICAgICAgICB7XG4gICAgICAgICAgdmFsdWU6ICdmZWVkYmFjaycsXG4gICAgICAgICAgbGFiZWw6ICdGZWVkYmFjaycsXG4gICAgICAgIH0sXG4gICAgICAgIHtcbiAgICAgICAgICB2YWx1ZTogJ2VmZmljaWVuY3knLFxuICAgICAgICAgIGxhYmVsOiAnRWZmaWNpZW5jeScsXG4gICAgICAgIH0sXG4gICAgICAgIHtcbiAgICAgICAgICB2YWx1ZTogJ2NvbnRyb2xsYWJpbGl0eScsXG4gICAgICAgICAgbGFiZWw6ICdDb250cm9sbGFiaWxpdHknLFxuICAgICAgICB9LFxuICAgICAgXSxcbiAgICB9LFxuICAgIHtcbiAgICAgIHZhbHVlOiAnbmF2aWdhdGlvbicsXG4gICAgICBsYWJlbDogJ05hdmlnYXRpb24nLFxuICAgICAgY2hpbGRyZW46IFtcbiAgICAgICAge1xuICAgICAgICAgIHZhbHVlOiAnc2lkZSBuYXYnLFxuICAgICAgICAgIGxhYmVsOiAnU2lkZSBOYXZpZ2F0aW9uJyxcbiAgICAgICAgfSxcbiAgICAgICAge1xuICAgICAgICAgIHZhbHVlOiAndG9wIG5hdicsXG4gICAgICAgICAgbGFiZWw6ICdUb3AgTmF2aWdhdGlvbicsXG4gICAgICAgIH0sXG4gICAgICBdLFxuICAgIH0sXG4gIF0sXG59XG5cbmNvbnN0IG9wdGlvbnMzID0gKCkgPT4ge1xuICByZXR1cm4gW3tcbiAgICB2YWx1ZTogJ2d1aWRlJyxcbiAgICBsYWJlbDogJ0d1aWRlJyxcbiAgICBjaGlsZHJlbjogW1xuICAgICAge1xuICAgICAgICB2YWx1ZTogJ2Rpc2NpcGxpbmVzJyxcbiAgICAgICAgbGFiZWw6ICdEaXNjaXBsaW5lcycsXG4gICAgICAgIGNoaWxkcmVuOiBbXG4gICAgICAgICAge1xuICAgICAgICAgICAgdmFsdWU6ICdjb25zaXN0ZW5jeScsXG4gICAgICAgICAgICBsYWJlbDogJ0NvbnNpc3RlbmN5JyxcbiAgICAgICAgICB9LFxuICAgICAgICAgIHtcbiAgICAgICAgICAgIHZhbHVlOiAnZmVlZGJhY2snLFxuICAgICAgICAgICAgbGFiZWw6ICdGZWVkYmFjaycsXG4gICAgICAgICAgfSxcbiAgICAgICAgICB7XG4gICAgICAgICAgICB2YWx1ZTogJ2VmZmljaWVuY3knLFxuICAgICAgICAgICAgbGFiZWw6ICdFZmZpY2llbmN5JyxcbiAgICAgICAgICB9LFxuICAgICAgICAgIHtcbiAgICAgICAgICAgIHZhbHVlOiAnY29udHJvbGxhYmlsaXR5JyxcbiAgICAgICAgICAgIGxhYmVsOiAnQ29udHJvbGxhYmlsaXR5JyxcbiAgICAgICAgICB9LFxuICAgICAgICBdLFxuICAgICAgfSxcbiAgICAgIHtcbiAgICAgICAgdmFsdWU6ICduYXZpZ2F0aW9uJyxcbiAgICAgICAgbGFiZWw6ICdOYXZpZ2F0aW9uJyxcbiAgICAgICAgY2hpbGRyZW46IFtcbiAgICAgICAgICB7XG4gICAgICAgICAgICB2YWx1ZTogJ3NpZGUgbmF2JyxcbiAgICAgICAgICAgIGxhYmVsOiAnU2lkZSBOYXZpZ2F0aW9uJyxcbiAgICAgICAgICB9LFxuICAgICAgICAgIHtcbiAgICAgICAgICAgIHZhbHVlOiAndG9wIG5hdicsXG4gICAgICAgICAgICBsYWJlbDogJ1RvcCBOYXZpZ2F0aW9uJyxcbiAgICAgICAgICB9LFxuICAgICAgICBdLFxuICAgICAgfSxcbiAgICBdLFxuICB9XVxufVxuXG5cbjwvc2NyaXB0PlxuIiwiZWxlbWVudC1wbHVzLmpzIjoiaW1wb3J0IEVsZW1lbnRQbHVzIGZyb20gJ2VsZW1lbnQtcGx1cydcbmltcG9ydCB7IGdldEN1cnJlbnRJbnN0YW5jZSB9IGZyb20gJ3Z1ZSdcblxubGV0IGluc3RhbGxlZCA9IGZhbHNlXG5hd2FpdCBsb2FkU3R5bGUoKVxuXG5leHBvcnQgZnVuY3Rpb24gc2V0dXBFbGVtZW50UGx1cygpIHtcbiAgaWYgKGluc3RhbGxlZCkgcmV0dXJuXG4gIGNvbnN0IGluc3RhbmNlID0gZ2V0Q3VycmVudEluc3RhbmNlKClcbiAgaW5zdGFuY2UuYXBwQ29udGV4dC5hcHAudXNlKEVsZW1lbnRQbHVzKVxuICBpbnN0YWxsZWQgPSB0cnVlXG59XG5cbmV4cG9ydCBmdW5jdGlvbiBsb2FkU3R5bGUoKSB7XG4gIGNvbnN0IHN0eWxlcyA9IFsnaHR0cHM6Ly9mYXN0bHkuanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L2Rpc3QvaW5kZXguY3NzJywgJ2h0dHBzOi8vZmFzdGx5LmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQGxhdGVzdC90aGVtZS1jaGFsay9kYXJrL2Nzcy12YXJzLmNzcyddLm1hcCgoc3R5bGUpID0+IHtcbiAgICByZXR1cm4gbmV3IFByb21pc2UoKHJlc29sdmUsIHJlamVjdCkgPT4ge1xuICAgICAgY29uc3QgbGluayA9IGRvY3VtZW50LmNyZWF0ZUVsZW1lbnQoJ2xpbmsnKVxuICAgICAgbGluay5yZWwgPSAnc3R5bGVzaGVldCdcbiAgICAgIGxpbmsuaHJlZiA9IHN0eWxlXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2xvYWQnLCByZXNvbHZlKVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdlcnJvcicsIHJlamVjdClcbiAgICAgIGRvY3VtZW50LmJvZHkuYXBwZW5kKGxpbmspXG4gICAgfSlcbiAgfSlcbiAgcmV0dXJuIFByb21pc2UuYWxsU2V0dGxlZChzdHlsZXMpXG59XG4iLCJ0c2NvbmZpZy5qc29uIjoie1xuICBcImNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogXCJFU05leHRcIixcbiAgICBcImpzeFwiOiBcInByZXNlcnZlXCIsXG4gICAgXCJtb2R1bGVcIjogXCJFU05leHRcIixcbiAgICBcIm1vZHVsZVJlc29sdXRpb25cIjogXCJCdW5kbGVyXCIsXG4gICAgXCJ0eXBlc1wiOiBbXCJlbGVtZW50LXBsdXMvZ2xvYmFsLmQudHNcIl0sXG4gICAgXCJhbGxvd0ltcG9ydGluZ1RzRXh0ZW5zaW9uc1wiOiB0cnVlLFxuICAgIFwiYWxsb3dKc1wiOiB0cnVlLFxuICAgIFwiY2hlY2tKc1wiOiB0cnVlXG4gIH0sXG4gIFwidnVlQ29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiAzLjNcbiAgfVxufVxuIiwiUGxheWdyb3VuZE1haW4udnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCBBcHAgZnJvbSAnLi9BcHAudnVlJ1xuaW1wb3J0IHsgc2V0dXBFbGVtZW50UGx1cyB9IGZyb20gJy4vZWxlbWVudC1wbHVzLmpzJ1xuc2V0dXBFbGVtZW50UGx1cygpXG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8QXBwIC8+XG48L3RlbXBsYXRlPlxuIiwiaW1wb3J0LW1hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge1xuICAgIFwidnVlXCI6IFwiaHR0cHM6Ly9mYXN0bHkuanNkZWxpdnIubmV0L25wbS9AdnVlL3J1bnRpbWUtZG9tQGxhdGVzdC9kaXN0L3J1bnRpbWUtZG9tLmVzbS1icm93c2VyLmpzXCIsXG4gICAgXCJAdnVlL3NoYXJlZFwiOiBcImh0dHBzOi8vZmFzdGx5LmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9zaGFyZWRAbGF0ZXN0L2Rpc3Qvc2hhcmVkLmVzbS1idW5kbGVyLmpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXNcIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvZGlzdC9pbmRleC5mdWxsLm1pbi5tanNcIixcbiAgICBcImVsZW1lbnQtcGx1cy9cIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvXCIsXG4gICAgXCJAZWxlbWVudC1wbHVzL2ljb25zLXZ1ZVwiOiBcImh0dHBzOi8vZmFzdGx5LmpzZGVsaXZyLm5ldC9ucG0vQGVsZW1lbnQtcGx1cy9pY29ucy12dWVAMi9kaXN0L2luZGV4Lm1pbi5qc1wiXG4gIH0sXG4gIFwic2NvcGVzXCI6IHt9XG59IiwiX28iOnt9fQ==)

### Steps to reproduce

cascader panel 在 popover 中使用，且 options 为 方法返回的或在模板中用数组包起来

### What is Expected?

一次点击触发子菜单

二次点击才会触发子菜单

### Additional comments

_(empty)_

micaiguai:
Fix #20269

### Issues
It's necessary to click twice to select a node in a cascader-panel when the cascader-panel is within a poper component.

### Causes
The cascader-panel will update after the parent focus-trap component updated(trigger after first click). The cascader-panel then compares the props, although the prop content is same as before, it's a new array, Causing cascader-panel re-init the nodes.

In the menu component, the key of the node is a uid that auto-increaments. So, after prop changed, the key of node no longer match. This mean the node clicked   first unmounted, the node clicked second is new.

### Solutions
Change the node key from uid to value.

However the value might not match the key type. I've checked the select, tree component, they use "any" type. Perhaps there is a better solution, please guide me.


micaiguai (commented):
Fix #20269
[Before](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtcG9wb3ZlclxuICAgIHZpc2libGVcbiAgICA6dGVsZXBvcnRlZD1cImZhbHNlXCJcbiAgICB3aWR0aD1cImF1dG9cIlxuICA+XG4gICAgPHRlbXBsYXRlICNyZWZlcmVuY2U+XG4gICAgICA8ZWwtYnV0dG9uPnRlc3Q8L2VsLWJ1dHRvbj5cbiAgICA8L3RlbXBsYXRlPlxuICAgIDxlbC1jYXNjYWRlci1wYW5lbFxuICAgICAgc3R5bGU9XCJ3aWR0aDogZml0LWNvbnRlbnRcIiA6b3B0aW9ucz1cIltvcHRpb25zMl1cIlxuICAgIC8+XG4gIDwvZWwtcG9wb3Zlcj5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5jb25zdCBvcHRpb25zMiA9IHtcbiAgdmFsdWU6ICdndWlkZScsXG4gIGxhYmVsOiAnR3VpZGUnLFxuICBjaGlsZHJlbjogW1xuICAgIHtcbiAgICAgIHZhbHVlOiAnZGlzY2lwbGluZXMnLFxuICAgICAgbGFiZWw6ICdEaXNjaXBsaW5lcycsXG4gICAgICBjaGlsZHJlbjogW1xuICAgICAgICB7XG4gICAgICAgICAgdmFsdWU6ICdjb25zaXN0ZW5jeScsXG4gICAgICAgICAgbGFiZWw6ICdDb25zaXN0ZW5jeScsXG4gICAgICAgIH0sXG4gICAgICAgIHtcbiAgICAgICAgICB2YWx1ZTogJ2ZlZWRiYWNrJyxcbiAgICAgICAgICBsYWJlbDogJ0ZlZWRiYWNrJyxcbiAgICAgICAgfSxcbiAgICAgICAge1xuICAgICAgICAgIHZhbHVlOiAnZWZmaWNpZW5jeScsXG4gICAgICAgICAgbGFiZWw6ICdFZmZpY2llbmN5JyxcbiAgICAgICAgfSxcbiAgICAgICAge1xuICAgICAgICAgIHZhbHVlOiAnY29udHJvbGxhYmlsaXR5JyxcbiAgICAgICAgICBsYWJlbDogJ0NvbnRyb2xsYWJpbGl0eScsXG4gICAgICAgIH0sXG4gICAgICBdLFxuICAgIH0sXG4gICAge1xuICAgICAgdmFsdWU6ICduYXZpZ2F0aW9uJyxcbiAgICAgIGxhYmVsOiAnTmF2aWdhdGlvbicsXG4gICAgICBjaGlsZHJlbjogW1xuICAgICAgICB7XG4gICAgICAgICAgdmFsdWU6ICdzaWRlIG5hdicsXG4gICAgICAgICAgbGFiZWw6ICdTaWRlIE5hdmlnYXRpb24nLFxuICAgICAgICB9LFxuICAgICAgICB7XG4gICAgICAgICAgdmFsdWU6ICd0b3AgbmF2JyxcbiAgICAgICAgICBsYWJlbDogJ1RvcCBOYXZpZ2F0aW9uJyxcbiAgICAgICAgfSxcbiAgICAgIF0sXG4gICAgfSxcbiAgXSxcbn1cbjwvc2NyaXB0PlxuIiwiZWxlbWVudC1wbHVzLmpzIjoiaW1wb3J0IEVsZW1lbnRQbHVzIGZyb20gJ2VsZW1lbnQtcGx1cydcbmltcG9ydCB7IGdldEN1cnJlbnRJbnN0YW5jZSB9IGZyb20gJ3Z1ZSdcblxubGV0IGluc3RhbGxlZCA9IGZhbHNlXG5hd2FpdCBsb2FkU3R5bGUoKVxuXG5leHBvcnQgZnVuY3Rpb24gc2V0dXBFbGVtZW50UGx1cygpIHtcbiAgaWYgKGluc3RhbGxlZCkgcmV0dXJuXG4gIGNvbnN0IGluc3RhbmNlID0gZ2V0Q3VycmVudEluc3RhbmNlKClcbiAgaW5zdGFuY2UuYXBwQ29udGV4dC5hcHAudXNlKEVsZW1lbnRQbHVzKVxuICBpbnN0YWxsZWQgPSB0cnVlXG59XG5cbmV4cG9ydCBmdW5jdGlvbiBsb2FkU3R5bGUoKSB7XG4gIGNvbnN0IHN0eWxlcyA9IFsnaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L2Rpc3QvaW5kZXguY3NzJywgJ2h0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQGxhdGVzdC90aGVtZS1jaGFsay9kYXJrL2Nzcy12YXJzLmNzcyddLm1hcCgoc3R5bGUpID0+IHtcbiAgICByZXR1cm4gbmV3IFByb21pc2UoKHJlc29sdmUsIHJlamVjdCkgPT4ge1xuICAgICAgY29uc3QgbGluayA9IGRvY3VtZW50LmNyZWF0ZUVsZW1lbnQoJ2xpbmsnKVxuICAgICAgbGluay5yZWwgPSAnc3R5bGVzaGVldCdcbiAgICAgIGxpbmsuaHJlZiA9IHN0eWxlXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2xvYWQnLCByZXNvbHZlKVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdlcnJvcicsIHJlamVjdClcbiAgICAgIGRvY3VtZW50LmJvZHkuYXBwZW5kKGxpbmspXG4gICAgfSlcbiAgfSlcbiAgcmV0dXJuIFByb21pc2UuYWxsU2V0dGxlZChzdHlsZXMpXG59XG4iLCJ0c2NvbmZpZy5qc29uIjoie1xuICBcImNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogXCJFU05leHRcIixcbiAgICBcImpzeFwiOiBcInByZXNlcnZlXCIsXG4gICAgXCJtb2R1bGVcIjogXCJFU05leHRcIixcbiAgICBcIm1vZHVsZVJlc29sdXRpb25cIjogXCJCdW5kbGVyXCIsXG4gICAgXCJ0eXBlc1wiOiBbXCJlbGVtZW50LXBsdXMvZ2xvYmFsLmQudHNcIl0sXG4gICAgXCJhbGxvd0ltcG9ydGluZ1RzRXh0ZW5zaW9uc1wiOiB0cnVlLFxuICAgIFwiYWxsb3dKc1wiOiB0cnVlLFxuICAgIFwiY2hlY2tKc1wiOiB0cnVlXG4gIH0sXG4gIFwidnVlQ29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiAzLjNcbiAgfVxufVxuIiwiUGxheWdyb3VuZE1haW4udnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCBBcHAgZnJvbSAnLi9BcHAudnVlJ1xuaW1wb3J0IHsgc2V0dXBFbGVtZW50UGx1cyB9IGZyb20gJy4vZWxlbWVudC1wbHVzLmpzJ1xuc2V0dXBFbGVtZW50UGx1cygpXG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8QXBwIC8+XG48L3RlbXBsYXRlPlxuIiwiaW1wb3J0LW1hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge1xuICAgIFwidnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AdnVlL3J1bnRpbWUtZG9tQGxhdGVzdC9kaXN0L3J1bnRpbWUtZG9tLmVzbS1icm93c2VyLmpzXCIsXG4gICAgXCJAdnVlL3NoYXJlZFwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9zaGFyZWRAbGF0ZXN0L2Rpc3Qvc2hhcmVkLmVzbS1idW5kbGVyLmpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXNcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvZGlzdC9pbmRleC5mdWxsLm1pbi5tanNcIixcbiAgICBcImVsZW1lbnQtcGx1cy9cIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvXCIsXG4gICAgXCJAZWxlbWVudC1wbHVzL2ljb25zLXZ1ZVwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQGVsZW1lbnQtcGx1cy9pY29ucy12dWVAMi9kaXN0L2luZGV4Lm1pbi5qc1wiXG4gIH0sXG4gIFwic2NvcGVzXCI6IHt9XG59IiwiX28iOnt9fQ==)
[After](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtcG9wb3ZlclxuICAgIHZpc2libGVcbiAgICA6dGVsZXBvcnRlZD1cImZhbHNlXCJcbiAgICB3aWR0aD1cImF1dG9cIlxuICA+XG4gICAgPHRlbXBsYXRlICNyZWZlcmVuY2U+XG4gICAgICA8ZWwtYnV0dG9uPnRlc3Q8L2VsLWJ1dHRvbj5cbiAgICA8L3RlbXBsYXRlPlxuICAgIDxlbC1jYXNjYWRlci1wYW5lbFxuICAgICAgc3R5bGU9XCJ3aWR0aDogZml0LWNvbnRlbnRcIiA6b3B0aW9ucz1cIltvcHRpb25zMl1cIlxuICAgIC8+XG4gIDwvZWwtcG9wb3Zlcj5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5jb25zdCBvcHRpb25zMiA9IHtcbiAgdmFsdWU6ICdndWlkZScsXG4gIGxhYmVsOiAnR3VpZGUnLFxuICBjaGlsZHJlbjogW1xuICAgIHtcbiAgICAgIHZhbHVlOiAnZGlzY2lwbGluZXMnLFxuICAgICAgbGFiZWw6ICdEaXNjaXBsaW5lcycsXG4gICAgICBjaGlsZHJlbjogW1xuICAgICAgICB7XG4gICAgICAgICAgdmFsdWU6ICdjb25zaXN0ZW5jeScsXG4gICAgICAgICAgbGFiZWw6ICdDb25zaXN0ZW5jeScsXG4gICAgICAgIH0sXG4gICAgICAgIHtcbiAgICAgICAgICB2YWx1ZTogJ2ZlZWRiYWNrJyxcbiAgICAgICAgICBsYWJlbDogJ0ZlZWRiYWNrJyxcbiAgICAgICAgfSxcbiAgICAgICAge1xuICAgICAgICAgIHZhbHVlOiAnZWZmaWNpZW5jeScsXG4gICAgICAgICAgbGFiZWw6ICdFZmZpY2llbmN5JyxcbiAgICAgICAgfSxcbiAgICAgICAge1xuICAgICAgICAgIHZhbHVlOiAnY29udHJvbGxhYmlsaXR5JyxcbiAgICAgICAgICBsYWJlbDogJ0NvbnRyb2xsYWJpbGl0eScsXG4gICAgICAgIH0sXG4gICAgICBdLFxuICAgIH0sXG4gICAge1xuICAgICAgdmFsdWU6ICduYXZpZ2F0aW9uJyxcbiAgICAgIGxhYmVsOiAnTmF2aWdhdGlvbicsXG4gICAgICBjaGlsZHJlbjogW1xuICAgICAgICB7XG4gICAgICAgICAgdmFsdWU6ICdzaWRlIG5hdicsXG4gICAgICAgICAgbGFiZWw6ICdTaWRlIE5hdmlnYXRpb24nLFxuICAgICAgICB9LFxuICAgICAgICB7XG4gICAgICAgICAgdmFsdWU6ICd0b3AgbmF2JyxcbiAgICAgICAgICBsYWJlbDogJ1RvcCBOYXZpZ2F0aW9uJyxcbiAgICAgICAgfSxcbiAgICAgIF0sXG4gICAgfSxcbiAgXSxcbn1cbjwvc2NyaXB0PlxuIiwiZWxlbWVudC1wbHVzLmpzIjoiaW1wb3J0IEVsZW1lbnRQbHVzIGZyb20gJ2VsZW1lbnQtcGx1cydcbmltcG9ydCB7IGdldEN1cnJlbnRJbnN0YW5jZSB9IGZyb20gJ3Z1ZSdcblxubGV0IGluc3RhbGxlZCA9IGZhbHNlXG5hd2FpdCBsb2FkU3R5bGUoKVxuXG5leHBvcnQgZnVuY3Rpb24gc2V0dXBFbGVtZW50UGx1cygpIHtcbiAgaWYgKGluc3RhbGxlZCkgcmV0dXJuXG4gIGNvbnN0IGluc3RhbmNlID0gZ2V0Q3VycmVudEluc3RhbmNlKClcbiAgaW5zdGFuY2UuYXBwQ29udGV4dC5hcHAudXNlKEVsZW1lbnRQbHVzKVxuICBpbnN0YWxsZWQgPSB0cnVlXG59XG5cbmV4cG9ydCBmdW5jdGlvbiBsb2FkU3R5bGUoKSB7XG4gIGNvbnN0IHN0eWxlcyA9IFsnaHR0cHM6Ly9wcmV2aWV3LTIyMDMwLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5jc3MnLCAnaHR0cHM6Ly9wcmV2aWV3LTIyMDMwLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvdGhlbWUtY2hhbGsvZGFyay9jc3MtdmFycy5jc3MnXS5tYXAoKHN0eWxlKSA9PiB7XG4gICAgcmV0dXJuIG5ldyBQcm9taXNlKChyZXNvbHZlLCByZWplY3QpID0+IHtcbiAgICAgIGNvbnN0IGxpbmsgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCdsaW5rJylcbiAgICAgIGxpbmsucmVsID0gJ3N0eWxlc2hlZXQnXG4gICAgICBsaW5rLmhyZWYgPSBzdHlsZVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdsb2FkJywgcmVzb2x2ZSlcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignZXJyb3InLCByZWplY3QpXG4gICAgICBkb2N1bWVudC5ib2R5LmFwcGVuZChsaW5rKVxuICAgIH0pXG4gIH0pXG4gIHJldHVybiBQcm9taXNlLmFsbFNldHRsZWQoc3R5bGVzKVxufVxuIiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IFwiRVNOZXh0XCIsXG4gICAgXCJqc3hcIjogXCJwcmVzZXJ2ZVwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwidHlwZXNcIjogW1wiZWxlbWVudC1wbHVzL2dsb2JhbC5kLnRzXCJdLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcImFsbG93SnNcIjogdHJ1ZSxcbiAgICBcImNoZWNrSnNcIjogdHJ1ZVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9ydW50aW1lLWRvbUBsYXRlc3QvZGlzdC9ydW50aW1lLWRvbS5lc20tYnJvd3Nlci5qc1wiLFxuICAgIFwiQHZ1ZS9zaGFyZWRcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvc2hhcmVkQGxhdGVzdC9kaXN0L3NoYXJlZC5lc20tYnVuZGxlci5qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9wcmV2aWV3LTIyMDMwLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5mdWxsLm1pbi5tanNcIixcbiAgICBcImVsZW1lbnQtcGx1cy9cIjogXCJ1bnN1cHBvcnRlZFwiLFxuICAgIFwiQGVsZW1lbnQtcGx1cy9pY29ucy12dWVcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlQDIvZGlzdC9pbmRleC5taW4uanNcIlxuICB9LFxuICBcInNjb3Blc1wiOiB7fVxufSIsIl9vIjp7InNob3dIaWRkZW4iOnRydWUsInN0eWxlU291cmNlIjoiaHR0cHM6Ly9wcmV2aWV3LTIyMDMwLWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5jc3MifX0=)

Dsaquel (approved):
Nice catch!

Can you add a relevant test case ?

micaiguai:
> Nice catch!
> 
> Can you add a relevant test case ?

Thanks for your suggestions. I've added the relevant test case.

Dsaquel (approved):
Thanks!

rzzf:
> The cascader-panel will update after the parent focus-trap component updated(trigger after first click). The cascader-panel then compares the props, although the prop content is same as before, it's a new array, Causing cascader-panel re-init the nodes.

It seems that even though `props.options` hasn’t actually changed, `initStore` is still re-executed.

https://github.com/element-plus/element-plus/blob/74387dc1a7f295dc4b39f423ba0868910686df19/packages/components/cascader-panel/src/index.vue#L366

I think it's meaningful to compare `props.options` using `isEqual`;
otherwise, it would trigger unnecessary update logic.

WDYT? @Dsaquel @micaiguai 

update:
The issue mentioned in #20269 did not exist before version 2.10.4.

packages/components/cascader-panel/__tests__/cascader-panel.test.tsx
```diff
@@ -1,9 +1,10 @@
-import { nextTick, ref } from 'vue'
+import { nextTick, ref, withModifiers } from 'vue'
 import { mount } from '@vue/test-utils'
 import { beforeEach, describe, expect, test, vi } from 'vitest'
 import { Check, Loading } from '@element-plus/icons-vue'
 import CascaderPanel from '../src/index.vue'
 import Node from '../src/node.vue'
+import { ElCascaderPanel } from '@element-plus/components'
```
rzzf:
```suggestion
```

packages/components/cascader-panel/__tests__/cascader-panel.test.tsx
```diff
@@ -1051,4 +1052,32 @@ describe('CascaderPanel.vue', () => {
...
+            value: '2',
+          },
+        ],
+      }
+      const foo = ref('0')
+
+      return (
+        <div onClick={withModifiers(() => (foo.value = '1'), ['capture'])}>
+          {foo.value}
+          <ElCascaderPanel options={[option]} />
```
rzzf:
```suggestion
          <CascaderPanel options={[option]} />
```

micaiguai:
> > The cascader-panel will update after the parent focus-trap component updated(trigger after first click). The cascader-panel then compares the props, although the prop content is same as before, it's a new array, Causing cascader-panel re-init the nodes.
> 
> It seems that even though `props.options` hasn’t actually changed, `initStore` is still re-executed.
> 
> https://github.com/element-plus/element-plus/blob/74387dc1a7f295dc4b39f423ba0868910686df19/packages/components/cascader-panel/src/index.vue#L366
> 
> I think it's meaningful to compare `props.options` using `isEqual`; otherwise, it would trigger unnecessary update logic.
> 
> WDYT? @Dsaquel @micaiguai
> 
> update: The issue mentioned in #20269 did not exist before version 2.10.4.

I have tried to compare `props.options` using `isEqual`. However, issue #21402 will still occur. This is because `oldVal` and `newVal` are variables pointint to the same reference address-so they will always be equal when only the content of the array is modified without changing the variable's address.

micaiguai:
> > The cascader-panel will update after the parent focus-trap component updated(trigger after first click). The cascader-panel then compares the props, although the prop content is same as before, it's a new array, Causing cascader-panel re-init the nodes.
> 
> It seems that even though `props.options` hasn’t actually changed, `initStore` is still re-executed.
> 
> https://github.com/element-plus/element-plus/blob/74387dc1a7f295dc4b39f423ba0868910686df19/packages/components/cascader-panel/src/index.vue#L366
> 
> I think it's meaningful to compare `props.options` using `isEqual`; otherwise, it would trigger unnecessary update logic.
> 
> WDYT? @Dsaquel @micaiguai
> 
> update: The issue mentioned in #20269 did not exist before version 2.10.4.

Thanks for your suggestions. However, I find it inconvient to implement the logic for storing old values in the existing code that watches `config` and `props.options`. Therefore, I've written new watcher code for storage purposes.

rzzf (approved):
LGTM

packages/components/cascader-panel/src/index.vue
```diff
@@ -363,9 +363,16 @@ watch(
   }
 )
 
-watch(() => props.options, initStore, {
-  deep: true,
-})
+watch(
+  () => props.options,
+  (newVal, oldVal) => {
+    if (newVal !== oldVal && isEqual(newVal, oldVal)) return
```
Dsaquel, line 369: `newVal !== oldVal` invert the condition to get the test cases passes but it is very hacky and maybe instable as `newVal` and `oldVal` should be the same object reference.

rzzf: > newVal and oldVal should be the same object reference.

When using literals in the template or manually changing the options reference, `newVal !== oldValue`.

Please See [Vue Play](https://play.vuejs.org/#eNp9VMlu2zAQ/ZWBLpZhRW7R5eDKQZMgh7RAG3S7VEGhSGOZCUUSFOXYEPTvHZKSLaepD7bIWd5sfNMGF0rFmwaDRZDUuWbKQI2mUeepYJWS2kALGlcRPGUmX0MHKy0rmJDHZG9xJSvVy+O5vVhAUqcil6I28Ig7WFqU8NV0kEllGB16ucAnuNA624XvpvGKcU6WcZWpMPwTAZvC8hzCFnh2j3wBLIJNxhukE3TTKUFyNMAMVjcF4fXIsbOJOYrSrD8ckllnouB4RZ8SyTp04G0q4NiRVL/jOD6S3aWiew5029TrGwp9AipWZDNK3yV6qMHn3VEZ0F9mM8rXRUrmfiY0DbqQUvHMIN0AkvvGGCngY85Z/rhMg3FhaXCe+wqpu6hR5JjMvcPLzjSi2Yy8sqKAz7g7bXxcOXnZAl3uIz/nWbCNgwBoW/cMOirLXhP3ZhZ9nwhz1Oy7NIC5jzz3AMl8VDpda7Pj9tgn5lpeZbpk4kyzcm0W8Pq92vomUguddRAFpqbJrVgZP9RS0IN3fmmQUyqMo/7qw6fBwiNaXca5fPrkZEY3GA3yfI354wvyh3prZWlwq7FGvaFJ7HWGMkTj1dffv+CWzntlJYuGk/UJ5TesJW9sjt7ssrFj0CM7l+2NIyUT5Y/6emtQ1ENRNlFr2Tn7NCCS2in8r/RDum/it86P+kldHAh+amO4ZRGBFJe4khp/qoJm92x5DDxSWiq7BwpcMYG39haOSbTwm4GSthxJhcMOPdmc7/BuIrB75FfGKTAv6Hvgo40kOW0DWfY2sFzSqvBmhBxR0gWi8l1ybEzFcfp9yH/wJu4pa8uyAvXEotHvFHMb3pMi4Qw2ZxSCCEBF0KojiC39D7UTExZEG1I7DVGtbUnVdcmcs54iDu2IIUH3F5bZ/BU=)

Dsaquel: Thanks for sharing,

After re-testing, even with the condition we will get unexpected watch triggers.

Taking a step back even if it's a regression i don't think we should handle object literal in the template.

rzzf: Okay, let's revert to https://github.com/element-plus/element-plus/pull/22030/commits/8e8f54ad36e2714bd640719d5d60531df68689f9
@micaiguai

Dsaquel: My though is just to guide the developer like https://github.com/element-plus/element-plus/pull/4689 but in the issue.

Dsaquel: The value of a node can be duplicated as well so the previous fix may not fix all cases.

rzzf: You make a great point!👍
My idea is to address the literal value issue with minimal cost, just like the current implementation. 
Indeed, it can’t completely prevent repeated execution when the reference remains the same.

Dsaquel: Okay i got your point, let's seek other people opinion's.

@element-plus/backers
Do you think we should handle specifically object literal in template for options prop ?
```vue
<template :options="[option]">
```

rzzf: > My though is just to guide the developer like #4689 but in the issue.

Sorry for my foolishness. 🙇

After careful consideration and investigation, I agree with this point of view. [This issue](https://github.com/vuejs/eslint-plugin-vue/issues/2516) already exists in Vue itself, and it’s better addressed through documentation or [ESLint rules](https://github.com/vuejs/eslint-plugin-vue/issues/3000) rather than being something we need to handle. 🫠

btea: My opinion is that it's best not to handle this, and ideally, developers should avoid this approach. Component libraries contain many similar configurations, and dealing with each one individually can feel exhausting.

By the way, @rzzf 's implementation of ESLint rules is excellent. 👍

rzzf:
In any case, thank you very much for your contribution!

PR #22051, opened 2025-09-05
fix: circular deps cause build size increase

warmthsea:
### Rel:

- https://github.com/element-plus/element-plus/pull/21905#issuecomment-3223508001

### After

image


btea:
What is the relationship of this circular dependency? `aria.ts` and `types.ts` do not seem to reference each other. 🤔

warmthsea:
> What is the relationship of this circular dependency? `aria.t` and `type.ts` do not seem to reference each other. 🤔

This change did fix the problem.


It's a bit confusing, but this does seem to create a loop.
utils/index.ts → dom/index.ts → aria.ts → types.ts → utils/index.ts

`isString` doesn't because it's form installed deps..

btea (approved):
The code looks fine, but the dependencies are a bit weird.

PR #22077, opened 2025-09-08
refactor(components): [checkout] adjust the TS type

lxKylin:

image

### DIFF
--- a/packages/components/checkbox/src/composables/use-checkbox-disabled.ts
+++ b/packages/components/checkbox/src/composables/use-checkbox-disabled.ts
@@ -15,12 +15,8 @@ export const useCheckboxDisabled = ({
     const max = checkboxGroup?.max?.value
     const min = checkboxGroup?.min?.value
     return (
-      (!isUndefined(max) &&
-        (model.value as any[]).length >= max &&
-        !isChecked.value) ||
-      (!isUndefined(min) &&
-        (model.value as any[]).length <= min &&
-        isChecked.value)
+      (!isUndefined(max) && model.value.length >= max && !isChecked.value) ||
+      (!isUndefined(min) && model.value.length <= min && isChecked.value)
     )
   })
 
--- a/packages/components/checkbox/src/composables/use-checkbox-disabled.ts
+++ b/packages/components/checkbox/src/composables/use-checkbox-disabled.ts
@@ -15,8 +15,12 @@ export const useCheckboxDisabled = ({
     const max = checkboxGroup?.max?.value
     const min = checkboxGroup?.min?.value
     return (
-      (!isUndefined(max) && model.value.length >= max && !isChecked.value) ||
-      (!isUndefined(min) && model.value.length <= min && isChecked.value)
+      (!isUndefined(max) &&
+        (model.value as any[]).length >= max &&
+        !isChecked.value) ||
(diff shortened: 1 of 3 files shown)


Dsaquel (commented):
The type of `model` in my editor is `WritableComputedRef<any>` and i don't have a type error, that's weird on vscode volar fails quite often 🤔 .

We can give `model` an explicit type:

https://github.com/lxKylin/element-plus/blob/a7e52f40a84f009bfbf86ae4f40e10c2077053b7/packages/components/checkbox/src/composables/use-checkbox-model.ts#L14
```diff
-  const model = computed({
+  const model = computed<any>({
```

lxKylin:
> The type of `model` in my editor is `WritableComputedRef<any>` and i don't have a type error, that's weird on vscode volar fails quite often 🤔 .
> 
> We can give `model` an explicit type:
> 
> https://github.com/lxKylin/element-plus/blob/a7e52f40a84f009bfbf86ae4f40e10c2077053b7/packages/components/checkbox/src/composables/use-checkbox-model.ts#L14
> 
> ```diff
> -  const model = computed({
> +  const model = computed<any>({
> ```

ok

Dsaquel (approved):
Thanks!

btea:
image

It seems that TS treats the value type as any, and no explicit declaration seems to be needed.

btea:
> The type of `model` in my editor is `WritableComputedRef<any>` and i don't have a type error, that's weird on vscode volar fails quite often 🤔 .
> 

I guess the ts version used by @lxKylin is inconsistent with our project.

lxKylin:
> > The type of `model` in my editor is `WritableComputedRef<any>` and i don't have a type error, that's weird on vscode volar fails quite often 🤔 .
> 
> I guess the ts version used by @lxKylin is inconsistent with our project.

image

This is consistent.

btea:
image

Check whether you are using the ts version of the project or the ts version of vscode.

lxKylin:
> image
> Check whether you are using the ts version of the project or the ts version of vscode.检查您使用的是项目的 ts 版本还是 ts 版本的 vscode。

image
It is indeed inconsistent.

btea:
If you click Select Version and select the workspace ts version, there should be no problem.

lxKylin:
> If you click Select Version and select the workspace ts version, there should be no problem.

Yes, there is no problem after selecting the workspace version. But I'm thinking that there might be others like me who haven't used the workspace version. So, would it be better to display the type as 'any' for ts?🤔

btea:
Because this is the TS version currently used in the source code repository, there are no issues with the source code under the current TS version. If we encounter problems when upgrading the TS version later, We will synchronously fix them. 

If there are similar problems caused by using an incompatible TS version in the project, we recommend switching to a compatible version. I think cross-version handling issues will lead to uncontrollable problems.

lxKylin:
> Because this is the TS version currently used in the source code repository, there are no issues with the source code under the current TS version. If we encounter problems when upgrading the TS version later, We will synchronously fix them.
> 
> If there are similar problems caused by using an incompatible TS version in the project, we recommend switching to a compatible version. I think cross-version handling issues will lead to uncontrollable problems.

👌

btea:
For the reasons above, I will close this PR. Thank you for your contribution though.

PR #22140, opened 2025-09-12
fix(components): [date-picker] allow plain array for modelValue

Dsaquel:
improve & close https://github.com/element-plus/element-plus/pull/20023

Allow without type error:

<ol>
  <li>

```vue
<template>
  <el-date-picker v-model="value" type="dates" />
</template>

<script setup lang="ts">
...

const value = ref<Date[]>([])
</script>
```
</li>

  <li>

```vue
<template>
  <el-date-picker v-model="value" type="dates" value-format="x" />
</template>

<script setup lang="ts">
...

const value = ref<number[]>([])
</script>
```
</li>
</ol>


warmthsea (approved):
LGTM

docs/en-US/component/date-picker-panel.md
```diff
@@ -60,7 +60,7 @@ Note, date time locale (month name, first day of the week ...) are also configur
...
 | --------------------- | ---------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------- |
-| model-value / v-model | binding value, if it is an array, the length should be 2                                                                     | ^[number] / ^[string] / ^[object]`Date \| [Date, Date] \| [string, string]`                                                                                    | ''         |
+| model-value / v-model | binding value, if it is an `range` picker, the length of the array should be 2                                               | ^[number] / ^[string] / ^[Date]`number[] \| string[] \| Date[]`                                                                                                | ''         |
```
btea:
```suggestion
| model-value / v-model | binding value, if it is an `range` picker, the length of the array should be 2                                               | ^[number] / ^[string] / ^[object]`number[] \| string[] \| Date[]`                                                                                                | ''         |
```
Is it more reasonable to use object here?

Dsaquel: ^[Date] is valid type and i think it's important to keep the enumeration number-string-Date in order to compare with the tooltip enumeration.
image

tolking:
```suggestion
| model-value / v-model | binding value, if it is an `range` picker, the length of the array should be 2                                               | ^[number] / ^[string] / ^[Date] / ^[array]`number[] \| string[] \| Date[]`                                                                                                | ''         |
```

Date and array should be distinguished.

Dsaquel: Sorry i don't understand, you want the tuple types back ?

tolking: ^[number] 
^[string] 
^[Date]
^[array]`number[] \| string[] \| Date[]`

Dsaquel: My bad, i just not seen the commit suggestion --'.

tolking (approved):

btea (approved):

PR #22147, opened 2025-09-14
feat(components): [tree-select] expose `treeRef` & `selectRef` instances

Dsaquel:
Rel https://github.com/element-plus/element-plus/pull/22105#discussion_r2340092218

The adding `tree` & `select` exposes will respect tree & select exposed instances.

It will be a breaking change when we'll remove the older methods exposed.


btea (approved):

packages/components/tree-select/src/tree-select.vue
```diff
@@ -66,6 +67,8 @@ export default defineComponent({
           'insertAfter',
         ]),
         ...pick(select.value, ['focus', 'blur', 'selectedLabel']),
+        tree: tree.value,
```
btea: Maybe it would be better to change to `treeRef`?

Dsaquel: Do you mean this ?
```diff
-        tree: tree.value,
-        select: select.value,
+        treeRef: tree.value,
+        selectRef: select.value,
```
Both are fine for me.

btea: Yes.

btea: refer to https://github.com/element-plus/element-plus/blob/25a25c6aec8badb9272c8574221489df05380603/packages/components/cascader/src/cascader.vue#L774

Dsaquel: Okay I'll change :+1:

btea (approved):

PR #22150, opened 2025-09-15
style(components): [checkbox & radio] correct cursor style in disabled state

rzzf:
Before

https://github.com/user-attachments/assets/62eeefa5-9cf7-4232-a667-8758794fe422

After

https://github.com/user-attachments/assets/0463d9e7-a0bb-439e-9361-bdb90a732773

### DIFF
--- a/packages/theme-chalk/src/checkbox.scss
+++ b/packages/theme-chalk/src/checkbox.scss
@@ -119,6 +119,8 @@ $checkbox-bordered-input-width: map.merge(
     position: relative;
 
     @include when(disabled) {
+      cursor: not-allowed;
+
       .#{$namespace}-checkbox__inner {
         background-color: getCssVar('checkbox-disabled-input-fill');
         border-color: getCssVar('checkbox-disabled-border-color');
--- a/packages/theme-chalk/src/radio.scss
+++ b/packages/theme-chalk/src/radio.scss
@@ -39,6 +39,10 @@ $radio-font-size: map.merge(
     }
   }
 
+  @include when(disabled) {
+    cursor: not-allowed;
+  }
+
   @include when(bordered) {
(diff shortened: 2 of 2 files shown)


btea:
Can you provide a sample code? I check the style on the official website is correct. 🤔

rzzf:
> Can you provide a sample code? I check the style on the official website is correct. 🤔

I can still reproduce this issue in the documentation, and it’s particularly noticeable in the [Radio](https://element-plus.org/en-US/component/radio.html#disabled) component: 

one case is when the mouse hovers over `el-radio`, and the other is when it hovers over `el-radio__input`.


https://github.com/user-attachments/assets/bae44e40-5251-44f6-b3c3-3f57caf11dbf

rzzf:
> Can you provide a sample code? I check the style on the official website is correct. 🤔

You may try increasing the browser zoom level to above 200%, which makes the issue easier to reproduce.

btea:
I have no problem opening it in my local browser. 🤔

cjn021:
I also reproduced this issue on the official website

1341757918288_ pic

btea:
> I also reproduced this issue on the official website

It looks like your mouse has just moved to the edge. If you move it a little further forward, it should be normal.

cjn021:
> > I also reproduced this issue on the official website
> 
> It looks like your mouse has just moved to the edge. If you move it a little further forward, it should be normal.

Yeah, the cursor turns into a pointer when it’s at the edge, which feels a bit strange and makes it seem like something is clickable

rzzf:
Checkbox and Radio components cursor behavior seems to be inconsistent, see screenshot

| Checkbox | Radio |
|-|-|
|image | image

Also mentioned by @Lensiq, the cursor changes to pointer when moving to the input edge. Is this expected?

PR #22024, opened 2025-09-03
feat(components): [tabs] support scrolling via mouse wheel

rzzf:
Currently, when there are many tabs, relying solely on the "previous" and "next" buttons for scrolling is not very user-friendly on large screens, as users have to move the cursor a long distance. I think supporting mouse wheel scrolling would provide a smoother experience.

img_v3_02pq_183f18c2-ea7e-4b7a-88ec-f0ae2de9f3dg

[...]

PR #22168, opened 2025-09-15
feat(components): [virtual-list] support horizontal scrolling with mouse wheel on Windows/ubuntu

### LINKED ISSUE #22370: [Component] [tabs] Tabs cannot scroll when positioned on the top or bottom .
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.22`
- Element Plus Version: `2.11.4`
- Browser / OS: `Chrome/139.0.0.0`
- Build Tool: `Vite`

### Related Component

- `el-tabs`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtdGFic1xuICAgIHYtbW9kZWw9XCJ0YWJzVmFsdWVcIlxuICA+XG4gICAgPGVsLXRhYi1wYW5lXG4gICAgICB2LWZvcj1cIml0ZW0gaW4gbVRhYnNcIlxuICAgICAgOmtleT1cIml0ZW0ubmFtZVwiXG4gICAgICA6bGFiZWw9XCJpdGVtLnRpdGxlXCJcbiAgICAgIDpuYW1lPVwiaXRlbS5uYW1lXCJcbiAgICA+XG4gICAgICB7eyBpdGVtLnRpdGxlIH19XG4gICAgPC9lbC10YWItcGFuZT5cbiAgPC9lbC10YWJzPlxuICA8ZWwtZGl2aWRlciAvPlxuICA8ZWwtdGFic1xuICAgIHYtbW9kZWw9XCJ0YWJzVmFsdWVcIlxuICAgIHRhYi1wb3NpdGlvbj1cImxlZnRcIlxuICAgIHN0eWxlPVwiaGVpZ2h0OiA0MDBweFwiXG4gID5cbiAgICA8ZWwtdGFiLXBhbmVcbiAgICAgIHYtZm9yPVwiaXRlbSBpbiBtVGFic1wiXG4gICAgICA6a2V5PVwiaXRlbS5uYW1lXCJcbiAgICAgIDpsYWJlbD1cIml0ZW0udGl0bGVcIlxuICAgICAgOm5hbWU9XCJpdGVtLm5hbWVcIlxuICAgID5cbiAgICAgIHt7IGl0ZW0udGl0bGUgfX1cbiAgICA8L2VsLXRhYi1wYW5lPlxuICA8L2VsLXRhYnM+XG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IGxhbmc9XCJ0c1wiIHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSAndnVlJ1xuXG5cbmxldCB0YWJJbmRleCA9IDJcbmNvbnN0IHRhYnNWYWx1ZSA9IHJlZigndGFiMScpXG5jb25zdCBtVGFicyA9IHJlZihbXG5dKVxuZm9yKGxldCBpID0gMTsgaSA8PSA1MDsgaSsrKSB7XG4gIG1UYWJzLnZhbHVlLnB1c2goe1xuICAgIHRpdGxlOiAnVGFiJyArIGksXG4gICAgbmFtZTogJ3RhYicgKyBpXG4gIH0pXG59XG48L3NjcmlwdD5cblxuPHN0eWxlPlxuLmRlbW8tdGFicyA+IC5lbC10YWJzX19jb250ZW50IHtcbiAgcGFkZGluZzogMzJweDtcbiAgY29sb3I6ICM2Yjc3OGM7XG4gIGZvbnQtc2l6ZTogMzJweDtcbiAgZm9udC13ZWlnaHQ6IDYwMDtcbn1cbjwvc3R5bGU+XG4iLCJlbGVtZW50LXBsdXMuanMiOiJpbXBvcnQgRWxlbWVudFBsdXMgZnJvbSAnZWxlbWVudC1wbHVzJ1xuaW1wb3J0IHsgZ2V0Q3VycmVudEluc3RhbmNlIH0gZnJvbSAndnVlJ1xuXG5sZXQgaW5zdGFsbGVkID0gZmFsc2VcbmF3YWl0IGxvYWRTdHlsZSgpXG5cbmV4cG9ydCBmdW5jdGlvbiBzZXR1cEVsZW1lbnRQbHVzKCkge1xuICBpZiAoaW5zdGFsbGVkKSByZXR1cm5cbiAgY29uc3QgaW5zdGFuY2UgPSBnZXRDdXJyZW50SW5zdGFuY2UoKVxuICBpbnN0YW5jZS5hcHBDb250ZXh0LmFwcC51c2UoRWxlbWVudFBsdXMpXG4gIGluc3RhbGxlZCA9IHRydWVcbn1cblxuZXhwb3J0IGZ1bmN0aW9uIGxvYWRTdHlsZSgpIHtcbiAgY29uc3Qgc3R5bGVzID0gWydodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvZGlzdC9pbmRleC5jc3MnLCAnaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L3RoZW1lLWNoYWxrL2RhcmsvY3NzLXZhcnMuY3NzJ10ubWFwKChzdHlsZSkgPT4ge1xuICAgIHJldHVybiBuZXcgUHJvbWlzZSgocmVzb2x2ZSwgcmVqZWN0KSA9PiB7XG4gICAgICBjb25zdCBsaW5rID0gZG9jdW1lbnQuY3JlYXRlRWxlbWVudCgnbGluaycpXG4gICAgICBsaW5rLnJlbCA9ICdzdHlsZXNoZWV0J1xuICAgICAgbGluay5ocmVmID0gc3R5bGVcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignbG9hZCcsIHJlc29sdmUpXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2Vycm9yJywgcmVqZWN0KVxuICAgICAgZG9jdW1lbnQuYm9keS5hcHBlbmQobGluaylcbiAgICB9KVxuICB9KVxuICByZXR1cm4gUHJvbWlzZS5hbGxTZXR0bGVkKHN0eWxlcylcbn1cbiIsInRzY29uZmlnLmpzb24iOiJ7XG4gIFwiY29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiBcIkVTTmV4dFwiLFxuICAgIFwianN4XCI6IFwicHJlc2VydmVcIixcbiAgICBcIm1vZHVsZVwiOiBcIkVTTmV4dFwiLFxuICAgIFwibW9kdWxlUmVzb2x1dGlvblwiOiBcIkJ1bmRsZXJcIixcbiAgICBcInR5cGVzXCI6IFtcImVsZW1lbnQtcGx1cy9nbG9iYWwuZC50c1wiXSxcbiAgICBcImFsbG93SW1wb3J0aW5nVHNFeHRlbnNpb25zXCI6IHRydWUsXG4gICAgXCJhbGxvd0pzXCI6IHRydWUsXG4gICAgXCJjaGVja0pzXCI6IHRydWVcbiAgfSxcbiAgXCJ2dWVDb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IDMuM1xuICB9XG59XG4iLCJQbGF5Z3JvdW5kTWFpbi52dWUiOiI8c2NyaXB0IHNldHVwPlxuaW1wb3J0IEFwcCBmcm9tICcuL0FwcC52dWUnXG5pbXBvcnQgeyBzZXR1cEVsZW1lbnRQbHVzIH0gZnJvbSAnLi9lbGVtZW50LXBsdXMuanMnXG5zZXR1cEVsZW1lbnRQbHVzKClcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxBcHAgLz5cbjwvdGVtcGxhdGU+XG4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvcnVudGltZS1kb21AbGF0ZXN0L2Rpc3QvcnVudGltZS1kb20uZXNtLWJyb3dzZXIuanNcIixcbiAgICBcIkB2dWUvc2hhcmVkXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AdnVlL3NoYXJlZEBsYXRlc3QvZGlzdC9zaGFyZWQuZXNtLWJ1bmRsZXIuanNcIixcbiAgICBcImVsZW1lbnQtcGx1c1wiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQGxhdGVzdC9kaXN0L2luZGV4LmZ1bGwubWluLm1qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzL1wiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQGxhdGVzdC9cIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJfbyI6e319)

### Steps to reproduce

I noticed the scrolling feature with the mouse wheel on the Tabs, so I tried it out. I found that mouse scrolling only works when the tabs are positioned on the left or right side. It doesn't work when the tabs are at the top or bottom.

### What is Expected?

Tabs can be scrolled with the mouse wheel when they are positioned at the top or bottom.

Tabs cannot be scrolled with the mouse wheel when they are positioned at the top or bottom.

### Additional comments

_(empty)_

keeplearning66:

image

resolve #22370

Currently, when there are many tabs in the horizontal direction, users can only scroll through the tabs using a **touchpad**, as the **mouse wheel** cannot be used for scrolling since `deltaX` is always `0`(On my computer, holding the `Shift` key doesn't enable horizontal scrolling of the tabs either). 
If users are accustomed to using a mouse while browsing the web, they may prefer scrolling with the mouse wheel. Therefore, directly supporting horizontal scrolling with the mouse wheel could provide a better experience.


rzzf:
> Currently, when there are many tabs in the horizontal direction, users can only scroll through the tabs using a **touchpad**, as the **mouse wheel** cannot be used for scrolling since `deltaX` is always `0`

#22024 includes support for mouse scrolling.

Note that you should hold down the `shift` key when scrolling horizontally, This is consistent with native scrolling.

[Demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8c3Ryb25nPk5hdGl2ZTwvc3Ryb25nPlxuICA8ZGl2IGNsYXNzPVwid3JhcFwiPlxuICAgIDxkaXYgY2xhc3M9XCJzY3JvbGwtYm94XCI+U2hpZnQgKyBNb3VzZSB3aGVlbDwvZGl2PlxuICA8L2Rpdj5cblxuXG4gIDxlbC1yYWRpby1ncm91cCB2LW1vZGVsPVwidGFiUG9zaXRpb25cIiBzdHlsZT1cIm1hcmdpbi1ib3R0b206IDMwcHhcIj5cbiAgICA8ZWwtcmFkaW8tYnV0dG9uIHZhbHVlPVwidG9wXCI+dG9wPC9lbC1yYWRpby1idXR0b24+XG4gICAgPGVsLXJhZGlvLWJ1dHRvbiB2YWx1ZT1cInJpZ2h0XCI+cmlnaHQ8L2VsLXJhZGlvLWJ1dHRvbj5cbiAgICA8ZWwtcmFkaW8tYnV0dG9uIHZhbHVlPVwiYm90dG9tXCI+Ym90dG9tPC9lbC1yYWRpby1idXR0b24+XG4gICAgPGVsLXJhZGlvLWJ1dHRvbiB2YWx1ZT1cImxlZnRcIj5sZWZ0PC9lbC1yYWRpby1idXR0b24+XG4gIDwvZWwtcmFkaW8tZ3JvdXA+XG5cbiAgPGVsLXRhYnNcbiAgICB2LW1vZGVsPVwiYWN0aXZlTmFtZVwiXG4gICAgY2xhc3M9XCJkZW1vLXRhYnNcIlxuICAgIHN0eWxlPVwiaGVpZ2h0OiA2MDBweFwiXG4gICAgOnRhYi1wb3NpdGlvbj1cIih0YWJQb3NpdGlvbiBhcyBhbnkpXCJcbiAgPlxuICAgIDxlbC10YWItcGFuZSB2LWZvcj1cImkgaW4gMzBcIiA6bGFiZWw9XCJgbGFiZWwke2l9YFwiIDpuYW1lPVwiaVwiPlxuICAgICAge3sgaSB9fVxuICAgIDwvZWwtdGFiLXBhbmU+XG4gIDwvZWwtdGFicz5cbjwvdGVtcGxhdGU+XG5cbjxzY3JpcHQgbGFuZz1cInRzXCIgc2V0dXA+XG5pbXBvcnQgeyByZWYgfSBmcm9tICd2dWUnXG5cbmNvbnN0IGFjdGl2ZU5hbWUgPSByZWYoMSlcbmNvbnN0IHRhYlBvc2l0aW9uID0gcmVmKCd0b3AnKVxuPC9zY3JpcHQ+XG5cbjxzdHlsZT5cbi53cmFwIHtcbiAgd2lkdGg6IDEwMCU7XG4gIG92ZXJmbG93LXg6IGF1dG87XG4gIGJvcmRlcjogMXB4IHNvbGlkICMwMDhjOGM7XG4gIG1hcmdpbi1ib3R0b206IDQwcHg7XG59XG5cbi5zY3JvbGwtYm94IHtcbiAgd2lkdGg6IDEwMDAwcHg7IFxuICBoZWlnaHQ6IDQwcHg7IFxufVxuXG4uZGVtby10YWJzID4gLmVsLXRhYnNfX2NvbnRlbnQge1xuICBwYWRkaW5nOiAzMnB4O1xuICBjb2xvcjogIzZiNzc4YztcbiAgZm9udC1zaXplOiAzMnB4O1xuICBmb250LXdlaWdodDogNjAwO1xufVxuPC9zdHlsZT5cbiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vcHJldmlldy0yMjAyNC1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzJywgJ2h0dHBzOi8vcHJldmlldy0yMjAyNC1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL3RoZW1lLWNoYWxrL2RhcmsvY3NzLXZhcnMuY3NzJ10ubWFwKChzdHlsZSkgPT4ge1xuICAgIHJldHVybiBuZXcgUHJvbWlzZSgocmVzb2x2ZSwgcmVqZWN0KSA9PiB7XG4gICAgICBjb25zdCBsaW5rID0gZG9jdW1lbnQuY3JlYXRlRWxlbWVudCgnbGluaycpXG4gICAgICBsaW5rLnJlbCA9ICdzdHlsZXNoZWV0J1xuICAgICAgbGluay5ocmVmID0gc3R5bGVcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignbG9hZCcsIHJlc29sdmUpXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2Vycm9yJywgcmVqZWN0KVxuICAgICAgZG9jdW1lbnQuYm9keS5hcHBlbmQobGluaylcbiAgICB9KVxuICB9KVxuICByZXR1cm4gUHJvbWlzZS5hbGxTZXR0bGVkKHN0eWxlcylcbn1cbiIsInRzY29uZmlnLmpzb24iOiJ7XG4gIFwiY29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiBcIkVTTmV4dFwiLFxuICAgIFwianN4XCI6IFwicHJlc2VydmVcIixcbiAgICBcIm1vZHVsZVwiOiBcIkVTTmV4dFwiLFxuICAgIFwibW9kdWxlUmVzb2x1dGlvblwiOiBcIkJ1bmRsZXJcIixcbiAgICBcInR5cGVzXCI6IFtcImVsZW1lbnQtcGx1cy9nbG9iYWwuZC50c1wiXSxcbiAgICBcImFsbG93SW1wb3J0aW5nVHNFeHRlbnNpb25zXCI6IHRydWUsXG4gICAgXCJhbGxvd0pzXCI6IHRydWUsXG4gICAgXCJjaGVja0pzXCI6IHRydWVcbiAgfSxcbiAgXCJ2dWVDb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IDMuM1xuICB9XG59XG4iLCJQbGF5Z3JvdW5kTWFpbi52dWUiOiI8c2NyaXB0IHNldHVwPlxuaW1wb3J0IEFwcCBmcm9tICcuL0FwcC52dWUnXG5pbXBvcnQgeyBzZXR1cEVsZW1lbnRQbHVzIH0gZnJvbSAnLi9lbGVtZW50LXBsdXMuanMnXG5zZXR1cEVsZW1lbnRQbHVzKClcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxBcHAgLz5cbjwvdGVtcGxhdGU+XG4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL3VucGtnLmNvbS9AdnVlL3J1bnRpbWUtZG9tQGxhdGVzdC9kaXN0L3J1bnRpbWUtZG9tLmVzbS1icm93c2VyLmpzXCIsXG4gICAgXCJAdnVlL3NoYXJlZFwiOiBcImh0dHBzOi8vdW5wa2cuY29tL0B2dWUvc2hhcmVkQGxhdGVzdC9kaXN0L3NoYXJlZC5lc20tYnVuZGxlci5qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9wcmV2aWV3LTIyMDI0LWVsZW1lbnQtcGx1cy5zdXJnZS5zaC9idW5kbGUvZGlzdC9pbmRleC5mdWxsLm1pbi5tanNcIixcbiAgICBcImVsZW1lbnQtcGx1cy9cIjogXCJ1bnN1cHBvcnRlZFwiLFxuICAgIFwiQGVsZW1lbnQtcGx1cy9pY29ucy12dWVcIjogXCJodHRwczovL3VucGtnLmNvbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJfbyI6eyJzaG93SGlkZGVuIjp0cnVlLCJzdHlsZVNvdXJjZSI6Imh0dHBzOi8vcHJldmlldy0yMjAyNC1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzIn19)

rzzf:
If horizontal Tabs need to support vertical scrolling, I suggest implementing it inside the Tabs component to avoid breaking the behavior of the `virtual-list`.

But we can wait for the members’ opinions first.😃

patienry:
Thank you for the reminder. On my computer, holding the `Shift` key doesn't enable horizontal scrolling of the tabs either (though swiping on the **touchpad** works).
I also compared it with several other component libraries, which support horizontal scrolling directly using the mouse wheel. That's why I created this PR. 😃

Dsaquel:
>On my computer, holding the Shift key doesn't enable horizontal scrolling of the tabs either (though swiping on the touchpad works).

Me too, shift + scroll doesn't work.

>I also compared it with several other component libraries, which support horizontal scrolling directly using the mouse wheel. That's why I created this PR. 😃

Honestly I'm little bit surprised other libraries like ant design handle this behavior. It feels unnatural to allow this type of scroll.

That said i don't really have an opinion on this.

packages/components/virtual-list/src/hooks/use-wheel.ts
```diff
@@ -34,7 +29,15 @@ const useWheel = (
...
-    const newOffset = e[LayoutKeys[layout.value]]
+    let newOffset = 0
+    const { deltaX, deltaY } = e
+    const absX = Math.abs(e.deltaX)
+    const absY = Math.abs(e.deltaY)
+    if (absX === absY) {
+      newOffset = layout.value === HORIZONTAL ? deltaX : deltaY
+    } else {
+      newOffset = absX > absY ? deltaX : deltaY
+    }
```
Dsaquel: I feel like we can only allow the vertical layout for tabs.

I don't see a scenario where we need horizontal scroll while using the wheel scroll.

https://github.com/element-plus/element-plus/blob/328523e2916a037693b62d0e568d81ccf2770db2/packages/components/tabs/src/tab-nav.tsx#L137-L139

```diff
-        layout: computed(() =>
-          isHorizontal.value ? 'horizontal' : 'vertical'
-        ),
+        layout: computed(() => 'vertical'
```

patienry: Thank you for your suggestion, but is it still necessary to use `computed` here? 
Can I write it directly like this: `layout: 'vertical'`?

Dsaquel: I checked and `layout` only accept computed :no_mouth:. Maybe it's better like this in order to not make to much refactoring.

patienry: Done, thank you for your review and check. 😃

Dsaquel (approved):
LGTM

patienry:
@Dsaquel Hello, I have found a problem. 🤔 

If we modify it directly like this: `layout: computed(() => 'vertical')`, it will affect the horizontal scroll behavior on the **touchpad**. When swiping left and right, horizontal scrolling will not work because the change is in `deltaX`, not `deltaY`.

So we might still need the following code to handle this issue.
image

I've thought of two ways to handle it.
1. Declare `handleWheel` in `tab-nav.tsx`, including the code above. This will not affect `useWheel`.
2. Add a property in `useWheel`, such as `isTabs`, with a default value of `false`. The handling logic from the above code will only be applied when it is set to `true`.

I'm not sure which approach is more suitable. Do you have any thoughts on this? 👀 👀

Dsaquel:
Nice catch, i need to hold and understand why we both cannot scroll using shift key.
Taking a step back, maybe there is a way to allow us scroll with shift key.

rzzf:
> Nice catch, i need to hold and understand why we both cannot scroll using shift key. Taking a step back, maybe there is a way to allow us scroll with shift key.

This appears to be caused by differences between systems or browsers.

This file might help address the issue: https://github.com/facebookarchive/fixed-data-table/blob/master/src/vendor_upstream/dom/normalizeWheel.js

P.S: The mousewheel directive also depends on it.
https://github.com/element-plus/element-plus/blob/459288c5a9d05e6283b72dd9b336fef3e91ac733/packages/directives/mousewheel/index.ts#L1

patienry:
Thank you, but the issue is still not resolved after I used `normalize-wheel-es`. 🤔 

Maybe some of the code in `tab-nav.tsx` is causing this issue, because I noticed that I can use shift + scroll to horizontally scroll in [Table(table-with-fixed-column)](https://element-plus.org/zh-CN/component/table#%E5%9B%BA%E5%AE%9A%E5%88%97), but I haven't figured out a better solution yet. 😢

rzzf:
> Maybe some of the code in `tab-nav.tsx` is causing this issue, because I noticed that I can use shift + scroll to horizontally scroll in [Table(table-with-fixed-column)](https://element-plus.org/zh-CN/component/table#%E5%9B%BA%E5%AE%9A%E5%88%97), but I haven't figured out a better solution yet. 😢

Can we try using the `v-mousewheel` directive like we do with the Table component? Sorry, I don’t have a Windows device, so I’m unable to test this myself.

https://github.com/element-plus/element-plus/blob/1f11db39db4d1fb1d3953428c6d2d7b515cdb04b/packages/components/table/src/table.vue#L36

patienry:
Thank you, but this seems unrelated to `v-mousewheel`, because even after I removed `v-mousewheel` from `table.vue`, I can still use shift + scroll to scroll horizontally. You can take a look at the comment below. 👀 👀

patienry:
I found a piece of code in `use-grid-wheel.ts`, which should be handling the horizontal scrolling situation in Windows.

```
// Special case for windows machine with shift key + wheel scrolling
if (e.shiftKey && y !== 0) {
   x = y
   y = 0
}
```

So I added this logic to `use-wheel.ts`. Do you think this is appropriate? 🤔

packages/components/virtual-list/src/hooks/use-wheel.ts
```diff
@@ -34,7 +29,14 @@ const useWheel = (
   const onWheel = (e: WheelEvent) => {
     cAF(frameHandle)
 
-    const newOffset = e[LayoutKeys[layout.value]]
+    let { deltaX, deltaY } = e
+    // Special case for windows machine with shift key + wheel scrolling
+    if (e.shiftKey && deltaY !== 0) {
```
rzzf, line 34: It works fine on macOS locally, but in the Playground, Shift + scroll doesn’t work — it only works when Shift is not pressed. I’m not sure if this is due to the Playground. 🤔

patienry: It’s strange, but I don’t have a macOS device, so I can’t do thorough debugging. 😢

rzzf: Sorry for the delayed feedback.

It’s possible that when I was testing in the playground earlier, the code I fetched wasn’t from the latest commit, which led to the incorrect conclusion.

I’ve tested it again, and everything works fine now.

Additionally, could you help test components like select-v2 and tree-v2 on Windows? We need to make sure that other components built with **build-list** are not affected.

patienry: ### Before

https://github.com/user-attachments/assets/239ab488-8e89-41df-89ba-6ee7fc3c5b28

### After

https://github.com/user-attachments/assets/0394665b-c413-45cd-bf17-b18ffa30e7fc

When I press `Shift`, vertical scrolling stops. However, in the previous version, pressing `Shift` would still allow vertical scrolling to continue (does this count as a bug? Because pressing `Shift` is typically meant to trigger horizontal scrolling).

rzzf: I think the behavior after the change is expected. Before the change, using Shift + scroll on macOS didn’t trigger scrolling because `deltaX` and `deltaY` were already reported correctly. The fact that it scrolls on Windows is an unexpected behavior.

cc @Dsaquel

rzzf (approved):

Dsaquel:
Unfortunately, with my linux system (Ubuntu 24) using shift + scroll to scroll horizontally doesn't work in local or playground.

rzzf:
> Unfortunately, with my linux system (Ubuntu 24) using shift + scroll to scroll horizontally doesn't work in local or playground.

On Firefox running in an Ubuntu 24 virtual machine, native scrolling doesn’t work for me, but it works correctly inside the Tabs component. Is there any configuration I need to adjust?


https://github.com/user-attachments/assets/26e0bae0-21ce-4640-8ee6-5c6725ab302a

Dsaquel:
>On Firefox running in an Ubuntu 24 virtual machine, native scrolling doesn’t work for me, but it works correctly inside the Tabs component. Is there any configuration I need to adjust?

I'm in a different desktop but still using Ubuntu 24 and both works now... :laughing:.

Dsaquel (approved):

PR #21826, opened 2025-08-19
fix(components): [cascader] treat `tagTooltipRef` as array

Dsaquel:
hotfix of https://github.com/element-plus/element-plus/pull/21732#issuecomment-3201388162

There are some refact in cascader that are nice to have, have to fix some issues in date picker before i'll jump into it.

PR #22191, opened 2025-09-16
refactor(components): [cascader] externalize tooltip tag from v-for

Dsaquel:
Ref https://github.com/element-plus/element-plus/pull/21826

follow select tagTooltipRef structure https://github.com/element-plus/element-plus/blob/dev/packages/components/select/src/select.vue#L94

### DIFF
--- a/packages/components/cascader/src/cascader.vue
+++ b/packages/components/cascader/src/cascader.vue
@@ -73,7 +73,7 @@
             nsCascader.is('validate', Boolean(validateState)),
           ]"
         >
-          <slot name="tag" :data="allPresentTags" :delete-tag="deleteTag">
+          <slot name="tag" :data="tags" :delete-tag="deleteTag">
             <el-tag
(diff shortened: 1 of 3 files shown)


Dsaquel:
:laughing: 
image

packages/components/cascader/src/cascader.vue
```diff
@@ -88,50 +88,58 @@
               <template v-if="tag.isCollapseTag === false">
                 <span>{{ tag.text }}</span>
               </template>
-              <template v-else>
```
btea, line 91: The judgment condition on line 88 seems to have no meaning.

Dsaquel: Good eyes :+1: , removed completely the property from `Tag` too because it was not used anymore.

It can break the typescript of very few users.

btea (approved):

tolking (approved):

PR #19839, opened 2025-02-11
fix(components): [date-picker] add missing data slot

Dsaquel:
close https://github.com/element-plus/element-plus/issues/19260
close #21373
close #13535
context pr #19481

PR #22192, opened 2025-09-16
fix(components): [date-picker] keyboard navigation

Dsaquel:
regression of https://github.com/element-plus/element-plus/pull/19839

The `keydown.down` was not working.

fix gif:
Peek 2025-09-16 21-01

### DIFF
--- a/packages/components/date-picker-panel/src/composables/use-basic-date-table.ts
+++ b/packages/components/date-picker-panel/src/composables/use-basic-date-table.ts
@@ -259,6 +259,13 @@ export const useBasicDateTable = (
     }
   }
 
+  const isSelectedCell = (cell: DateCell) => {
+    return (
+      (!unref(hasCurrent) && cell?.text === 1 && cell.type === 'normal') ||
+      cell.isCurrent
+    )
+  }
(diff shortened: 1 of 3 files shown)


packages/components/date-picker-panel/src/composables/use-basic-date-table.ts
```diff
@@ -259,6 +259,13 @@ export const useBasicDateTable = (
     }
   }
 
+  const isSelectedCell = (cell: DateCell) => {
+    return (
+      (!unref(hasCurrent) && cell?.text === 1 && cell.type === 'normal') ||
```
btea, line 264: Why add `cell?.text === 1` here?

Dsaquel: I took the previous code deleted in https://github.com/element-plus/element-plus/pull/19839, My though is that  `cell?.text === 1` is a fallback if there are no date already selected date in the current table after doing `keydown.down`.

btea (approved):

tolking (approved):

PR #20628, opened 2025-05-06
feat(components): [virtual-list] add touch event handling for scrolling

warmthsea:
- fix:  https://github.com/element-plus/element-plus/issues/20624

[...]

PR #22195, opened 2025-09-17
chore: add `consistent-type-imports` eslint rules to .ts file

warmthsea:
### Rel

https://github.com/element-plus/element-plus/pull/20628#discussion_r2352019607

Same as .vue:
https://github.com/element-plus/element-plus/blob/chore/eslint-ts-import-type/internal/eslint-config/index.js#L125-L131

screenshots


packages/components/select/src/select.ts
```diff
@@ -6,7 +6,7 @@ import {
   useSizeProp,
 } from '@element-plus/hooks'
 import {
-  EmitFn,
+  type EmitFn,
```
btea: This one should probably be moved down as well.

btea (approved):

packages/components/select/src/select.ts
warmthsea: In fact, it is affected by this:
But it didn't work, possibly due to a version issue.
```
{ disallowTypeAnnotations: true },
```



There are several similar code, and since most of them are imported separately, I manually changed them as well.

PR #22197, opened 2025-09-17
ci: split test into separate lint

warmthsea:
### Before

🔄 Lint & TypeCheck 
        ⏬      
🔄 Test

Test CI long execution time.

https://github.com/element-plus/element-plus/blob/ci/test-separate-lint/.github/workflows/test-unit.yml#L18-L25

### After

No impact on each other, time is greatly shortened.

🔄 Lint & TypeCheck  ↔️  🔄 Test


btea (approved):
It looks good, even if lint or typecheck fails it doesn't seem to affect the running of the test case.

Issue #22216, opened 2025-09-17
[Component] The styles of `el-button` become abnormal when `el-popconfirm`/`el-popover`/`el-tooltip` are used together inside an `el-button-group`


cloydlau:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.21`
- Element Plus Version: `2.11.2`
- Browser / OS: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/140.0.0.0 Safari/537.36`
- Build Tool: `Vite`

### Related Component

- `el-button-group`
- `el-popconfirm`
- `el-popover`
- `el-tooltip`
- `el-button`

### Reproduction Link

[...]

PR #22229, opened 2025-09-18
docs(components): [popconfirm] set type definition for virtual-ref

cloydlau:

Relative issue: https://github.com/element-plus/element-plus/issues/22216

### DIFF
--- a/packages/components/popconfirm/src/popconfirm.ts
+++ b/packages/components/popconfirm/src/popconfirm.ts
@@ -1,7 +1,10 @@
 import { buttonTypes } from '@element-plus/components/button'
 import { QuestionFilled } from '@element-plus/icons-vue'
 import { buildProps, iconPropType } from '@element-plus/utils'
-import { useTooltipContentProps } from '@element-plus/components/tooltip'
+import {
+  useTooltipContentProps,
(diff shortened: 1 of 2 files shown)


rzzf:
In fact, both Popconfirm and Popover are internally built on top of Tooltip: `<el-tooltip v-bind="$attrs" >`.

I think the missing props are not limited to `virtual-triggering` and `virtual-ref`. In theory, except for the props they define or override themselves, all props provided by Tooltip should be usable.

Currently, if users are not aware of this, they might not realize that tooltip’s props can actually be passed through—similar to what happened in #22216.

I would expect to have type hints in the IDE or clear documentation about this.

What do you think?
cc @btea

btea:
@cszhjh That makes sense. 👍

Dsaquel:
After seeing https://github.com/element-plus/element-plus/pull/22314, let's just add the type definition of `virtual-triggering` and `virtual-ref` for the moment, we'll figure it out adding the other properties later.

Dsaquel:
Wanted to keep the commits history clean but after rebasing, git returned me only the final commit. Strange :thinking: .

Dsaquel (approved):

tolking (approved):

PR #22252, opened 2025-09-19
refactor: remove `tooltip-v2` `visual-hidden` components

warmthsea:
This is an unfinished component that shouldn't be in the `master/dev` branch, so I'm removing it.


btea (approved):
It seems that it is indeed not used.

Dsaquel (approved):
Indeed not used 👍 .

PR #22258, opened 2025-09-20
feat(components): [color-picker-panel] hue-slider a11y

tolking:

1. Refactoring the hue-slider component and improving accessibility features
2. Add Home End PageDown PageUp key to control slider position
3. Add a prompt for the current color when using the [screen reader](https://developer.mozilla.org/en-US/docs/Glossary/Screen_reader)

---

The hue slider is opposite in direction to the regular slider.

When controlled through the keyboard, I manually reverse it. But there will still be unreasonable behavior (press the ArrowUp key, but the value decreases).

Perhaps only by changing the design can this issue be completely fixed.


Dsaquel (approved):

packages/components/color-picker-panel/src/props/slider.ts
warmthsea, line 1: It seems inappropriate to change the name here as there is a duplicate. WDYT?

https://github.com/element-plus/element-plus/blob/feat/color-hue-slider/packages/components/slider/src/slider.ts#L30

image

packages/components/color-picker-panel/src/composables/use-slider.ts
```diff
@@ -39,87 +46,111 @@ export const useAlphaSlider = (props: AlphaSliderProps) => {
...
+        props.color.set(key, key === 'hue' ? maxValue : minValue)
+        break
+      case EVENT_CODE.end:
+        props.color.set(key, key === 'hue' ? minValue : maxValue)
+        break
+      case EVENT_CODE.pageDown:
+        incrementPosition(-4 * reverse)
+        break
+      case EVENT_CODE.pageUp:
+        incrementPosition(4 * reverse)
```
btea, line 104: Should the step value here be made configurable? 🤔

tolking: It seems unnecessary, average users won't need this.

Here `4` is the configuration of the reference ElSlider.

btea (approved):

PR #22277, opened 2025-09-21
style(theme-chalk): same height across pickers for range date pickers

Dsaquel:
| Before | After |
|--------|-------|
| image| image |


btea (approved):
looks great 👍

warmthsea (approved):

PR #22337, opened 2025-09-26
fix(components): [date-picker] should not pick a value when change view

### LINKED ISSUE #7842: [Bug Report] [Component] [date-picker] FocusTrap breaks nested dialog "focus on tab"
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.33`
- Element Plus Version: `2.2.1`
- Browser / OS: `Chrome/MacOS`
- Build Tool: `Vite`

### Related Component

- `el-date-picker`

### Reproduction Link

Select a date by selecting the year from the year table, month from the month table, day from the day table:
[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtZGF0ZS1waWNrZXIgOm1vZGVsLXZhbHVlPVwiZGF0ZVwiIEB1cGRhdGU6bW9kZWwtdmFsdWU9XCJ1cGRhdGVEYXRlXCIgLz5cblx0PG9sPlxuICAgIDxsaSB2LWZvcj1cIihldmVudCwgaW5kZXgpIGluIGV2ZW50c1wiIDprZXk9XCJpbmRleFwiPjxwcmU+e3sgZXZlbnQgfX08L3ByZT48L2xpPiAgXG4gIDwvb2w+XG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IGxhbmc9XCJ0c1wiIHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSAndnVlJ1xuXG5jb25zdCBldmVudHMgPSByZWYoW10pXG5cbmNvbnN0IGRhdGUgPSByZWYoJycpXG5jb25zdCB1cGRhdGVEYXRlID0gKHZhbHVlKSA9PiB7XG4gIGRhdGUudmFsdWUgPSB2YWx1ZVxuICBldmVudHMudmFsdWUucHVzaCh2YWx1ZSlcbiAgY29uc29sZS5sb2coJ0VtaXR0ZWQ6JywgdmFsdWUpXG59XG48L3NjcmlwdD5cbiIsImltcG9ydF9tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHt9XG59IiwiX28iOnt9fQ==)

### Steps to reproduce

Select a date by selecting the year from the years table, then the month from the months table, then the day from the days table.

1. Start by selecting the year label at the top. 
  

2. Selecting 2023 will emit next year's current date

3. Selecting June will emit next year's first of June

4. Selecting the day will finally emit the correct day. That's the only emit that should happen in this case.

### What is Expected?

Only one "update:model-value" event should be emitted after selecting the day.

There are three "update:model-value" events: one after each selection: year, month, day. 

### Additional comments

I believe the regression was introduced in https://github.com/element-plus/element-plus/pull/7598

More specifically: [here](https://github.com/element-plus/element-plus/pull/7598/files#diff-ce64860e98ab8af028e85764b0975f255abd40120c8544de36e9105b18966759R438-R442
) and [here](https://github.com/element-plus/element-plus/pull/7598/files#diff-ce64860e98ab8af028e85764b0975f255abd40120c8544de36e9105b18966759R454-R458
)
Before creating a PR that removes some of the code linked above, @opengraphica could you share some info about the decision to add that code? And also, thank you so much for making these components more accessible! 🙇

Dsaquel:
fix https://github.com/element-plus/element-plus/issues/7842, fix https://github.com/element-plus/element-plus/discussions/21815

When switching the view in `type="date"`, the date-picker pick a value.
Before:
Peek 2025-09-26 13-32

It is very annoying so i removed this behavior.

This is actually not the right way to fix the issue because now the focus leave the date-picker.
Other actions like "changing view with arrows" also don't support the fallback focus too and only keyboard users may notice the small difference so i think it's a fine trade off.

The right fix would involve a lot of refactor even if we think it is not. So i decided just removed this annoying behavior for now.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved unintended auto-focus behavior during date selection actions
  * Improved date and month selection logic in date picker views
  * Fixed view navigation flow between year, month, and date selections

* **Accessibility**
  * Enhanced focus indicators for date picker cells with improved keyboard navigation visibility

### DIFF
--- a/packages/components/date-picker-panel/src/composables/use-basic-date-table.ts
+++ b/packages/components/date-picker-panel/src/composables/use-basic-date-table.ts
@@ -204,7 +204,13 @@ export const useBasicDateTable = (
     }
   )
 
-  const focus = async () => unref(currentCellRef)?.focus()
+  let suppressNextAutoFocus = false
+
+  const focus = async () => {
+    suppressNextAutoFocus = true
+    unref(currentCellRef)?.focus()
+    suppressNextAutoFocus = false
+  }
 
   const isCurrent = (cell: DateCell): boolean => {
     return (
@@ -268,12 +274,18 @@ export const useBasicDateTable = (
 
   const isSelectedCell = (cell: DateCell) => {
     return (
-      (!unref(hasCurrent) && cell?.text === 1 && isNormalDay(cell.type)) ||
-      cell.isCurrent
+      cell.isCurrent ||
+      (!unref(hasCurrent) &&
+        isNormalDay(cell.type) &&
+        cellMatchesDate(cell, props.date))
     )
   }
 
   const handleFocus = (event: FocusEvent) => {
+    if (suppressNextAutoFocus) {
+      suppressNextAutoFocus = false
+      return
+    }
     if (focusWithClick || unref(hasCurrent) || props.selectionMode !== 'date')
       return
     handlePickDate(event, true)
--- a/packages/components/date-picker-panel/src/date-picker-com/panel-date-pick.vue
+++ b/packages/components/date-picker-panel/src/date-picker-com/panel-date-pick.vue
@@ -505,7 +505,8 @@ const handleYearPick = async (
     innerDate.value = getValidDateOfYear(data, lang.value, disabledDate)
     currentView.value = 'month'
     if (['month', 'year', 'date', 'week'].includes(selectionMode.value)) {
-      // TODO: fallback focus into basic month table
+      await nextTick()
+      handleFocusPicker()
     }
   }
   handlePanelChange('year')
--- a/packages/components/date-picker-panel/src/date-picker-com/panel-date-pick.vue
+++ b/packages/components/date-picker-panel/src/date-picker-com/panel-date-pick.vue
@@ -484,9 +484,7 @@ const handleMonthPick = async (
     )
     currentView.value = 'date'
     if (['month', 'year', 'date', 'week'].includes(selectionMode.value)) {
-      emit(innerDate.value, true)
-      await nextTick()
-      handleFocusPicker()
+      // TODO: fallback focus into basic date table
     }
   }
   handlePanelChange('month')
@@ -507,9 +505,7 @@ const handleYearPick = async (
     innerDate.value = getValidDateOfYear(data, lang.value, disabledDate)
     currentView.value = 'month'
     if (['month', 'year', 'date', 'week'].includes(selectionMode.value)) {
-      emit(innerDate.value, true)
-      await nextTick()
-      handleFocusPicker()
+      // TODO: fallback focus into basic month table
     }
   }
   handlePanelChange('year')
--- a/packages/components/date-picker-panel/src/date-picker-com/panel-date-pick.vue
+++ b/packages/components/date-picker-panel/src/date-picker-com/panel-date-pick.vue
@@ -484,7 +484,8 @@ const handleMonthPick = async (
     )
     currentView.value = 'date'
     if (['month', 'year', 'date', 'week'].includes(selectionMode.value)) {
-      // TODO: fallback focus into basic date table
+      await nextTick()
+      handleFocusPicker()
     }
   }
   handlePanelChange('month')
--- a/packages/components/date-picker/__tests__/date-picker.test.ts
+++ b/packages/components/date-picker/__tests__/date-picker.test.ts
@@ -477,37 +477,36 @@ describe('DatePicker', () => {
     expect(input.element.value).toBe('2023-02-04')
   })
 
-  it('should not pick a value when changing picker view', async () => {
-    const wrapper = _mount(`<el-date-picker v-model="value" />`, () => ({
-      value: '',
-      disabledDate(time) {
-        const dayTime = new Date(2023, 1, 4).getTime()
-        return time.getTime() < dayTime
-      },
-    }))
+  it('should not emit during year > month > date view navigation', async () => {
+    const onUpdateModelValue = vi.fn()
+    const wrapper = _mount(
+      `<el-date-picker
+        v-model="value"
+        @update:modelValue="onUpdateModelValue"
+    />`,
+      () => ({ value: '2024-06-15', onUpdateModelValue })
+    )
     const input = wrapper.find('input')
     input.trigger('blur')
     input.trigger('focus')
     await nextTick()
-    const yearLabel: HTMLElement = document.querySelectorAll(
-      '.el-date-picker__header-label'
-    )[0]
-    yearLabel.click()
+    ;(
+      document.querySelectorAll(
+        '.el-date-picker__header-label'
+      )[0] as HTMLElement
+    ).click()
     await nextTick()
-    const yearCells = document.querySelectorAll('.el-date-table-cell__text')
-    const year2023 = [...yearCells].find((item) => item.innerHTML === '2023')
-    year2023.click()
+    ;[...document.querySelectorAll('.el-date-table-cell__text')]
+      .find((el) => el.innerHTML === '2025')!
+      .click()
     await nextTick()
-    const monthLabel: HTMLElement = document.querySelectorAll(
-      '.el-date-picker__header-label'
-    )[1]
-    monthLabel.click()
+    document.querySelectorAll('.el-date-table-cell__text')[2].click()
     await nextTick()
-    const monthCells = document.querySelectorAll('.el-date-table-cell__text')
-    const februaryCell = monthCells[1]
-    februaryCell.click()
+    expect(onUpdateModelValue).not.toHaveBeenCalled()
+
+    document.querySelector('td.available .el-date-table-cell__text')!.click()
     await nextTick()
-    expect(input.element.value).toBe('')
+    expect(onUpdateModelValue).toHaveBeenCalledTimes(1)
   })
 
   it('should work when using disabledDate prop and daterange type', async () => {
@@ -1852,7 +1851,8 @@ describe('DatePicker keyboard events', () => {
     await nextTick()
     triggerEvent(input, 'keydown', EVENT_CODE.down)
     await nextTick()
-    expect(document.querySelector('.current')?.textContent).toBe('1')
+    const focusedCell = document.querySelector('td[tabindex="0"]')
+    expect(focusedCell.classList.contains('available')).toBe(true)
   })
 })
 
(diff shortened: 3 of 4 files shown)


tolking:
It seems to have affected the accessibility. After selecting the year through the keyboard, the focus did not move to the month.

- before

https://github.com/user-attachments/assets/f83a3a01-bdde-406d-a321-1ac551e3fdbc

- after

https://github.com/user-attachments/assets/a4a95017-c563-48e7-a6e8-9928ca161669

Dsaquel:
I can't see the videos provided but i see what you're referring to. [shortcut play](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtZGF0ZS1waWNrZXIgdHlwZT1cImRhdGVcIiA6bW9kZWwtdmFsdWU9XCJkYXRlXCIgQHVwZGF0ZTptb2RlbC12YWx1ZT1cInVwZGF0ZURhdGVcIiAvPlxuXHQ8b2w+XG4gICAgPGxpIHYtZm9yPVwiKGV2ZW50LCBpbmRleCkgaW4gZXZlbnRzXCIgOmtleT1cImluZGV4XCI+PHByZT57eyBldmVudCB9fTwvcHJlPjwvbGk+XG4gIDwvb2w+XG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IGxhbmc9XCJ0c1wiIHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSAndnVlJ1xuXG5jb25zdCBldmVudHMgPSByZWYoW10pXG5cbmNvbnN0IGRhdGUgPSByZWYoJycpXG5jb25zdCB1cGRhdGVEYXRlID0gKHZhbHVlKSA9PiB7XG4gIGRhdGUudmFsdWUgPSB2YWx1ZVxuICBldmVudHMudmFsdWUucHVzaCh2YWx1ZSlcbiAgY29uc29sZS5sb2coJ0VtaXR0ZWQ6JywgdmFsdWUpXG59XG48L3NjcmlwdD5cbiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vcHJldmlldy0yMjMzNy1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzJywgJ2h0dHBzOi8vcHJldmlldy0yMjMzNy1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL3RoZW1lLWNoYWxrL2RhcmsvY3NzLXZhcnMuY3NzJ10ubWFwKChzdHlsZSkgPT4ge1xuICAgIHJldHVybiBuZXcgUHJvbWlzZSgocmVzb2x2ZSwgcmVqZWN0KSA9PiB7XG4gICAgICBjb25zdCBsaW5rID0gZG9jdW1lbnQuY3JlYXRlRWxlbWVudCgnbGluaycpXG4gICAgICBsaW5rLnJlbCA9ICdzdHlsZXNoZWV0J1xuICAgICAgbGluay5ocmVmID0gc3R5bGVcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignbG9hZCcsIHJlc29sdmUpXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2Vycm9yJywgcmVqZWN0KVxuICAgICAgZG9jdW1lbnQuYm9keS5hcHBlbmQobGluaylcbiAgICB9KVxuICB9KVxuICByZXR1cm4gUHJvbWlzZS5hbGxTZXR0bGVkKHN0eWxlcylcbn1cbiIsInRzY29uZmlnLmpzb24iOiJ7XG4gIFwiY29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiBcIkVTTmV4dFwiLFxuICAgIFwianN4XCI6IFwicHJlc2VydmVcIixcbiAgICBcIm1vZHVsZVwiOiBcIkVTTmV4dFwiLFxuICAgIFwibW9kdWxlUmVzb2x1dGlvblwiOiBcIkJ1bmRsZXJcIixcbiAgICBcInR5cGVzXCI6IFtcImVsZW1lbnQtcGx1cy9nbG9iYWwuZC50c1wiXSxcbiAgICBcImFsbG93SW1wb3J0aW5nVHNFeHRlbnNpb25zXCI6IHRydWUsXG4gICAgXCJhbGxvd0pzXCI6IHRydWUsXG4gICAgXCJjaGVja0pzXCI6IHRydWVcbiAgfSxcbiAgXCJ2dWVDb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IDMuM1xuICB9XG59XG4iLCJQbGF5Z3JvdW5kTWFpbi52dWUiOiI8c2NyaXB0IHNldHVwPlxuaW1wb3J0IEFwcCBmcm9tICcuL0FwcC52dWUnXG5pbXBvcnQgeyBzZXR1cEVsZW1lbnRQbHVzIH0gZnJvbSAnLi9lbGVtZW50LXBsdXMuanMnXG5zZXR1cEVsZW1lbnRQbHVzKClcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxBcHAgLz5cbjwvdGVtcGxhdGU+XG4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL2Zhc3RseS5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvcnVudGltZS1kb21AbGF0ZXN0L2Rpc3QvcnVudGltZS1kb20uZXNtLWJyb3dzZXIuanNcIixcbiAgICBcIkB2dWUvc2hhcmVkXCI6IFwiaHR0cHM6Ly9mYXN0bHkuanNkZWxpdnIubmV0L25wbS9AdnVlL3NoYXJlZEBsYXRlc3QvZGlzdC9zaGFyZWQuZXNtLWJ1bmRsZXIuanNcIixcbiAgICBcImVsZW1lbnQtcGx1c1wiOiBcImh0dHBzOi8vcHJldmlldy0yMjMzNy1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwidW5zdXBwb3J0ZWRcIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly9mYXN0bHkuanNkZWxpdnIubmV0L25wbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJfbyI6eyJzaG93SGlkZGVuIjp0cnVlLCJzdHlsZVNvdXJjZSI6Imh0dHBzOi8vcHJldmlldy0yMjMzNy1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzIn19).

tolking:
The focus is not moving to the date when the selected month

Dsaquel:
Yes this is the regression expected.
>This is actually not the right way to fix the issue because now the focus leave the date-picker.
Other actions like "changing view with arrows" also don't support the fallback focus too and only keyboard users may notice the small difference so i think it's a fine trade off.

>The right fix would involve a lot of refactor even if we think it is not. So i decided just removed this annoying behavior for now.

tolking:
The `changing view with arrows` will not affect the focus, as the focus is still on the arrow.

I am more inclined towards the current Controlled Components solution. For form elements, user actions should be timely.

If the displayed value does not match the bound value, it will be even more uncomfortable (like #22004)

Dsaquel:
If i understand correctly you mean that the current behavior is normal (emitting the value while changing the view) ?

tolking:
> If i understand correctly you mean that the current behavior is normal (emitting the value while changing the view) ?

yes

Dsaquel:
Ok now i understand more your point on this:
>I am more inclined towards the current Controlled Components solution. For form elements, user actions should be timely.

>If the displayed value does not match the bound value, it will be even more uncomfortable (like https://github.com/element-plus/element-plus/issues/22004)

But IMO, selecting a view don't mean to pick a value, for a user perspective when selecting a view you don't want to pick the first one available, you don't want at all a value being picked.

For example type="daterange" is doing the logic wanted even if there are a lack of accessibility.

tolking:
The biggest difference between `daterange` and `date` is that `daterange` requires selecting two values and cannot respond to user input in real-time. So it becomes a Uncontrolled Components, using `leftDate` and `rightDate` to record the values selected by the user.

Dsaquel:
>user input in real-time

I'm sorry but selecting a year/month while changing the view is not selecting the end date, it's a side effect that is just terrible in a UX and not wanted.

Even the person who added this functionality assume this need a refactor https://github.com/element-plus/element-plus/issues/7842#issuecomment-1141515874.

[Ant design date-picker](https://antdv.com/components/date-picker) is doing the behavior wanted for example.

Dsaquel:
@tolking PTAL.

Have fixed the behavior https://github.com/element-plus/element-plus/pull/22337#issuecomment-3338674815 regression (keyboard).

Note that i have also changed this: when doing keydown.down after open in a date-picker type="date", the focus will be on the current day, this fit with type="month" and type="year".

PR #22366, opened 2025-09-29
fix(components): [upload] Invalid CSS variable

FrontEndDog:
fix #22357


Dsaquel (commented):
It is intended to have a variable at this level, the developer should override this variable at the component level.
If we accept this change we should do it in the entire code base where we inject the scss variables in the component level.

Like select:
https://github.com/element-plus/element-plus/blob/efef336db8c0a7d16117f315f31308fd0a37cfa6/packages/theme-chalk/src/select.scss#L15-L17

Image

I assume @keeplearning66 is right [here](https://github.com/element-plus/element-plus/issues/22357#issuecomment-3343103412).

tolking:
I remember doing this now is to reduce the style volume when importing on demand

FrontEndDog:
> It is intended to have a variable at this level, the developer should override this variable at the component level. If we accept this change we should do it in the entire code base where we inject the scss variables in the component level.
> 
> Like select:
> 
> https://github.com/element-plus/element-plus/blob/efef336db8c0a7d16117f315f31308fd0a37cfa6/packages/theme-chalk/src/select.scss#L15-L17
> 
> Image
> I assume @keeplearning66 is right [here](https://github.com/element-plus/element-plus/issues/22357#issuecomment-3343103412).

I have already raised this variable to the component level. Thanks for your review.

packages/theme-chalk/src/upload.scss
```diff
@@ -77,14 +77,12 @@
...
-    @include set-css-var-value(('upload', 'picture-card', 'size'), 148px);
-
     background-color: getCssVar('fill-color', 'lighter');
     border: 1px dashed getCssVar('border-color', 'darker');
     border-radius: 6px;
     box-sizing: border-box;
-    width: getCssVar('upload', 'picture-card', 'size');
-    height: getCssVar('upload', 'picture-card', 'size');
+    width: getCssVar('upload-picture-card-size');
+    height: getCssVar('upload-picture-card-size');
```
tolking: There seems to be no need to modify it here

FrontEndDog: Done.

FrontEndDog: I have made some other modifications since the last review.
Retained the original variables `--el-upload-list-picture-card-size` and `--el-upload-picture-card-size`
The default value for `--el-upload-picture-card-size` is 148px, and the default value for `--el-upload-list-picture-card-size` is `var(--el-upload-picture-card-size)`;

tolking (approved):

Dsaquel (approved):
👍

PR #22382, opened 2025-09-29
fix(components): [input-number] arrow key repeat and disabled-scientific not working

rzzf:
1. Arrow up/down keys trigger value changes twice
2. The `disabled-scientific` attribute not working

I’m not sure why the change was made from keydown to keyup, but for me, ensuring the feature works correctly on PC is more important.

### DIFF
--- a/packages/components/input-number/src/input-number.vue
+++ b/packages/components/input-number/src/input-number.vue
@@ -57,7 +57,6 @@
       :validate-event="false"
       :inputmode="inputmode"
       @keydown="handleKeydown"
-      @keyup="handleKeydown"
       @blur="handleBlur"
       @focus="handleFocus"
       @input="handleInput"
--- a/packages/components/input-number/src/input-number.vue
+++ b/packages/components/input-number/src/input-number.vue
@@ -57,7 +57,6 @@
       :validate-event="false"
       :inputmode="inputmode"
       @keydown="handleKeydown"
-      @keyup="handleKeyup"
       @blur="handleBlur"
       @focus="handleFocus"
       @input="handleInput"
@@ -233,14 +232,6 @@ const handleKeydown = (event: KeyboardEvent | Event) => {
       decrease()
       break
     }
-  }
-}
-const handleKeyup = (event: KeyboardEvent | Event) => {
-  const key = getEventKey(event as KeyboardEvent)
-
-  if (props.disabledScientific && ['e', 'E'].includes(key)) {
-    event.preventDefault()
-    return
   }
 }
 const increase = () => {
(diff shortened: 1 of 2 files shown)


snowbitx:
It seems to be caused by the fix for this bug.     [pr](https://github.com/element-plus/element-plus/pull/22228/files#diff-0ae90cff73c6316f25336e807f41c1ecaca3dfbe4210f80273a4310aa89ac5a8L59)

tolking:
~~I tested the change event without any issues~~

Macos 26 + Chrome 140 / Safari 26

---

Only by using Keyup on Android can the input 'e' and 'E' characters be obtained

rzzf:
> ~I tested the change event without any issues~
> 
> Macos 26 + Chrome 140 / Safari 26
> 
> Only by using Keyup on Android can the input 'e' and 'E' characters be obtainedAndroid

Thanks for explaining. 
On macOS 15.4.1 with Chrome 140, using keyup results in these two issues.

https://github.com/user-attachments/assets/a4b88701-c96f-4371-8ebc-9e84425ad806

Since mobile devices do not have arrow keys, I kept the keyup event while adding keydown, and ignored the arrow key logic in keyup to ensure compatibility. WDYT?

packages/components/input-number/src/input-number.vue
```diff
@@ -56,6 +56,7 @@
       :aria-label="ariaLabel"
       :validate-event="false"
       :inputmode="inputmode"
+      @keydown="handleKeydown"
       @keyup="handleKeydown"
```
tolking:
```suggestion
      @keydown="handleKeydown"
```

Let us restore `keydown` here. The built-in Keyboard control and `handleKeydown` are duplicated when `<input type=number />`

rzzf: > Only by using Keyup on Android can the input 'e' and 'E' characters be obtained

Will this cause the return of Android issues?

tolking: Of course, you can also try to handle it. Since the value has already been entered at keyup, you should also clear any invalid characters like E/e at that time.

rzzf: I’m not sure whether calling `preventDefault` in the keyup event works on Android devices. 
I also haven’t found a good way to disable exponential notation, since in `input[type=number]`, the input.value always becomes an empty string when an invalid value is entered. 
Although using `input[type=text]` with` inputmode="numeric"` and `pattern` can achieve this, it inevitably introduces more complexity.

So, I’ve decided to remove the keyup event for now.

tolking (approved):

PR #22410, opened 2025-10-03
[fix](components): [table] resolve column misalignment when scrolling to rightmost#22351

xiaocainiao633:
When the table enables native scrollbar, the header and body columns are misaligned 
due to the scrollbar width occupation. This fix adds a placeholder cell in the header 
to compensate for the scrollbar width, ensuring layout consistency.

- Calculate scrollbar width dynamically based on browser characteristics
- Add header placeholder cell synchronized with scrollbar width
- Update placeholder position on scroll and window resize events
- Only enable fix logic when nativeScrollbar prop is true


xiaocainiao633:
image
image

xiaocainiao633:
image
image

PR #22432, opened 2025-10-09
feat(components): [table] accessibility enhancement

### LINKED ISSUE #14301: [table-v2] [Accessibility WCAG] - el-table-v2
Bug Type: **`Other`**

## Environment

- Vue Version: `3.3.4`
- Element Plus Version: `2.3.14`
- Browser / OS: `110.0.5481.177 (arm64) / MacOs`
- Build Tool: `Vite`

### Related Component

- `el-table-v2`

### Reproduction Link

[Docs](https://element-plus.org/en-US/component/table-v2.html)

### Steps to reproduce

With a WCAG validator.

- Use a WCAG chrome extension like WAVE : https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh

- In chrome, dev tools, lighthouse, you can check the Accessibility

### What is Expected?

Should not give accessibility error or warning

In Canada WCAG 2.0 is now mandatory. Big enterprise are now allowed to use library if they are not 100% WCAG 2.0 compliant

On the tablev2 display page, table-with-selections trigger Accessibility error on Google Lighthouse.

https://element-plus.org/en-US/component/table-v2.html#table-with-selections

Error : Form elements do not have associated labels
https://dequeuniversity.com/rules/axe/4.7/label

Example :

div.el-table-v2__row-cell > label.el-checkbox > span.el-checkbox__input > input#el-id-4491-121
<input class="el-checkbox__original" type="checkbox" aria-hidden="false" value="false" id="el-id-4491-121">
div > div.el-table-v2__row > div.el-table-v2__row-cell > label.el-checkbox
<label class="el-checkbox">

### Additional comments

Should always pass wcag and section508 accessibility. You could provide accessibility default in the component. And maybe add some properties to override them. In some juridictions, the law force us to follow accessibility rules, so it's very important.

Thanks for the good work

tolking:

fix #14301

### table

- Improve the accessibility labels for `<el-table-column type="selection" />`.
- Improve the accessibility labels for `<el-table-column type="expand">` and the TreeTable.
- Improve the accessibility labels for sortable
- Improve the accessibility labels for filterable

---

### checkbox

- When the `<label>` tag is used and the `inputId` exists, the `for` attribute is missing.

## Breaking changes

The expand buttons changed from `div` element to `button` element


packages/components/table/src/config.ts
```diff
@@ -87,7 +81,8 @@ export const cellForced = {
         },
         onClick: (event: Event) => event.stopPropagation(),
         modelValue: store.isSelected(row),
-        ariaLabel: column.label,
+        value: row,
```
btea: `value` is only used when inside a checkbox-group and no binding seems to be needed here.

tolking:
```ts
  const hasOwnLabel = computed<boolean>(() => {
    return !!slots.default || !isPropAbsent(actualValue.value)
  })
```

This is mainly to avoid generating the `inputId`

btea: But the input has an extra value binding, which looks a bit strange.

image

packages/theme-chalk/src/table.scss
```diff
@@ -79,12 +79,24 @@
...
+    margin: 0;
+    width: 19px;
+    height: 19px;
+    background-color: transparent;
+    border: none;
+    border-radius: getCssVar('border-radius', 'base');
+    outline: none;
+
+    &:focus-visible {
+      outline: 2px solid getCssVar('table', 'focus-border-color');
```
Dsaquel: I understand we have a lot of 2px outline when focus-visible but the result will show a outline-top truncated.

Wouldn't it be better to set it to one 1px ?
| Before | After |
|--------|-------|
| Image | Image |

```suggestion
      outline: 1px solid getCssVar('table', 'focus-border-color');
```

packages/components/table/src/config.ts
tolking: I have removed the code, now it will generate inputId

btea (approved):
LGTM

packages/theme-chalk/src/table.scss
Dsaquel: With the current modification, the style don't display correctly in some table
image

You can get the reproduction here https://preview-22432-element-plus.surge.sh/en-US/component/table#tree-data-and-lazy-mode.

Dsaquel (approved):
LGTM

PR #22435, opened 2025-10-10
chore: upgrade pnpm to v10

FrontEndDog:


package.json
```diff
@@ -120,7 +120,7 @@
   },
   "volta": {
     "node": ">=20",
-    "pnpm": "9.5.0"
+    "pnpm": "10.18.2"
```
btea: The configuration instructions here can be removed. The `packageManager` field already exists, and the corresponding pnpm version will be automatically downloaded when executing pnpm-related commands.

FrontEndDog: Done！ I have made some other modifications. Could you please take a look again when you have time.

btea (approved):

.github/workflows/publish-pr-commit-pkg.yml
```diff
@@ -18,15 +18,26 @@ jobs:
...
-      - run: npm i -g --force corepack && corepack enable
       - uses: actions/setup-node@v4
         with:
           node-version: 20
-          cache: 'pnpm'
 
       - name: Setup pnpm
         uses: pnpm/action-setup@v4
 
+      - name: Get pnpm store directory
```
Dsaquel: Is the cache really necessary ?

Image [link](https://pnpm.io/continuous-integration).

FrontEndDog: I remember that I used cache before because after upgrading to V10, there was an [error](https://github.com/element-plus/element-plus/actions/runs/18677548080) in `Github/workflows/publish-pr-commit-pkg.yml`,That's why I used another caching method.
I found an error from the link you provided. The reason is that the `Setup node` and `Setup pnpm` are written incorrectly.
Now, I have fixed it.

Dsaquel: Didn't saw the difference, nice catch :+1: .

Dsaquel (approved):

PR #22447, opened 2025-10-12
ci: fix eslint warning

Dsaquel:
After multiple eslint run, eslint can print this warning:
>ESLintPoorConcurrencyWarning: You may disable concurrency or use a numeric concurrency setting to improve performance.

After upgrading we don't get it anymore.


btea (approved):
Although I still haven't found the reason.

PR #22452, opened 2025-10-13
docs: fix typecheck

Dsaquel:
There are 2 typescript issues in the docs directory:

- In this file https://github.com/element-plus/element-plus/blob/dev/docs/.vitepress/vitepress/components/vp-app.vue, there is a conflict with the vue version 3.4 vs 3.2 (vitepress vue version against our locale vue version) image
After several ways to resolve this type conflict i only found the solution where we have to install 3.2 without losing types in the examples directory.

- In this file https://github.com/element-plus/element-plus/blob/dev/docs/.vitepress/vitepress/components/vp-skip-link.vue#L38, the element `<a>` and other html tag have a react type, i found that docsearch/js under the hood have react installed that corrupt our template vue files. image image
There is a good thread about this where i took this solution https://github.com/vuejs/language-tools/discussions/592#discussioncomment-2163181.


warmthsea (approved):
LGTM

makedopamine:
Installing Vue 3.4.31 in the docs package would be better. That resolves both issues without other changes.

Dsaquel:
Setting vue in the docs package to `^3.4.31` or `=3.4.31` will make the example directory lost type.

image

Dsaquel:
Along with the fact that the tests aren't passing.

makedopamine:
@Dsaquel I added a commit to your PR.  I've tested it loccally and it seems to work fine. Please confirm it works on your side as well🙇‍♂️

Dsaquel:
>I added a commit to your PR. I've tested it loccally and it seems to work fine. Please confirm it works on your side as well🙇‍♂️

Looks very good, thanks you very much :heart: .

docs/tsconfig.json
```diff
@@ -12,9 +12,9 @@
     "skipLibCheck": true,
     "lib": ["ESNext", "DOM"],
     "paths": {
-      "react": ["./stub/types__react"],
```
Dsaquel: Just found that removing this line cause some type issues in the examples/table-v2 complaining about jsx, [file ref](https://github.com/element-plus/element-plus/blob/dev/docs/examples/table-v2/filter.vue).

If adding the line again the type error is gone, if you are ok let's bring it back ? @makedopamine .

makedopamine: >Just found that removing this line cause some type issues in the examples/table-v2 complaining about jsx, [file ref](https://github.com/element-plus/element-plus/blob/dev/docs/examples/table-v2/filter.vue).

Resolved via https://github.com/element-plus/element-plus/pull/22452/commits/267ea933aefa065a00b60082f36b765a3d637719, please take a look and see if it fixes this issue.

Dsaquel: Indeed fixed, thanks :bow:

btea (approved):

rzzf (approved):

PR #22477, opened 2025-10-15
fix(components): [cascader/autocomplete] make debounce prop reactive

keeplearning66:

relate [#22471 (comment)](https://github.com/element-plus/element-plus/pull/22471#pullrequestreview-3337497615)
`debounce` prop should be reactive


patienry:
{17CEC278-9B89-43F1-96AD-E07344016A62}

After I use `useDebounceFn`, one of the test cases in `autocomplete.test.tsx` fails. 😢

btea:
It seems that this is because the `useDebounceFn` function handles the logic differently when the second parameter is `0`.

patienry:
Thank you. So, would this cause a **breaking change**? 🤔

btea:
I just tried it again, and it's a bit strange. During the actual code running process, the input box focus will be triggered once, and the enter key will also be triggered once. It seems reasonable to trigger it twice.

btea:
It is indeed caused by the `useDebounceFn ` function.

https://github.com/vueuse/vueuse/blob/main/packages/shared/utils/filters.ts#L87

When the time is set to `0`, the function call is treated as a microtask. After waiting for the `nextTick` event to execute in the test case, the microtask has already been executed and the `fetchSuggestions` function is called once. When the `enter` event is triggered, the same logic is executed again, so it is called twice.
https://github.com/element-plus/element-plus/blob/b45e094ce3fc0841fbff7965990dff0f231ae839/packages/components/autocomplete/__tests__/autocomplete.test.tsx#L162-L165
If you change the `<= 0` in the corresponding line of code in vueuse above to `<0`, the test case will be fine.

btea:
> Thank you. So, would this cause a **breaking change**? 🤔

Although the impact is not significant, when debounce is set to 0, the triggering time of the function is indeed different. It was a macro task before, and a micro task after the modification.

But I think it's okay, we can modify the test case. cc @warmthsea

patienry:
Thank you very much for your detailed explanation. I make the changes for now to ensure the test case passes.

I also noticed that `slider` has `debounce` prop, but it’s not working. It seems like an incomplete feature. Do we need this functionality? If so, I’m happy to create a new PR to finish it. If not, I can remove it from the documentation.

btea:
It really looks useless and can be removed from the document.

btea (approved):

patienry:
> It really looks useless and can be removed from the document.

Okay, completed in #22488.

warmthsea (approved):

PR #22480, opened 2025-10-15
ci: add release version to commit message

warmthsea:
After: website deploy 2.11.4

like https://github.com/element-plus/element-plus/blob/chore--update-commit/.github/workflows/publish-pr-commit-pkg.yml#L36-L40

Before: https://github.com/element-plus/element-plus/commits/gh-pages/

image


btea (approved):
Overall it seems great to me. Copilot left a suggestion for compatible error handling, which doesn't seem necessary.

Dsaquel (approved):
LGTM

PR #22487, opened 2025-10-15
chore: remove csstype

Dsaquel:
In 3.2 i had typecheck error if i removed this package.

After the vue upgrade i don't think we need csstype anymore.


btea (approved):

FrontEndDog (approved):
LGTM

PR #22081, opened 2025-09-09
fix(components): [dropdown]: first hover target show outline

warmthsea:
[...]

PR #22489, opened 2025-10-16
fix(components): [Dropdown] focus style is incorrect

FrontEndDog:
[Reproduce docs link](https://element-plus.org/zh-CN/component/dropdown)

Open the official document. Hover the mouse over any dropdown, Unexpected focus styles may appear.
image
image


btea:
It looks like a duplicate of #22081. We can move the discussion there.

FrontEndDog:
duplicate of https://github.com/element-plus/element-plus/pull/22081.

PR #22491, opened 2025-10-16
test: image stable loading in CI

warmthsea:
This image test case often fails.

https://github.com/element-plus/element-plus/actions/runs/18548953046/job/52872411027?pr=22490

83ecc03f-cdef-4841-a081-7e911354a72a


rzzf:
Looks like the Avatar might have a similar issue too.

https://github.com/element-plus/element-plus/actions/runs/18551673276/job/52880160475

packages/components/image/__tests__/image.test.tsx
```diff
@@ -24,6 +24,27 @@ async function doubleWait() {
...
+
+    if (condition()) return
+    await new Promise((resolve) => setTimeout(resolve, interval))
+  }
+
+  if (!condition()) {
+    throw new Error(`Condition not met within ${timeout}ms timeout`)
+  }
+}
+
```
FrontEndDog:
```suggestion
function waitUntil(condition, timeout = 3000, interval = 100) {
  const start = Date.now()
  return new Promise(async (resolve, reject) => {
    async function check() {
      await nextTick()
      await flushPromises()
      if (condition()) return resolve()
      if (Date.now() - start >= timeout) return reject(`Condition not met within ${timeout}ms timeout`)
      setTimeout(check, interval)
    }
    check()
  })
}
```

FrontEndDog (changes requested):

packages/components/image/__tests__/image.test.tsx
warmthsea: Thanks, I feel like the current Fn is clear and solves the problem, and it doesn't seem like there's any need to change it.

FrontEndDog (approved):
LGTM

PR #22499, opened 2025-10-16
feat(components): [tree-select] add instance type and improve test cleanup logic

warmthsea:
#### Rel:

This CI multiple execution errors.
https://github.com/element-plus/element-plus/actions/runs/18553354152/job/52885097548?pr=22491

image

Removed `vShow` because it caused VueWrapper<SelectInstance> type error.
Actually, it's useless.

image


packages/components/tree-select/__tests__/tree-select.test.tsx
```diff
@@ -1,13 +1,44 @@
...
+    let id = 1
+    while (id < 1000) {
+      try {
+        window.cancelAnimationFrame(id)
+      } catch {
+        // Ignore errors
+      }
+      id++
+    }
+  }
```
btea: You seem to want to clear all calls to `requestAniamtionFrame`, but this approach does not seem reasonable. Perhaps intercepting `cancelAnimationFrame` and `requestAniamtionFrame` is more reasonable.

```ts
const frameIds = new Set();
const originalRequestAnimationFrame = global.requestAnimationFrame;
global.requestAnimationFrame = function(cb) {
  const id = originalRequestAnimationFrame(() => {
    frameIds.delete(id)
  })
  frameIds.add(id)
  return id
}

function cancelAllAnimationFrames() {
  frameIds.forEach(id =>global.cancelAnimationFrame(id))
  frameIds.clear()
}
```

warmthsea: 👍

packages/components/tree-select/src/instance.ts
```diff
@@ -0,0 +1,7 @@
+import { SelectInstance } from '@element-plus/components/select'
+import { TreeInstance } from '@element-plus/components/tree'
```
Dsaquel:
```suggestion
import type { SelectInstance } from '@element-plus/components/select'
import type { TreeInstance } from '@element-plus/components/tree'
```

Dsaquel (approved):
I have no idea if this will work, let's see.

PR #22501, opened 2025-10-16
chore: remove concurrently

Dsaquel:
Removing the concurrently package as it is implemented in only one command and it can be replaced, [ref](https://stackoverflow.com/questions/3004811/how-do-you-run-multiple-programs-in-parallel-from-a-bash-script).

Before and after; postinstall it does take the same time (4.5s).

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "docs:crowdin-credentials": "pnpm run -C docs crowdin-credentials",
     "stub": "pnpm run -r --parallel stub",
     "prepare": "husky",
-    "postinstall": "pnpm stub && concurrently \"pnpm gen:version\" \"pnpm run -C internal/metadata dev\""
+    "postinstall": "pnpm stub && pnpm gen:version & pnpm run -C internal/metadata dev &"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


btea (approved):
👍 
I learned it, I didn't expect that I could use only one `&`

package.json
```diff
@@ -40,7 +40,7 @@
     "docs:crowdin-credentials": "pnpm run -C docs crowdin-credentials",
     "stub": "pnpm run -r --parallel stub",
     "prepare": "husky",
-    "postinstall": "pnpm stub && concurrently \"pnpm gen:version\" \"pnpm run -C internal/metadata dev\""
+    "postinstall": "pnpm stub && pnpm gen:version & pnpm run -C internal/metadata dev &"
```
tolking, line 43: Is the last & sign necessary?

Dsaquel: I'm not entirely sure about this but my shell (fish) understand that it is a job:
image

tolking (approved):

package.json
tolking: Works correctly on the zsh shell (macOS)

PR #22526, opened 2025-10-18
refactor(components): [tree] improve emit to get type hints

rzzf:
Before
before

After
after

### DIFF
--- a/packages/components/tree/src/tree.ts
+++ b/packages/components/tree/src/tree.ts
@@ -3,62 +3,58 @@ import { isBoolean } from '@element-plus/utils'
 import type { ComponentInternalInstance } from 'vue'
 import type { EmitFn } from '@element-plus/utils'
 import type Node from './model/node'
-import type { CheckedInfo, NodeDropType, TreeNodeData } from './tree.type'
+import type { CheckedInfo, NodeDropType } from './tree.type'
 
 /* eslint-disable @typescript-eslint/no-unused-vars */
 export const treeEmits = {
-  'check-change': (
-    data: TreeNodeData,
-    checked: boolean,
-    indeterminate: boolean
-  ) => data && isBoolean(checked) && isBoolean(indeterminate),
-  'current-change': (data: TreeNodeData | null, node: Node | null) => true,
+  'check-change': (data: any, checked: boolean, indeterminate: boolean) =>
+    data && isBoolean(checked) && isBoolean(indeterminate),
+  'current-change': (data: any | null, node: Node | null) => true,
   'node-click': (
-    data: TreeNodeData,
+    data: any,
     node: Node,
     nodeInstance: ComponentInternalInstance | null,
     event: MouseEvent
   ) => data && node && event instanceof Event,
   'node-contextmenu': (
     event: Event,
-    data: TreeNodeData,
+    data: any,
     node: Node,
     nodeInstance: ComponentInternalInstance | null
   ) => event instanceof Event && data && node,
   'node-collapse': (
-    data: TreeNodeData,
+    data: any,
     node: Node,
     nodeInstance: ComponentInternalInstance | null
   ) => data && node,
   'node-expand': (
-    data: TreeNodeData,
+    data: any,
     node: Node,
     nodeInstance: ComponentInternalInstance | null
   ) => data && node,
-  check: (data: TreeNodeData, checkedInfo: CheckedInfo) => data && checkedInfo,
-  'node-drag-start': (node: Node, event: DragEvent) =>
-    node && event instanceof DragEvent,
+  check: (data: any, checkedInfo: CheckedInfo) => data && checkedInfo,
+  'node-drag-start': (node: Node, event: DragEvent) => node && event,
   'node-drag-end': (
     draggingNode: Node,
     dropNode: Node | null,
-    dropType: NodeDropType | null,
+    dropType: NodeDropType,
     event: DragEvent
-  ) => draggingNode && event instanceof DragEvent,
+  ) => draggingNode && event,
   'node-drop': (
     draggingNode: Node,
     dropNode: Node,
-    dropType: Exclude<NodeDropType, 'none'> | null,
+    dropType: Exclude<NodeDropType, 'none'>,
     event: DragEvent
-  ) => draggingNode && dropNode && event instanceof DragEvent,
+  ) => draggingNode && dropNode && event,
   'node-drag-leave': (
     draggingNode: Node,
     oldDropNode: Node,
     event: DragEvent
-  ) => draggingNode && oldDropNode && event instanceof DragEvent,
+  ) => draggingNode && oldDropNode && event,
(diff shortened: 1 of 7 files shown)


packages/components/tree/src/tree.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import type { ComponentInternalInstance } from 'vue'
+import type { EmitFn } from '@element-plus/utils'
+import type Node from './model/node'
+import type { CheckedInfo, NodeDropType, TreeNodeData } from './tree.type'
+
+/* eslint-disable @typescript-eslint/no-unused-vars */
+export const treeEmits = {
+  'check-change': (
+    data: TreeNodeData,
```
Dsaquel: TreeNodeData will have a conflict with other interface created by the developer.

Perhaps we can leave these fields to `any` until a fully generic support (?)

packages/components/tree/src/tree.ts
```diff
@@ -0,0 +1,65 @@
...
+    nodeInstance: ComponentInternalInstance | null
+  ) => data && node,
+  check: (data: TreeNodeData, checkedInfo: CheckedInfo) => data && checkedInfo,
+  'node-drag-start': (node: Node, event: DragEvent) =>
+    node && event instanceof DragEvent,
+  'node-drag-end': (
+    draggingNode: Node,
+    dropNode: Node | null,
+    dropType: NodeDropType | null,
+    event: DragEvent
```
Dsaquel: In this example https://github.com/element-plus/element-plus/blob/dev/docs/examples/tree/draggable.vue#L22 we pass `DragEvents`, it can be deleted.

packages/components/tree/src/tree.ts
rzzf: Thanks, I didn't notice. 👍

packages/components/tree/src/tree.ts
rzzf: This appears to be an error in the example. `DragEvents` seems to be used internally only, while the emitted event is still the native `DragEvent`.

https://github.com/element-plus/element-plus/blob/1cbe108a78bda58d3fe2d1527bad3d391713b279/packages/components/tree/src/model/useDragNode.ts#L81

Dsaquel: I agree with you, i wasn't clear, we can replace `DragEvents` to the native `DragEvent` in the draggable example file. Because we continue having type error in this file.


Something like:

```diff
diff --git a/docs/examples/tree/draggable.vue b/docs/examples/tree/draggable.vue
--- a/docs/examples/tree/draggable.vue
+++ b/docs/examples/tree/draggable.vue
@@ -19,38 +19,29 @@
 <script lang="ts" setup>
 import type {
   AllowDropType,
-  DragEvents,
   NodeDropType,
   RenderContentContext,
 } from 'element-plus'
 
 type Node = RenderContentContext['node']
 
-const handleDragStart = (node: Node, ev: DragEvents) => {
+const handleDragStart = (node: Node, ev: DragEvent) => {
   console.log('drag start', node)
 }
-const handleDragEnter = (
-  draggingNode: Node,
-  dropNode: Node,
-  ev: DragEvents
-) => {
+const handleDragEnter = (draggingNode: Node, dropNode: Node, ev: DragEvent) => {
   console.log('tree drag enter:', dropNode.label)
 }
-const handleDragLeave = (
-  draggingNode: Node,
-  dropNode: Node,
-  ev: DragEvents
-) => {
+const handleDragLeave = (draggingNode: Node, dropNode: Node, ev: DragEvent) => {
   console.log('tree drag leave:', dropNode.label)
 }
-const handleDragOver = (draggingNode: Node, dropNode: Node, ev: DragEvents) => {
+const handleDragOver = (draggingNode: Node, dropNode: Node, ev: DragEvent) => {
   console.log('tree drag over:', dropNode.label)
 }
 const handleDragEnd = (
   draggingNode: Node,
-  dropNode: Node,
+  dropNode: Node | null,
   dropType: NodeDropType,
-  ev: DragEvents
+  ev: DragEvent
 ) => {
   console.log('tree drag end:', dropNode && dropNode.label, dropType)
 }
@@ -58,7 +49,7 @@ const handleDrop = (
   draggingNode: Node,
   dropNode: Node,
   dropType: NodeDropType,
-  ev: DragEvents
+  ev: DragEvent
 ) => {
   console.log('tree drop:', dropNode.label, dropType)
 }
```

rzzf: I see, I misunderstood earlier.

Dsaquel (approved):
LGTM

packages/components/tree/src/tree.ts
```diff
@@ -0,0 +1,61 @@
...
+  ) => draggingNode && dropNode && event,
+  'node-drag-leave': (
+    draggingNode: Node,
+    oldDropNode: Node,
+    event: DragEvent
+  ) => draggingNode && oldDropNode && event,
+  'node-drag-enter': (draggingNode: Node, dropNode: Node, event: DragEvent) =>
+    draggingNode && dropNode && event,
+  'node-drag-over': (draggingNode: Node, dropNode: Node, event: DragEvent) =>
```
btea:
```suggestion
  'node-drag-over': (draggingNode: Node, dropNode: Node, eve: DragEvent) =>
```

Image

The name of the event-related parameter is changed to `eve`, otherwise there seems to be duplicate parameter names. refer to #17545

packages/components/tree/src/tree.ts
```diff
@@ -0,0 +1,61 @@
...
+    draggingNode && dropNode && event,
+}
+/* eslint-enable @typescript-eslint/no-unused-vars */
+
+export type TreeEmits = EmitFn<typeof treeEmits>
```
btea, line 58: I see that the emit type exported by other components does not have a nested `EmitFn`. Is this necessary?

rzzf: Used in `util`.

https://github.com/element-plus/element-plus/pull/22526/files#diff-10174ba11a775bb2063c67cf5bee3ef565caf74c20f6ad8082213d29e503f28aR25

Other components also have similar situations.
image

packages/components/tree/src/tree.ts
rzzf: Done, Thanks🙇.

packages/components/tree/src/tree.ts
btea: It's a bit confusing, some components are used and some are not. 🤔

rzzf: > It's a bit confusing, some components are used and some are not. 🤔

Do you want me to remove the `TreeEmits` type?

btea: Don’t worry, I’ll look back at history later to see if I can find the specific reason for doing this.

btea: I can't find any specific record of adding it, as it was overwritten by the refactoring pr. 

However, it seems that `TreeEmits` is used as a parameter type declaration when emit is passed as a function parameter.

btea (approved):

PR #22531, opened 2025-10-19
chore: replace ns.is condition

Dsaquel:
Some of `ns.is` can be rewritten better as it accept directly the condition in the second parameter.

Thanks to [ast-grep](https://github.com/ast-grep/ast-grep):
Create file like https://github.com/element-plus/element-plus/commit/fa56c226738a25ca2022f261193f3209a65821c2 and:
```sh
sg --pattern '{ [$NS.is($CLASS)]: $$$CONDITION }'
```


btea (approved):
👍

rzzf (approved):
Cool 👍

PR #22557, opened 2025-10-22
fix(components): [anchor] Anchor custom container scrolling exception

SpanManX:
fix issue #22543


rzzf (commented):
This doesn’t seem like a reasonable approach, as the current change prevents the URL from including the hash when an anchor link is clicked.

SpanManX:
> This doesn’t seem like a reasonable approach, as the current change prevents the URL from including the hash when an anchor link is clicked.

If I add this `history.pushState(null, '', props.href as string)`, is that alright?

rzzf:
> If I add this `history.pushState(null, '', props.href as string)`, is that alright?

This could introduce new issues. for example, how can users prevent it from being added to the hash?

I think we need to figure out how the original problem occurred.

IMO, changing `href` to `data-*` is a feature, not a fix for the problem.

PR #22028, opened 2025-09-04
fix(components): [alert] show immediately when `showAfter` not needed

rzzf:
fix #22012

[...]

Issue #5844, opened 2022-02-08
[Feature Request] Auto Close option for Alert


narasimhajupally:
### Existing Component
Yes

### Component Name
el-alert

### Description
Can we have an attribute for auto-close in el-alert component. 
Thanks for the awesome library

PR #22560, opened 2025-10-23
Revert "feat(components): [alert] add open and auto-close functionality with delay (#20533,#22028)"

warmthsea:
### Rel:

- https://github.com/element-plus/element-plus/pull/20533
- https://github.com/element-plus/element-plus/pull/22028

### Why?

**User event should provide immediate feedback. This is not what a UI library should do.**

Refer to other UI libraries: 

https://mui.com/material-ui/react-alert/#transitions
https://vuetifyjs.com/en/components/alerts/#closable
https://ant.design/components/alert-cn#alert-demo-closable
https://www.naiveui.com/zh-CN/os-theme/components/alert#closable.vue
https://daisyui.com/components/alert/
https://chakra-ui.com/docs/components/alert

The [original question](https://github.com/element-plus/element-plus/issues/5844) should be handled manually by the user through `v-if`, not by the code library.

### Caused other issues:

A default delay is added after clicking Close.

https://github.com/element-plus/element-plus/blob/dev/packages/hooks/use-delayed-toggle/index.ts#L18-L21

For a more obvious demo, I changed it to 1000ms.
screenshots


rzzf:
> The https://github.com/element-plus/element-plus/issues/5844 should be handled manually by the user through v-if, not by the code library.

I agree with you.

I think alerts are generally meant for urgent situations where the user needs to be notified immediately. 
Therefore, the `show-after` property doesn’t make much sense to me, and I’d prefer to let developers implement these three properties on their own.

rzzf (approved):

tolking (approved):

Dsaquel:
I'm ok to remove it, but we should at least set a message in the doc **when** it is removed, currently the deprecated tag only announce in the doc that the feature is obsolete.

Also if we want to push the remove smoothly perhaps it is better to warn the developer and in a next minor release remove the feature.

warmthsea:
> I'm ok to remove it, but we should at least set a message in the doc **when** it is removed, currently the deprecated tag only announce in the doc that the feature is obsolete.
> 
> Also if we want to push the remove smoothly perhaps it is better to warn the developer and in a next minor release remove the feature.

This is a nearly useless feature that causes other issues.
I think almost no one uses it, maybe less than 0.1% of users?

The following is a GitHub search result: 0
https://github.com/search?q=el-alert+show-after&type=code
https://github.com/search?q=el-alert+auto-close&type=code


So I think it can be removed directly without waiting for the next minor release. Don't spend too much time on this issue and put more energy into other PRs.

Dsaquel:
If I reword: In perspective user side i **don't** know when the feature's are deleted.

Let's see what other people think @element-plus/backers.

btea:
This is really unnecessary. When to display it can be completely controlled by the developer's code logic.

btea (approved):

Dsaquel (approved):

btea:
I think we could merge this PR and then add a note about it in the logs. There's no need to keep dragging it out.  🤔

Dsaquel:
Yes it has been a long time.

I let you merge, i don't know if you want to merge it as is or delete the feature once.

warmthsea:
Yes, it's been a bit long.

I've changed the exact version number comment. I think we don't need to wait any longer.

PR #22564, opened 2025-10-23
docs(components): [tour] correct value

cjn021:


btea:
It looks like there is a conflict that needs to be addressed.

rzzf (approved):
Thanks!

patienry:
The default values of `mask`, `showArrow`, and `showClose` in the source code are `true`. Why modify them? 🤔

rzzf:
> The default values of `mask`, `showArrow`, and `showClose` in the source code are `true`. Why modify them? 🤔

These three properties have a default value of `undefined` in the Step, and the Tour defines properties with the same names. 

When determining whether a particular step requires a `mask`, `step.props.mask` takes precedence. Therefore, I think it’s reasonable to indicate in the documentation that these properties have no default values.

https://github.com/element-plus/element-plus/blob/031282014a1119cbb87e258169ba1165794aedf1/packages/components/tour/src/step.vue#L116

cjn021:
> It looks like there is a conflict that needs to be addressed.

thanks, I have already solved it.

patienry:
image

But when I don't set these three properties, the component's behavior is consistent with the default value being `true`.

rzzf:
> But when I don't set these three properties, the component's behavior is consistent with the default value being `true`.

Since `tour.props.mask` has a default value of `true`, it will be used when a step does not specify its own `mask` value.

https://github.com/element-plus/element-plus/blob/031282014a1119cbb87e258169ba1165794aedf1/packages/components/tour/src/tour.vue#L77

https://github.com/element-plus/element-plus/blob/031282014a1119cbb87e258169ba1165794aedf1/packages/components/tour/src/tour.ts#L65-L67

Note: The default values of the three properties modified in this PR belong to tour-step.

You can check the default values in step.ts
https://github.com/element-plus/element-plus/blob/dev/packages/components/tour/src/step.ts

patienry:
My mistake, I misread it. I thought the property being modified was in `tour.vue` in the PR. Sorry for taking up your time.

rzzf:
> My mistake, I misread it. I thought the property being modified was in `tour.vue` in the PR. Sorry for taking up your time.

Don’t worry, it’s people like you that make EP better and better 👍.

warmthsea (approved):
LGTM

PR #22615, opened 2025-10-29
chore: update puppeteer to V24

FrontEndDog:
update puppeteer to V24  and simultaneously fixed the warning in the following image when 'pnpm test:ssr'.

image

### DIFF
--- a/.github/workflows/test-ssr.yml
+++ b/.github/workflows/test-ssr.yml
@@ -31,6 +31,9 @@ jobs:
       - name: Install dependencies
         run: pnpm i --frozen-lockfile
 
+      - name: Install Chromium for Puppeteer
+        run: npx puppeteer browsers install chrome
+
       - name: Local build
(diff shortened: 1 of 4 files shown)


Dsaquel:
There was the same upgrade in the past but it caused an error in other ci pr's.

I don't know what the error but maybe a new version of pupeteer version can solve it (?).

refs https://github.com/element-plus/element-plus/pull/19500 https://github.com/element-plus/element-plus/pull/19521 https://github.com/element-plus/element-plus/pull/19610

btea:
@Dsaquel Thank you for listing those links. I vaguely remember that the problem initially arose because CI was caching an older version. (However, I'm not entirely sure 😄 )

However, this upgrade seems to be working fine. Perhaps we can merge them first, and then address any issues encountered during release.

btea (approved):

Dsaquel (approved):

PR #22668, opened 2025-11-01
fix(components): [date-picker] keyboard navigation on first of the month

Dsaquel:
rel https://github.com/element-plus/element-plus/pull/22662

The first day of the month is a type "today" and not "normal".

Wanted to isolate this test case but it seems like [setSystemTime](https://v2.vitest.dev/api/vi.html#vi-setsystemtime) doesn't work as expected unfortunately.


btea (approved):

btea:
I'm not sure if it's because I'm using it incorrectly, but using `setSystemTime` causes other test cases to fail.

Dsaquel:
>I'm not sure if it's because I'm using it incorrectly, but using setSystemTime causes other test cases to fail.

Same for me it is a bit weird.

Issue #21560, opened 2025-07-31
[All] Element plus files are unable to access process.env in vitest browser mode


Starlord48:
Bug Type: **`Other`**

## Environment

- Vue Version: `3.5.16`
- Element Plus Version: `2.10.4`
- Browser / OS: `Chrome`
- Build Tool: `Vite`

### Related Component

- `All`

### Reproduction Link

[Link]()

### Additional comments

error log:
ReferenceError: process is not defined
 - /packages/components/tooltip/src/content.vue:94:2
 - /node_modules/@vue/reactivity/dist/reactivity.esm-bundler.js:391:27
 - /node_modules/@vue/reactivity/dist/reactivity.esm-bundler.js:1643:4
 - /node_modules/@vue/reactivity/dist/reactivity.esm-bundler.js:1488:28
 - /packages/components/tooltip/src/content.vue:107:9
 - /node_modules/@vue/reactivity/dist/reactivity.esm-bundler.js:391:27
 - /node_modules/@vue/reactivity/dist/reactivity.esm-bundler.js:1643:4
 - /node_modules/@vue/reactivity/dist/reactivity.esm-bundler.js:1488:28
 - /node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/deps/element-plus.js?v=727dd4c2:6852:15
 - /node_modules/@vue/runtime-core/dist/runtime-core.esm-bundler.js:692:12

btea:
本地测试`vitest --ui`并没有问题。

Starlord48:
So it will not work on browser mode and there is no plan to support it?

btea:
Please provide a link to reproduce and steps to do so.

Starlord48:
https://github.com/Starlord48/element-plus-env-issue-reproduction


please refer to readme file for issue reproduction

btea:
https://github.com/element-plus/element-plus/blob/e479006c4cba05d6d419abe629f07f0728cdc892/packages/components/tooltip/src/content.vue#L89-L95

This is intentionally used in conjunction with the test environment and will be replaced during packaging, so there is no problem.

If you need to use the `playwright` test component, just install the released version.

btea:
Sorry, I noticed that the replacement is only done when the full package is built, I will fix that.

Starlord48:
Thanks for the support

Starlord48:
@btea when is the next patch release planned

btea:
[...]

PR #22675, opened 2025-11-03
chore: only show warnings in dev mode

warmthsea:
### Rel:

- https://github.com/element-plus/element-plus/pull/22657
- https://github.com/element-plus/element-plus/pull/21584

### Diff

|  Env (warn show)      | Before | After |
| ------ | ------- | ----- |
| User dev        |   yes ✅      |     yes  ✅     |
| User prod       |   **yes**❌  |    **no** ✅ |

- Before:

 | Before | After |
| ------- | ----- |
|    image        |    image    |

### Original problem ✅

- https://github.com/element-plus/element-plus/issues/21560

image

### DIFF
--- a/packages/components/tooltip/src/content.vue
+++ b/packages/components/tooltip/src/content.vue
@@ -94,7 +94,7 @@ const persistentRef = computed(() => {
   // For testing, we would always want the content to be rendered
   // to the DOM, so we need to return true here.
   if (
-    (!process || process.env.NODE_ENV === 'test') &&
+    (typeof process === 'undefined' || process.env.NODE_ENV === 'test') &&
     !process?.env.RUN_TEST_WITH_PERSISTENT
   ) {
     return true
--- a/packages/components/tooltip/src/content.vue
+++ b/packages/components/tooltip/src/content.vue
@@ -93,14 +93,13 @@ const transitionClass = computed(() => {
 const persistentRef = computed(() => {
   // For testing, we would always want the content to be rendered
   // to the DOM, so we need to return true here.
-  if (typeof process === 'undefined') {
-    return true
-  }
-  if (
-    process.env.NODE_ENV === 'test' &&
-    !process?.env.RUN_TEST_WITH_PERSISTENT
-  ) {
-    return true
+  if (typeof process !== 'undefined') {
+    if (
+      process.env.NODE_ENV === 'test' &&
+      !process.env.RUN_TEST_WITH_PERSISTENT
+    ) {
+      return true
+    }
   }
   return props.persistent
 })
--- a/packages/components/tooltip/src/content.vue
+++ b/packages/components/tooltip/src/content.vue
@@ -93,10 +93,11 @@ const transitionClass = computed(() => {
 const persistentRef = computed(() => {
   // For testing, we would always want the content to be rendered
   // to the DOM, so we need to return true here.
-  if (process.env.NODE_ENV === 'test') {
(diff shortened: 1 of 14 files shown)


packages/components/tooltip/src/content.vue
```diff
@@ -93,10 +93,14 @@ const transitionClass = computed(() => {
...
-      return true
-    }
+  if (typeof process === 'undefined') {
+    return true
+  }
+  if (
+    process.env.NODE_ENV === 'test' &&
+    !process?.env.RUN_TEST_WITH_PERSISTENT
+  ) {
+    return true
```
btea: I believe that if the `process` does not exist, the original logic should be followed instead of directly returning `true`.
```ts
if (
  typeof process !== 'undefined'
  && ...
  && ...
)
```

warmthsea: This change will cause the original problem to report an error.

https://github.com/element-plus/element-plus/pull/22675/commits/f776553fae885252d6d0ca65bcddd7bf1f71f35b

```
 typeof process !== 'undefined'
  && ... 
```
image

btea: The logic in your screenshot is different from the modification I suggested above. 

Your modification returns a fixed value in the browser and doesn't follow the logic below. The browser should correctly follow the logic below.

btea: It seems that modifications to other files should also be made in accordance with the above suggestions.

packages/utils/error.ts
```diff
@@ -14,9 +14,11 @@ export function throwError(scope: string, m: string): never {
 export function debugWarn(err: Error): void
 export function debugWarn(scope: string, message: string): void
 export function debugWarn(scope: string | Error, message?: string): void {
-  const error: Error = isString(scope)
-    ? new ElementPlusError(`[${scope}] ${message}`)
-    : scope
-  // eslint-disable-next-line no-console
-  console.warn(error)
+  if (process.env.NODE_ENV !== 'production') {
```
btea:
```suggestion
  if (typeof process !== 'undefined' && process.env.NODE_ENV !== 'production') {
```
This should also be added; otherwise, the error message "process does not exist" might still occur.

warmthsea: 👍  Adding security checks to all `process` seems like the best approach.

btea:
https://github.com/element-plus/element-plus/blob/5d1e59ed88059fe54ff0686776490008f2811839/internal/build/src/tasks/full-bundle.ts#L70-L72
During a full build, the original logic will replace `process.env.NODE_ENV`, and the related code will be removed from the final build result. 

Do you think we need to replace the newly added conditional logic code during a full build? 🤔

warmthsea:
> ull-bundle.ts

I think that's fine now, as the existing problems have been resolved.

The code in `full-bundle.ts` did not introduce any new problems.

btea:
In this way, we might be able to remove the plugin that replaces the relevant characters as well.

warmthsea:
> In this way, we might be able to remove the plugin that replaces the relevant characters as well.

Let me compare the build outputs.

warmthsea:
> In this way, we might be able to remove the plugin that replaces the relevant characters as well.

image


```ts
replace({ 
   'process.env.NODE_ENV': '"production"', 
 }), 
```

The code inside `Floating-ui` can only be handled in this way.

btea (approved):

Dsaquel (approved):

PR #22734, opened 2025-11-09
refactor(components): [image-viewer] use `useLockscreen` for scroll lock

rzzf:
I think the width provided by `useLockscreen` is necessary to prevent page shifting when the vertical scrollbar appears.

https://github.com/user-attachments/assets/623e1b65-7d9b-472e-8294-14e2b6104430


Dsaquel (approved):
Well found 👍

btea (approved):

PR #22761, opened 2025-11-11
ci: unable to add comment on PR from fork

rzzf:
Use `pull_request_target` to trigger the workflow in the base repository context.

There should be no security issues in this workflow.
This workflow will not take effect in this PR because changing it to `pull_request_target` prevents it from running in a PR. Please see below for reference.

Rel https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

> This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. **This event allows your workflow to do things like label or comment on pull requests from forks. Avoid using this event if you need to build or run code from the pull request.**
 
Reasons for failure:
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#using-the-permissions-key-for-forked-repositories
https://github.com/orgs/community/discussions/25358

Before:
Failed: https://github.com/element-plus/element-plus/actions/runs/19269312673/workflow?pr=22759

After:
https://github.com/rzzf/element-plus/pull/2#issuecomment-3517537046

### DIFF
--- a/.github/workflows/pr-conflict.yml
+++ b/.github/workflows/pr-conflict.yml
@@ -1,17 +1,27 @@
 name: Comment on PR with conflict Label
+# Note that this `pull_request_target` is vulnerable, it grants write access to a fork repo
+# DO NOT ADD ANY CHECKOUT/CACHING in this workflow
 
 on:
-  pull_request:
+  pull_request_target:
     types: [labeled, synchronize]
+
+permissions:
+  pull-requests: write
 
 jobs:
   comment:
     runs-on: ubuntu-latest
     steps:
       - name: Check for specific label
-        if: contains(github.event.pull_request.labels.*.name, 'conflict pending')
+        if: |
+          contains(github.event.pull_request.labels.*.name, 'conflict pending') &&
+          github.repository_owner == 'element-plus'
         uses: actions-cool/maintain-one-comment@v3.2.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           body: |
             @${{ github.event.pull_request.user.login }} This PR has conflicts, please resolve them.
+            <!-- ELEMENT_PLUS_CONFLICT_COMMENT -->
+          body-include: '<!-- ELEMENT_PLUS_CONFLICT_COMMENT -->'
+          number: ${{ github.event.pull_request.number }}
--- a/.github/workflows/pr-conflict.yml
+++ b/.github/workflows/pr-conflict.yml
@@ -19,7 +19,6 @@ jobs:
           github.repository_owner == 'element-plus'
         uses: actions-cool/maintain-one-comment@v3.2.0
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
           body: |
             @${{ github.event.pull_request.user.login }} This PR has conflicts, please resolve them.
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-conflict.yml
```diff
@@ -1,17 +1,27 @@
...
     runs-on: ubuntu-latest
     steps:
       - name: Check for specific label
-        if: contains(github.event.pull_request.labels.*.name, 'conflict pending')
+        if: |
+          contains(github.event.pull_request.labels.*.name, 'conflict pending') &&
+          github.repository_owner == 'element-plus'
         uses: actions-cool/maintain-one-comment@v3.2.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
```
Dsaquel: Great investigation, have you considered removing
```
token: ${{ secrets.GITHUB_TOKEN }}
```

I see that it doesn't seems like it is required. WDYT ?

rzzf: Good catch. 👍

btea (approved):

.github/workflows/pr-conflict.yml
Dsaquel: Now because we removed the secret and pass the permission to `write` is `pull_request_target` really necessary in this case ?

rzzf: I don’t think removing the Token makes much difference. 
Even with the `pull_request` event and `permissions: write`, PRs coming from forked repositories still don’t have permission to add comments. That’s exactly how the **pr-comment** workflow behaves.
(See: https://github.com/element-plus/element-plus/blob/dev/.github/workflows/pr-comment.yml
)

Remove the token and downgrade the workflow to pull_request: https://github.com/rzzf/element-plus/actions/runs/19293675401/workflow

Token Rel: https://docs.github.com/en/actions/tutorials/authenticate-with-github_token#using-the-github_token-in-a-workflow

> An action can access the GITHUB_TOKEN through the github.token context even if the workflow does not explicitly pass the GITHUB_TOKEN to the action.

As far as I know, the simplest way to allow PRs from forked repositories to add labels or comments is to use `pull_request_target`. (The premise is that no CHECKOUT or CACHING operations have been performed.)

Alternatively, you can follow the approach used by workflows like **pr-docs-build, pr-docs-deploy, and pr-docs-start**, where a `pull_request` workflow triggers a `pull_request_target` workflow to perform the write operations.
There are other methods too, but they tend to be less convenient.

rzzf: We can try this in this PR

Update:

https://github.com/element-plus/element-plus/actions/runs/19295037488/job/55174769033?pr=22761
image

Dsaquel: Interesting, ok now it's clear for me. Thanks <3

Dsaquel (approved):
Indeed pull_request_target is needed 👍 

Thanks for the explanation.

rzzf:
@btea @Dsaquel Thank you for taking the time to help me track this issue. I really appreciate it! 🙇

PR #22774, opened 2025-11-12
fix(components): [form] modify useFormDisabled hook

jeff-fe:
优化 useFormDisabled hook


rzzf:
Could you explain why these changes were made? It seems that many test cases have failed due to the current modifications.

PR #22806, opened 2025-11-14
ci: fix lint-commit-message not working for PR from fork

rzzf:
Rel #19947, #22761

The reason why `github.event.pull_request.user.login` cannot retrieve the username in a post workflow is that this workflow is triggered by a `workflow_run` event rather than directly by a `pull_request` event. 
The payload of `workflow_run` does not include the `pull_request` object, so `github.event.pull_request.user.login` is always empty. 
To obtain the PR author, you need to pass the information via **artifacts**, use the API, or carry it from the original workflow.

[workflow_run payloads](https://docs.github.com/en/webhooks/webhook-events-and-payloads#workflow_run)

Security Reference: https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/#:~:text=Together%20with%20the,g.%20API%20tokens).

> Together with the pull_request_target, a new trigger workflow_run was introduced to enable scenarios that require building the untrusted code and also need write permissions to update the PR with e.g. code coverage results or other test results. To do this in a secure manner, the untrusted code must be handled via the pull_request trigger so that it is isolated in an unprivileged environment. The workflow processing the PR should then store any results like code coverage or failed/passed tests in artifacts and exit. The following workflow then starts on workflow_run where it is granted write permission to the target repository and access to repository secrets, so that it can download the artifacts and make any necessary modifications to the repository or interact with third party services that require repository secrets (e.g. API tokens).

### DIFF
--- a/.github/workflows/lint-commit-message-post.yml
+++ b/.github/workflows/lint-commit-message-post.yml
@@ -53,7 +53,7 @@ jobs:
     permissions:
       pull-requests: write
     steps:
-      - name: successfully
+      - name: Successfully
         if: ${{ needs.result.outputs.succeeded == 'true' }}
         uses: actions-awesome/pr-helper@1.1.1
         with:
@@ -63,7 +63,7 @@ jobs:
           body-filter: '<!-- ELEMENT_PLUS_COMMIT_LINT -->'
           pr-number: ${{ needs.result.outputs.pr }}
 
-      - name: failed
+      - name: Failed
         if: ${{ needs.result.outputs.succeeded != 'true' }}
         uses: actions-awesome/pr-helper@1.1.1
         with:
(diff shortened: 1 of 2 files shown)


Dsaquel:
After creating a PR from a fork, the post workflow doesn't seems to run https://github.com/element-plus/element-plus/actions?query=workflow%3ALint+Commit+Message+Post+.

rzzf:
> After creating a PR from a fork, the post workflow doesn't seems to run https://github.com/element-plus/element-plus/actions?query=workflow%3ALint+Commit+Message+Post+.

Yes, workflow_run behaves the same as pull_request_target: it does not run inside the PR but instead runs on the base branch. Since the current dev branch doesn’t contain this workflow, it won’t trigger — this is a security restriction.

You can test it in your own repository, just like I did in mine. https://github.com/rzzf/element-plus/pull/6

Make sure your base branch includes this workflow.
https://github.com/rzzf/element-plus/blob/dev/.github/workflows/lint-commit-message-post.yml

Dsaquel:
Make sense, thanks for explanation :+1:

Dsaquel (approved):

btea (approved):

PR #22846, opened 2025-11-18
feat(components): [formItem] add disabled prop

jeff-fe:


l246804:
支持

btea:
@jeff-fe  Thank you for your contribution. It seems to overlap with #18739. I will continue to work on it, and your help in reviewing it is welcome.

jeff-fe:
@btea 3q, i close pr

PR #22879, opened 2025-11-20
fix(components): [table-column] reactively update filterable state

### LINKED ISSUE #22877: [Component] [table-column] [components] [table-cloumn] Filter icon does not appear or disappear when filters or filter-method is added or removed dynamically
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.20`
- Element Plus Version: `2.11.8`
- Browser / OS: `Chrome 142.0.7444.176 /  MacOS Sequoia 15.7.1`
- Build Tool: `Vite`

### Related Component

- `el-table-column`

### Reproduction Link

[CodeSandbox](https://codesandbox.io/p/devbox/filters-36762j)

### Steps to reproduce

1. Create an ElTable with an ElTableColumn.
2. Initially, render the column without filters or filter-method props (so the column is not filterable).
3. Dynamically add the filters or filter-method prop to the column.
4. Alternatively, start with a filterable column and dynamically remove the filters prop.
5. Observe whether the filter icon appears or disappears in the column header.

### What is Expected?

The filter icon should dynamically appear when filters or filter-method is added, and disappear when they are removed. The filterable state should correctly reflect the presence of these props.

The filter icon's visibility does not change.
When adding filters, the icon remains hidden because the internal filterable flag is not updated to true.
When removing filters, the icon may remain visible (but non-functional) because filterable is not updated to false.

### Additional comments

_(empty)_

tuzixiangs:
This PR fixes an issue where the filter icon in ElTableColumn does not update dynamically when the filters or filter-method props are added or removed after the initial render.

Currently, the filterable property is only calculated during the component initialization. If a user dynamically toggles the filters prop (e.g., changing from null to an array), the watcher updates the filters value in columnConfig, but fails to re-evaluate the filterable flag. As a result, the filter icon remains hidden (or visible) incorrectly.

This change adds logic to the watcher in `watcher-helper.ts` to recalculate filterable whenever filters or filter-method changes, ensuring the UI stays in sync with the props.

Fixes #22877

### DIFF
--- a/packages/components/table/src/table-column/watcher-helper.ts
+++ b/packages/components/table/src/table-column/watcher-helper.ts
@@ -76,6 +76,12 @@ function useWatcher<T extends DefaultRow>(
           () => props_[columnKey],
           (newVal) => {
             instance.columnConfig.value[key as never] = newVal
+            if (key === 'filters' || key === 'filterMethod') {
+              instance.columnConfig.value['filterable'] = !!(
+                instance.columnConfig.value['filters'] ||
+                instance.columnConfig.value['filterMethod']
+              )
+            }
           }
         )
       }


tolking (approved):

Dsaquel (approved):
👍

PR #22883, opened 2025-11-20
ci: test

rzzf:
<h1>Do not merge!!!</h1>

### DIFF
--- a/packages/components/select/src/useSelect.ts
+++ b/packages/components/select/src/useSelect.ts
@@ -608,7 +608,7 @@ export const useSelect = (props: SelectProps, emit: SelectEmits) => {
       | OptionPublicInstance[]
       | SelectStates['selected']
   ) => {
-    const targetOption = isArray(option) ? option[0] : option
+    const targetOption = isArray(option) ? option[0] : option // Cause conflict
     let target = null
 
     if (!isNil(targetOption?.value)) {


rzzf:
Oh, the reason it ran successfully in my repository was simply that I removed the check, not that I actually replaced it with `mergeable`. At the moment, it seems that even using `mergeable` might not work, as it may not get the result that quickly. I'll investigate further.

https://github.com/rzzf/element-plus/actions/runs/19541427365/workflow

Dsaquel:
@rzzf Maybe `mergeable` need to have an approval ? :thinking:

rzzf:
> @rzzf Maybe `mergeable` need to have an approval ? 🤔

Maybe so. I think we can reverse the logic: as long as `mergeable_state` is not `dirty`, it indicates that there are no code conflicts.

Dsaquel:
~~`UNSTABLE` can mean that it is also in conflict no ?~~

>**Mergeable** with non-passing commit status

It will probably indicate that it has no code conflict. But other enum like `UNKNOWN`, `BEHIND` i have a doubt.

rzzf:
> ~`UNSTABLE` can mean that it is also in conflict no ?~
> 
> > **Mergeable** with non-passing commit status
> 
> It will probably indicate that it has no code conflict. But other enum like `UNKNOWN`, `BEHIND` i have a doubt.

unknown indicates the auto-merge result is not ready.
behind indicates the branch is behind the base branch.

GPT told me that mergeable and mergeable_state are computed asynchronously and require waiting until auto-merge completes to get accurate values. ~It also suggested a small trick: polling until the status becomes clear. 😃~

```yaml
# Can't work
run: |
  while [ "${{ github.event.pull_request.mergeable_state }}" == "unknown" ]; do
    echo "Waiting for GitHub to compute mergeable_state..."
    sleep 5
  done
```

rzzf:
I haven’t found a way to resolve the asynchronous state issue yet; I’ll take some time tomorrow to look into it more deeply.

Dsaquel:
Ok, thanks for your investigation 🫡

PR #22096, opened 2025-09-09
chore: upgrade to Vue 3.5

makedopamine:
Note to reviewers: It might be easier to review this commit by commit.

PR #22884, opened 2025-11-20
fix(components): [table] fix header HMR and keep v-if fallback

tuzixiangs:

1. The Issue Currently, the `header` slot of `ElTableColumn` does not trigger HMR updates. When developers modify the content of the header slot during development, the table header remains unchanged until a full reload.

2. Root Cause Analysis The issue stems from a conflict between Vue's renderSlot optimization and Element Plus's Table architecture.
- Architecture: `ElTableColumn` defines a `renderHeader` function (closure) and passes it to `ElTableHeader` via the Table Store.
- Vue Optimization: The `header` slot is often compiled as `STABLE` by Vue. When `renderSlot` is called inside `renderHeader`, it detects the `STABLE` flag and generates a VNode with `PatchFlags.STABLE_FRAGMENT`.
- The Conflict: Since `ElTableHeader` invokes `renderHeader` directly (not as a component slot), Vue's patch algorithm sees the `STABLE_FRAGMENT` flag and skips updating the children, causing HMR to fail.

3. Why not just bypass `renderSlot`? Simply replacing `renderSlot` with `h(Fragment, slots.header(scope))` solves the HMR issue but introduces a regression (reverting the fix from #15035). If a user writes `<template #header v-if="false">`, `slots.header` still exists but returns comment nodes. Without the validity check provided by `renderSlot`, the table would render empty comments instead of falling back to the `label` prop.

4. The Solution This PR implements a robust solution that supports both HMR and correct fallback behavior:

- Refactor: Extracted `ensureValidVNode` helper to `table/src/util.ts` (mirroring Vue's internal logic).
- Implementation: In `renderHeader`, we manually check the slot result using `ensureValidVNode`.

  - If valid: Render `h(Fragment, result)`. This creates a fresh Fragment without the `STABLE` flag, allowing Vue to patch updates correctly (HMR works).

  - If invalid (e.g., comments from `v-if="false"`): Fallback to rendering `column.label`.

Related Issue

- Fixes #21892

### DIFF
--- a/packages/components/table/src/table-column/render-helper.ts
+++ b/packages/components/table/src/table-column/render-helper.ts
@@ -1,5 +1,6 @@
 import {
   Comment,
+  Fragment,
   computed,
   getCurrentInstance,
   h,
@@ -16,7 +17,7 @@ import {
   getDefaultClassName,
   treeCellPrefix,
 } from '../config'
-import { parseMinWidth, parseWidth } from '../util'
+import { ensureValidVNode, parseMinWidth, parseWidth } from '../util'
 
 import type { ComputedRef, RendererNode, Slots, VNode } from 'vue'
 import type { TableColumn, TableColumnCtx } from './defaults'
@@ -122,7 +123,17 @@ function useRender<T extends DefaultRow>(
       column.renderHeader = (scope) => {
         // help render
         instance.columnConfig.value['label']
-        return renderSlot(slots, 'header', scope, () => [column.label])
+
+        if (slots.header) {
+          const slotResult = slots.header(scope)
+          // Manual valid check to support v-if fallback
+          // and bypass renderSlot to support HMR
+          if (ensureValidVNode(slotResult)) {
+            return h(Fragment, slotResult)
+          }
+        }
+
+        return h('span', column.label)
       }
     }
 
--- a/packages/components/table/src/table-column/render-helper.ts
+++ b/packages/components/table/src/table-column/render-helper.ts
@@ -2,6 +2,7 @@ import {
   Comment,
   Fragment,
   computed,
+  createTextVNode,
   getCurrentInstance,
   h,
   ref,
@@ -133,7 +134,7 @@ function useRender<T extends DefaultRow>(
           }
         }
 
-        return h('span', column.label)
+        return createTextVNode(column.label)
       }
     }
 
--- a/packages/components/table/src/util.ts
+++ b/packages/components/table/src/util.ts
@@ -1,4 +1,4 @@
-import { createVNode, isVNode, render } from 'vue'
+import { Comment, Fragment, createVNode, isVNode, render } from 'vue'
 import { flatMap, get, isNull, merge } from 'lodash-unified'
 import {
   ensureArray,
@@ -19,7 +19,7 @@ import ElTooltip, {
 
 import type { DefaultRow, Table, TreeProps } from './table/defaults'
 import type { TableColumnCtx } from './table-column/defaults'
-import type { CSSProperties, VNode } from 'vue'
+import type { CSSProperties, VNode, VNodeArrayChildren } from 'vue'
 
 export type TableOverflowTooltipOptions = Partial<
   Pick<
@@ -675,4 +675,22 @@ export const ensurePosition = (
   if (!Number.isNaN(style[key])) {
     style[key] = `${style[key]}px` as any
   }
+}
+
+export function ensureValidVNode(
+  vnodes: VNodeArrayChildren
+): VNodeArrayChildren | null {
+  return vnodes.some((child) => {
+    if (!isVNode(child)) return true
+    if (child.type === Comment) return false
+    if (
+      child.type === Fragment &&
+      !ensureValidVNode(child.children as VNodeArrayChildren)
+    ) {
+      return false
+    }
+    return true
+  })
+    ? vnodes
+    : null
 }


packages/components/table/src/table-column/render-helper.ts
```diff
@@ -122,7 +123,17 @@ function useRender<T extends DefaultRow>(
...
+        if (slots.header) {
+          const slotResult = slots.header(scope)
+          // Manual valid check to support v-if fallback
+          // and bypass renderSlot to support HMR
+          if (ensureValidVNode(slotResult)) {
+            return h(Fragment, slotResult)
+          }
+        }
+
+        return h('span', column.label)
```
btea:
```suggestion
        return column.label
```
We should keep the tag structure consistent with the one we rendered earlier.

tuzixiangs: Returning a plain string causes renderHeader to be inferred as `string | VNode` , which violates the declared signature `TableColumnCtx['renderHeader'] => VNode` . To keep the existing tag structure unchanged while satisfying the type constraint, I switched the fallback to a text `VNode: createTextVNode(column.label)` . The rest of the logic remains the same. Please take another look and let me know if this meets your expectations.

btea: look good to me 👍

btea (approved):

Dsaquel:
The change looks good to me and it does indeed fix the issue.

But I also tested the issue with the preview of https://github.com/element-plus/element-plus/pull/22096 and it seems to solve it.

I'm not sure now if we should handle this fix. 

cc @btea @makedopamine

tuzixiangs:
> The change looks good to me and it does indeed fix the issue.
> 
> But I also tested the issue with the preview of #22096 and it seems to solve it.
> 
> I'm not sure now if we should handle this fix.
> 
> cc @btea @makedopamine

Hi @Dsaquel, thank you for reviewing this PR!

Regarding your concern about PR #22096 potentially solving this issue, I'd like to clarify:

Vue 3.5 Doesn't Fix the Root Cause
I've compared the renderSlot implementation between Vue 3.2 and 3.5.  The STABLE_FRAGMENT logic is identical:

```
// Both Vue 3.2 and 3.5 use the exact same condition
validSlotContent && (slots as RawSlots). _ === SlotFlags. STABLE
?  PatchFlags. STABLE_FRAGMENT
: PatchFlags. BAIL
```

Local Testing Confirms
I've tested locally using the build artifacts from PR #22096, and the HMR issue still persists.  This PR's fix is still necessary.

Looking forward to your feedback and thoughts on this analysis!

btea:
Indeed, #22096 does not seem to have solved this problem.

Dsaquel:
I re-tested and #22096 looks good but if you have tested #22096 in codesandbox it will in fact don't work.

That said with your change it looks like it does work in codesandbox too :thinking: .

tuzixiangs:
> I re-tested and #22096 looks good but if you have tested #22096 in codesandbox it will in fact don't work.
> 
> That said with your change it looks like it does work in codesandbox too 🤔 .

Thanks for the re-testing!  Yeah, I noticed something interesting - I can't reproduce this bug in Element Playground at all (tried different versions), but it consistently shows up in CodeSandbox and local projects .

Not sure if it's something specific to the Playground environment, but since developers actually work in CodeSandbox/local setups, I think this fix might still be useful for those cases.

What do you think?

Dsaquel:
>Thanks for the re-testing! Yeah, I noticed something interesting - I can't reproduce this bug in Element Playground at all (tried different versions), but it consistently shows up in CodeSandbox and local projects .

The playground will reload the sandbox every save so there is no hmr there.

>Not sure if it's something specific to the Playground environment, but since developers actually work in CodeSandbox/local setups, I think this fix might still be useful for those cases.

In local setup it works in my environment, environments like codesandbox or other it can fail.
I'm ok to merge to fix that :+1: .

packages/components/table/src/util.ts
```diff
@@ -676,3 +676,21 @@ export const ensurePosition = (
     style[key] = `${style[key]}px` as any
   }
 }
+
+export function ensureValidVNode(
+  vnodes: VNodeArrayChildren
+): VNodeArrayChildren | null {
+  return vnodes.some((child) => {
+    if (!isVNode(child)) return true
```
Dsaquel, line 684: Can you just explain why we don't return false here ?

tuzixiangs: Great question! This `ensureValidVNode` function is directly copied from Vue 3's `renderSlot` implementation ([source](https://github.com/vuejs/core/blob/main/packages/runtime-core/src/helpers/renderSlot.ts#L103-L119)).

From my understanding of Vue's design, the return type is `VNodeArrayChildren | null` rather than `boolean`:
1. **Purpose**: Validates whether a VNode array contains valid (non-comment, non-empty-fragment) nodes
2. **Return semantics**: 
   - Returns the **original `vnodes` array** if valid nodes exist
   - Returns `null` if all nodes are invalid
   - Not `false` because the caller expects either the array or `null` for conditional rendering
 
This design aligns with Vue's internal slot rendering logic. Let me know if you'd like me to add a JSDoc comment clarifying this!

Dsaquel: Thanks for the explanation :bow: .

Dsaquel (approved):

PR #22907, opened 2025-11-24
improvement(components): [splitter] use import emits

warmthsea:
Similar to other components.

Also https://github.com/element-plus/element-plus/actions/runs/19628925766/job/56203799580?pr=22096

### DIFF
--- a/packages/components/splitter/src/splitter.ts
+++ b/packages/components/splitter/src/splitter.ts
@@ -23,5 +23,6 @@ export const splitterEmits = {
   resizeEnd: (index: number, sizes: number[]) => true,
   collapse: (index: number, type: 'start' | 'end', sizes: number[]) => true,
 }
+/* eslint-enable @typescript-eslint/no-unused-vars */
 
 export type SplitterEmits = typeof splitterEmits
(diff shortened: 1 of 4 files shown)


packages/components/splitter/src/splitter.ts
```diff
@@ -15,3 +15,13 @@ export const splitterProps = buildProps({
...
 export type SplitterInstance = InstanceType<typeof Splitter> & unknown
+
+/* eslint-disable @typescript-eslint/no-unused-vars */
+export const splitterEmits = {
+  resizeStart: (index: number, sizes: number[]) => true,
+  resize: (index: number, sizes: number[]) => true,
+  resizeEnd: (index: number, sizes: number[]) => true,
+  collapse: (index: number, type: 'start' | 'end', sizes: number[]) => true,
+}
+
```
rzzf, line 27:
```suggestion
/* eslint-enable @typescript-eslint/no-unused-vars */

```

warmthsea: No eslint comments are needed here.

Image

rzzf: Yeah, but please note that we should **re-enable** the ESLint rule here.

warmthsea: > we should **re-enable** the ESLint rule here.

Could you explain in more detail? I don't quite understand what you mean.

warmthsea: I see, I don't think I've ever paid attention to this before.

rzzf (approved):

PR #22908, opened 2025-11-24
feat(components): [menu] apply popper-class and popper-style to the tooltip of the title

keeplearning66:

relate #22888 

1. I also noticed `popper-offset` prop. Should we apply it to the tooltip of the title as well?
2. The description in the documentation about **Popper Offset** says **Menu with popperOffset will override Submenu's popper-offset.** Isn't this reversed? It should be **Submenu with popperOffset will override Menu's popper-offset**, right? 🤔


rzzf (approved):

Dsaquel:
>The description in the documentation about Popper Offset says Menu with popperOffset will override Submenu's popper-offset. Isn't this reversed? It should be Submenu with popperOffset will override Menu's popper-offset, right?

It is, would you mind to update this too ?

>I also noticed popper-offset prop. Should we apply it to the tooltip of the title as well?

I think it will be dangerous to edit that. Perhaps adding a new prop for this better.

patienry:
> It is, would you mind to update this too ?

Okay, it's updated.

> I think it will be dangerous to edit that. Perhaps adding a new prop for this better.

If we need to add a new prop, maybe we could wait until a user requests it before handling it. The name of the prop might also need some discussion. WDYT ? 👀 👀

Dsaquel:
>If we need to add a new prop, maybe we could wait until a user requests it before handling it. The name of the prop might also need some discussion. WDYT ? 👀 👀

You right, we can wait until a developer request :+1: .

Dsaquel (approved):

btea (approved):

PR #22922, opened 2025-11-25
feat(components): [select-v2] add `option-extra-width` prop

feynmx:
**Summary**  
- Add `optionExtraWidth` prop (number or `(option) => number`) so SelectV2 can account for slot-rendered adornments when auto-calculating dropdown width.  
- Extend `calculateLabelMaxWidth` to include the declared extra width per option, and react to prop changes when `fit-input-width` is false.  
- Document the workflow with a “Slot Width Compensation” section, API entry, and demo (`select-v2/slot-extra-width`) showing icons inside option slots.

**Testing**  
- [ ] `pnpm test select-v2` (or relevant unit suite)  
- [ ] `pnpm docs:build` (docs compile)  
- [ ] Manually verified the new demo and SelectV2 behavior in dev server

**Notes**  
- Default value is `0`, so existing users are unaffected.  
- Demo illustrates why `fit-input-width="false"` + `option-extra-width="24"` keeps icon + label fully visible.

### DIFF
--- a/docs/en-US/component/select-v2.md
+++ b/docs/en-US/component/select-v2.md
@@ -231,6 +231,16 @@ select-v2/custom-width
 
 :::
 
+## Slot Width Compensation
+
+When you're using the default slot to render extra content (icons, tags, etc.) before the label, the automatically calculated dropdown width might still only consider the text. Use `option-extra-width` to hint how many extra pixels should be reserved so the dropdown can fully display the slot content.
+
+::::demo
+
+select-v2/slot-extra-width
+
+::::
+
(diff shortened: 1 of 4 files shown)


keeplearning66 (commented):
Would directly using `fit-input-width` to customize the width meet your requirements? 🤔 
I feel that with `fit-input-width` available, `option-extra-width` seems a bit redundant, since you can directly use `fit-input-width` to set the overall estimated width of the `option`.
https://element-plus.org/en-US/component/select-v2#custom-width

PR #22929, opened 2025-11-26
feat(components): [timeline] support `mode` prop

keeplearning66:

Make `Timeline` support `mode` prop, allowing the relative position of the timeline and content to be set.

## Summary by CodeRabbit

* **New Features**
  * Added a Timeline "mode" option with four layout modes (start/alternate/alternate-reverse/end) to control timeline/content positioning.

* **Documentation**
  * Added a Mode section, updated API table, and a live example demonstrating mode switching.
  * Removed fixed max-width from multiple timeline examples to use default sizing.

* **Style**
  * Updated timeline styles to support the four responsive layouts.

* **Tests**
  * Added tests covering all four modes.


rzzf:
This is a nice feature. Regarding the prop name, I think mode might be a better choice. I referenced the naming conventions used by [antd](https://ant.design/components/timeline-cn#timeline), [arco](https://arco.design/react/components/timeline#%E7%BA%B5%E5%90%91%E6%97%B6%E9%97%B4%E8%BD%B4), and [tdesign](https://tdesign.tencent.com/vue-next/components/timeline?tab=demo#%E5%B8%83%E5%B1%80%E6%96%B9%E5%BC%8F).

Additionally, are you interested in continuing the implementation of the **alternating** mode?

patienry:
> This is a nice feature. Regarding the prop name, I think mode might be a better choice. I referenced the naming conventions used by [antd](https://ant.design/components/timeline-cn#timeline), [arco](https://arco.design/react/components/timeline#%E7%BA%B5%E5%90%91%E6%97%B6%E9%97%B4%E8%BD%B4), and [tdesign](https://tdesign.tencent.com/vue-next/components/timeline?tab=demo#%E5%B8%83%E5%B1%80%E6%96%B9%E5%BC%8F).

Thank you for your suggestion, I have already made the changes.

> Additionally, are you interested in continuing the implementation of the alternating mode?

The `alternate` mode is already supported.

patienry:
[demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtcmFkaW8tZ3JvdXAgdi1tb2RlbD1cIm1vZGVcIj5cbiAgICA8ZWwtcmFkaW8tYnV0dG9uIGxhYmVsPVwic3RhcnRcIiB2YWx1ZT1cInN0YXJ0XCIgLz5cbiAgICA8ZWwtcmFkaW8tYnV0dG9uIGxhYmVsPVwiYWx0ZXJuYXRlXCIgdmFsdWU9XCJhbHRlcm5hdGVcIiAvPlxuICAgIDxlbC1yYWRpby1idXR0b24gbGFiZWw9XCJhbHRlcm5hdGUtcmV2ZXJzZVwiIHZhbHVlPVwiYWx0ZXJuYXRlLXJldmVyc2VcIiAvPlxuICAgIDxlbC1yYWRpby1idXR0b24gbGFiZWw9XCJlbmRcIiB2YWx1ZT1cImVuZFwiIC8+XG4gIDwvZWwtcmFkaW8tZ3JvdXA+XG5cbiAgPGVsLXRpbWVsaW5lIHN0eWxlPVwibWFyZ2luLXRvcDogMTZweFwiIDptb2RlPVwibW9kZVwiPlxuICAgIDxlbC10aW1lbGluZS1pdGVtXG4gICAgICB2LWZvcj1cIihhY3Rpdml0eSwgaW5kZXgpIGluIGFjdGl2aXRpZXNcIlxuICAgICAgOmtleT1cImluZGV4XCJcbiAgICAgIDp0aW1lc3RhbXA9XCJhY3Rpdml0eS50aW1lc3RhbXBcIlxuICAgID5cbiAgICAgIHt7IGFjdGl2aXR5LmNvbnRlbnQgfX1cbiAgICA8L2VsLXRpbWVsaW5lLWl0ZW0+XG4gIDwvZWwtdGltZWxpbmU+XG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IGxhbmc9XCJ0c1wiIHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSAndnVlJ1xuXG5pbXBvcnQgdHlwZSB7IFRpbWVsaW5lUHJvcHMgfSBmcm9tICdlbGVtZW50LXBsdXMnXG5cbmNvbnN0IGFjdGl2aXRpZXMgPSBbXG4gIHtcbiAgICBjb250ZW50OiAnRXZlbnQgc3RhcnQnLFxuICAgIHRpbWVzdGFtcDogJzIwMTgtMDQtMTUnLFxuICB9LFxuICB7XG4gICAgY29udGVudDogJ0FwcHJvdmVkJyxcbiAgICB0aW1lc3RhbXA6ICcyMDE4LTA0LTEzJyxcbiAgfSxcbiAge1xuICAgIGNvbnRlbnQ6ICdTdWNjZXNzJyxcbiAgICB0aW1lc3RhbXA6ICcyMDE4LTA0LTExJyxcbiAgfSxcbl1cblxuY29uc3QgbW9kZSA9IHJlZjxUaW1lbGluZVByb3BzWydtb2RlJ10+KCdzdGFydCcpXG48L3NjcmlwdD5cbiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vcHJldmlldy0yMjkyOS1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzJywgJ2h0dHBzOi8vcHJldmlldy0yMjkyOS1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL3RoZW1lLWNoYWxrL2RhcmsvY3NzLXZhcnMuY3NzJ10ubWFwKChzdHlsZSkgPT4ge1xuICAgIHJldHVybiBuZXcgUHJvbWlzZSgocmVzb2x2ZSwgcmVqZWN0KSA9PiB7XG4gICAgICBjb25zdCBsaW5rID0gZG9jdW1lbnQuY3JlYXRlRWxlbWVudCgnbGluaycpXG4gICAgICBsaW5rLnJlbCA9ICdzdHlsZXNoZWV0J1xuICAgICAgbGluay5ocmVmID0gc3R5bGVcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignbG9hZCcsIHJlc29sdmUpXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2Vycm9yJywgcmVqZWN0KVxuICAgICAgZG9jdW1lbnQuYm9keS5hcHBlbmQobGluaylcbiAgICB9KVxuICB9KVxuICByZXR1cm4gUHJvbWlzZS5hbGxTZXR0bGVkKHN0eWxlcylcbn1cbiIsInRzY29uZmlnLmpzb24iOiJ7XG4gIFwiY29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiBcIkVTTmV4dFwiLFxuICAgIFwianN4XCI6IFwicHJlc2VydmVcIixcbiAgICBcIm1vZHVsZVwiOiBcIkVTTmV4dFwiLFxuICAgIFwibW9kdWxlUmVzb2x1dGlvblwiOiBcIkJ1bmRsZXJcIixcbiAgICBcInR5cGVzXCI6IFtcImVsZW1lbnQtcGx1cy9nbG9iYWwuZC50c1wiXSxcbiAgICBcImFsbG93SW1wb3J0aW5nVHNFeHRlbnNpb25zXCI6IHRydWUsXG4gICAgXCJhbGxvd0pzXCI6IHRydWUsXG4gICAgXCJjaGVja0pzXCI6IHRydWVcbiAgfSxcbiAgXCJ2dWVDb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IDMuM1xuICB9XG59XG4iLCJQbGF5Z3JvdW5kTWFpbi52dWUiOiI8c2NyaXB0IHNldHVwPlxuaW1wb3J0IEFwcCBmcm9tICcuL0FwcC52dWUnXG5pbXBvcnQgeyBzZXR1cEVsZW1lbnRQbHVzIH0gZnJvbSAnLi9lbGVtZW50LXBsdXMuanMnXG5zZXR1cEVsZW1lbnRQbHVzKClcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxBcHAgLz5cbjwvdGVtcGxhdGU+XG4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvcnVudGltZS1kb21AbGF0ZXN0L2Rpc3QvcnVudGltZS1kb20uZXNtLWJyb3dzZXIuanNcIixcbiAgICBcIkB2dWUvc2hhcmVkXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AdnVlL3NoYXJlZEBsYXRlc3QvZGlzdC9zaGFyZWQuZXNtLWJ1bmRsZXIuanNcIixcbiAgICBcImVsZW1lbnQtcGx1c1wiOiBcImh0dHBzOi8vcHJldmlldy0yMjkyOS1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwidW5zdXBwb3J0ZWRcIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJfbyI6eyJzaG93SGlkZGVuIjp0cnVlLCJzdHlsZVNvdXJjZSI6Imh0dHBzOi8vcHJldmlldy0yMjkyOS1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzIn19)

packages/components/timeline/src/tokens.ts
```diff
@@ -1 +1,10 @@
-export const TIMELINE_INJECTION_KEY = 'timeline'
+import type { InjectionKey, Slots } from 'vue'
+import type { TimelineProps } from './timeline'
+
+export interface TimelineContext {
+  props: TimelineProps
+  slots: Slots
+}
+
+export const timelineInjectionKey: InjectionKey<TimelineContext> =
```
rzzf: It’s better to keep the original name, otherwise it might introduce a BC. I’m also not sure if anyone is using it.

packages/components/timeline/src/timeline-item.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <li :class="[ns.b(), { [ns.e('center')]: center }]">
+  <li :class="[ns.b(), { [ns.e('center')]: center }, timelineItemModeKls]">
```
rzzf: I think they can be combined.

packages/components/timeline/src/timeline.ts
```diff
@@ -29,8 +46,12 @@ const Timeline = defineComponent({
      *   so i can't reverse the slots, when i use 'v-for' directive.
      */
 
+    const timelineModeKls = computed(() => ns.is(props.mode))
+
     return () => {
-      return h('ul', { class: [ns.b()] }, [renderSlot(slots, 'default')])
+      return h('ul', { class: [ns.b(), timelineModeKls.value] }, [
```
rzzf: I think the classes here can be combined as well.

rzzf (commented):
The timeline currently uses a `<ul>`, which comes with a default browser `padding-left: 40px`.

<details>
  <summary>Image</summary>
  Image
</details>

So when the mode is not left, the layout looks a bit odd. Do you have any ideas?

<details>
  <summary>Image</summary>
  Image
</details>

patienry:
> So when the mode is not left, the layout looks a bit odd. Do you have any ideas?

Do you think the current result is satisfactory? 👀 👀 

{01B61B5B-F2C1-47D0-B1A4-365ADAD4B168}

{AAF2B018-0052-4979-8C01-1508B6AFEC30}

packages/components/timeline/src/tokens.ts
patienry: Okay, I've kept the original `TIMELINE_INJECTION_KEY`.

packages/components/timeline/src/timeline.ts
patienry: Done, thank you!

packages/components/timeline/src/timeline-item.vue
patienry: Done, thank you!

packages/theme-chalk/src/timeline.scss
```diff
@@ -43,4 +43,28 @@
       height: calc(50% - 10px);
     }
   }
+
+  @include when(right){
+    padding-left: 0;
+    padding-right: 40px;
```
rzzf, line 75: I think relying on default styles is not reliable, as developers may use tools like `normalize.css` to override them. 
In this case, if `left` or `alternate` is used, the original issue could resurface. 

We might need to explicitly override the default styles, but this could introduce a breaking change. Let’s wait and see what others suggest.

patienry: image


What if we explicitly set `padding` in different modes? Do you think that would be appropriate? 🤔

rzzf: Yeah, this is exactly what I mean, but it introduces a breaking change.
Additionally, in `alternate` mode, the left and right padding should be consistent.

rzzf: Maybe we can make this change, and merge it in the next minor release.

patienry: Okay, I have made the changes, and the version number is set to `2.12.0`.

packages/components/timeline/src/timeline-item.vue
```diff
@@ -39,15 +39,22 @@
...
 import { timelineItemProps } from './timeline-item'
+import { TIMELINE_INJECTION_KEY } from './tokens'
+
+import type { TimelineProvider } from './tokens'
 
 defineOptions({
   name: 'ElTimelineItem',
 })
 
+const { props: timelineProps } = inject<TimelineProvider>(
```
rzzf, line 56: nit: It’s better to place the inject statement below the macro definitions.

rzzf (approved):
look good to me.

packages/components/timeline/src/timeline-item.vue
patienry: Done, thank you!

Dsaquel (approved):

patienry:
Should the version number of this PR be updated to `2.14.0`? 👀 👀

Dsaquel:
I don't think the feature need a breaking change label as it preserve the visual aspect (`padding-left: 40px`).

rzzf:
> I don't think the feature need a breaking change label as it preserve the visual aspect (`padding-left: 40px`).

For users who rely on reset.css, normalize.css(for example [Documentation](https://element-plus.org/en-US/component/timeline#basic-usage)), or similar style reset solutions, would this be considered a breaking change?

image

Dsaquel:
Didn't know about reset.css and it has indeed a padding reset however normalize.css don't reset the ul padding.

I think its ok to merge in `2.13.1`, what do you think ?

rzzf:
> I think its ok to merge in `2.13.1`, what do you think ?

I agree. 👍

docs/en-US/component/timeline.md
```diff
@@ -61,9 +61,10 @@ timeline/reverse
 
 ### Timeline Attributes
 
-| Name              | Description           | Type       | Default |
-| ----------------- | --------------------- | ---------- | ------- |
-| reverse ^(2.11.9) | whether reverse order | ^[boolean] | false   |
+| Name              | Description                               | Type                                      | Default |
+| ----------------- | ----------------------------------------- | ----------------------------------------- | ------- |
+| reverse ^(2.11.9) | whether reverse order                     | ^[boolean]                                | false   |
+| mode ^(2.13.0)    | relative position of timeline and content | ^[enum]`'left' \| 'alternate' \| 'right'` | left    |
```
rzzf:
```suggestion
| mode ^(2.13.1)    | relative position of timeline and content | ^[enum]`'left' \| 'alternate' \| 'right'` | left    |
```

patienry:
Updated, thank you both!

patienry:
I've encountered an issue. I plan to add an example for `mode`, but there's a reset style in the document environment (which most users' environments likely handle this way), causing the `alternate` mode styles to be incorrect. Currently, the `alternate` mode styles are calculated based on `content-box`. Would it be suitable to set `box-sizing: border-box;` in `.el-timeline-item__wrapper` and modify the style calculation? (I’ve noticed that setting it in `.el-timeline` or `.el-timeline-item` gets overridden in the end.) Do you have any thoughts or suggestions on this? 👀 🤔  @rzzf @Dsaquel 

2e8a4c70ab73fd5b97517e349b4f1698

f6212c00a633c2d2561af764febf2627

rzzf:
It seems that only the width and padding of `el-timeline-item__wrapper` are affected. 
If possible, we could remove the `padding` so it works correctly in both modes. 
If that’s hard to implement, we could also explicitly set `box-sizing: content-box;`. Either approach works for me.

patienry:
> we could also explicitly set box-sizing: content-box;

Okay, this change is smaller. I also added an example.

docs/examples/timeline/mode.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <el-radio-group v-model="mode">
+    <el-radio-button label="left" value="left" />
+    <el-radio-button label="alternate" value="alternate" />
+    <el-radio-button label="right" value="right" />
+  </el-radio-group>
+
+  <el-timeline class="mt-4" style="max-width: 600px" :mode="mode">
```
rzzf:
```suggestion
  <el-timeline class="mt-4" :mode="mode">
```

This style can be removed to provide a clearer reference.

packages/theme-chalk/src/timeline.scss
```diff
@@ -43,4 +43,34 @@
       height: calc(50% - 10px);
     }
   }
+
+  @include when(left) {
+    padding-left: 40px;
```
rzzf, line 69: It would be better to also initialize `padding-right: 0` here. In some locales with right-to-left layouts, the default `ul` padding appears on the right side.

Ref https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/padding-inline-start

docs/examples/timeline/mode.vue
patienry: Done, thank you! I also updated the other examples for consistency.

packages/theme-chalk/src/timeline.scss
patienry: Done, thank you!

rzzf (approved):
Thank you! This looks good to me now.

Dsaquel:
The content-box solution looks good to me too :+1: .

packages/theme-chalk/src/timeline.scss
```diff
@@ -43,4 +43,35 @@
...
+  @include when(right) {
+    padding-left: 0;
+    padding-right: 40px;
+  }
+
+  @include when(alternate) {
+    padding-left: 20px;
+    padding-right: 20px;
+
+    .#{$namespace}-timeline-item:nth-child(2n + 1) {
```
btea, line 93: We could add another configuration to determine whether the style applies to the 2n application or the 2n+1 application. 🤔

patienry: Do you mean adding a prop like `alternate-first-position` to achieve the effect below? 🤔 

<table border="1">
  <tr>
    <th>alternate-first-position=right</th>
    <th>alternate-first-position=left</th>
  </tr>
  <tr>
    <td>
{6B886F59-FDB3-40D0-B04F-3E853F68E8D6}
</td>
    <td>
90583CB4-34AC-46d7-AC03-33D26A4F5E59
</td>
  </tr>
</table>

btea: Yes, absolutely right. Do you think it's necessary? Let's see what other people think. cc @rzzf @Dsaquel

patienry: Although other component libraries don't seem to have a similar prop, this could indeed improve flexibility. Let's wait for the thoughts of other team members. 👀 👀 

*Perhaps we can also assess its necessity based on feedback from future users.*

rzzf: I agree. I think providing this property won’t cause any issues.

btea: I'm thinking that if we were to add this feature later, we might have to split the current `alternate` configuration into two. To avoid affecting existing users, we might retain `alternate` and add a new value, which could make the two values ​​appear asymmetrical. 

Perhaps this update would directly support this, preventing similar situations from occurring in the future.

patienry: > Perhaps this update would directly support this, preventing similar situations from occurring in the future.

Then, would it be appropriate to make `mode` support these four values? 🤔 

A314537F-1B7A-48ff-B50A-664F00B9D414

btea: I think it's okay. 👍

patienry: Okay, I have made the changes. [demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtcmFkaW8tZ3JvdXAgdi1tb2RlbD1cIm1vZGVcIj5cbiAgICA8ZWwtcmFkaW8tYnV0dG9uIGxhYmVsPVwibGVmdFwiIHZhbHVlPVwibGVmdFwiIC8+XG4gICAgPGVsLXJhZGlvLWJ1dHRvbiBsYWJlbD1cImFsdGVybmF0ZS1sZWZ0XCIgdmFsdWU9XCJhbHRlcm5hdGUtbGVmdFwiIC8+XG4gICAgPGVsLXJhZGlvLWJ1dHRvbiBsYWJlbD1cImFsdGVybmF0ZS1yaWdodFwiIHZhbHVlPVwiYWx0ZXJuYXRlLXJpZ2h0XCIgLz5cbiAgICA8ZWwtcmFkaW8tYnV0dG9uIGxhYmVsPVwicmlnaHRcIiB2YWx1ZT1cInJpZ2h0XCIgLz5cbiAgPC9lbC1yYWRpby1ncm91cD5cblxuICA8ZWwtdGltZWxpbmUgc3R5bGU9XCJtYXJnaW4tdG9wOiAxNnB4O1wiIDptb2RlPVwibW9kZVwiPlxuICAgIDxlbC10aW1lbGluZS1pdGVtXG4gICAgICB2LWZvcj1cIihhY3Rpdml0eSwgaW5kZXgpIGluIGFjdGl2aXRpZXNcIlxuICAgICAgOmtleT1cImluZGV4XCJcbiAgICAgIDp0aW1lc3RhbXA9XCJhY3Rpdml0eS50aW1lc3RhbXBcIlxuICAgID5cbiAgICAgIHt7IGFjdGl2aXR5LmNvbnRlbnQgfX1cbiAgICA8L2VsLXRpbWVsaW5lLWl0ZW0+XG4gIDwvZWwtdGltZWxpbmU+XG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IGxhbmc9XCJ0c1wiIHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSAndnVlJ1xuXG5jb25zdCBhY3Rpdml0aWVzID0gW1xuICB7XG4gICAgY29udGVudDogJ0V2ZW50IHN0YXJ0JyxcbiAgICB0aW1lc3RhbXA6ICcyMDE4LTA0LTE1JyxcbiAgfSxcbiAge1xuICAgIGNvbnRlbnQ6ICdBcHByb3ZlZCcsXG4gICAgdGltZXN0YW1wOiAnMjAxOC0wNC0xMycsXG4gIH0sXG4gIHtcbiAgICBjb250ZW50OiAnU3VjY2VzcycsXG4gICAgdGltZXN0YW1wOiAnMjAxOC0wNC0xMScsXG4gIH0sXG5dXG5cbmNvbnN0IG1vZGUgPSByZWY8J2xlZnQnIHwgJ2FsdGVybmF0ZS1sZWZ0JyB8ICdhbHRlcm5hdGUtcmlnaHQnIHwgJ3JpZ2h0Jz4oXG4gICdsZWZ0J1xuKVxuPC9zY3JpcHQ+XG4iLCJlbGVtZW50LXBsdXMuanMiOiJpbXBvcnQgRWxlbWVudFBsdXMgZnJvbSAnZWxlbWVudC1wbHVzJ1xuaW1wb3J0IHsgZ2V0Q3VycmVudEluc3RhbmNlIH0gZnJvbSAndnVlJ1xuXG5sZXQgaW5zdGFsbGVkID0gZmFsc2VcbmF3YWl0IGxvYWRTdHlsZSgpXG5cbmV4cG9ydCBmdW5jdGlvbiBzZXR1cEVsZW1lbnRQbHVzKCkge1xuICBpZiAoaW5zdGFsbGVkKSByZXR1cm5cbiAgY29uc3QgaW5zdGFuY2UgPSBnZXRDdXJyZW50SW5zdGFuY2UoKVxuICBpbnN0YW5jZS5hcHBDb250ZXh0LmFwcC51c2UoRWxlbWVudFBsdXMpXG4gIGluc3RhbGxlZCA9IHRydWVcbn1cblxuZXhwb3J0IGZ1bmN0aW9uIGxvYWRTdHlsZSgpIHtcbiAgY29uc3Qgc3R5bGVzID0gWydodHRwczovL3ByZXZpZXctMjI5MjktZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS9kaXN0L2luZGV4LmNzcycsICdodHRwczovL3ByZXZpZXctMjI5MjktZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS90aGVtZS1jaGFsay9kYXJrL2Nzcy12YXJzLmNzcyddLm1hcCgoc3R5bGUpID0+IHtcbiAgICByZXR1cm4gbmV3IFByb21pc2UoKHJlc29sdmUsIHJlamVjdCkgPT4ge1xuICAgICAgY29uc3QgbGluayA9IGRvY3VtZW50LmNyZWF0ZUVsZW1lbnQoJ2xpbmsnKVxuICAgICAgbGluay5yZWwgPSAnc3R5bGVzaGVldCdcbiAgICAgIGxpbmsuaHJlZiA9IHN0eWxlXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2xvYWQnLCByZXNvbHZlKVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdlcnJvcicsIHJlamVjdClcbiAgICAgIGRvY3VtZW50LmJvZHkuYXBwZW5kKGxpbmspXG4gICAgfSlcbiAgfSlcbiAgcmV0dXJuIFByb21pc2UuYWxsU2V0dGxlZChzdHlsZXMpXG59XG4iLCJ0c2NvbmZpZy5qc29uIjoie1xuICBcImNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogXCJFU05leHRcIixcbiAgICBcImpzeFwiOiBcInByZXNlcnZlXCIsXG4gICAgXCJtb2R1bGVcIjogXCJFU05leHRcIixcbiAgICBcIm1vZHVsZVJlc29sdXRpb25cIjogXCJCdW5kbGVyXCIsXG4gICAgXCJ0eXBlc1wiOiBbXCJlbGVtZW50LXBsdXMvZ2xvYmFsLmQudHNcIl0sXG4gICAgXCJhbGxvd0ltcG9ydGluZ1RzRXh0ZW5zaW9uc1wiOiB0cnVlLFxuICAgIFwiYWxsb3dKc1wiOiB0cnVlLFxuICAgIFwiY2hlY2tKc1wiOiB0cnVlXG4gIH0sXG4gIFwidnVlQ29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiAzLjNcbiAgfVxufVxuIiwiUGxheWdyb3VuZE1haW4udnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCBBcHAgZnJvbSAnLi9BcHAudnVlJ1xuaW1wb3J0IHsgc2V0dXBFbGVtZW50UGx1cyB9IGZyb20gJy4vZWxlbWVudC1wbHVzLmpzJ1xuc2V0dXBFbGVtZW50UGx1cygpXG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8QXBwIC8+XG48L3RlbXBsYXRlPlxuIiwiaW1wb3J0LW1hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge1xuICAgIFwidnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AdnVlL3J1bnRpbWUtZG9tQGxhdGVzdC9kaXN0L3J1bnRpbWUtZG9tLmVzbS1icm93c2VyLmpzXCIsXG4gICAgXCJAdnVlL3NoYXJlZFwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9zaGFyZWRAbGF0ZXN0L2Rpc3Qvc2hhcmVkLmVzbS1idW5kbGVyLmpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXNcIjogXCJodHRwczovL3ByZXZpZXctMjI5MjktZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS9kaXN0L2luZGV4LmZ1bGwubWluLm1qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzL1wiOiBcInVuc3VwcG9ydGVkXCIsXG4gICAgXCJAZWxlbWVudC1wbHVzL2ljb25zLXZ1ZVwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQGVsZW1lbnQtcGx1cy9pY29ucy12dWVAMi9kaXN0L2luZGV4Lm1pbi5qc1wiXG4gIH0sXG4gIFwic2NvcGVzXCI6IHt9XG59IiwiX28iOnsic2hvd0hpZGRlbiI6dHJ1ZSwic3R5bGVTb3VyY2UiOiJodHRwczovL3ByZXZpZXctMjI5MjktZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS9kaXN0L2luZGV4LmNzcyJ9fQ==) There's just one small issue: according to the effect described in the [comment](https://github.com/element-plus/element-plus/pull/22929#discussion_r2666750321), when `mode` is `alternate-left`, the content of the first item is on the **left** side of the timeline, while when `mode` is `left`, the content is on the **right** side of the timeline. I’m wondering if this could cause any confusion for users? 👀

btea: I think that's enough for now.

btea (approved):

packages/theme-chalk/src/timeline.scss
patienry: Okay. 😃

Dsaquel: The added modes are really great,

>Okay, I have made the changes. demo There's just one small issue: according to the effect described in the https://github.com/element-plus/element-plus/pull/22929#discussion_r2666750321, when mode is alternate-left, the content of the first item is on the left side of the timeline, while when mode is left, the content is on the right side of the timeline. I’m wondering if this could cause any confusion for users? 👀

Perhaps inversing `left` and `right` make more sense in this context ?
The default of the `mode` would be `right` in this case.

Dsaquel: Just checked [material-ui](https://mui.com/material-ui/react-timeline/#basic-timeline) has their timeline with default to right; [api](https://mui.com/material-ui/api/timeline/).

Dsaquel:
Just a quick follow-up question:
Currently, in alternate and right modes, the timeline itself moves to the center or right of the page. Is it possible/feasible to only change the item placement without moving the timeline itself ?

packages/theme-chalk/src/timeline.scss
patienry: >Perhaps inversing left and right make more sense in this context ?
The default of the mode would be right in this case.
Just checked [material-ui](https://mui.com/material-ui/react-timeline/#basic-timeline) has their timeline with default to right; [api](https://mui.com/material-ui/api/timeline/).

For the current default behavior, since the **entire timeline component** is visually on the **left** side of the container, it might make more sense for the default `mode` to be set to `left`. However, for the `alternate` mode, it could be beneficial to consider using terms like `alternate` and `alternate-reverse` as seen in material-ui, as this might help reduce any potential ambiguity. 🤔 WDYT? @btea @rzzf

patienry:
>Currently, in alternate and right modes, the timeline itself moves to the center or right of the page. Is it possible/feasible to only change the item placement without moving the timeline itself ?

The current approach is to provide more space for the **content** area. For example, when `mode` is set to `right`, the **content** area can utilize the remaining space on the left side. This approach is commonly adopted by most UI component libraries, such as [ant-design](https://ant.design/components/timeline#timeline-demo-end) and [naiveui](https://www.naiveui.com/en-US/os-theme/components/timeline#item-placement.vue). 🤔

packages/theme-chalk/src/timeline.scss
rzzf: `Alternate-reverse` looks really nice. 

As for the arrangement direction, I don’t really have a specific reference in mind. Material Design doesn’t seem to have a dedicated timeline component or guideline. So I think the approaches taken by MUI and mainstream Chinese UI libraries each have their own advantages. In any case, I’m fine with either solution.

Dsaquel:
>The current approach is to provide more space for the content area. For example, when mode is set to right, the content area can utilize the remaining space on the left side. This approach is commonly adopted by most UI component libraries, such as [ant-design](https://ant.design/components/timeline#timeline-demo-end) and [naiveui](https://www.naiveui.com/en-US/os-theme/components/timeline#item-placement.vue). 🤔

I'm ok with that then :+1:

packages/theme-chalk/src/timeline.scss
Dsaquel: Sorry for changing again the opinion of the design api but if we add `alternate` and `alternate-reverse` in `mode` it would be complicated to add the orientation of the timeline (`vertical` and `horizontal`) later.

Perhaps being sync with antd make more sense.

patienry: That's okay, it's better to discuss and come up with the most suitable solution. 👍 I'll finish the revisions tonight.

Currently, the latest version of [ant-design](https://ant.design/components/timeline#timeline) (6.1.4) has the values for `mode` as `start | alternate | end`, while previous versions used `left | alternate | right`. Should we align with the latest version and set `mode` to `start | alternate | end`?

Dsaquel: I think so, and adding the `reverse` prop too to resolve the behavior of `alternate-left`/`alternate-right`.

WDYT ? cc @btea @rzzf

patienry: > adding the reverse prop too to resolve the behavior of alternate-left/alternate-right.

Do you mean adding a new prop? The timeline component already has `reverse` prop.

> if we add alternate and alternate-reverse in mode it would be complicated to add the orientation of the timeline (vertical and horizontal) later.

I thought there was no need to resolve the behavior of `alternate-left/alternate-right` anymore. 😂

Dsaquel: >Do you mean adding a new prop? The timeline component already has reverse prop.

You right i forgot.

>>if we add alternate and alternate-reverse in mode it would be complicated to add the orientation of the timeline (vertical and horizontal) later.

>I thought there was no need to resolve the behavior of alternate-left/alternate-right anymore. 😂

Wanted to mean by that the orientation will be difficult in the current naming.
We can add then `alternate` `alternate-reverse` in `mode`.

patienry: Okay, it's been modified. [demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtcmFkaW8tZ3JvdXAgdi1tb2RlbD1cIm1vZGVcIj5cbiAgICA8ZWwtcmFkaW8tYnV0dG9uIGxhYmVsPVwic3RhcnRcIiB2YWx1ZT1cInN0YXJ0XCIgLz5cbiAgICA8ZWwtcmFkaW8tYnV0dG9uIGxhYmVsPVwiYWx0ZXJuYXRlXCIgdmFsdWU9XCJhbHRlcm5hdGVcIiAvPlxuICAgIDxlbC1yYWRpby1idXR0b24gbGFiZWw9XCJhbHRlcm5hdGUtcmV2ZXJzZVwiIHZhbHVlPVwiYWx0ZXJuYXRlLXJldmVyc2VcIiAvPlxuICAgIDxlbC1yYWRpby1idXR0b24gbGFiZWw9XCJlbmRcIiB2YWx1ZT1cImVuZFwiIC8+XG4gIDwvZWwtcmFkaW8tZ3JvdXA+XG5cbiAgPGVsLXRpbWVsaW5lIHN0eWxlPVwibWFyZ2luLXRvcDogMTZweFwiIDptb2RlPVwibW9kZVwiPlxuICAgIDxlbC10aW1lbGluZS1pdGVtXG4gICAgICB2LWZvcj1cIihhY3Rpdml0eSwgaW5kZXgpIGluIGFjdGl2aXRpZXNcIlxuICAgICAgOmtleT1cImluZGV4XCJcbiAgICAgIDp0aW1lc3RhbXA9XCJhY3Rpdml0eS50aW1lc3RhbXBcIlxuICAgID5cbiAgICAgIHt7IGFjdGl2aXR5LmNvbnRlbnQgfX1cbiAgICA8L2VsLXRpbWVsaW5lLWl0ZW0+XG4gIDwvZWwtdGltZWxpbmU+XG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IGxhbmc9XCJ0c1wiIHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSAndnVlJ1xuXG5pbXBvcnQgdHlwZSB7IFRpbWVsaW5lUHJvcHMgfSBmcm9tICdlbGVtZW50LXBsdXMnXG5cbmNvbnN0IGFjdGl2aXRpZXMgPSBbXG4gIHtcbiAgICBjb250ZW50OiAnRXZlbnQgc3RhcnQnLFxuICAgIHRpbWVzdGFtcDogJzIwMTgtMDQtMTUnLFxuICB9LFxuICB7XG4gICAgY29udGVudDogJ0FwcHJvdmVkJyxcbiAgICB0aW1lc3RhbXA6ICcyMDE4LTA0LTEzJyxcbiAgfSxcbiAge1xuICAgIGNvbnRlbnQ6ICdTdWNjZXNzJyxcbiAgICB0aW1lc3RhbXA6ICcyMDE4LTA0LTExJyxcbiAgfSxcbl1cblxuY29uc3QgbW9kZSA9IHJlZjxUaW1lbGluZVByb3BzWydtb2RlJ10+KCdzdGFydCcpXG48L3NjcmlwdD5cbiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vcHJldmlldy0yMjkyOS1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzJywgJ2h0dHBzOi8vcHJldmlldy0yMjkyOS1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL3RoZW1lLWNoYWxrL2RhcmsvY3NzLXZhcnMuY3NzJ10ubWFwKChzdHlsZSkgPT4ge1xuICAgIHJldHVybiBuZXcgUHJvbWlzZSgocmVzb2x2ZSwgcmVqZWN0KSA9PiB7XG4gICAgICBjb25zdCBsaW5rID0gZG9jdW1lbnQuY3JlYXRlRWxlbWVudCgnbGluaycpXG4gICAgICBsaW5rLnJlbCA9ICdzdHlsZXNoZWV0J1xuICAgICAgbGluay5ocmVmID0gc3R5bGVcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignbG9hZCcsIHJlc29sdmUpXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2Vycm9yJywgcmVqZWN0KVxuICAgICAgZG9jdW1lbnQuYm9keS5hcHBlbmQobGluaylcbiAgICB9KVxuICB9KVxuICByZXR1cm4gUHJvbWlzZS5hbGxTZXR0bGVkKHN0eWxlcylcbn1cbiIsInRzY29uZmlnLmpzb24iOiJ7XG4gIFwiY29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiBcIkVTTmV4dFwiLFxuICAgIFwianN4XCI6IFwicHJlc2VydmVcIixcbiAgICBcIm1vZHVsZVwiOiBcIkVTTmV4dFwiLFxuICAgIFwibW9kdWxlUmVzb2x1dGlvblwiOiBcIkJ1bmRsZXJcIixcbiAgICBcInR5cGVzXCI6IFtcImVsZW1lbnQtcGx1cy9nbG9iYWwuZC50c1wiXSxcbiAgICBcImFsbG93SW1wb3J0aW5nVHNFeHRlbnNpb25zXCI6IHRydWUsXG4gICAgXCJhbGxvd0pzXCI6IHRydWUsXG4gICAgXCJjaGVja0pzXCI6IHRydWVcbiAgfSxcbiAgXCJ2dWVDb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IDMuM1xuICB9XG59XG4iLCJQbGF5Z3JvdW5kTWFpbi52dWUiOiI8c2NyaXB0IHNldHVwPlxuaW1wb3J0IEFwcCBmcm9tICcuL0FwcC52dWUnXG5pbXBvcnQgeyBzZXR1cEVsZW1lbnRQbHVzIH0gZnJvbSAnLi9lbGVtZW50LXBsdXMuanMnXG5zZXR1cEVsZW1lbnRQbHVzKClcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxBcHAgLz5cbjwvdGVtcGxhdGU+XG4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvcnVudGltZS1kb21AbGF0ZXN0L2Rpc3QvcnVudGltZS1kb20uZXNtLWJyb3dzZXIuanNcIixcbiAgICBcIkB2dWUvc2hhcmVkXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AdnVlL3NoYXJlZEBsYXRlc3QvZGlzdC9zaGFyZWQuZXNtLWJ1bmRsZXIuanNcIixcbiAgICBcImVsZW1lbnQtcGx1c1wiOiBcImh0dHBzOi8vcHJldmlldy0yMjkyOS1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwidW5zdXBwb3J0ZWRcIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJfbyI6eyJzaG93SGlkZGVuIjp0cnVlLCJzdHlsZVNvdXJjZSI6Imh0dHBzOi8vcHJldmlldy0yMjkyOS1lbGVtZW50LXBsdXMuc3VyZ2Uuc2gvYnVuZGxlL2Rpc3QvaW5kZXguY3NzIn19)

btea: > I think so, and adding the `reverse` prop too to resolve the behavior of `alternate-left`/`alternate-right`.
> 
> WDYT ? cc @btea @rzzf

Look great.

Dsaquel: Perfect :ok_hand:

Dsaquel (approved):

btea (approved):

packages/theme-chalk/src/timeline.scss
rzzf: Perfect 👍

rzzf (approved):

rzzf:
I think we can merge this PR now.

PR #22979, opened 2025-12-01
docs: allow passing extra packages for ep-playground

Dsaquel:
[ref](https://github.com/element-plus/element-plus/pull/22971)

tldr: In some examples, the playground crashes because it doesn't have vueuse as dependencies.

Workarounds seems a bit complicated to show for developers, perhaps it is more comfortable to make vueuse and other packages controlled in this repository.
We would assume the version would be the latest as @keeplearning66 pointed [here](https://github.com/element-plus/element-plus-playground/pull/334#issue-3600366042).


btea (approved):

rzzf (approved):
👍

PR #23002, opened 2025-12-04
chore: remove type-fest

Dsaquel:
I think we can get rid of type-fest as it is used only one time.

### DIFF
--- a/packages/utils/objects.ts
+++ b/packages/utils/objects.ts
@@ -1,12 +1,11 @@
 import { get, set } from 'lodash-unified'
 
-import type { Entries } from 'type-fest'
 import type { Arrayable } from '.'
 
 export const keysOf = <T extends object>(arr: T) =>
   Object.keys(arr) as Array<keyof T>
 export const entriesOf = <T extends object>(arr: T) =>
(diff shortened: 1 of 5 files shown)


btea (approved):
LGTM

packages/utils/objects.ts
```diff
@@ -1,12 +1,11 @@
 import { get, set } from 'lodash-unified'
 
-import type { Entries } from 'type-fest'
 import type { Arrayable } from '.'
 
 export const keysOf = <T extends object>(arr: T) =>
   Object.keys(arr) as Array<keyof T>
 export const entriesOf = <T extends object>(arr: T) =>
-  Object.entries(arr) as Entries<T>
+  Object.entries(arr) as [keyof typeof arr, (typeof arr)[keyof typeof arr]][]
```
rzzf: Would it be better to use generics here? 🤔

```ts
[keyof T, T[keyof T]][]
```

Dsaquel: yes it's more shorter :+1: .

rzzf (approved):
👍

PR #23019, opened 2025-12-07
chore: refine process.env judgment

Dsaquel:
Just noticed now that in dev mode we don't get warnings anymore:

In browser context the condition `typeof process !== 'undefined'` will be always false. Resulting with no warnings in dev mode along with potentials breaks that happens in some part where the condition is involved.

https://github.com/element-plus/element-plus/blob/0e1c46fe903ca31718e7e22c8a15811a54653ac5/packages/utils/error.ts#L17

After reverting the fixes, the issue https://github.com/element-plus/element-plus/issues/21560 should not to be related in our side. 

I checked in the vitest side and it looks like `process` should be mocked in browser mode, [solution](https://github.com/element-plus/element-plus/issues/21560).

### DIFF
--- a/internal/build/src/tasks/modules.ts
+++ b/internal/build/src/tasks/modules.ts
@@ -44,6 +44,9 @@ const plugins: Plugin[] = [
     loaders: {
       '.vue': 'ts',
     },
+    define: {
+      'process.env.NODE_ENV': '"production"',
+    },
   }),
 ]
 
--- a/internal/build/src/tasks/modules.ts
+++ b/internal/build/src/tasks/modules.ts
@@ -44,9 +44,6 @@ const plugins: Plugin[] = [
     loaders: {
       '.vue': 'ts',
     },
-    define: {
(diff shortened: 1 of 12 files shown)


btea:
I haven't verified it yet, but based on the code logic, the logic for checking if the process exists should be consistent with the subsequent logic. 🤔

Dsaquel:
>the logic for checking if the process exists should be consistent with the subsequent logic.

The process checking is in fact not right for this reason:
In browser context the condition typeof process !== 'undefined' will be always false. Resulting with no warnings in dev mode along with potentials breaks that happens in some part where the condition is involved.

btea:
You are right. When actually using it, the development tool will first transform the code, for example, directly converting `process.env.NODE_ENV !== 'production'` to `true`.

I feel like I did something stupid, and then I came full circle. :sweat_smile:

btea (approved):

Dsaquel:
No worries, i haven't checked carefully too :smiley: .

PR #23029, opened 2025-12-08
fix(components): [data-picker] the panel `disabled` state should the same as component

jeff-fe:
fix https://github.com/element-plus/element-plus/issues/23030

### DIFF
--- a/packages/components/date-picker/__tests__/date-picker.test.ts
+++ b/packages/components/date-picker/__tests__/date-picker.test.ts
@@ -2674,7 +2674,7 @@ describe('MonthRange', () => {
       expect(formItem.attributes().role).toBe('group')
     })
 
-    it('form is disabled, but the datePicker is not.', async () => {
+    it('should give its own disabled prop higher priority within a form', async () => {
       const wrapper = _mount(
         `<el-form :disabled="true">
           <el-form-item>
--- a/packages/components/date-picker/__tests__/date-picker.test.ts
+++ b/packages/components/date-picker/__tests__/date-picker.test.ts
@@ -16,7 +16,7 @@ import zhCn from '@element-plus/locale/lang/zh-cn'
 import enUs from '@element-plus/locale/lang/en'
 import 'dayjs/locale/zh-cn'
 import { EVENT_CODE } from '@element-plus/constants'
-import { ElFormItem } from '@element-plus/components/form'
+import { ElForm, ElFormItem } from '@element-plus/components/form'
 import DatePicker from '../src/date-picker'
 import DatePickerRange from '@element-plus/components/date-picker-panel/src/date-picker-com/panel-date-range.vue'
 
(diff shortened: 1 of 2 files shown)


btea:
Could you add a test case?

btea (approved):

packages/components/date-picker/src/date-picker.tsx
```diff
@@ -72,7 +73,11 @@ export default defineComponent({
         >
           {{
             default: (scopedProps: /**FIXME: remove any type */ any) => (
-              <ElDatePickerPanel border={false} {...scopedProps}>
+              <ElDatePickerPanel
+                disabled={disabled.value}
```
rzzf: I think simply passing through the `disabled` prop is sufficient, since each type of panel component already uses `useFormDisabled` internally. The current handling here seems a bit redundant.
What do you think? @btea @jeff-fe 

See
https://github.com/element-plus/element-plus/blob/d215c1c1455ce3932fb50b32813b7a98d3f73169/packages/components/date-picker-panel/src/date-picker-panel.tsx#L72-L75

https://github.com/element-plus/element-plus/blob/dev/packages/components/date-picker-panel/src/panel-utils.ts

```suggestion
                disabled={props.disabled}
```

packages/components/date-picker/__tests__/date-picker.test.ts
```diff
@@ -2672,6 +2673,26 @@ describe('MonthRange', () => {
...
+      )
+
+      await nextTick()
+      const datePickerInput = wrapper.find('.el-input__inner')
+      expect(datePickerInput.attributes('disabled')).toBeUndefined()
+
+      datePickerInput.trigger('focus')
+      await nextTick()
+      const panel = document.querySelector('.el-picker-panel')
+      expect(panel?.classList.contains('is-disabled')).toBeFalsy()
```
rzzf, line 2694: This test still passes even before the change.

rzzf (changes requested):

packages/components/date-picker/src/date-picker.tsx
jeff-fe: @rzzf 谢谢，我对这块理解不深，如何才能合并你的代码

rzzf: 可以继续在这个分支提交新的代码。😀

> 如何才能合并你的代码
image

jeff-fe:
@rzzf 已更新

packages/components/date-picker/__tests__/date-picker.test.ts
```diff
@@ -2672,6 +2673,26 @@ describe('MonthRange', () => {
       const formItem = wrapper.find('[data-test-ref="item"]')
       expect(formItem.attributes().role).toBe('group')
     })
+
+    it('form is disabled, but the datePicker is not.', async () => {
```
rzzf:
```suggestion
    it('should give its own disabled prop higher priority within a form', async () => {
```

rzzf (approved):
@jeff-fe Thank you!

packages/components/date-picker/src/date-picker.tsx
btea: @rzzf That's true. 👍

PR #23142, opened 2025-12-17
style(components): [form/form-item] separate form and form-item styles

rzzf:
fix #23141

[...]

PR #23153, opened 2025-12-18
style(components): separate sub component styles

rzzf:
https://github.com/element-plus/element-plus/pull/23142#pullrequestreview-3589877398

e.g. components like menu-item, menu-item-group, dropdown-item, and dropdown-menu are inherently required to be wrapped by their corresponding root components. Using them standalone will result in errors, so no additional changes are necessary.


Dsaquel (approved):
👍

btea (approved):

warmthsea (approved):

PR #23161, opened 2025-12-19
fix(components): [table] ts-plugin(4023) error

rzzf:
https://github.com/element-plus/element-plus/actions/runs/20015988731/job/58432046686?pr=23013

### DIFF
--- a/packages/components/table/src/table.vue
+++ b/packages/components/table/src/table.vue
@@ -196,13 +196,14 @@ import { TABLE_INJECTION_KEY } from './tokens'
 import { hColgroup } from './h-helper'
 import { useScrollbar } from './composables/use-scrollbar'
 
+import type { ObjectDirective } from 'vue'
 import type { Table } from './table/defaults'
 
 let tableIdSeed = 1
 export default defineComponent({
   name: 'ElTable',
   directives: {
-    Mousewheel,
+    Mousewheel: Mousewheel as ObjectDirective,
   },
   components: {
     TableHeader,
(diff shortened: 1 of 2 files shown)


btea (approved):

makedopamine:
It's werid. I tried checking out [8c4aaec83](https://github.com/element-plus/element-plus/actions/runs/20015988731/job/58432046686?pr=23013#step:5:495), and the build completed successfully in local. It seems that the error CI reported is not due to the two non-exported types.

rzzf:
> It's werid. I tried checking out [8c4aaec83](https://github.com/element-plus/element-plus/actions/runs/20015988731/job/58432046686?pr=23013#step:5:495), and the build completed successfully in local. It seems that the error CI reported is not due to the two non-exported types.

This is a bit strange. on the dev branch, I encounter the same error both during local builds and in VS Code. However, once I export these two types, everything works as expected.

image

makedopamine:
Did you run into the same issue while building or in VS Code before this PR? @btea

Dsaquel:
It is surely related to the difference of the build of the dev branch vs build of the PR, [ref](https://github.com/element-plus/element-plus/pull/23058#discussion_r2606728987).

btea:
> Did you run into the same issue while building or in VS Code before this PR? @btea

Actually, it works fine on my local machine. However, the address provided by @rzzf did throw an error, and since exporting two variables doesn't seem to have any side effects, I merged them.

btea:
> It is surely related to the difference of the build of the dev branch vs build of the PR, [ref](https://github.com/element-plus/element-plus/pull/23058#discussion_r2606728987).

That's very likely the reason.

PR #23174, opened 2025-12-20
fix(components): [date-picker] trigger clear evt in footer slot

Dsaquel:
fix #23173


rzzf (commented):
Maybe we should also update the documentation for date-picker-panel?

rzzf (approved):
👍

PR #23195, opened 2025-12-23
fix(components): [select/select-v2] show empty slot when remote search empty

YXY-cell:
Fixes #23191

### What is this change?

Fixes the issue where the dropdown menu prevents the empty slot from showing when remote search returns empty.

### Why is this change needed?

When `remote` is enabled and search results are empty, the dropdown menu logic was incorrectly hiding the menu, making the `#empty` slot (e.g., "Add Option" button) invisible to users. This fix adds a condition to check for the empty slot.

## Summary by CodeRabbit

* **Tests**
  * Added tests verifying select components (v1 and v2) display a custom empty-state template when remote searches return no results.

* **Bug Fixes**
  * Fixed dropdown visibility so custom empty-state templates are shown correctly during remote search flows when no options are found.

### DIFF
--- a/packages/components/select/__tests__/select.test.ts
+++ b/packages/components/select/__tests__/select.test.ts
@@ -4239,4 +4239,55 @@ describe('Select', () => {
     await input.trigger('blur')
     expect(handleVisibleChange).toHaveBeenCalledTimes(2)
   })
+
+  test('should show empty slot when remote search returns empty', async () => {
+    const wrapper = mount({
+      components: {
+        'el-select': Select,
+        'el-option': Option,
+      },
+      template: `
+        <el-select
+          v-model="value"
+          filterable
+          remote
+          :remote-method="remoteMethod"
+        >
+          <el-option
+            v-for="item in options"
+            :key="item.value"
+            :label="item.label"
+            :value="item.value"
+          />
+          <template #empty>
+            <div class="custom-empty">NO DATA</div>
+          </template>
+        </el-select>
+      `,
+      setup() {
+        const value = ref('')
+        const options = ref<any[]>([])
+        const remoteMethod = vi.fn((query) => {
+          if (query.includes('empty')) {
+            options.value = []
+          } else {
+            options.value = [{ value: '1', label: 'Option 1' }]
+          }
+        })
+        return { value, options, remoteMethod }
+      },
+    })
+    const select = wrapper.findComponent(Select)
+    const input = wrapper.find('input')
+    await input.trigger('click')
+    await input.setValue('empty')
+    await nextTick()
+    await nextTick()
+    expect((select.vm as any).expanded).toBe(true)
+    const emptyText = document.querySelector('.custom-empty')
+    expect(emptyText).not.toBeNull()
+    expect(emptyText?.textContent).toBe('NO DATA')
+    const popper = document.querySelector('.el-select__popper') as HTMLElement
+    expect(popper.style.display).not.toBe('none')
+  })
 })
--- a/packages/components/select/__tests__/select.test.ts
+++ b/packages/components/select/__tests__/select.test.ts
@@ -4240,9 +4240,9 @@ describe('Select', () => {
     expect(handleVisibleChange).toHaveBeenCalledTimes(2)
   })
 
-  test('should show empty slot when remote search returns empty', async () => {
+  test('should show empty slot correctly in remote search scenarios', async () => {
     vi.useFakeTimers()
-    wrapper = mount({
+    const wrapper = mount({
       components: {
         'el-select': Select,
         'el-option': Option,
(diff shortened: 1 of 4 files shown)


btea (approved):

YXY-cell:
如何申请合并这个代码

btea:
Once approved by other reviewers, the data can be merged.

packages/components/select/__tests__/select.test.ts
```diff
@@ -4239,4 +4239,55 @@ describe('Select', () => {
...
+        return { value, options, remoteMethod }
+      },
+    })
+    const select = wrapper.findComponent(Select)
+    const input = wrapper.find('input')
+    await input.trigger('click')
+    await input.setValue('empty')
+    await nextTick()
+    await nextTick()
+    expect((select.vm as any).expanded).toBe(true)
```
rzzf: Even without applying this change, this test case can still pass, so the test itself needs to be adjusted. Additionally, dropdownMenuVisible should be asserted here.

rzzf (changes requested):

packages/components/select/__tests__/select.test.ts
YXY-cell: OK，I found the question.This is due to the test is too fast to react.I update this PR

YXY-cell: As a newcomer in the open-source field, I would like to ask you, is the 'resolve conversation' mainly clicked by the project maintainer or by the person who submitted the PR?This is the first time I commit PR

btea: Either is fine. If you think the problem is solved, you can click on it.

packages/components/select/__tests__/select.test.ts
```diff
@@ -4239,4 +4239,65 @@ describe('Select', () => {
...
+    await nextTick()
+    expect(vm.states.options.size).toBe(0)
+    expect(vm.states.query).toBe('empty')
+    expect(vm.dropdownMenuVisible).toBe(true)
+    expect(vm.expanded).toBe(true)
+    const emptyText = document.querySelector('.custom-empty')
+    expect(emptyText).not.toBeNull()
+    expect(emptyText?.textContent).toBe('NO DATA')
+    const popper = document.querySelector('.el-select__popper') as HTMLElement
+    expect(popper.style.display).not.toBe('none')
```
rzzf: This test file is quite large, so I’d like to simplify it as much as possible. Some of these assertions seem redundant; maybe it’s sufficient to assert only `options.size`, `dropdownMenuVisible`, and `custom-empty`.

rzzf: Additionally, this test doesn’t seem to cover the original issue: the case where `input.value` is empty.

Update:

Maybe we should test both cases as well.

<details>
<summary>demo</summary>

```ts
const wrapper = _mount(
    `
    <el-select
      v-model="value"
      filterable
      remote
      :remote-method="remoteMethod"
    >
      <el-option
        v-for="item in options"
        :key="item.value"
        :label="item.label"
        :value="item.value"
      />
      <template #empty>
        <div class="custom-empty">NO DATA</div>
      </template>
    </el-select>
  `,
    () => ({
      value: '',
      options: [{ value: '1', label: 'Option 1' }],
    }),
    {
      methods: {
        remoteMethod(query: string) {
          if (!query || query === 'empty') {
            this.options = []
          } else {
            this.options = [{ value: '1', label: 'Option 1' }]
          }
        },
      },
    }
  )
  const select = wrapper.findComponent(Select)
  const input = wrapper.find('input')
  const vm = select.vm as any

  await input.trigger('click')
  expect(vm.states.options.size).toBe(0)
  expect(vm.dropdownMenuVisible).toBe(true)
  expect(document.querySelector('.custom-empty')).not.toBeNull()

  await input.setValue('a')
  expect(vm.states.options.size).toBe(1)
  expect(vm.dropdownMenuVisible).toBe(true)
  expect(document.querySelector('.custom-empty')).toBeNull()

  await input.setValue('empty')
  expect(vm.states.options.size).toBe(0)
  expect(vm.dropdownMenuVisible).toBe(true)
  expect(document.querySelector('.custom-empty')).not.toBeNull()
})
```
</details>

YXY-cell: I have a bit of trouble understanding why it is said that there is no cover for the initial situation?I solve this issue by letting the developer to set empty slot,if the developer set it,the menu would be visible no mater the select have options

YXY-cell: So I simulated setting up an empty slot here, cleared the options, and then conducted a search.          if (query.includes('empty')) {
            options.value = []
          }like here。then， expect(vm.dropdownMenuVisible).toBe(true) to see that if this could be seen

YXY-cell: Oh，I understand，You are right，I‘ll recover this soon

rzzf: Yes, but the original issue is that when `input.value` is empty, the dropdown will not be shown because `isRemoteSearchEmpty` remains true.

On the other hand, if `input.value` has a value (for example, "empty"), `isRemoteSearchEmpty` will always be false, so this logic works as expected and the dropdown is displayed.

https://github.com/element-plus/element-plus/blob/c40a2c10264abcd7b674e10d15216295c7609930/packages/components/select/src/useSelect.ts#L173-L175

YXY-cell: OK，I recovered these

rzzf (approved):
looks good to me, thank you!

YXY-cell:
> looks good to me, thank you!

happy new year！Thank for your guide

rzzf:
> happy new year！Thank for your guide

happy new year ❤️

YXY-cell:
> Once approved by other reviewers, the data can be merged.

I’m sorry to bother you, and I hope I’m not interrupting your vacation. Since I’m a beginner, I’d like to ask you a question: generally speaking, how many reviewers need to approve for the code to be merged?

Dsaquel (commented):
It looks like select-v2 has the same problem, can you fix it too ?

YXY-cell:
> It looks like select-v2 has the same problem, can you fix it too ?

I ‘ll try it later,thank for your appreciation！

YXY-cell:
> It looks like select-v2 has the same problem, can you fix it too ?

Should I create a new issue and submit a new PR, or make the changes directly on this existing PR?

Dsaquel:
>Should I create a new issue and submit a new PR, or make the changes directly on this existing PR?

You can make changes in this PR, we often update select and select-v2 together for related fixes.

YXY-cell:
> > Should I create a new issue and submit a new PR, or make the changes directly on this existing PR?
> 
> You can make changes in this PR, we often update select and select-v2 together for related fixes.

Hello，I recovered select-v2，can you check it please?thanks

Dsaquel (approved):
Thank you and happy new year :smiley: .

YXY-cell:
> Thank you and happy new year 😃 .谢谢你，新年快乐 😃。

Thanks！Happy new year❤️

PR #22808, opened 2025-11-14
fix(utils): [addUnit] handle 0 value correctly

rzzf:
fix #22807

PR #23211, opened 2025-12-25
feat(components): [avatar-group] new component

keeplearning66:

Relate #23128, add `avatar-group` component. It has following props.

1. `size`: control the size of avatars in this avatar-group
2. `shape`: control the shape of avatars in this avatar-group
3. `collapse-avatars`: whether to collapse avatars
4. `collapse-avatars-tooltip`: whether show all collapsed avatars when mouse hover text of the collapse-avatar. To use this, `collapse-avatars` must be true
5. `max-collapse-avatars`: the max avatars number to be shown. To use this, `collapse-avatars` must be true
6. `effect`: tooltip theme
7. `placement`: placement of tooltip
8. `popper-class`: custom class name for tooltip
9. `popper-style`: custom style for tooltip
10. `collapse-class`: custom class name for the collapse-avatar
11. `collapse-style`: custom style for the collapse-avatar

[demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZGl2IGNsYXNzPVwibS00XCI+XG4gICAgPHA+ZGVmYXVsdDwvcD5cbiAgICA8ZWwtYXZhdGFyLWdyb3VwPlxuICAgICAgPGVsLWF2YXRhciB2LWZvcj1cIm51bWJlciBpbiA1XCIgOmtleT1cIm51bWJlclwiIDpzcmM9XCJjaXJjbGVVcmxcIiAvPlxuICAgIDwvZWwtYXZhdGFyLWdyb3VwPlxuICA8L2Rpdj5cbiAgPGRpdiBjbGFzcz1cIm0tNFwiPlxuICAgIDxwPnVzZSBjb2xsYXBzZS1hdmF0YXJzPC9wPlxuICAgIDxlbC1hdmF0YXItZ3JvdXAgY29sbGFwc2UtYXZhdGFycz5cbiAgICAgIDxlbC1hdmF0YXIgdi1mb3I9XCJudW1iZXIgaW4gNVwiIDprZXk9XCJudW1iZXJcIiA6c3JjPVwiY2lyY2xlVXJsXCIgLz5cbiAgICA8L2VsLWF2YXRhci1ncm91cD5cbiAgPC9kaXY+XG4gIDxkaXYgY2xhc3M9XCJtLTRcIj5cbiAgICA8cD51c2UgY29sbGFwc2UtY2xhc3MgYW5kIGNvbGxhcHNlLXN0eWxlPC9wPlxuICAgIDxlbC1hdmF0YXItZ3JvdXBcbiAgICAgIGNvbGxhcHNlLWF2YXRhcnNcbiAgICAgIDpjb2xsYXBzZS1zdHlsZT1cInsgJ2JhY2tncm91bmQtY29sb3InOiAnI2Q5ZWNmZicgfVwiXG4gICAgICBjb2xsYXBzZS1jbGFzcz1cIm15LWNvbGxhcHNlLWF2YXRhclwiXG4gICAgPlxuICAgICAgPGVsLWF2YXRhciB2LWZvcj1cIm51bWJlciBpbiA1XCIgOmtleT1cIm51bWJlclwiIDpzcmM9XCJjaXJjbGVVcmxcIiAvPlxuICAgIDwvZWwtYXZhdGFyLWdyb3VwPlxuICA8L2Rpdj5cbiAgPGRpdiBjbGFzcz1cIm0tNFwiPlxuICAgIDxwPnVzZSBtYXgtY29sbGFwc2UtYXZhdGFyczwvcD5cbiAgICA8ZWwtYXZhdGFyLWdyb3VwIGNvbGxhcHNlLWF2YXRhcnMgOm1heC1jb2xsYXBzZS1hdmF0YXJzPVwiM1wiPlxuICAgICAgPGVsLWF2YXRhciB2LWZvcj1cIm51bWJlciBpbiA1XCIgOmtleT1cIm51bWJlclwiIDpzcmM9XCJjaXJjbGVVcmxcIiAvPlxuICAgIDwvZWwtYXZhdGFyLWdyb3VwPlxuICA8L2Rpdj5cbiAgPGRpdiBjbGFzcz1cIm0tNFwiPlxuICAgIDxwPnVzZSBjb2xsYXBzZS1hdmF0YXJzLXRvb2x0aXA8L3A+XG4gICAgPGVsLWF2YXRhci1ncm91cFxuICAgICAgY29sbGFwc2UtYXZhdGFyc1xuICAgICAgOm1heC1jb2xsYXBzZS1hdmF0YXJzPVwiM1wiXG4gICAgICBjb2xsYXBzZS1hdmF0YXJzLXRvb2x0aXBcbiAgICA+XG4gICAgICA8ZWwtYXZhdGFyIHYtZm9yPVwibnVtYmVyIGluIDVcIiA6a2V5PVwibnVtYmVyXCIgOnNyYz1cImNpcmNsZVVybFwiIC8+XG4gICAgPC9lbC1hdmF0YXItZ3JvdXA+XG4gIDwvZGl2PlxuPC90ZW1wbGF0ZT5cblxuPHNjcmlwdCBsYW5nPVwidHNcIiBzZXR1cD5cbmNvbnN0IGNpcmNsZVVybCA9XG4gICdodHRwczovL2N1YmUuZWxlbWVjZG4uY29tLzMvN2MvM2VhNmJlZWM2NDM2OWMyNjQyYjkyYzY3MjZmMWVwbmcucG5nJ1xuPC9zY3JpcHQ+XG5cbjxzdHlsZT5cbi5teS1jb2xsYXBzZS1hdmF0YXIge1xuICBjb2xvcjogIzQwOWVmZjtcbn1cbjwvc3R5bGU+XG4iLCJlbGVtZW50LXBsdXMuanMiOiJpbXBvcnQgRWxlbWVudFBsdXMgZnJvbSAnZWxlbWVudC1wbHVzJ1xuaW1wb3J0IHsgZ2V0Q3VycmVudEluc3RhbmNlIH0gZnJvbSAndnVlJ1xuXG5sZXQgaW5zdGFsbGVkID0gZmFsc2VcbmF3YWl0IGxvYWRTdHlsZSgpXG5cbmV4cG9ydCBmdW5jdGlvbiBzZXR1cEVsZW1lbnRQbHVzKCkge1xuICBpZiAoaW5zdGFsbGVkKSByZXR1cm5cbiAgY29uc3QgaW5zdGFuY2UgPSBnZXRDdXJyZW50SW5zdGFuY2UoKVxuICBpbnN0YW5jZS5hcHBDb250ZXh0LmFwcC51c2UoRWxlbWVudFBsdXMpXG4gIGluc3RhbGxlZCA9IHRydWVcbn1cblxuZXhwb3J0IGZ1bmN0aW9uIGxvYWRTdHlsZSgpIHtcbiAgY29uc3Qgc3R5bGVzID0gWydodHRwczovL3ByZXZpZXctMjMyMTEtZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS9kaXN0L2luZGV4LmNzcycsICdodHRwczovL3ByZXZpZXctMjMyMTEtZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS90aGVtZS1jaGFsay9kYXJrL2Nzcy12YXJzLmNzcyddLm1hcCgoc3R5bGUpID0+IHtcbiAgICByZXR1cm4gbmV3IFByb21pc2UoKHJlc29sdmUsIHJlamVjdCkgPT4ge1xuICAgICAgY29uc3QgbGluayA9IGRvY3VtZW50LmNyZWF0ZUVsZW1lbnQoJ2xpbmsnKVxuICAgICAgbGluay5yZWwgPSAnc3R5bGVzaGVldCdcbiAgICAgIGxpbmsuaHJlZiA9IHN0eWxlXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2xvYWQnLCByZXNvbHZlKVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdlcnJvcicsIHJlamVjdClcbiAgICAgIGRvY3VtZW50LmJvZHkuYXBwZW5kKGxpbmspXG4gICAgfSlcbiAgfSlcbiAgcmV0dXJuIFByb21pc2UuYWxsU2V0dGxlZChzdHlsZXMpXG59XG4iLCJ0c2NvbmZpZy5qc29uIjoie1xuICBcImNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogXCJFU05leHRcIixcbiAgICBcImpzeFwiOiBcInByZXNlcnZlXCIsXG4gICAgXCJtb2R1bGVcIjogXCJFU05leHRcIixcbiAgICBcIm1vZHVsZVJlc29sdXRpb25cIjogXCJCdW5kbGVyXCIsXG4gICAgXCJ0eXBlc1wiOiBbXCJlbGVtZW50LXBsdXMvZ2xvYmFsLmQudHNcIl0sXG4gICAgXCJhbGxvd0ltcG9ydGluZ1RzRXh0ZW5zaW9uc1wiOiB0cnVlLFxuICAgIFwiYWxsb3dKc1wiOiB0cnVlLFxuICAgIFwiY2hlY2tKc1wiOiB0cnVlXG4gIH0sXG4gIFwidnVlQ29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiAzLjNcbiAgfVxufVxuIiwiUGxheWdyb3VuZE1haW4udnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCBBcHAgZnJvbSAnLi9BcHAudnVlJ1xuaW1wb3J0IHsgc2V0dXBFbGVtZW50UGx1cyB9IGZyb20gJy4vZWxlbWVudC1wbHVzLmpzJ1xuc2V0dXBFbGVtZW50UGx1cygpXG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8QXBwIC8+XG48L3RlbXBsYXRlPlxuIiwiaW1wb3J0LW1hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge1xuICAgIFwidnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AdnVlL3J1bnRpbWUtZG9tQGxhdGVzdC9kaXN0L3J1bnRpbWUtZG9tLmVzbS1icm93c2VyLmpzXCIsXG4gICAgXCJAdnVlL3NoYXJlZFwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9zaGFyZWRAbGF0ZXN0L2Rpc3Qvc2hhcmVkLmVzbS1idW5kbGVyLmpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXNcIjogXCJodHRwczovL3ByZXZpZXctMjMyMTEtZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS9kaXN0L2luZGV4LmZ1bGwubWluLm1qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzL1wiOiBcInVuc3VwcG9ydGVkXCIsXG4gICAgXCJAZWxlbWVudC1wbHVzL2ljb25zLXZ1ZVwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQGVsZW1lbnQtcGx1cy9pY29ucy12dWVAMi9kaXN0L2luZGV4Lm1pbi5qc1wiXG4gIH0sXG4gIFwic2NvcGVzXCI6IHt9XG59IiwiX28iOnsic2hvd0hpZGRlbiI6dHJ1ZSwic3R5bGVTb3VyY2UiOiJodHRwczovL3ByZXZpZXctMjMyMTEtZWxlbWVudC1wbHVzLnN1cmdlLnNoL2J1bmRsZS9kaXN0L2luZGV4LmNzcyJ9fQ==)

## Summary by CodeRabbit

* **New Features**
  * Avatar Group: display multiple avatars with optional collapsing, configurable max visible, tooltip for collapsed items, and group-level size/shape inheritance. Individual avatar defaults for size/shape removed (group values apply when present).

* **Documentation**
  * Added Avatar Group docs and example demonstrating collapse modes, tooltip content, and custom styling.

* **Style**
  * Theme updates for avatar-group spacing, overlap, and collapse spacing.

* **Tests**
  * Added tests covering rendering, collapse behavior, and tooltip.

* **Chores**
  * Registered Avatar Group in public components and global typings.


keeplearning66:
### Playground

5DD47EDE-DF38-4943-AA94-578C6B004FAB

### Local

16AC3107-D4DA-4b43-9AA6-C4B34FF75C7A

It looks like there is an issue with the styles on the Playground.

packages/theme-chalk/src/avatar-group.scss
```diff
@@ -0,0 +1,20 @@
...
+
+@include b(avatar-group) {
+  display: flex;
+
+  .#{$namespace}-avatar {
+    border: 1px solid getCssVar('border-color-extra-light');
+  }
+
+  .#{$namespace}-avatar:not(:first-child) {
+    margin-left: -8px;
```
rzzf: Would it be better to use a variable here? Personally, I feel the distance between the avatars is a bit too large.

packages/components/avatar/src/avatar.vue
```diff
@@ -31,24 +32,30 @@ defineOptions({
 const props = defineProps(avatarProps)
 const emit = defineEmits(avatarEmits)
 
+const avatarGroupContext = inject(avatarGroupContextKey, undefined)
+
 const ns = useNamespace('avatar')
 
 const hasLoadError = ref(false)
 
+const size = computed(() => props.size || avatarGroupContext?.size || '')
```
rzzf: It would be better to use `??` here — we should respect `0`, right?

ref #22808

```suggestion
const size = computed(() => props.size ?? avatarGroupContext?.size ?? '')
```

packages/components/avatar/src/avatar-group.tsx
```diff
@@ -0,0 +1,140 @@
+import { defineComponent, provide, reactive, toRef } from 'vue'
```
rzzf: What is the reason for using TSX instead of Vue?

rzzf (commented):
Thank you for your contribution. 👍
I like this new component.

rzzf:
image

There is a bug here.

packages/components/avatar/src/avatar-group.tsx
keeplearning66: Because it involves modifying the content of `slots`, I personally feel that using **TSX** would be more convenient. If using **Vue**, it can also handle this situation easily. I can refactor it. 😃

patienry:
> image
> There is a bug here.

08E0580E-92D7-4e15-A317-34E2BB131412

I couldn't reproduce it. 👀 👀

rzzf:
> I couldn't reproduce it. 👀 👀

20251226111852_rec_

This seems to be related to hydration. You can reproduce it locally by running `pnpm run docs:build` and then `pnpm run docs:serve`.

packages/components/avatar/src/avatar-group.tsx
patienry: This involves handling the content of `slots`. I think using **TSX** would be more convenient. If **Vue** can also handle it easily, I can spend some time refactoring it. 😃

rzzf: Thanks for the explanation — there’s no need to rush into refactoring. Sorry I haven’t had a chance to look closely at the slot handling yet; we can also wait and see what others think.

rzzf:
@coderabbitai review

patienry:
> This seems to be related to hydration.

I found that some other components also have this issue, but it doesn't affect the functionality. For example, with `select` component: if you enter `select` and then refresh the page, you'll see the same error message. However, the functionality of `avatar-group` component is impacted, and I haven't found a suitable solution yet. 😢 

70EDD566-4168-46de-B8BC-132F30CD1107

packages/theme-chalk/src/avatar-group.scss
patienry: > Would it be better to use a variable here?

It's done. Does this look okay now? 👀 👀 

>Personally, I feel the distance between the avatars is a bit too large.

Actually, I'm not sure what distance would be more in line with **element-plus** specifications. 😂

packages/components/avatar/src/avatar.vue
patienry: Done, thank you!

packages/theme-chalk/src/avatar-group.scss
rzzf: Don’t worry, we can wait for feedback from the designer, other team members, or the community.

btea:
I can reproduce the problem mentioned by @rzzf r. The tooltip's content renders correctly the first time the mouse hovers over it, but the content doesn't display the second time the mouse hovers over it.

packages/components/avatar/src/avatar-group.tsx
```diff
@@ -0,0 +1,67 @@
...
+                  size={props.size}
+                  shape={props.shape}
+                  class={props.collapseClass}
+                  style={props.collapseStyle}
+                >
+                  + {hiddenAvatars.length}
+                </ElAvatar>
+              ),
+              content: () => (
+                <div class={ns.e('collapse-avatars')}>{hiddenAvatars}</div>
```
btea: It is due to the tooltip; we can modify the hiddenAvatars rendering logic.

```tsx
{hiddenAvatars.map((node, idx) =>
  isVNode(node) ? cloneVNode(node, { key: node.key ?? idx }) : node
)}
```

patienry: Thank you very much for your guidance. Everything is fine after the modification. ❤️ 

I've noticed that other components using `tooltip` are not affected. Could this be related to my modification of the `slot` content using **TSX**? Also, I can only reproduce the issue using `pnpm docs:build` and `pnpm docs:serve`. The problem doesn't occur with `pnpm docs:dev` or in the **Playground**. 👀 🤔

btea: I was able to reproduce this issue in the documentation I accessed while running `pnpm docs:dev`. I'm not entirely sure what the root cause is, but it seems to be a problem with the dynamic rendering logic of TSX. 🤔

patienry: Okay, at least there are no problems now. 😃 Thank you!

ringgoyao:
image
是否能添加处理font-size的逻辑，如果组件能提供这个处理就好了

patienry:
> 是否能添加处理font-size的逻辑，如果组件能提供这个处理就好了

I think this feature would be better completed in a new PR to avoid having too much content in one PR, which would increase the review difficulty. I will try to complete it when I have time (maybe other contributors will complete it). 😃

packages/components/avatar/src/avatar-group-props.ts
```diff
@@ -0,0 +1,70 @@
...
+   */
+  collapseAvatars: Boolean,
+  /**
+   * @description whether show all collapsed avatars when mouse hover text of the collapse-avatar. To use this, `collapse-avatars` must be true
+   */
+  collapseAvatarsTooltip: Boolean,
+  /**
+   * @description the max avatars number to be shown. To use this, `collapse-avatars` must be true
+   */
+  maxCollapseAvatars: {
```
rzzf, line 29: Is this `Avatars` necessary? I think it might be a bit redundant.

rzzf (approved):
looks good to me, thank you!

packages/components/avatar/src/avatar-group-props.ts
patienry: 6F39FB91-1331-4e60-81F0-8B3E53CE0964

4E344D91-1A06-42b0-B206-9645A67D933C

This naming is to be consistent with the `props` of other components related to collapse tags, such as `Select`, `Cascader`,  and `InputTag`. 🤔 However, simplifying it a bit is also acceptable for me. 😂

btea (approved):

packages/components/avatar/src/avatar-group-props.ts
btea: I think it's reasonable to keep the configuration consistent with other components for now.

rzzf: Okay.

packages/theme-chalk/src/common/var.scss
```diff
@@ -1523,6 +1523,15 @@ $avatar-size: map.merge(
   $avatar-size
 );
 
+$avatar-group: () !default;
+$avatar-group: map.merge(
+  (
+    'item-gap': -8px,
+    'collapse-item-gap': 4px,
```
Dsaquel, line 1530: Would it be more convenient to configure the gap in props ?

Dsaquel (commented):
Just added one question, overall it looks very neat 👍

packages/theme-chalk/src/common/var.scss
patienry: Personally, I feel that it's quite convenient to configure through **CSS variables**. Currently, `avatar-group` already has 11 props. If we add two more props, compared to `button-group`, `checkbox-group` and `radio-group`, doesn't `avatar-group` have a bit too many props? 👀 However, if everyone thinks that adding two props is more appropriate, I wouldn't mind adding them. 😃

Dsaquel: In this component group my concern is:
- Switching gaps from avatar component from another is a little bit more convenient with props instead of switching scss var locally to the component.
- chanigng gaps dynamically is more convenient too.

In other component group we don't really handle something like that.
Another note: I got the idea from the offset props in https://github.com/element-plus/element-plus/pull/22555.

But it's already good enough we pass to props if developers want to.

Dsaquel (approved):

PR #23247, opened 2025-12-29
fix(theme-chalk): [Table] fixed-right border missing in high DPI

YXY-cell:
### What is this PR for?
Fixes #23229

This PR fixes an issue where the left border of the `fixed-right` column disappears or becomes blurry on high DPI screens (e.g., 4K monitors or 125%/150% browser zoom).

### Why this happens?
The original `border-left` implementation is susceptible to sub-pixel rendering issues. When the browser calculates layout at non-integer pixel ratios, the background of the fixed column often covers the 1px border.

### How to fix?
Using `box-shadow` instead of `border-left` is a robust solution (widely used in other UI libraries like Ant Design). It projects a shadow to the left, which ensures the visual border remains visible regardless of sub-pixel positioning errors.
image

## Summary by CodeRabbit

* **Style**
  * Improved table styling for cases where a right-fixed column is also the first column: the left border is removed and a subtle left shadow (matching the table border color) is added for clearer visual separation. This is purely visual with no behavioral or structural impact.


btea:
cc @element-plus/backers

rzzf:
This is indeed an issue. It’s easy to reproduce on macOS by reducing the browser zoom level and then resizing the window. However, this fix is not correct: the `border` still appears even when **no border** is specified.

YXY-cell:
> This is indeed an issue. It’s easy to reproduce on macOS by reducing the browser zoom level and then resizing the window. However, this fix is not correct: the `border` still appears even when **no border** is specified.这确实是一个问题。在 macOS 上，通过降低浏览器缩放级别然后调整窗口大小，很容易重现此问题。然而，这个修复是不正确的：即使未指定边框， `border` 仍然出现。

Fixed. Now the style only applies when the border prop is present.

PR #23265, opened 2026-01-01
feat(components): [upload] add `directory` prop

rzzf:
resolves #23093

## Summary by CodeRabbit

* **New Features**
  * Upload now supports directory uploads via a new boolean prop (default: false), enabling folder selection, drag-and-drop of directories, and flattening contained files.

* **Documentation**
  * Added documentation and an interactive example demonstrating directory upload usage and API details.

* **Tests**
  * Added a test validating directory drop/upload behavior and the emitted file list.

### DIFF
--- a/docs/en-US/component/upload.md
+++ b/docs/en-US/component/upload.md
@@ -81,6 +81,16 @@ upload/drag-and-drop
 
 :::
 
+## Upload Directory ^(2.13.1)
+
+Enable folder upload via the `directory` prop.
+
+:::demo
+
+upload/directory
+
+:::
+
 ## Manual Upload
 
 :::demo
@@ -121,6 +131,7 @@ upload/manual
 | http-request                  | override default xhr behavior, allowing you to implement your own upload-file's request.                                                                                              | ^[Function]`(options: UploadRequestOptions) => XMLHttpRequest \| Promise<unknown>`                                                         | ajaxUpload [see](https://github.com/element-plus/element-plus/blob/dev/packages/components/upload/src/ajax.ts#L55) |
 | disabled                      | whether to disable upload.                                                                                                                                                            | ^[boolean]                                                                                                                                 | false                                                                                                              |
 | limit                         | maximum number of uploads allowed.                                                                                                                                                    | ^[number]                                                                                                                                  | —                                                                                                                  |
+| directory ^(2.13.1)           | whether to support uploading directory.                                                                                                                                               | ^[boolean]                                                                                                                                 | false                                                                                                              |
 
 ### Slots
 
--- a/docs/en-US/component/upload.md
+++ b/docs/en-US/component/upload.md
@@ -85,7 +85,7 @@ upload/drag-and-drop
 
 Enable folder upload via the `directory` prop.
 
-:::demo
+:::demo After enabling it, only folders can be selected, and after selecting a folder, the files within the folder will be flattened.
 
 upload/directory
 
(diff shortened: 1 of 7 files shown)


packages/components/upload/src/upload-dragger.vue
```diff
@@ -40,14 +41,58 @@ const ns = useNamespace('upload')
...
+    while (readEntries.length > 0) {
+      entries.push(...readEntries)
+      readEntries = await getEntries()
+    }
+
+    const files = await Promise.all(entries.map(getAllFiles))
+    return flatten(files)
+  }
+  return []
+}
```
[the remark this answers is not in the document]
rzzf: I’m not sure how to handle errors in a more elegant way. Besides just logging the error, do you have any suggestions?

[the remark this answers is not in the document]
rzzf: I can’t use `Promise.allSettled` for now because the component library’s lib target is below ES2020.

Dsaquel (approved):
Looks neat 👍

docs/en-US/component/upload.md
```diff
@@ -121,6 +131,7 @@ upload/manual
...
 | limit                         | maximum number of uploads allowed.                                                                                                                                                    | ^[number]                                                                                                                                  | —                                                                                                                  |
+| directory ^(2.13.1)           | whether to support uploading directory.                                                                                                                                               | ^[boolean]                                                                                                                                 | false                                                                                                              |
```
btea: Perhaps we can describe it more precisely: after enabling it, only folders can be selected, and after selecting a folder, the files within the folder will be flattened.

rzzf: That makes sense.

btea (approved):

packages/components/upload/src/upload-dragger.vue
```diff
@@ -40,14 +41,68 @@ const ns = useNamespace('upload')
...
+        return new Promise((resolve, reject) =>
+          dirReader.readEntries(resolve, reject)
+        )
+      }
+
+      const entries: FileSystemEntry[] = []
+      let readEntries = await getEntries()
+      while (readEntries.length > 0) {
+        entries.push(...readEntries)
+        readEntries = await getEntries()
```
btea, line 71: Is this the reason for the repeated calls here?

https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryReader/readEntries#:~:text=On%20Chrome%2077%2C%20readEntries()%20will%20only%20return%20the%20first%20100%20FileSystemEntry%20instances.%20In%20order%20to%20obtain%20all%20of%20the%20instances%2C%20readEntries()%20must%20be%20called%20multiple%20times.

rzzf: yeah.

rzzf: I noticed that we no longer support Chrome versions below 85, so let’s remove this loop.

image

rzzf: I’m not sure whether this is an issue with the MDN documentation or a limitation of `webkitGetAsEntry` itself. In practice, when dragging a directory containing more than 100 files, calling the read logic only once will hit a 100-entry limit. 
You can easily reproduce and test this by creating 101 empty files using the following shell command: `for i in {1..101}; do touch "$i"; done`

https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem/webkitGetAsEntry#javascript:~:text=Note%3A%20To%20read%20all%20files%20in%20a%20directory%2C%20readEntries%20needs%20to%20be%20called%20repeatedly%20until%20it%20returns%20an%20empty%20array.%20In%20Chromium%2Dbased%20browsers%2C%20the%20following%20example%20will%20only%20return%20a%20max%20of%20100%20entries.

btea: I tried it, and even in the latest version of Chrome, it still only returns 100 files at a time.

rzzf: Yeah, so I think this loop needs to be kept. What do you think?

btea: Agree 👍

btea: > Is this the reason for the repeated calls here?
> 
> https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryReader/readEntries#:~:text=On%20Chrome%2077%2C%20readEntries()%20will%20only%20return%20the%20first%20100%20FileSystemEntry%20instances.%20In%20order%20to%20obtain%20all%20of%20the%20instances%2C%20readEntries()%20must%20be%20called%20multiple%20times.

I believe the version mentioned here is a documentation error, and I have created a pull request to correct it. I am awaiting official review and update from MDN.

btea (approved):

packages/components/upload/src/upload-dragger.vue
rzzf: > I believe the version mentioned here is a documentation error, and I have created a pull request to correct it. I am awaiting official review and update from MDN.

cool! 👍

PR #23295, opened 2026-01-04
ci: remove unnecessary comment

rzzf:
I’ve subscribed to the issues in this repository and receive notification emails almost every day. 
This can be somewhat distracting for both the issue author and other subscribed developers. When the same problem comes up again, people will naturally open a new issue or leave a comment on the fix PR. Therefore, I think the downsides of this comment outweigh its benefits.

## Summary by CodeRabbit

* **Chores**
  * Removed the automatic notification message from the issue-locking workflow, simplifying the issue management process.

### DIFF
--- a/.github/workflows/issue-inactive.yml
+++ b/.github/workflows/issue-inactive.yml
@@ -39,10 +39,6 @@ jobs:
           issue-state: closed
           labels: 'inactive'
           inactive-day: 30
-          body: |
-            This issue has been automatically locked since there has not been any recent activity after it was closed. Please open a new issue for related bugs.
-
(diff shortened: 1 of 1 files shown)


btea (approved):
This email should only be received by developers who participated in the relevant pull request. However, I think removing the comments is also acceptable.

rzzf:
> This email should only be received by developers who participated in the relevant pull request. However, I think removing the comments is also acceptable.

This might be a bit tricky. I’m not sure whether the workflow payload provides a list of users. If we actually need this in the future, we can revisit it then. Thank you! ❤️

PR #23172, opened 2025-12-20
fix(components): [input] correct event type for v-model.number

ixyzorg:
Please make sure these boxes are checked before submitting your PR, thank you!

- [x] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).
- [x] Make sure you are merging your commits to `dev` branch.
- [x] Add some descriptions and refer to relative issues for your PR.

**BugDesc**
- When we use v-model.number on the ElInput component, the value inside the input and change event handlers is of type string. We expect it to be of type number. The official documentation also mentions this point: in the function signatures of the input and change events, the value parameter should be of type number or string.

https://github.com/user-attachments/assets/ff8c7723-aee8-417d-a259-c71912471dcf

**Fix**
- fixed correct event type for v-model.number
- modified the function signature. Due to the change in the function signature, other components or tests need to be updated to be compatible with the TypeScript types.

https://github.com/user-attachments/assets/a0130a84-3ecf-4d14-8e27-dbebadcf2a28

PR #23366, opened 2026-01-12
refactor(components): [input] use type-based definitions

rzzf:
Try refactoring runtime declarations into type-based declarations.

rel #23172

## Summary by CodeRabbit

* **New Features**
  * Added placeholder attribute for the Input component.
  * Added word-limit-position attribute (values: "inside" or "outside", default "inside").
  * Model-modifier flags for lazy/number/trim are now presence-based (flags rather than booleans).
  * Default clear icon updated to a CircleClose icon.


packages/components/input/src/input.ts
```diff
@@ -10,16 +10,17 @@ import { useAriaProps, useSizeProp } from '@element-plus/hooks'
...
   HTMLAttributes,
   StyleValue,
 } from 'vue'
+import type { ComponentSize } from '@element-plus/constants'
 
 export type InputModelModifiers = {
-  lazy?: boolean
-  number?: boolean
-  trim?: boolean
+  lazy?: true
```
btea, line 21: Why was it changed to only support `true`?

rzzf: This change is due to the fact that the type of [defineModel](https://vuejs.org/api/sfc-script-setup.html#usage-with-typescript) is [`true | undefined`](https://github.com/vuejs/core/blob/623bfb29a23c36bf935e7faefa64d1baa69d465c/packages/runtime-core/src/apiSetupHelpers.ts#L243-L247). It might be better to align with the official implementation, even though there is no practical difference in the outcome.

packages/components/input/src/input.vue
```diff
@@ -213,8 +221,20 @@ defineOptions({
...
 })
-const props = defineProps(inputProps)
-const emit = defineEmits(inputEmits)
+const props = withDefaults(defineProps<InputProps<M>>(), {
+  disabled: undefined,
+  modelValue: '',
+  modelModifiers: () => ({}) as M,
+  type: 'text',
+  autocomplete: 'off',
+  clearIcon: IconCircleClose,
```
Dsaquel: imported components should not be converted as a proxy.
```suggestion
  clearIcon: markRaw(IconCircleClose),
```

packages/element-plus/package.json
```diff
@@ -93,7 +93,8 @@
     "lodash-es": "^4.17.21",
     "lodash-unified": "^1.0.3",
     "memoize-one": "^6.0.0",
-    "normalize-wheel-es": "^1.2.0"
+    "normalize-wheel-es": "^1.2.0",
+    "vue-component-type-helpers": "catalog:"
```
Dsaquel: Not sure if we need it in package.json of the bundled package, even lodash types are listed in `dependencies`, its a bit surprising 🤔

I feel like it should be at the most in devDependencies, [ref](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file).

packages/components/package.json
```diff
@@ -18,7 +18,8 @@
     "@element-plus/utils": "workspace:^",
     "@popperjs/core": "catalog:",
     "@vueuse/core": "catalog:",
-    "lodash-unified": "catalog:"
+    "lodash-unified": "catalog:",
+    "vue-component-type-helpers": "catalog:"
```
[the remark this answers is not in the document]
Dsaquel: @coderabbitai ignore

I think this comment is pertinent.

Dsaquel (commented):
The result looks good, just added some comments.

packages/element-plus/package.json
rzzf: If it is installed as a `devDep`, downstream projects installing element-plus will not install vue-component-type-helpers, which leads to an incorrect `InputInstance` type. In this case, should we move it to `peerDependencies`?

Dsaquel: >If it is installed as a devDep, downstream projects installing element-plus will not install vue-component-type-helpers, which leads to an incorrect InputInstance type

I have tested that locally using local path when installing EP, after doing `pnpm why` none of the dependencies of EP are listed but i tested with a real release and it's indeed installed.

Dsaquel:
According to CR review https://github.com/element-plus/element-plus/pull/23366#discussion_r2680943233, `package.json` and `packages/components/package.json` should have `vue-component-type-helpers` in devDep IMO.

packages/components/package.json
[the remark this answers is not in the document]
rzzf: done

rzzf:
> According to CR review [#23366 (comment)](https://github.com/element-plus/element-plus/pull/23366#discussion_r2680943233), `package.json` and `packages/components/package.json` should have `vue-component-type-helpers` in devDep IMO.

You’re right. During testing, it didn’t take effect due to pnpm caching, so I added this dependency for them. After retesting, everything works well now.

packages/components/input/src/input.ts
```diff
@@ -227,9 +231,15 @@ export const inputProps = buildProps({
    */
   name: String,
 } as const)
-export type InputProps = ExtractPropTypes<typeof inputProps>
+
+/**
+ * @deprecated Removed after 3.0.0
```
rzzf: Would it be more appropriate to remove this in version 3.0, or is it acceptable to remove it in a minor release? I’m not entirely sure.

packages/element-plus/package.json
Dsaquel: in this package.json we should set the number version. The developers will not access to our workspace.yaml file.

rzzf: I installed the version mentioned here https://github.com/element-plus/element-plus/pull/23366#issuecomment-3736971724, and it correctly parses numeric version numbers. If this is still considered an issue, does that mean VueUse has the same problem as well?

<details>
<summary>Screenshot</summary>
image
</details>

btea: This should be fine, the final version that is packaged and released will be the definitive version.

packages/components/input/src/input.ts
```diff
@@ -247,4 +257,159 @@ export const inputEmits = {
   compositionupdate: (evt: CompositionEvent) => evt instanceof CompositionEvent,
   compositionend: (evt: CompositionEvent) => evt instanceof CompositionEvent,
 }
-export type InputEmits = typeof inputEmits
+
+export interface InputProps<
+  M extends InputModelModifiers = InputModelModifiers,
```
btea: It seems this isn't a public type that will be used everywhere, so perhaps we don't need to declare generics here. 🤔

packages/element-plus/package.json
Dsaquel: I used again `pnpm link` instead of https://github.com/element-plus/element-plus/pull/23366#issuecomment-3736971724 :facepalm: .

packages/components/input/src/input.ts
Dsaquel: Both is fine for me.

packages/components/input/src/input.ts
rzzf: This is intended to address the case where, when using `v-model`, we need to infer whether the developer has applied the `number modifier`. Without doing so, it seems difficult to correctly handle the parameter types of the `input` and `change` events.

Please see: https://github.com/element-plus/element-plus/pull/23172#pullrequestreview-3601104395

Could we also choose not to address this issue?

rzzf: After discussion, the Input component does not need generics. For scenarios that require handling numbers, using `input-number` is recommended.

btea (approved):

Dsaquel (approved):

PR #23374, opened 2026-01-13
fix(components): [upload] `webkitdirectory="false"` being preserved during hydration

rzzf:

This can be reproduced by refreshing the documentation page: https://element-plus.org/en-US/component/upload

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed file upload component to properly handle directory selection states, ensuring correct behavior when directory mode is not active.

### DIFF
--- a/packages/components/upload/src/upload-content.vue
+++ b/packages/components/upload/src/upload-content.vue
@@ -31,7 +31,7 @@
       :disabled="disabled"
       :multiple="multiple"
       :accept="accept"
-      :webkitdirectory="directory"
+      :webkitdirectory="directory || undefined"
       type="file"
       @change="handleChange"
       @click.stop


Dsaquel:
I don't think its something to get fixed/done, other attributes can have their html attributes changed due to hydration too.

rzzf:
> I don't think its something to get fixed/done, other attributes can have their html attributes changed due to hydration too.

Although [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory#value) states that its value is a boolean, `webkitdirectory` seems to behave differently from other attributes: in **Chromium**, it takes effect as long as the attribute is present on the element, regardless of its value. I’m not very familiar with SSR, but wouldn’t this behavior be considered a bug?

<details>
<summary>Screenshot</summary>

image

</details>

Dsaquel:
Thanks for explanation, i'm not familiar with ssr too, i would say the issue is not related in our side but it has to be fixed :+1: .

Dsaquel (approved):

btea (approved):

PR #23376, opened 2026-01-13
fix(components): [notification] use reactive positioning to fix stacking

webfanzc:
close #23209

## Summary by CodeRabbit

* **New Features**
  * Notifications now expose their rendered bottom position for consumers.

* **Improvements**
  * Per-position reactive queues for more reliable stacking, spacing and dynamic offsets.
  * Offsets and placement update reactively when notification size or layout changes, including animation-aware positioning.

* **Tests**
  * Added jsdom-friendly tests and mocks to simulate resize events and verify stacking and offset updates.


packages/components/notification/src/notify.ts
```diff
@@ -19,18 +19,46 @@ import type {
   NotifyFn,
 } from './notification'
 
-// This should be a queue but considering there were `non-autoclosable` notifications.
```
rzzf, line 22: It looks like the meaning of the variables hasn’t changed. Please restore these comments.

packages/components/notification/src/notify.ts
```diff
@@ -19,18 +19,46 @@ import type {
...
+  id: string,
+  offset: number,
+  position: NotificationOptions['position']
+) => {
+  const list = notifications[position] || []
+  const idx = list.findIndex(({ vm }) => vm.component!.props.id === id)
+  return idx > 0 ? GAP_SIZE : offset
 }
 
-// the gap size between each notification
```
rzzf, line 33: And here as well.

packages/components/notification/src/notify.ts
```diff
@@ -132,38 +163,24 @@ export function close(
...
-  const len = orientedNotifications.length
-  if (len < 1) return
-  // starting from the removing item.
-  for (let i = idx; i < len; i++) {
-    // new position equals the current offsetTop minus removed height plus 16px(the gap size between each item)
-    const { el, component } = orientedNotifications[i].vm
-    const pos =
-      Number.parseInt(el!.style[verticalPos], 10) - removedHeight - GAP_SIZE
-    component!.props.offset = pos
-  }
```
rzzf, line 159: Why was the logic here removed?

packages/components/notification/src/notify.ts
```diff
@@ -132,38 +163,24 @@ export function close(
...
+  userOnClose?.(vm)
 }
 
 export function closeAll(): void {
   // loop through all directions, close them at once.
-  for (const orientedNotifications of Object.values(notifications)) {
-    orientedNotifications.forEach(({ vm }) => {
-      // same as the previous close method, we'd like to make sure lifecycle gets handle properly.
+  for (const positionInstances of Object.values(notifications)) {
+    positionInstances.forEach(({ vm }) => {
```
rzzf: Please revert this change, as it doesn’t introduce any difference.

rzzf (changes requested):
Thanks for the fix. I have a few suggestions.

packages/components/notification/src/notify.ts
webfanzc: Sorry, removing the `userOnClose` call was a mistake. I intended to remove the manual offset calculation logic below since offset is now computed reactively via `getLastOffset` and `getOffsetOrSpace`. The `userOnClose` call should remain and I've restored it. Thanks for catching that!

webfanzc:
> Thanks for the fix. I have a few suggestions.

@rzzf Thanks for the review! I've addressed all the issues you pointed out
Please review again, thanks!

packages/components/notification/src/notify.ts
rzzf: The original code does not seem to conflict with the new offset control approach. It is intended to ensure that notifications below can move to the correct position during the closing animation.

<details>

<summary>video</summary>

<h3>Before</h3>

https://github.com/user-attachments/assets/65d93f44-79df-4162-bc3f-7613fb3d9a77

<h3>Current</h3>

https://github.com/user-attachments/assets/d392ccbb-1a63-4951-a485-a2342e978fce


</details>

packages/components/notification/src/notification.vue
```diff
@@ -40,12 +41,13 @@
...
 <script lang="ts" setup>
 import { computed, onMounted, ref } from 'vue'
-import { useEventListener, useTimeoutFn } from '@vueuse/core'
+import { useEventListener, useResizeObserver, useTimeoutFn } from '@vueuse/core'
 import { TypeComponentsMap, getEventCode } from '@element-plus/utils'
 import { EVENT_CODE } from '@element-plus/constants'
 import { ElIcon } from '@element-plus/components/icon'
 import { useGlobalComponentSettings } from '@element-plus/components/config-provider'
 import { notificationEmits, notificationProps } from './notification'
+import { getLastOffset, getOffsetOrSpace } from './notify'
```
rzzf: It would be better to extract the parts that are used by both `notification.vue` and `notify.ts` into a separate file; otherwise, this can very easily lead to circular dependencies. 

Similar message https://github.com/element-plus/element-plus/blob/dev/packages/components/message/src/instance.ts

packages/components/notification/src/notification.vue
```diff
@@ -132,9 +144,13 @@ onMounted(() => {
...
 useEventListener(document, 'keydown', onKeydown)
 
+useResizeObserver(notificationRef, () => {
+  height.value = notificationRef.value!.getBoundingClientRect().height
+})
+
 defineExpose({
   visible,
-  /** @description close notification */
+  bottom,
```
rzzf, line 177: Could you please restore the comment here?

webfanzc: Got it, I'll fix this. Sorry for the oversight!

packages/components/notification/src/notify.ts
webfanzc: I've restored the logic here. As you mentioned, it's indeed intended to ensure notifications below can move to the correct position during the closing animation. Thanks for the clarification!

packages/components/notification/src/instance.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export const getOffsetOrSpace = (
+  id: string,
+  offset: number,
+  position: NotificationOptions['position']
+) => {
+  const list = notifications[position] || []
+  const idx = list.findIndex(({ vm }) => vm.component!.props.id === id)
+  return idx > 0 ? 16 : offset
```
rzzf:
```suggestion
  return idx > 0 ? GAP_SIZE : offset
```

I think `GAP_SIZE` should also be moved into this file.

packages/components/notification/__tests__/notify.test.tsx
```diff
@@ -97,6 +127,22 @@ describe('Notification on command', () => {
...
+        height:
+          this.style?.height !== undefined
+            ? Number.parseFloat(this.style.height) || height
+            : height,
+      }
+    }
+    cleanups.push(() => {
+      Element.prototype.getBoundingClientRect = origGetBoundingClientRect
+    })
+
```
rzzf: This test case is intended to test the `updateOffsets` method. We should add a new test case to cover the current issue.

rzzf (commented):
Sorry, I forgot to submit it.

packages/components/notification/__tests__/notify.test.tsx
rzzf: The new offset method seems to change the animation behavior when multiple notifications are opened at once, which also causes this test case to need adjustment. There might be an issue somewhere.

webfanzc: Sorry for any confusion. Regarding updateOffsets, after implementing reactive position calculation, I believe it no longer serves a practical purpose, so I've directly modified the relevant code in the original test cases.
Following your suggestion, I've created a separate test case for the stacking test. I'm mocking getBoundingClientRect because the component's height is now set through useResizeObserver, which requires manual simulation in the test environment.
Thank you for your feedback.

rzzf: I know, but what I want to point out is that the initial position of the notify seems to be incorrect when the animation starts.

[demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZGl2PjwvZGl2PlxuPC90ZW1wbGF0ZT5cblxuPHNjcmlwdCBsYW5nPVwidHNcIiBzZXR1cD5cbmltcG9ydCB7IGgsIG5leHRUaWNrLCBvbk1vdW50ZWQgfSBmcm9tICd2dWUnXG5pbXBvcnQgeyBFbE5vdGlmaWNhdGlvbiB9IGZyb20gJ2VsZW1lbnQtcGx1cydcblxub25Nb3VudGVkKGFzeW5jICgpID0+IHtcbiAgZm9yIChsZXQgaSA9IDA7IGkgPCAxMDsgaSsrKSB7XG4gICAgRWxOb3RpZmljYXRpb24oe1xuICAgICAgdGl0bGU6ICdUaXRsZScsXG4gICAgICBtZXNzYWdlOiBoKCdpJywgeyBzdHlsZTogJ2NvbG9yOiB0ZWFsJyB9LCAnVGhpcyBpcyBhIHJlbWluZGVyJyksXG4gICAgfSlcbiAgICBhd2FpdCBuZXh0VGljaygpXG4gIH1cbn0pXG48L3NjcmlwdD5cbiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQGxhdGVzdC9kaXN0L2luZGV4LmNzcycsICdodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvdGhlbWUtY2hhbGsvZGFyay9jc3MtdmFycy5jc3MnXS5tYXAoKHN0eWxlKSA9PiB7XG4gICAgcmV0dXJuIG5ldyBQcm9taXNlKChyZXNvbHZlLCByZWplY3QpID0+IHtcbiAgICAgIGNvbnN0IGxpbmsgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCdsaW5rJylcbiAgICAgIGxpbmsucmVsID0gJ3N0eWxlc2hlZXQnXG4gICAgICBsaW5rLmhyZWYgPSBzdHlsZVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdsb2FkJywgcmVzb2x2ZSlcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignZXJyb3InLCByZWplY3QpXG4gICAgICBkb2N1bWVudC5ib2R5LmFwcGVuZChsaW5rKVxuICAgIH0pXG4gIH0pXG4gIHJldHVybiBQcm9taXNlLmFsbFNldHRsZWQoc3R5bGVzKVxufVxuIiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IFwiRVNOZXh0XCIsXG4gICAgXCJqc3hcIjogXCJwcmVzZXJ2ZVwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwidHlwZXNcIjogW1wiZWxlbWVudC1wbHVzL2dsb2JhbC5kLnRzXCJdLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcImFsbG93SnNcIjogdHJ1ZSxcbiAgICBcImNoZWNrSnNcIjogdHJ1ZVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9ydW50aW1lLWRvbUBsYXRlc3QvZGlzdC9ydW50aW1lLWRvbS5lc20tYnJvd3Nlci5qc1wiLFxuICAgIFwiQHZ1ZS9zaGFyZWRcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvc2hhcmVkQGxhdGVzdC9kaXN0L3NoYXJlZC5lc20tYnVuZGxlci5qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L1wiLFxuICAgIFwiQGVsZW1lbnQtcGx1cy9pY29ucy12dWVcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlQDIvZGlzdC9pbmRleC5taW4uanNcIlxuICB9LFxuICBcInNjb3Blc1wiOiB7fVxufSIsIl9vIjp7fX0=)

<details>
<summary>Video</summary>

<h3>Before</h3>

https://github.com/user-attachments/assets/27345538-9bc0-4f39-986f-7f7b8201d9d8

<h3>Current</h3>

https://github.com/user-attachments/assets/efb22bee-cd60-4e3e-8e56-17741b0995ff


</details>

webfanzc: Okay, let me check what's causing this problem.

rzzf: Thank you for the back-and-forth discussion.

webfanzc: I've looked into this issue carefully. Here's the situation:

The original implementation calculates the offset when creating the notification and passes it in. This causes multiple notifications to stack together when called synchronously, because we can't get the `offsetHeight` of previous elements during synchronous calls (Vue renders asynchronously).

If I restore the logic here to calculate offsetHeight at creation time and pass it in:
 Async calls: Can restore the same animation behavior as the original, since the original offset was calculated at creation time
 Sync calls: Will show my current animation behavior

So I'm not sure whether I should fix it this way, or keep the current code. Also, the message component seems to have the same issue.
Should I submit this fix? Or do you have a better suggestion?

webfanzc: I've pushed my changes. As I mentioned before, there might be some slight inconsistency in the animation behavior between sync and async calls now. Let me know if you have any concerns about this.

rzzf: Thank you for the detailed explanation. 🙇 
Personally, I think that if enabling synchronous opening comes at the cost of changing the visual behavior, that is probably not what I expect. 
In addition, I don’t think using `nextTick()` in scenarios with multiple calls would be a significant burden for developers.

cc @element-plus/backers What do you think?

webfanzc: I agree that calling `nextTick()` isn't a significant burden. However, if developers don't carefully review the source code implementation, synchronous calls might cause some confusion, even though such scenarios are relatively uncommon.

If this change isn't adopted, perhaps we could improve the documentation to remind developers that this method shouldn't be called synchronously? This would help prevent potential confusion while maintaining the current behavior.

rzzf: > If this change isn't adopted, perhaps we could improve the documentation to remind developers that this method shouldn't be called synchronously? This would help prevent potential confusion while maintaining the current behavior.

I think a clear document description is a very good solution. 👍

webfanzc: Been thinking about this a bit more.

What if we just keep it simple - drop the current changes and update the docs to suggest async calls, then call a debounced `updateOffsets` after creating notification? 

That way, even if devs don't read the docs carefully and call it sync, no weird stacking happens.

webfanzc: Hey! I've opened a new PR with a different approach. I think this one looks better - just using a debounced `updateOffsets` call and updating the docs. Way simpler than all that reactive positioning stuff.

Even if someone calls notifications sync, the debounce will fix the positions automatically. Plus we're telling folks in the docs to use async calls for the best experience.

I'll close this PR since the new one has the better solution. Let me know what you think!
The new PR #23500

Issue #23399, opened 2026-01-17
Refactor Component Props to type-based declarations


rzzf:
## Plans

Components are gradually being migrated from runtime-based props declarations to [type-based](https://vuejs.org/api/sfc-script-setup.html#type-only-props-emit-declarations) props declarations.

- remove `buildProps` / `buildProp`
- [Further improve the DX](https://github.com/element-plus/element-plus/discussions/20239)

**If anyone wants to help on this one you can simply ping me and tell me which component you intend to work on.**

## Progress

- [x] affix
  - https://github.com/element-plus/element-plus/pull/23400
- [x] alert
  - https://github.com/element-plus/element-plus/pull/23401
- [x] anchor / anchor-link
  - https://github.com/element-plus/element-plus/pull/23403
- [x] autocomplete
  - https://github.com/element-plus/element-plus/pull/23410
- [x] avatar
  - https://github.com/element-plus/element-plus/pull/23457
- [x] backtop
  - https://github.com/element-plus/element-plus/pull/23435
- [x] badge
  - https://github.com/element-plus/element-plus/pull/23414
- [x] breadcrumb / breadcrumb-item
  - https://github.com/element-plus/element-plus/pull/23414
- [x] button / button-group
  - https://github.com/element-plus/element-plus/pull/23414
- [x] calendar
  - https://github.com/element-plus/element-plus/pull/23419
- [x] card
  - https://github.com/element-plus/element-plus/pull/23416
- [x] carousel / carousel-item
  - https://github.com/element-plus/element-plus/pull/23430
- [x] cascader / cascader-panel
  - https://github.com/element-plus/element-plus/pull/23478
- [x] check-tag
  - https://github.com/element-plus/element-plus/pull/23424
- [ ] checkbox / checkbox-button / checkbox-group
  - https://github.com/element-plus/element-plus/pull/23480
- [x] col
  - https://github.com/element-plus/element-plus/pull/23438
- [x] collapse / collapse-item
  - https://github.com/element-plus/element-plus/pull/23443
- [x] color-picker
  - https://github.com/element-plus/element-plus/pull/23488
- [x] color-picker-panel
  - https://github.com/element-plus/element-plus/pull/23456
- [x] container / aside / footer/header
  - https://github.com/element-plus/element-plus/pull/23429
- [x] countdown
  - https://github.com/element-plus/element-plus/pull/23447
- [x] descriptions
    - https://github.com/element-plus/element-plus/pull/23434
- [x] dialog
[...]

PR #23414, opened 2026-01-17
refactor(components): [badge/breadcrumb/button] use type-based definitions

snowbitx:

ref:#23399.  badge & breadcrumb / breadcrumb-item & button / button-group

## Summary by CodeRabbit

* **Refactor**
  * Strengthened public TypeScript interfaces across Badge, Breadcrumb, Button, and ButtonGroup; components now use explicit typed props with inline defaults.

* **Chores**
  * Marked legacy prop type aliases and public typings as deprecated to guide migration to the new interfaces.

* **User-facing**
  * Components now expose explicit runtime defaults (badge offset [0, 0], breadcrumb separator '/', button defaults including direction, type, nativeType and loading icon) with preserved behavior.

* **Documentation**
  * Badge docs: offset default updated to [0, 0].

### DIFF
--- a/docs/en-US/component/badge.md
+++ b/docs/en-US/component/badge.md
@@ -59,18 +59,18 @@ badge/offset
 
 ### Attributes
 
-| Name                 | Description                                                                   | Type                                                                | Default |
-| -------------------- | ----------------------------------------------------------------------------- | ------------------------------------------------------------------- | ------- |
-| value                | display value.                                                                | ^[string] / ^[number]                                               | ''      |
-| max                  | maximum value, shows `{max}+` when exceeded. Only works if value is a number. | ^[number]                                                           | 99      |
-| is-dot               | if a little dot is displayed.                                                 | ^[boolean]                                                          | false   |
-| hidden               | hidden badge.                                                                 | ^[boolean]                                                          | false   |
-| type                 | badge type.                                                                   | ^[enum]`'primary' \| 'success' \| 'warning' \| 'danger' \| 'info'`  | danger  |
-| show-zero ^(2.6.0)   | Whether to show badge when value is zero.                                     | ^[boolean]                                                          | true    |
-| color ^(2.6.3)       | background color of the dot                                                   | ^[string]                                                           |         |
-| offset ^(2.7.0)      | offset of badge                                                               | ^[array]`[number, number]`                                          | [0, 0]  |
-| badge-style ^(2.7.1) | custom style of badge                                                         | ^[string] / ^[object]`CSSProperties \| CSSProperties[] \| string[]` | {}      |
-| badge-class ^(2.7.1) | custom class of badge                                                         | ^[string]                                                           | —       |
+| Name                 | Description                                                                   | Type                                                               | Default |
+| -------------------- | ----------------------------------------------------------------------------- | ------------------------------------------------------------------ | ------- |
+| value                | display value.                                                                | ^[string] / ^[number]                                              | ''      |
+| max                  | maximum value, shows `{max}+` when exceeded. Only works if value is a number. | ^[number]                                                          | 99      |
+| is-dot               | if a little dot is displayed.                                                 | ^[boolean]                                                         | false   |
+| hidden               | hidden badge.                                                                 | ^[boolean]                                                         | false   |
+| type                 | badge type.                                                                   | ^[enum]`'primary' \| 'success' \| 'warning' \| 'danger' \| 'info'` | danger  |
+| show-zero ^(2.6.0)   | Whether to show badge when value is zero.                                     | ^[boolean]                                                         | true    |
+| color ^(2.6.3)       | background color of the dot                                                   | ^[string]                                                          |         |
+| offset ^(2.7.0)      | offset of badge                                                               | ^[array]`[number, number]`                                         | [0, 0]  |
+| badge-style ^(2.7.1) | custom style of badge                                                         | ^[object]`CSSProperties`                                           | —       |
+| badge-class ^(2.7.1) | custom class of badge                                                         | ^[string]                                                          | —       |
 
 ### Slots
 
--- a/docs/en-US/component/badge.md
+++ b/docs/en-US/component/badge.md
@@ -59,18 +59,18 @@ badge/offset
 
 ### Attributes
 
-| Name                 | Description                                                                   | Type                                                               | Default |
-| -------------------- | ----------------------------------------------------------------------------- | ------------------------------------------------------------------ | ------- |
-| value                | display value.                                                                | ^[string] / ^[number]                                              | ''      |
-| max                  | maximum value, shows `{max}+` when exceeded. Only works if value is a number. | ^[number]                                                          | 99      |
-| is-dot               | if a little dot is displayed.                                                 | ^[boolean]                                                         | false   |
-| hidden               | hidden badge.                                                                 | ^[boolean]                                                         | false   |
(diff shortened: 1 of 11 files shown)


packages/components/badge/src/badge.ts
```diff
@@ -1,7 +1,53 @@
...
+   */
+  showZero?: boolean
+  /**
+   * @description customize dot background color
+   */
+  color?: string
+  /**
+   * @description CSS style of badge
+   */
+  badgeStyle?: StyleValue
```
patienry, line 37: <table border="1">
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
Image
</td>
    <td>
Image
</td>
  </tr>
</table>

After the modification, the default value of `badgeStyle` in **devtools** has changed. To keep it consistent with the previous version, we may need to explicitly set the default value. 🤔 WDYT?

snowbitx: > Before	After
> Image	Image
> After the modification, the default value of `badgeStyle` in **devtools** has changed. To keep it consistent with the previous version, we may need to explicitly set the default value. 🤔 WDYT?

Thanks for the review. I see see

packages/components/badge/src/badge.vue
```diff
@@ -26,15 +26,24 @@
...
-const props = defineProps(badgeProps)
+const props = withDefaults(defineProps<BadgeProps>(), {
+  value: '',
+  max: 99,
+  isDot: false,
+  hidden: false,
+  type: 'danger',
+  showZero: true,
+  offset: () => [0, 0],
+  badgeStyle: false,
```
rzzf:
```suggestion
  value: '',
  max: 99,
  type: 'danger',
  showZero: true,
  offset: () => [0, 0],
```

The default value for a Boolean type is false, so we can ignore it.

packages/components/breadcrumb/src/breadcrumb-item.vue
```diff
@@ -21,15 +21,18 @@ import { getCurrentInstance, inject } from 'vue'
...
 import type { Router } from 'vue-router'
 
 defineOptions({
   name: 'ElBreadcrumbItem',
 })
 
-const props = defineProps(breadcrumbItemProps)
+const props = withDefaults(defineProps<BreadcrumbItemProps>(), {
+  to: '',
+  replace: false,
```
rzzf:
```suggestion
```

It is the same here.

packages/components/button/src/button-group.vue
```diff
@@ -7,13 +7,16 @@
...
 import { buttonGroupContextKey } from './constants'
 
+import type { ButtonGroupProps } from './button-group'
+
 defineOptions({
   name: 'ElButtonGroup',
 })
-const props = defineProps(buttonGroupProps)
+const props = withDefaults(defineProps<ButtonGroupProps>(), {
+  direction: 'horizontal',
```
rzzf, line 18:
```suggestion
  direction: 'horizontal',
  type: '',
```

rzzf (changes requested):

packages/components/badge/src/badge.ts
snowbitx: > Before	After
> Image	Image
> After the modification, the default value of `badgeStyle` in **devtools** has changed. To keep it consistent with the previous version, we may need to explicitly set the default value. 🤔 WDYT?

@keeplearning66   Hi, I believe the style prop here should be consistent with input, so I updated the default value and the documentation.
Please take another look, and I can make further changes if needed.

image

rzzf: @snowbitx  We don't need to modify it temporarily; this is not the main content.

snowbitx: > @snowbitx We don't need to modify it temporarily; this is not the main content.

Sure, thanks for the review.
I noticed that the previous default value for the style here was false, which feels a bit unusual to me. What do you think?
The input style definition looks more complete, so I followed that approach.
Would you prefer keeping false or switching to undefined for now?

rzzf: > Sure, thanks for the review.
I noticed that the previous default value for the style here was false, which feels a bit unusual to me. What do you think?
The input style definition looks more complete, so I followed that approach.

okay, this is also good

packages/components/badge/src/badge.vue
```diff
@@ -26,15 +26,22 @@
 import { computed } from 'vue'
 import { useNamespace } from '@element-plus/hooks'
 import { addUnit, isNumber } from '@element-plus/utils'
-import { badgeProps } from './badge'
 
+import type { BadgeProps } from './badge'
 import type { StyleValue } from 'vue'
```
rzzf:
```suggestion
import type { StyleValue } from 'vue'
import type { BadgeProps } from './badge'
```

packages/components/breadcrumb/src/breadcrumb-item.vue
```diff
@@ -21,15 +21,17 @@ import { getCurrentInstance, inject } from 'vue'
 import ElIcon from '@element-plus/components/icon'
 import { useNamespace } from '@element-plus/hooks'
 import { breadcrumbKey } from './constants'
-import { breadcrumbItemProps } from './breadcrumb-item'
 
+import type { BreadcrumbItemProps } from './breadcrumb-item'
 import type { Router } from 'vue-router'
```
rzzf:
```suggestion
import type { Router } from 'vue-router'
import type { BreadcrumbItemProps } from './breadcrumb-item'
```

docs/en-US/component/badge.md
```diff
@@ -59,18 +59,18 @@ badge/offset
...
+| type                 | badge type.                                                                   | ^[enum]`'primary' \| 'success' \| 'warning' \| 'danger' \| 'info'`  | danger  |
+| show-zero ^(2.6.0)   | Whether to show badge when value is zero.                                     | ^[boolean]                                                          | true    |
+| color ^(2.6.3)       | background color of the dot                                                   | ^[string]                                                           |         |
+| offset ^(2.7.0)      | offset of badge                                                               | ^[array]`[number, number]`                                          | —       |
+| badge-style ^(2.7.1) | custom style of badge                                                         | ^[string] / ^[object]`CSSProperties \| CSSProperties[] \| string[]` | {}      |
+| badge-class ^(2.7.1) | custom class of badge                                                         | ^[string]                                                           | —       |
```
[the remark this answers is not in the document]
rzzf: @snowbitx  This seems reasonable, can you also fix it?

rzzf (approved):
Thanks for all the back-and-forth

keeplearning66 (approved):
Thank you!

packages/components/badge/src/badge.ts
btea: I think the default value in the documentation doesn't need to be changed. A default value `undefiend` can be set in the code.

snowbitx: > I think the default value in the documentation doesn't need to be changed. A default value `undefiend` can be set in the code.

Got it. I’ve reverted the `badgeStyle` changes. The behavior now matches the documentation, with the default value set to `undefined`.

btea (approved):

patienry:
242650CF-176D-4f4f-83EF-B1F925C0923F

relate [comment](https://github.com/element-plus/element-plus/pull/23414#discussion_r2701122322)
We might need to explicitly set the default value of `badgeStyle` to `undefined`, because after merging this PR, the default value of `badgeStyle` in **DevTools** is shown as `false`, which is different from the previous version. 🤔

PR #23418, opened 2026-01-17
refactor(components): [card/calendar/carousel] use type-based definitions

snowbitx:

ref: #23399  #23414 

## Summary by CodeRabbit

* **Documentation**
  * Reformatted Card API attributes table and updated body-style type/default.

* **New Features**
  * Card: header/footer defaults, headerClass/bodyClass/footerClass added; body-style accepts object/array/string and defaults to {}.
  * Carousel & CarouselItem: standardized defaults for many props (including interval, initial index, loop, card scale) and defaulted name/label.
  * Calendar: controllerType now defaults to "button".

* **Deprecations**
  * Legacy prop-type exports marked deprecated; migrate to new public interfaces.

### DIFF
--- a/packages/components/carousel/src/use-carousel.ts
+++ b/packages/components/carousel/src/use-carousel.ts
@@ -55,8 +55,12 @@ export const useCarousel = (
   )
 
   const hasLabel = computed(() => {
-    return items.value.some((item) => item.props.label.toString().length > 0)
+    return items.value.some(
+      (item) => (item.props.label ?? '').toString().length > 0
+    )
   })
+
+  const interval = computed(() => props.interval ?? 3000)
 
   const isCardType = computed(() => props.type === 'card')
   const isVertical = computed(() => props.direction === 'vertical')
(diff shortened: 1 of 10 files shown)


btea:
Please do not mix modifications from multiple components together in the future.

packages/components/carousel/src/use-carousel.ts
```diff
@@ -55,9 +55,13 @@ export const useCarousel = (
   )
 
   const hasLabel = computed(() => {
-    return items.value.some((item) => item.props.label.toString().length > 0)
+    return items.value.some(
```
btea, line 58: This change seems unnecessary, and we shouldn't modify code logic in a code refactoring pull request.

snowbitx: > This change seems unnecessary, and we shouldn't modify code logic in a code refactoring pull request.

This change switches from runtime prop validation to type-based validation, so the props are now defined as optional in the interface.
Because of this, we need to handle the `undefined` case in the implementation; otherwise, `pnpm run typecheck` fails due to possibly undefined errors.
Would there be a better approach for this in the codebase?

snowbitx:
> Please do not mix modifications from multiple components together in the future.

Thanks for the review.
My intention was to group similar changes together, but I understand the concern.
I’ll follow this guideline and keep future changes split by component.

packages/components/carousel/src/use-carousel.ts
snowbitx: > This change seems unnecessary, and we shouldn't modify code logic in a code refactoring pull request.

Hi, as this PR overlaps with changes from other contributors, I’ve decided to close it.
We can revisit and discuss this in a new [PR](https://github.com/element-plus/element-plus/pull/23430) later.

PR #23445, opened 2026-01-18
refactor(components): [notification] use type-based definitions

E66Crisp:
ref  #23399

## Summary by CodeRabbit

* **Improvements**
  * Clearer, explicit typing and a unified prop surface for notifications.
  * Added explicit notification types and four corner positions (now including bottom-right).
  * Default values declared for all props and a refined default close icon for better dismissal UX.
  * Positioning/offset handling updated to support all corner positions.
  * Legacy prop alias marked as deprecated to guide migration.

* **Tests**
  * Simplified tests: removed assertion of console warning for invalid icon types.

### DIFF
--- a/packages/components/notification/src/notification.vue
+++ b/packages/components/notification/src/notification.vue
@@ -39,7 +39,7 @@
 </template>
 
 <script lang="ts" setup>
-import { computed, onMounted, ref } from 'vue'
+import { computed, markRaw, onMounted, ref } from 'vue'
 import { useEventListener, useTimeoutFn } from '@vueuse/core'
 import { TypeComponentsMap, getEventCode } from '@element-plus/utils'
 import { EVENT_CODE } from '@element-plus/constants'
@@ -66,7 +66,7 @@ const props = withDefaults(defineProps<NotificationProps>(), {
   showClose: true,
   title: '',
   type: '',
-  closeIcon: Close,
+  closeIcon: markRaw(Close),
 })
 defineEmits(notificationEmits)
 
--- a/packages/components/notification/src/notification.vue
+++ b/packages/components/notification/src/notification.vue
@@ -45,15 +45,29 @@ import { TypeComponentsMap, getEventCode } from '@element-plus/utils'
 import { EVENT_CODE } from '@element-plus/constants'
 import { ElIcon } from '@element-plus/components/icon'
 import { useGlobalComponentSettings } from '@element-plus/components/config-provider'
-import { notificationEmits, notificationProps } from './notification'
+import { notificationEmits } from './notification'
+import { Close } from '@element-plus/icons-vue'
 
 import type { CSSProperties } from 'vue'
+import type { NotificationProps } from './notification'
 
 defineOptions({
   name: 'ElNotification',
 })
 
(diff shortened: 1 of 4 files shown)


E66Crisp:
Hello Reviewer, regarding the content about `icon?: string | Component`, is it necessary to export it as a type in packages\utils\vue\icon.ts? For example, add `export type IconComponentType = string | Component` in the icon.ts file.

rzzf:
> Hello Reviewer, regarding the content about `icon?: string | Component`, is it necessary to export it as a type in packages\utils\vue\icon.ts? For example, add `export type IconComponentType = string | Component` in the icon.ts file.

It might be needed in the future, but I don’t think it’s necessary for now.

E66Crisp:
> > Hello Reviewer, regarding the content about `icon?: string | Component`, is it necessary to export it as a type in packages\utils\vue\icon.ts? For example, add `export type IconComponentType = string | Component` in the icon.ts file.
> 
> It might be needed in the future, but I don’t think it’s necessary for now.

Yes, refactoring the code does not include this part of the modification.

E66Crisp:
Hello Reviewer, I'm encountering a test failure after refactoring from `buildProps` to `defineProps<NotificationProps>()`. 

The test expects `console.warn` to be called for invalid prop values, but runtime validation was lost after the migration. I noticed `Message` component has a similar test but doesn't check for `console.warn` after the same migration.

Should I:
1. Remove the `console.warn` expectation from the test (similar to `Message`), or
2. Add manual runtime validation to preserve the developer experience?

Looking forward to your guidance. Thanks!

rzzf:
> 1. Remove the `console.warn` expectation from the test (similar to `Message`), or

We can first comment out the assertion instead of deleting it.

> 2. Add manual runtime validation to preserve the developer experience?

See https://github.com/element-plus/element-plus/pull/23458

E66Crisp:
> > 1. Remove the `console.warn` expectation from the test (similar to `Message`), or
> 
> We can first comment out the assertion instead of deleting it.
> 
> > 2. Add manual runtime validation to preserve the developer experience?
> 
> See #23458

Got it. I've commented out this section of code. tks.

packages/components/notification/src/notification.vue
```diff
@@ -45,15 +45,29 @@ import { TypeComponentsMap, getEventCode } from '@element-plus/utils'
...
+  duration: 4500,
+  id: '',
+  message: '',
+  offset: 0,
+  onClick: () => undefined,
+  position: 'top-right',
+  showClose: true,
+  title: '',
+  type: '',
+  closeIcon: Close,
```
btea:
```suggestion
  closeIcon: markRaw(Close),
```

E66Crisp: It seems #23455 has the same case. Should they all be unified to use markRaw?

btea: There is a difference between the two: either set it as a function that directly returns the icon component, or use the `markRaw` tag.

E66Crisp: tks, i got it

rzzf (approved):

btea (approved):

PR #23483, opened 2026-01-19
refactor(components): [form/form-item] use type-based definitions

snowbitx:

ref #23399 

## Summary by CodeRabbit

* **Refactor**
  * Converted Form and FormItem public prop typings to explicit TypeScript interfaces (FormProps, FormMetaProps, FormItemProps) with deprecation notices for prior types.
  * Switched components to typed props with defaults for several fields (e.g., labelPosition, showMessage, required, inlineMessage, labelWidth).
  * Preserved runtime behavior and backward compatibility while clarifying exported prop shapes.

### DIFF
--- a/packages/components/form/src/form-item.ts
+++ b/packages/components/form/src/form-item.ts
@@ -1,7 +1,7 @@
-import { ExtractPublicPropTypes } from 'vue'
 import { componentSizes } from '@element-plus/constants'
 import { buildProps, definePropType } from '@element-plus/utils'
 
+import type { ExtractPublicPropTypes } from 'vue'
 import type { ComponentSize } from '@element-plus/constants'
 import type { Arrayable } from '@element-plus/utils'
 import type { FormItemRule } from './types'
--- a/packages/components/form/src/form-item.ts
+++ b/packages/components/form/src/form-item.ts
@@ -1,3 +1,4 @@
+import { ExtractPublicPropTypes } from 'vue'
 import { componentSizes } from '@element-plus/constants'
 import { buildProps, definePropType } from '@element-plus/utils'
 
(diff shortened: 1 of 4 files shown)


packages/components/form/src/form-item.ts
```diff
@@ -90,5 +144,3 @@ export const formItemProps = buildProps({
     values: componentSizes,
   },
 } as const)
-export type FormItemProps = ExtractPropTypes<typeof formItemProps>
-export type FormItemPropsPublic = ExtractPublicPropTypes<typeof formItemProps>
```
patienry, line 94: Image

I think it's better not to delete `FormItemPropsPublic` directly here, because `FormItemPropsPublic` is exported, and we are not sure if any users are using it. The same applies to `FormPropsPublic` and `FormMetaPropsPublic` below. WDYT?

btea, line 94: For types like `FormItemProps` that we need to reassign, we can simply modify them. For types like `FormItemPropsPublic`, we can temporarily keep them by adding the `deprecated` flag.

snowbitx: done thank you

snowbitx: done thank you

packages/components/form/src/form-item.ts
```diff
@@ -1,7 +1,8 @@
+import { ExtractPublicPropTypes } from 'vue'
```
patienry: It’s better to use `import type { ExtractPublicPropTypes } from 'vue'`, the same applies to `form.ts`.

snowbitx: Auto-imports can be such a trap. 😅

keeplearning66 (approved):
Thank you!

btea (approved):

rzzf (approved):

PR #23513, opened 2026-01-23
fix(components): [autocomplete] reset ignoreFocusEvent on outside click

E66Crisp:

fix: #23512
Reset ignoreFocusEvent flag when clicking outside the autocomplete, regardless of suggestionVisible state, to ensure dropdown can open normally on next focus after clicking non-selectable areas in dropdown.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved outside-click and blur handling for the autocomplete so suggestions and focus transitions no longer close unexpectedly; focus-aware logic prevents incorrect dismissals in edge cases.

* **Chores**
  * No changes to public APIs or visible interactions; users should see fewer unexpected dismissals or focus glitches.

### DIFF
--- a/packages/components/autocomplete/src/autocomplete.vue
+++ b/packages/components/autocomplete/src/autocomplete.vue
@@ -295,7 +295,6 @@ const handleBlur = (evt: FocusEvent) => {
       ignoreFocusEvent = true
       return
     }
-    ignoreFocusEvent = false
     activated.value && close()
     emit('blur', evt)
   })
@@ -398,8 +397,13 @@ const getSuggestionContext = () => {
   return [suggestion, suggestionList] as const
 }
 
-const stopHandle = onClickOutside(listboxRef, () => {
-  handleBlur(new FocusEvent('blur'))
+const stopHandle = onClickOutside(listboxRef, (event: FocusEvent) => {
+  // Prevent closing if focus is inside popper content
+  if (popperRef.value?.isFocusInsideContent()) return
+  if (suggestionVisible.value) {
+    ignoreFocusEvent = false
+    handleBlur(new FocusEvent('blur', event))
+  }
 })
 
 const handleKeydown = (e: KeyboardEvent | Event) => {
--- a/packages/components/autocomplete/src/autocomplete.vue
+++ b/packages/components/autocomplete/src/autocomplete.vue
@@ -295,6 +295,7 @@ const handleBlur = (evt: FocusEvent) => {
       ignoreFocusEvent = true
       return
     }
+    ignoreFocusEvent = false
     activated.value && close()
     emit('blur', evt)
   })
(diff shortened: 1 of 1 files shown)


rzzf (changes requested):
Can you add a test case?

packages/components/autocomplete/src/autocomplete.vue
```diff
@@ -400,7 +400,10 @@ const getSuggestionContext = () => {
 const stopHandle = onClickOutside(listboxRef, () => {
   // Prevent closing if focus is inside popper content
   if (popperRef.value?.isFocusInsideContent()) return
-  suggestionVisible.value && close()
+  if (suggestionVisible.value) {
+    ignoreFocusEvent = false
+    close()
+  }
 })
```
rzzf: After carefully reviewing the related fix PRs, I think we may still need to manually trigger the blur event.

Starting from #19522, clicking on the edge of the dropdown menu or its scrollbar no longer hides the dropdown. 
As a result, when the dropdown is later closed by clicking outside, the blur event is not triggered.

Reproduction steps:

1. Show the dropdown
2. Click on the edge of the dropdown or the scrollbar
3. Focus moves from the input to the dropdown content
4. Click outside to close the dropdown
5. blur is not triggered again

rzzf: I think is to manually trigger a blur event based on https://github.com/element-plus/element-plus/pull/23513/changes/ee21ccd74d799c3f4b23d92444df1c65a69e6821.
The current implementation is not correct, as it causes the blur event to be triggered one extra time.

```ts
const stopHandle = onClickOutside(listboxRef, (event: FocusEvent) => {
  // Prevent closing if focus is inside popper content
  if (popperRef.value?.isFocusInsideContent()) return
  if (suggestionVisible.value) {
    ignoreFocusEvent = false
    handleBlur(new FocusEvent('blur', event))
  }
})
```

packages/components/autocomplete/src/autocomplete.vue
```diff
@@ -397,10 +397,13 @@ const getSuggestionContext = () => {
...
 
-const stopHandle = onClickOutside(listboxRef, () => {
+const stopHandle = onClickOutside(listboxRef, (event: FocusEvent) => {
   // Prevent closing if focus is inside popper content
   if (popperRef.value?.isFocusInsideContent()) return
-  suggestionVisible.value && close()
+  if (suggestionVisible.value) {
+    ignoreFocusEvent = false
+    handleBlur(new FocusEvent('blur', event))
+  }
```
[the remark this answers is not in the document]
rzzf: In fact, `handleBlur` and the `setTimeout` inside it are executed before the `onClickOutside` callback. 
As a result, before `handleBlur` is called within `onClickOutside`, if the `@blur` handler on the input has already processed the blur logic and set `suggestionVisible` to `false`, then `handleBlur` in `onClickOutside` will not be invoked.

However, adding this conditional check shouldn’t cause any side effects. I think we could also handle it this way. What do you think? @E66Crisp

rzzf (approved):

Dsaquel (approved):

PR #23532, opened 2026-01-26
refactor(components): [time-picker] correct return types for disabled

snowbitx:

Before:
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Time-picker: improved handling when minute/second disabled-value providers return no values, preventing errors while preserving existing behavior of selectors and lists; increases robustness of disabled-time calculations.

* **Refactor**
  * Internal list-generation logic adjusted to gracefully accept optional absence of disabled entries.

### DIFF
--- a/packages/components/time-picker/src/common/props.ts
+++ b/packages/components/time-picker/src/common/props.ts
@@ -25,13 +25,13 @@ export type GetDisabledMinutes = (
   hour: number,
   role: string,
   comparingDate?: Dayjs
-) => number[]
+) => number[] | undefined
 export type GetDisabledSeconds = (
   hour: number,
   minute: number,
   role: string,
   comparingDate?: Dayjs
-) => number[]
+) => number[] | undefined
 
 export const timePickerDefaultProps = buildProps({
   /**
--- a/packages/components/time-picker/src/composables/use-time-panel.ts
+++ b/packages/components/time-picker/src/composables/use-time-panel.ts
@@ -30,7 +30,7 @@ export const useTimePanel = ({
     let result = date
     ;(['hour', 'minute', 'second'] as const).forEach((type) => {
       if (availableTimeGetters[type]) {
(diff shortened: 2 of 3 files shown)


btea:
Based on your screenshot example, the conditional statement in the code doesn't handle all cases, resulting in `undefined` being returned outside of the conditional statement.

I think it's more reasonable to strictly restrict the function return value to an array containing numbers on the user side; we shouldn't need to modify it.

snowbitx:
> Based on your screenshot example, the conditional statement in the code doesn't handle all cases, resulting in `undefined` being returned outside of the conditional statement.
> 
> I think it's more reasonable to strictly restrict the function return value to an array containing numbers on the user side; we shouldn't need to modify it.

Thanks for the review!
I think the usage shown in the example is reasonable.
Users should only need to focus on the condition itself, not on ensuring every function explicitly returns an array.
If we don’t adjust the type definition, users would be forced to add `return []` to every function, which feels unnecessary and adds extra burden.
What do you think?
image

patienry:
Thank you for your contribution. 🙏 
Personally, I feel that if a function has a return value in other conditional statements, it is a relatively common practice to return a default value when none of the conditional statements are matched.
Additionally, I have checked some other component libraries, and the current handling methods here are consistent.

btea:
In conclusion, we have decided to close this PR. However, we still appreciate your contribution.

snowbitx:
> Thank you for your contribution. 🙏 Personally, I feel that if a function has a return value in other conditional statements, it is a relatively common practice to return a default value when none of the conditional statements are matched. Additionally, I have checked some other component libraries, and the current handling methods here are consistent.

Got it. thank you!

snowbitx:
> In conclusion, we have decided to close this PR. However, we still appreciate your contribution.


> Thanks for the kind words. I'm still learning and improving as well.🙏

PR #23539, opened 2026-01-28
fix(theme-chalk): [card] ensure height works correctly in flex layout

rzzf:
fix #23538

## Summary by CodeRabbit

* **Style**
  * Adjusted card component layout to use an explicit grow setting for flex behavior, refining how cards resize inside flexible containers for more predictable layout and spacing.

### DIFF
--- a/packages/theme-chalk/src/card.scss
+++ b/packages/theme-chalk/src/card.scss
@@ -35,7 +35,7 @@
   }
 
   @include e(body) {
-    flex-grow: 1;
+    flex-grow: 1; // #23538
     overflow: auto;
     padding: getCssVar('card', 'padding');
   }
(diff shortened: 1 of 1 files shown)


btea (approved):

packages/theme-chalk/src/card.scss
```diff
@@ -35,7 +35,7 @@
   }
 
   @include e(body) {
-    flex: 1;
+    flex-grow: 1;
```
btea: We can put the issue number here. It's not convenient to add test cases here, to prevent future coverage and regression issues. 🤔

Dsaquel (approved):
It seems like `flex-basis: 0%` has a really weird behavior.

PR #23544, opened 2026-01-28
feat(components): [drawer] supplement type for `DrawerInstance`

snowbitx:

I have extracted all APIs exposed via defineExpose from the documentation and added them to the Developer Guide. All entries have been verified to provide proper TypeScript hints, making it easier for developers to discover and reference the available types.
image

image

## Summary by CodeRabbit

* **Documentation**
  * Added a "Component instance types" guide with TypeScript examples and a detailed reference table to help obtain type hints for component refs.

* **New Features**
  * Enhanced TypeScript support for the Drawer component to improve IDE autocompletion and developer experience when working with component references.

### DIFF
--- a/docs/en-US/guide/dev-guide.md
+++ b/docs/en-US/guide/dev-guide.md
@@ -71,7 +71,7 @@ Components that currently use `defineExpose`:
 | Component        | Instance type(s)                                |
 | ---------------- | ----------------------------------------------- |
 | Button           | `ButtonInstance`, `ButtonGroupInstance`         |
-| Scrollbar        | `ScrollbarPublicExpose`                         |
+| Scrollbar        | `ScrollbarInstance`                             |
 | Splitter         | `SplitterInstance`, `SplitterPanelInstance`     |
 | Autocomplete     | `AutocompleteInstance`                          |
 | Cascader         | `CascaderInstance`                              |
--- a/docs/en-US/guide/dev-guide.md
+++ b/docs/en-US/guide/dev-guide.md
@@ -85,6 +85,7 @@ Components that currently use `defineExpose`:
 | Mention          | `MentionInstance`                               |
 | Rate             | `RateInstance`                                  |
 | Switch           | `SwitchInstance`                                |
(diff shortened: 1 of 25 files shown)


snowbitx (commented):
submit

packages/components/drawer/src/drawer.ts
```diff
@@ -76,3 +77,11 @@ export const drawerEmits = {
...
 
 export type DrawerEmits = typeof drawerEmits
+
+export type DrawerExpose = {
+  handleClose: () => void
+  afterEnter: () => void
+  afterLeave: () => void
+}
+
+export type DrawerInstance = InstanceType<typeof Drawer> & DrawerExpose
```
patienry: Do we need to export `DrawerExpose` separately here? Maybe we can handle it the same way as other components, for example, `export type DrawerInstance = InstanceType<typeof Drawer> & unknown`. 🤔

snowbitx: Thanks for the suggestion. I followed the approach used in this [PR](https://github.com/element-plus/element-plus/pull/22387/changes), and after testing, I found that using `InstanceType<typeof Drawer> & unknown` does not affect the TypeScript hints. I’ve updated the implementation accordingly.

docs/en-US/guide/dev-guide.md
```diff
@@ -51,6 +51,67 @@ will start the local development environment.
 
 Modify `App.vue` file per your needs to get things work.
 
+## Component instance types
```
Dsaquel: I don't think it is really necessary to add this section. If it's to remind the developers to not forget the Instance type when doing a new component perhaps they will refer to the existing components.

snowbitx: Got it, thanks for the clarification. I initially added this documentation because I thought there were many components missing TypeScript instance hints. I’ve reverted this part of the changes now.

keeplearning66 (approved):
Thank you!

btea (approved):

PR #23556, opened 2026-01-29
feat(components): [select/select-v2] add `tag-tooltip` config object prop

LostElkByte:
[select/select-v2] Description
This PR introduces the tag-tooltip configuration object for both el-select and el-select-v2 components. It provides granular control over the Tooltip that appears when hovering over collapsed tags (in collapse-tags mode), addressing issues where the tooltip might be clipped or misaligned in complex layouts (e.g., inside dialogs or custom containers).

Key Changes

Unified API: Added tag-tooltip to both el-select and el-select-v2 to maintain consistency across select components.
Decoupled Styles: Allows independent control of the tag tooltip's styles, resolving the issue where popper-class and popper-style were previously shared between the dropdown and the tag tooltip.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable tag-tooltip for Select collapse-tags (append target, placement, fallback placements, effect, classes/styles, teleporting, popper options, show/hide delays, auto-close, offset, transition).
  * New public configuration type for the tag-tooltip API.

* **Documentation**
  * Select docs updated with a dedicated tag-tooltip section, attribute table, examples and guidance for custom container positioning and fallback behavior.

* **Tests**
  * Tests updated to verify tag-tooltip integration and that tooltip content is appended to the configured target.


LostElkByte:
I think it's a pretty common need to set a specific container for the "show-all-tags" tooltip when using collapse-tags-tooltip. We should provide a direct attribute for this instead of making users deal with popper-options.

btea:
lgtm. The prop name is suitable, but it feels too long. What do you guys think? cc @kooriookami @rzzf @Dsaquel @keeplearning66

LostElkByte:
> lgtm. The prop name is suitable, but it feels too long. What do you guys think? cc @kooriookami @rzzf @Dsaquel @keeplearning66

We can simplify the prop name to tooltipAppendTo and drop collapseTags. In the Select component, the tooltip is only triggered when collapseTags is enabled, so there is no ambiguity or alternative context.

btea:
This seems suitable; let's modify it accordingly first.

LostElkByte:
I force-pushed this branch to rebase it onto the latest `dev` in order to keep a clean, linear history.

During the rebase, this also includes the rename from `collapseTagsTooltipAppendTo` to `tooltipAppendTo`,
which is part of the same change and aligns the prop naming with the updated API.
No behavior or functionality is changed. @btea

rzzf:
`tooltipAppendTo` looks good.

packages/components/select/__tests__/select.test.ts
```diff
@@ -1505,10 +1505,20 @@ describe('Select', () => {
...
+        <el-select
+          v-model="selectedList"
+          multiple
+          collapseTags
+          collapse-tags-tooltip
+          tooltip-append-to=".append-target"
+          placeholder="请选择"
+        >
+          <el-option v-for="item in options" :key="item.value" :label="item.label" :value="item.value">
+          </el-option>
```
rzzf:
```suggestion
          <el-option v-for="item in options" :key="item.value" :label="item.label" :value="item.value" />
```

rzzf (commented):
LGTM, I have a small structural change here.

LostElkByte:
> LGTM, I have a small structural change here.

Thanks for the suggestion!  
Applied the self-closing change as suggested 👍

rzzf (approved):

keeplearning66 (commented):
I’m wondering if it’s necessary to have a single prop to centrally manage the **tag tooltip** props (e.g. `tagTooltipProps: { appendTo: useTooltipContentProps.appendTo }`). The reason is that in the future, users might want to separately control other props of the **tag tooltip** (e.g. `popperClass`, `popperStyle`). If we keep adding individual props each time, the component’s props might become a bit bloated. 🤔

LostElkByte:
> I’m wondering if it’s necessary to have a single prop to centrally manage the **tag tooltip** props (e.g. `tagTooltipProps: { appendTo: useTooltipContentProps.appendTo }`). The reason is that in the future, users might want to separately control other props of the **tag tooltip** (e.g. `popperClass`, `popperStyle`). If we keep adding individual props each time, the component’s props might become a bit bloated. 🤔

I’d like to share my thoughts on the tagTooltip configuration design with you, and hope to get your valuable feedback on this proposal. Specifying a target element for the tagTooltip is actually a frequently encountered requirement in daily development. Since many developers may not be familiar with Popper.js configuration options, placing this common setting solely within  tagTooltipProps  could raise the usage barrier for them, which may not be the most user-friendly approach.
 
My suggestion is to expose a top-level prop for this high-frequency need (e.g., the appendTo configuration for the tagTooltip), and keep  tagTooltipProps  as a supplementary option for less commonly used customizations. This approach could balance the ease of use for most developers and the flexibility for advanced customization scenarios — it directly addresses the core needs of regular developers with lower learning costs, while still reserving full customizability for those who have more in-depth customization requirements.

LostElkByte:
> I’m wondering if it’s necessary to have a single prop to centrally manage the **tag tooltip** props (e.g. `tagTooltipProps: { appendTo: useTooltipContentProps.appendTo }`). The reason is that in the future, users might want to separately control other props of the **tag tooltip** (e.g. `popperClass`, `popperStyle`). If we keep adding individual props each time, the component’s props might become a bit bloated. 🤔

For a number of reasons – such as visual aesthetics requirements, and the occasional issue where the tagTooltip loses its right border in certain browsers like Chrome – testers and the product team have specified that the tooltip should be confined within the main content area, rather than the entire viewport. This is an issue I’ve received countless feedback on over the years, as many team members have no idea how to configure popper-options to fix it. Just last week, this problem cropped up again: one of our team members couldn’t work out the correct configuration, and even AI failed to deliver an elegant solution. Consequently, following AI’s guidance, they resorted to an extremely inelegant workaround – overriding the default slot and manually adding the appendTo attribute to a custom el-tooltip component – just to resolve the issue.
 
This is why I’m proposing to expose this configuration as a dedicated top-level prop: it addresses a recurring pain point that has long plagued developers who find Popper.js configuration daunting. Moreover, custom scenarios involving properties like popperClass and popperStyle are far less common in daily development than appendTo; in fact, appendTo is already implemented as a top-level prop in the native el-tooltip component itself, which further justifies this design choice.🤯

btea:
@keeplearning66 Thank you for your thoughtful consideration; normally we should indeed consider it that way. However, in our current implementation, configuring `popperClass` and `popperStyle` applies to the dropdown and the floating area of ​​the tag tooltip.
https://github.com/element-plus/element-plus/blob/065b6b023e27f0630c536264e8b676256a3711f2/packages/components/select/src/select.vue#L101-L102

@LostElkByte Thank you for providing sufficient reasons.

While reviewing the source code and answering the questions above, I suddenly realized that tag tooltips do indeed have many configurable options. Perhaps we could provide a separate object called tagTooltip to allow users to configure `placement`, `fallback-placements`, `teleported`, `appendTo`, and other settings. 🤔

btea:
> This is why I’m proposing to expose this configuration as a dedicated top-level prop: it addresses a recurring pain point that has long plagued developers who find Popper.js configuration daunting. Moreover, custom scenarios involving properties like popperClass and popperStyle are far less common in daily development than appendTo; in fact, appendTo is already implemented as a top-level prop in the native el-tooltip component itself, which further justifies this design choice.🤯

Yes, it does work very well. However, in our source code implementation, it's actually a nested tooltip component, with the tag tooltip being one of the nested layers. Therefore, it's more reasonable to centralize the configuration of the nested layers.

LostElkByte:
> 我在想是否有必要有一个道具来集中管理**标签工具提示**道具（例如 `tagTooltipProps: { appendTo: useTooltipContentProps.appendTo }` ）。原因是未来用户可能希望单独控制**标签工具提示**的其他道具（例如 `popperClass`、`popperStyle`）。如果我们每次都添加单个道具，组件的道具可能会有些臃肿。🤔

I understand your point, and I agree.

LostElkByte:
> @keeplearning66感谢您的细致考虑;通常我们确实应该这样看待。然而，在我们当前的实现中，配置 `popperClass` 和 `popperStyle` 适用于标签工具提示的下拉菜单和浮动区域。
> 
> https://github.com/element-plus/element-plus/blob/065b6b023e27f0630c536264e8b676256a3711f2/packages/components/select/src/select.vue#L101-L102
> 
> @LostElkByte感谢你提供了充分的理由。
> 
> 在复习源代码并回答上述问题时，我突然意识到标签工具提示确实有很多可配置的选项。也许我们可以提供一个名为 tagTooltip 的独立对象，允许用户配置`放置 `、` 备选放置 `、` 传送 `、` 附加`及其他设置。🤔

I understand your point and agree.

LostElkByte:
1. Introduce tag-tooltip-props object
Instead of adding individual top-level props, we should introduce a single tag-tooltip-props object.
Why: This avoids "prop bloating" at the component's top level and provides scalability for future tooltip-related configurations (e.g., show-after, offset, effect, etc.) without needing further API changes.

2. Precedence / Fallback Logic
To ensure both flexibility and consistency, the tooltip attributes should follow this priority:
Priority 1 (Highest): Specific fields defined within the tag-tooltip-props object.
Example: :tag-tooltip-props="{ appendTo: 'body' }" will mount the tooltip to the body.
Priority 2 (Inheritance/Fallback): If a specific field is not defined in the object, it should fall back to the existing top-level props of el-select.
Example: popper-class and popper-style. If tag-tooltip-props.popperClass is missing, it should inherit the Select's popper-class to maintain UI consistency by default.
Priority 3 (Baseline): If neither of the above is provided, the default values of the el-tooltip component will be used.

3. Backward Compatibility
Non-breaking: This approach is completely seamless for existing users. Since we implemented the fallback mechanism, any existing global styling applied via popper-class will continue to work exactly as before.
Opt-in Precision: The new prop is purely "opt-in." Only developers who need fine-grained control over the tag tooltip (e.g., to solve the dialog overflow issue) need to use it.

Based on our discussion, I’ve summarized a refined approach for this PR. I’d love to hear your thoughts and suggestions on this. @btea @keeplearning66  @rzzf

btea:
@LostElkByte Your summary is very clear and complete. 👍

docs/en-US/component/select.md
```diff
@@ -271,6 +272,21 @@ select/custom-label
...
+| popper-options      | [popper.js](https://popper.js.org/docs/v2/) parameters                                                               | ^[object]refer to [popper.js](https://popper.js.org/docs/v2/) doc                                                                                                           | —                                  |
+| show-after          | delay of appearance, in millisecond                                                                                  | ^[number]                                                                                                                                                                   | —                                  |
+| offset              | offset of the Tooltip                                                                                                | ^[number]                                                                                                                                                                   | —                                  |
+
```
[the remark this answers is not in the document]
btea: @LostElkByte Could you add text descriptions to the document as Rabbit suggested?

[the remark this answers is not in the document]
LostElkByte: > @LostElkByte你能像 Rabbit 建议的那样，给文档添加文字描述吗？

Any field omitted in `tag-tooltip` follows a fallback mechanism: it first inherits from the corresponding prop of the `el-select` component (e.g., `append-to`, `effect`, `popper-class`, `popper-style`, `teleported`, `popper-options`), and then falls back to the default value of the underlying `el-tooltip` component. Note that Select-level `append-to`, `effect`, `popper-class`, `popper-style`, `teleported`, and `popper-options` apply to both the Select dropdown and the collapse-tags tooltip unless specifically overridden here. Fields `placement` and `fallback-placements` have their own defaults (`'bottom'` and `['bottom', 'top', 'right', 'left']` respectively), while `show-after` and `offset` are passed directly to the tooltip.

Is this wording acceptable, or should I simplify it further?

LostElkByte:
I have added the tag-tooltip prop. Please take a look and let me know if there are any errors or suggestions. I will address them tomorrow. @btea @keeplearning66 @rzzf

packages/components/select/src/select.ts
```diff
@@ -27,6 +27,22 @@ import type {
...
+/**
+ * @description Tag tooltip configuration interface
+ */
+export interface TagTooltipProps {
+  appendTo?: string | HTMLElement
+  placement?: Placement
+  fallbackPlacements?: Placement[]
+  effect?: PopperEffect
+  popperClass?: string
+  popperStyle?: CSSProperties
```
rzzf:
```suggestion
  popperStyle?: string | CSSProperties
```

packages/components/select/src/select.vue
```diff
@@ -95,13 +95,23 @@
...
+                "
+                :effect="tagTooltip.effect ?? effect"
+                :placement="tagTooltip.placement ?? 'bottom'"
+                :popper-class="tagTooltip.popperClass ?? popperClass"
+                :popper-style="tagTooltip.popperStyle ?? popperStyle"
+                :teleported="tagTooltip.teleported ?? teleported"
+                :append-to="tagTooltip.appendTo ?? appendTo"
+                :popper-options="tagTooltip.popperOptions ?? popperOptions"
+                :show-after="tagTooltip.showAfter"
+                :offset="tagTooltip.offset"
```
rzzf: Maybe we can keep the original list of props and then use `v-bind="tagTooltip"`. What do you think? Vue should automatically merge props with the same name.

LostElkByte: I'm considering using v-bind for tag-tooltip to allow full customization. My main concern is that some props (e.g., content, manual, visible) might break the internal tag logic if overridden by the user. I plan to exclude these 'conflict-heavy' props and only pass through the 'safe' ones. Does this approach sound good to you?

Proposed Blacklist:

content: The tooltip must always display the tag's label (especially when truncated). Allowing users to override this would defeat the purpose of the tag tooltip.

visible / v-model: In a multi-select scenario, each tag needs its own visibility state. An external visible prop would cause state conflicts across all tags.

virtual-ref / virtual-triggering: We need to bind the tooltip to the actual Tag DOM element to ensure correct positioning. Bypassing this via virtual refs would break the internal link.

manual: Tag tooltips should react to hover/overflow automatically. Manual mode would require the user to manage the state of every individual tag, which isn't feasible here.

disabled: (Internal Priority) el-select needs to disable the tooltip when text is not overflowing. We should merge this with the user's disabled prop to ensure we don't show empty tooltips.

onUpdate:visible: To prevent v-model side effects from interfering with the internal display logic.

By filtering these, we can safely allow users to customize all other aspects like placement, effect, show-after, offset, and popper-class without risking component instability."

Or, shall we not deal with it and allow the users more freedom instead?

rzzf: > I'm considering using v-bind for tag-tooltip to allow full customization. My main concern is that some props (e.g., content, manual, visible) might break the internal tag logic if overridden by the user.

Yeah, that makes sense, let’s leave it as is for now.

btea (approved):

rzzf (approved):

docs/en-US/component/select.md
```diff
@@ -271,6 +272,41 @@ select/custom-label
...
 | disabled          | specify which key of node object is used as the node's disabled | ^[string] | disabled |
 
+### tag-tooltip
+
+:::tip Fallback Mechanism
+
+Properties in tag-tooltip follow this priority order:
+
+1. Explicitly defined fields within the tag-tooltip object.
+2. Shared props inherited from el-select (e.g., effect, teleported, popper-class, popper-options).
```
patienry:
```suggestion
2. Shared props inherited from el-select (e.g. effect, teleported, popper-class, popper-options).
```

docs/en-US/component/select.md
```diff
@@ -271,6 +272,41 @@ select/custom-label
...
+| placement           | position of Tooltip                                                                                                  | ^[enum]`'top' \| 'top-start' \| 'top-end' \| 'bottom' \| 'bottom-start' \| 'bottom-end' \| 'left' \| 'left-start' \| 'left-end' \| 'right' \| 'right-start' \| 'right-end'` | bottom                             |
+| fallback-placements | list of possible positions for Tooltip [popper.js](https://popper.js.org/docs/v2/modifiers/flip/#fallbackplacements) | ^[array]`Placement[]`                                                                                                                                                       | ['bottom', 'top', 'right', 'left'] |
+| effect              | Tooltip theme, built-in theme: `dark` / `light`                                                                      | ^[enum]`'dark' \| 'light'`                                                                                                                                                  | —                                  |
```
patienry:
```suggestion
| effect              | Tooltip theme, built-in theme: `dark` / `light`                                                                      | ^[enum]`'dark' \| 'light'` / ^[string]                                                                                                                                                  | —                                  |
```

patienry: I think CoderabbitAI's [comment](https://github.com/element-plus/element-plus/pull/23556#discussion_r2762663747) is reasonable. It’s best to keep consistency here.

LostElkByte: > I think CoderabbitAI's [comment](https://github.com/element-plus/element-plus/pull/23556#discussion_r2762663747) is reasonable. It’s best to keep consistency here.

Good catch! Thank you for pointing this out. I've committed the suggestion. Much appreciated!

keeplearning66 (approved):
Thank you! We usually make changes to both `select` and `select-v2`. Would you be willing to make the same changes to `select-v2`? 🙏

LostElkByte:
> 谢谢！我们通常会对两者进行更改。`select`和`select-v2`.你愿意对. 做出同样的修改吗？`select-v2`? 🙏

Sure! Thanks for all the guidance. Please bear with me – I'll update the select-v2 next week.

LostElkByte:
There's a bit of an issue. I'm currently investigating it.

LostElkByte:
> There's a bit of an issue. I'm currently investigating it.

Use states.cachedOptions.length instead of modelValue.length for tag-tooltip
conditions to ensure proper DOM timing and prevent teleport target errors.
 
Changes:
- Tag-tooltip v-if condition: modelValue.length → states.cachedOptions.length  
- Collapse count display: modelValue.length → states.cachedOptions.length
 
Fixes: TypeError: Cannot read properties of null (reading 'insertBefore')

packages/components/select-v2/src/select.vue
```diff
@@ -92,16 +92,31 @@
               </div>
 
               <el-tooltip
-                v-if="collapseTags && modelValue.length > maxCollapseTags"
+                v-if="
+                  collapseTags && states.cachedOptions.length > maxCollapseTags
+                "
```
patienry, line 97: The changes here are unrelated to the current PR's topic. If there is indeed an issue, you can create a **new** PR to address the modification.

LostElkByte: I believe these changes are interrelated; otherwise, a timing issue occurs when binding tag-tooltip.appendTo.

Specifically, when using the following configuration:
```vue
<el-select-v2
  :tag-tooltip="{ appendTo: '.select-v2-box' }"
  ...
/>
```
It currently throws an error: 
```
TypeError: Cannot read properties of null (reading 'insertBefore') Failed to locate Teleport target with selector ".select-v2-box".
```

This modification is designed to fix this issue. Upon investigation, I found that the standard Select component uses states.selected.length to control the tooltip's rendering, which naturally avoids this timing conflict. My change aligns SelectV2 with the Select component's implementation logic.


While this can be bypassed by manually setting the selector in onMounted like so:
```ts
const tagBox = ref('')
onMounted(() => {
  tagBox.value = '.select-v2-box'
})
```

I believe the usage of tag-tooltip.appendTo in SelectV2 should be consistent with Select out of the box, without requiring extra boilerplate code from the user.

Could you please evaluate whether I should keep this fix in the current PR or revert it and submit it as a separate PR? 
Thanks for your time and maintenance! Looking forward to your guidance.

docs/en-US/component/select.md
```diff
@@ -271,6 +272,41 @@ select/custom-label
 | options ^(2.11.0) | specify which key of node object is used as the node's children | ^[string] | options  |
 | disabled          | specify which key of node object is used as the node's disabled | ^[string] | disabled |
 
+### tag-tooltip
```
Dsaquel:
```suggestion
### tag-tooltip ^(2.13.3)
```

docs/en-US/component/select-v2.md
```diff
@@ -302,6 +303,41 @@ select-v2/custom-width
 | options   | specify which key of node object is used as the node's children | ^[string] | options  |
 | disabled  | specify which key of node object is used as the node's disabled | ^[string] | disabled |
 
+### tag-tooltip
```
Dsaquel:
```suggestion
### tag-tooltip ^(2.13.3)
```

packages/components/select-v2/src/select.vue
Dsaquel: Looks like the change is right, i don't see any friction between `states.cachedOptions` and `modelValue.length`.

Dsaquel (commented):
Thanks, looks great 😃

Dsaquel (approved):
Thanks!

packages/components/select-v2/src/select.vue
patienry: Okay, since this issue is related to `tag-tooltip`, let's keep it in the current PR. Thank you!

PR #23572, opened 2026-01-30
fix(components): [switch] prevent click event from bubbling

ixyzorg:

Desc:
In scenarios like a card where clicking the card navigates to details, clicking the Switch to toggle a state would also trigger the card click if bubbling is not stopped. This ensures the Switch has independent interaction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved click event handling in the switch component to prevent unintended event propagation to parent elements, ensuring clicks are properly isolated and don't trigger actions outside the component.

### DIFF
--- a/packages/components/switch/src/switch.vue
+++ b/packages/components/switch/src/switch.vue
@@ -1,5 +1,5 @@
 <template>
-  <div :class="switchKls" @click.prevent="switchValue">
+  <div :class="switchKls" @click.prevent.stop="switchValue">
     <input
       :id="inputId"
       ref="input"


packages/components/switch/src/switch.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div :class="switchKls" @click.prevent="switchValue">
+  <div :class="switchKls" @click.prevent.stop="switchValue">
```
rzzf, line 2: I’m not quite sure what issue this change is actually fixing. 
If the goal is simply to prevent the switch click event from bubbling, developers can already achieve this by using the `@click.stop` directive on el-switch. 
Different scenarios usually come with different requirements, so I don’t think this is something we should handle at the component library level.

ixyzorg: Thx

PR #23587, opened 2026-02-02
fix(components): [color-picker] add colorFormat prop validation

E66Crisp:

add type hints for colorFormat prop

befor:
image

after:
image

## Summary by CodeRabbit

* **Refactor**
  * Color picker components now require and validate a constrained set of color format options, improving consistency of color inputs.

* **Documentation**
  * API docs reformatted and updated to list an expanded set of accepted color formats (more granular HEX/RGB variations, HSL, HSV, etc.) and streamlined attribute tables.


packages/components/color-picker-panel/src/color-picker-panel.ts
```diff
@@ -22,7 +25,7 @@ export interface ColorPickerPanelProps {
   /**
    * @description color format of v-model
    */
-  colorFormat?: string
+  colorFormat?: ColorFormat
```
btea: Image
It seems we shouldn't restrict its type; we can leave it as is.

E66Crisp: Thanks for your reply. I’m not imposing restrictions on the types here; instead, I hope there will be prompts for the optional values when using the color-format attribute. According to the color-format attribute in the documentation, we currently only support four type options.

E66Crisp: Sorry for the misunderstanding. You're right that the previous change restricted the supported color formats. 😥
I've updated the implementation to:

- Remove the colorFormats array constant and runtime validation
- Re-export ColorFormat as a union type from TinyColor's ColorFormats for better IDE autocomplete
- Support all TinyColor color formats: 'rgb', 'prgb', 'hex', 'hex3', 'hex4', 'hex6', 'hex8', 'name', 'hsl', 'hsv'

This approach avoids maintaining duplicate definitions while providing clear type hints. Thank you for your review.

Dsaquel (commented):
LGTM, can you update the doc in order to include the new extra format ?

https://github.com/E66Crisp/element-plus/blob/feat-color-picker/docs/en-US/component/color-picker.md?plain=1#L53
https://github.com/E66Crisp/element-plus/blob/feat-color-picker/docs/en-US/component/color-picker-panel.md?plain=1#L64

E66Crisp:
> LGTM, can you update the doc in order to include the new extra format ?
> 
> https://github.com/E66Crisp/element-plus/blob/feat-color-picker/docs/en-US/component/color-picker.md?plain=1#L53 https://github.com/E66Crisp/element-plus/blob/feat-color-picker/docs/en-US/component/color-picker-panel.md?plain=1#L64

Done. Updated the `color-format` type in both ColorPicker and ColorPickerPanel docs to align with `ColorFormats` from `@ctrl/tinycolor`.

Dsaquel (approved):
Thanks!

PR #23605, opened 2026-02-04
chore: update the package URL in npmx

rzzf:
https://npmx.dev/element-plus is redirected to https://npmx.dev/package/element-plus.
I think it would be better to use the final URL directly, to avoid potential 404 issues in the future.

## Summary by CodeRabbit

* **Documentation**
  * Updated a documentation link to point to the correct package resource page.


btea (approved):
👍

PR #23634, opened 2026-02-08
feat(components): [input-otp] new component

### LINKED ISSUE #7865: [Feature Request] [New Component] Add Support for OTP Input
Feature type: **`New Component`**

## Details

- New component name: `OTP Input`

## Description

As there are more and more websites use 2fa, it is important to add a OTP Input component which will make the page much more beautiful. Vuetify has provided this (URL:https://vuetifyjs.com/en/components/otp-input/), it is very useful. Thank you.

### Additional comments

_(empty)_

rzzf:
close #7865

[demos](https://preview-23634-element-plus.surge.sh/en-US/component/input-otp)

## Summary by CodeRabbit

* **New Features**
  * Added a configurable Input OTP component: adjustable length, sizes, three variants, per-field validation, keyboard navigation, paste support, auto-focus, and disabled/readonly states.

* **Documentation**
  * Complete docs and live examples: basic usage, custom length, variants, sizes, disabled/readonly, and validation demos.

* **Localization**
  * Added input-OTP translation keys for 60+ languages.

* **Tests**
  * Comprehensive unit tests covering rendering and interactions.

* **Style**
  * New theme variables and component styles for OTP inputs.


packages/locale/lang/no.ts
```diff
@@ -192,5 +192,8 @@ export default {
       rightArrow: 'Neste bilde',
       indicator: 'Bytt bilde til indeks {index}',
     },
+    inputOTP: {
+      defaultLabel: 'Please enter OTP character {index}', // to be translated
+    },
```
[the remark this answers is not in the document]
rzzf: At the moment, we don’t have contributors dedicated to translations.

Except for English, the content in other languages is currently copied as-is.

We can revisit this once someone helps with the translations in the future.

packages/locale/lang/id.ts
```diff
@@ -194,5 +194,9 @@ export default {
       rightArrow: 'Carousel arrow right', // to be translated
       indicator: 'Carousel switch to index {index}', // to be translated
     },
+    inputOTP: {
+      groupLabel: 'input-otp', // to be translated
+      defaultLabel: 'Please enter OTP character {index}', // to be translated
+    },
```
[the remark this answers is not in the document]
rzzf: @coderabbitai In the upcoming review, please ignore all checks related to i18n files.

packages/components/input-otp/src/input-otp.vue
```diff
@@ -0,0 +1,211 @@
...
+
+const updateModelValue = () => {
+  const value = innerValue.value.join('').slice(0, length.value)
+  if (value !== props.modelValue) {
+    emit(UPDATE_MODEL_EVENT, value)
+    if (value.length === 0 || value.length === length.value) {
+      emit(CHANGE_EVENT, value)
+    }
+  }
+}
```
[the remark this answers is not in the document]
rzzf: This is intentional.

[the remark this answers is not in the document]
rzzf: Remember this.

packages/theme-chalk/src/input-otp.scss
```diff
@@ -0,0 +1,173 @@
...
+      }
+    }
+
+    @include when(disabled) {
+      @include e(input-field) {
+        &:hover::after {
+          background-color: getCssVar('border-color');
+        }
+      }
+    }
```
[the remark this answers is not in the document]
rzzf: When disabled, the input does not receive focus.

xingyixiang:
The input field is missing the error state style.

packages/components/input-otp/src/input-otp.vue
```diff
@@ -0,0 +1,218 @@
...
+    "
+    :aria-labelledby="isLabeledByFormItem ? formItem?.labelId : undefined"
+  >
+    <label
+      v-for="(_, index) in length"
+      :key="index"
+      :class="ns.e('input-field')"
+    >
+      <input
+        v-bind="$attrs"
```
Dsaquel: Adding `$attrs` to sub-input cause `@focus`, `@blur`, etc... trigger multiple time.
Maybe it's better to trigger these events once. WDYT ?

docs/en-US/component/input-otp.md
```diff
@@ -0,0 +1,92 @@
...
+| id                    | Native `id` attribute                                                                                                                      | ^[string]                                             | —          |
+| aria-label ^(a11y)    | Native `aria-label` attribute                                                                                                              | ^[string]                                             | —          |
+
+### InputOtp Events
+
+| Name              | Description                       | Type                                   |
+| ----------------- | --------------------------------- | -------------------------------------- |
+| update:modelValue | triggers when value updates       | ^[Function]`(value: string) => void`   |
+| change            | triggers when the value changes   | ^[Function]`(value: string) => void`   |
+| input             | triggers when input value changes | ^[Function]`(value: string[]) => void` |
```
Dsaquel: We can have a shortcut event called `finish` like vuetify that triggers when the opt field reach the length. WDYT ?

docs/en-US/component/input-otp.md
```diff
@@ -0,0 +1,92 @@
...
+| model-value / v-model | The value of the OTP fields. Since numbers must not have leading zeros, `modelValue` is allowed to be a number only during initialization. | ^[string] / ^[number]                                 | undefined  |
+| length                | The OTP fields length, range [4, 8]                                                                                                        | ^[number]                                             | 6          |
+| validate              | Custom validator function                                                                                                                  | ^[Function]`(char: string, index: number) => boolean` | () => true |
+| inputmode             | Native `inputmode` attribute                                                                                                               | ^[string]                                             | —          |
+| variant               | The variant of the OTP fields                                                                                                              | ^[enum]`'outlined' \| 'filled' \| 'underlined'`       | 'outlined' |
```
Dsaquel: perhaps we can rename this prop `type` instead ? This will fit more the design api of the library.

docs/examples/input-otp/basic.vue
```diff
@@ -0,0 +1,12 @@
+<template>
```
Dsaquel, line 1: After inputing a lot in opt-input my brower start slowing down and then it freeze. Do you have the same problem ?

You can try input/remove quickly in every otp-input and probably you will have the same issue.

packages/components/input-otp/src/input-otp.vue
rzzf: <del>I can’t reproduce a case where it triggers twice. `handleFocus` fires twice because there is also a click event. However, we can explicitly add these two event listeners and emit them manually. I think this would be better than attaching the events through `attrs`.</del>

Each individual input can listen to `focus` and `blur`, which might provide more flexibility. Perhaps we could also introduce a `focusAll`-like event to handle overall `focus` and `blur` states?

docs/examples/input-otp/basic.vue
rzzf: Ah, I was able to reproduce it, `handleFocus` is stuck in an infinite loop.

docs/en-US/component/input-otp.md
rzzf: That makes sense.

docs/en-US/component/input-otp.md
rzzf: This seems to overlap somewhat with the functionality of the `change` event. Perhaps `change` could be renamed to `finish` and triggered only when the input is completed. 
Currently, the `change` event is triggered only when the input is cleared or completed. WDYT?

docs/examples/input-otp/basic.vue
rzzf: fixed by [`fa4dfef`](https://github.com/element-plus/element-plus/pull/23634/commits/fa4dfef3e227723d0182bf34ef0df8ce07607ad7)

packages/components/input-otp/src/input-otp.vue
```diff
@@ -0,0 +1,238 @@
...
+  const pasteData = event.clipboardData?.getData('text') ?? ''
+  if (!pasteData) return
+
+  event.preventDefault()
+
+  const targetIndex = getFirstIndex(index)
+  const chars = castValues(pasteData, targetIndex)
+  const focusIndex = Math.min(targetIndex + chars.length, length.value - 1)
+
+  chars.forEach((char, i) => (innerValue.value[targetIndex + i] = char))
```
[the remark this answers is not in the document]
rzzf: `forEach` does not use the return value, so this will not have any effect. I think this approach is a bit cleaner.

docs/examples/input-otp/basic.vue
Dsaquel: fixed :smiley: :+1:

docs/en-US/component/input-otp.md
Dsaquel: Thanks for noticing it, looking at the events again and maybe we should rethink about it:

`@input` and `@change` should behave like `el-input`.

Right now `@input` emits `string[]` while `@update:modelValue` emits `string` — both fire at the same time but with different output. I'd either drop `@input` or have it emit `string`.

For `@change`, firing only when empty or fully complete is not super ideal — it should fire on blur when focus leaves the component and the value actually changed.

When the user complete the input we add a `@complete/finish` event instead.

To summarize:
- `@update:modelValue` → stays the same
- `@change` → fires on blur, if value changed
- `@complete/finish` → fires when all fields are filled
- `@input` → either drop or emit `string`

What do you think ?

packages/components/input-otp/src/input-otp.vue
Dsaquel: I think `@focus` and `@blur` should stay as a parent-level abstraction, it represents when the user enters/leaves the component as a whole which is the most common use case.

If we need per-input focus tracking we could add a custom event for that, but I don't see a relevant usage for it right now.

What do you think ?

docs/en-US/component/input-otp.md
rzzf: That makes a lot of sense. 👍
I’ve updated it.

packages/components/input-otp/src/input-otp.vue
rzzf: > I think `@focus` and `@blur` should stay as a parent-level abstraction, it represents when the user enters/leaves the component as a whole which is the most common use case.

Yeah, I agree 👍, I’ve updated it.

> If we need per-input focus tracking we could add a custom event for that, but I don't see a relevant usage for it right now.

I think if there are clear use cases in the future, we can reconsider this, and it's also very easy for developers to implement on their own.

packages/components/input-otp/src/input-otp.vue
```diff
@@ -0,0 +1,250 @@
...
+        :inputmode="inputmode"
+        autocomplete="one-time-code"
+        :aria-label="t('el.inputOTP.defaultLabel', { index: index + 1 })"
+        @click="handleFocus"
+        @focus="handleFocus"
+        @blur="handleBlur"
+        @paste="handlePaste($event, index)"
+        @keydown="handleKeydown($event, index)"
+        @input="handleInput($event, index)"
+      />
```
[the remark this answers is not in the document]
rzzf: Maybe we shouldn’t pass `$attrs` down to the inputs.

packages/components/input-otp/src/input-otp.vue
```diff
@@ -0,0 +1,250 @@
...
+
+  const focusIndex = Math.min(targetIndex + (forward ? 1 : 0), length.value - 1)
+
+  innerValue.value[targetIndex] = value
+  inputRefs.value[focusIndex]?.focus()
+  if (focusIndex === length.value - 1) {
+    inputRefs.value[focusIndex]?.select()
+  }
+  updateModelValue()
+}
```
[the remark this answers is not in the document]
rzzf: I’m not able to reproduce this issue at the moment. Could you provide an example?

packages/components/input-otp/src/input-otp.vue
```diff
@@ -0,0 +1,252 @@
...
+      nextInputRef?.focus()
+      break
+    default:
+      preventDefault = false
+  }
+
+  if (preventDefault) {
+    event.preventDefault()
+  }
+}
```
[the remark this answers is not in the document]
rzzf: I wanted to ask — is it possible for you to do a full review at once? It seems like some comments that could have been caught earlier are coming up little by little. 😅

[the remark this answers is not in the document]
rzzf: You can pause the review for now and resume when I ask for it.

Dsaquel:
Looks good to me,

Just one thing:
image
The filled example look not very "stylish", i don't know how it should be supposed to look but maybe we should add some border or box-shadow. I don't really know honestly :laughing:

packages/components/input-otp/src/input-otp.vue
Dsaquel: Yeah with the expose it's maybe enough for now.

Anyway, thank you for you quick update :+1:

docs/en-US/component/input-otp.md
Dsaquel: Thank you, it's perfect :+1:

rzzf:
I’m not entirely sure how to further refine it, so feel free to adjust it if you have better ideas. And happy Chinese New Year! ❤️

Dsaquel:
>I’m not entirely sure how to further refine it, so feel free to adjust it if you have better ideas.

Maybe other reviewers will.

happy Chinese New Year to you too! :heart:

Dsaquel (approved):

packages/locale/lang/zh-cn.ts
```diff
@@ -191,5 +191,9 @@ export default {
       rightArrow: '下一张幻灯片',
       indicator: '幻灯片切换至索引 {index}',
     },
+    inputOTP: {
+      groupLabel: 'input-otp', // to be translated
```
btea:
```suggestion
      groupLabel: '一次性密码输入框', // to be translated
```

packages/components/input-otp/src/input-otp.ts
```diff
@@ -0,0 +1,70 @@
...
+   */
+  readonly?: boolean
+  /**
+   * @description Same as `id` in native input
+   */
+  id?: string
+  /**
+   * @description Whether to trigger form validation
+   */
+  validateEvent?: boolean
```
btea, line 61: Perhaps we can add a `separator` configuration.

rzzf: added

btea (approved):
👍

xingyixiang:
https://vuetifyjs.com/en/components/otp-input/#masked

This function is still very commonly used

rzzf:
> [vuetifyjs.com/en/components/otp-input#masked](https://vuetifyjs.com/en/components/otp-input/#masked)
> 
> This function is still very commonly used

@xingyixiang Thanks for paying attention to this component.

I didn’t intend to introduce too many props in this PR at first, since that would make the review process more difficult. However, I’m certain that I will add a `mask` prop to this component — either in this PR or shortly after it gets merged.

packages/components/input-otp/src/input-otp.vue
```diff
@@ -0,0 +1,268 @@
...
+      </slot>
+    </template>
+  </div>
+</template>
+
+<script lang="ts" setup>
+import { computed, ref, watch } from 'vue'
+import { clamp } from '@vueuse/core'
+import { useLocale, useNamespace } from '@element-plus/hooks'
+import { debugWarn, getEventCode, isFunction, rAF } from '@element-plus/utils'
```
patienry:
```suggestion
import { NOOP, getEventCode, isFunction, rAF } from '@element-plus/utils'
```

packages/components/input-otp/src/input-otp.vue
```diff
@@ -0,0 +1,268 @@
...
+  })
+}
+
+const handleBlur = (event: FocusEvent) => {
+  const { relatedTarget } = event
+  if (!inputRefs.value.includes(relatedTarget as any)) {
+    isFocused.value = false
+    emit('blur', event)
+    if (props.validateEvent) {
+      formItem?.validate?.('blur').catch((err) => debugWarn(err))
```
patienry:
```suggestion
      formItem?.validate?.('blur').catch(NOOP)
```

packages/components/input-otp/src/input-otp.vue
```diff
@@ -0,0 +1,268 @@
...
+  return result
+}
+
+watch(
+  () => props.modelValue,
+  () => {
+    innerValue.value = initialValue.value
+
+    if (props.validateEvent) {
+      formItem?.validate?.('change').catch((err) => debugWarn(err))
```
patienry:
```suggestion
      formItem?.validate?.('change').catch(NOOP)
```

packages/components/input-otp/src/input-otp.vue
patienry: Should this be consistent with #23663? 🤔

keeplearning66 (commented):
Looks great! 👍

packages/components/input-otp/src/input-otp.vue
rzzf: sure

keeplearning66 (approved):
👍

docs/examples/input-otp/custom-length.vue
```diff
@@ -0,0 +1,21 @@
...
+const otp1 = ref('')
+const otp2 = ref('')
+</script>
+
+<style scoped>
+.demo-input-otp {
+  display: flex;
+  flex-direction: column;
+  gap: 20px;
+}
```
micaiguai: Perhaps we need avoid using `flex-gap` in all example. If we don't support Chrome <= 83, you can ignore this suggestion.

Refer: https://caniuse.com/flexbox-gap


```suggestion
.demo-input-otp {
  display: grid;
  gap: 20px;
}
```

micaiguai:
Perhaps we need to restrict input to numeric values 0-9 by default.

Refer: https://vuetifyjs.com/en/components/otp-input/

packages/components/input-otp/src/input-otp.vue
```diff
@@ -0,0 +1,283 @@
...
+
+import type { InputOtpProps } from './input-otp'
+
+defineOptions({
+  name: 'ElInputOtp',
+})
+
+const props = withDefaults(defineProps<InputOtpProps>(), {
+  length: 6,
+  validator: () => true,
```
micaiguai:
```suggestion
  validator: () => (char: string) => /^\d$/.test(char),
```

rzzf:
> Perhaps we need to restrict input to numeric values 0-9 by default.
> 
> Refer: [vuetifyjs.com/en/components/otp-input](https://vuetifyjs.com/en/components/otp-input/)

OTP may consist of digits, letters, or a mix of uppercase and lowercase characters depending on the use case. Defaulting to digits would make the component more business-specific and reduce its generality.

If you want, you can: https://element-plus.org/en-US/guide/custom-defaults

micaiguai:
> > Perhaps we need to restrict input to numeric values 0-9 by default.
> > Refer: [vuetifyjs.com/en/components/otp-input](https://vuetifyjs.com/en/components/otp-input/)
> 
> OTP may consist of digits, letters, or a mix of uppercase and lowercase characters depending on the use case. Defaulting to digits would make the component more business-specific and reduce its generality.
> 
> If you want, you can: https://element-plus.org/en-US/guide/custom-defaults

Thanks for the explanation. What you said makes a lot of sense.

packages/components/input-otp/src/input-otp.vue
```diff
@@ -0,0 +1,284 @@
...
+
+const separators = computed(() => {
+  const { separator } = props
+  const _separator = isFunction(separator) ? separator : () => separator
+  return Array.from({ length: props.length - 1 }, (_, i) => _separator(i))
+})
+
+const innerValue = ref<string[]>(initialValue.value)
+const isFocused = ref(false)
+const inputRefs = useTemplateRef('inputRefs')
```
rzzf: Waiting for #23814 to be merged.

docs/examples/input-otp/sizes.vue
```diff
@@ -0,0 +1,38 @@
...
+      <p>Small</p>
+      <el-input-otp v-model="otp1" size="small" />
+    </div>
+    <div>
+      <p>Default</p>
+      <el-input-otp v-model="otp2" size="default" />
+    </div>
+    <div>
+      <p>Large</p>
+      <el-input-otp v-model="otp3" size="large" />
```
micaiguai:
```suggestion
      <p>Large</p>
      <el-input-otp v-model="otp1" size="large" />
    </div>
    <div>
      <p>Default</p>
      <el-input-otp v-model="otp2" size="default" />
    </div>
    <div>
      <p>Small</p>
      <el-input-otp v-model="otp3" size="small" />
```

micaiguai (commented):
Perhaps we maintain a consistent order.
- Follow the unified order specified in the doc: [There are three sizes available: `large`, `default`, and `small`.](https://github.com/element-plus/element-plus/pull/23634/changes#diff-5c3e1539519af8e64d6706540391c617af591a8aa1f1fccc69a4e1431923c472R40)
- Other components already display the large > default > small order.

kooriookami:
Cool Component!

packages/components/input-otp/src/input-otp.vue
rzzf: This requires devs to upgrade Vue to ^3.5.0, which I think is a breaking change.
But #23814 doesn’t seem ready yet, so I’ve reverted this change for now.

PR #23663, opened 2026-02-15
refactor(components): remove console noises from failed form validation

### LINKED ISSUE #19591: [Component] [form-item, input] Validation errors trigger unhelpful console warnings
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.13`
- Element Plus Version: `2.9.2`
- Browser / OS: `Chrome 131.0.6778.205 / macOS 15.2`
- Build Tool: `Vite`

### Related Component

- `el-form-item`
- `el-input`

### Reproduction Link

[Github Repo](https://github.com/EwenDC/Element-Plus-Validate-Bug)

### Steps to reproduce

1. Clone the repository.
2. Run the commands `npm i` then `npm run dev` in the cloned repository.
3. Navigate to `http://localhost:5173/` in a web browser.
4. Open the browser console.
5. Clear all text from the "Greeting" input then click away.

### What is Expected?

No message appears in the browser console.

Element Plus (not Async-Validator) is logging the "fields" array from the Async-Validator response to the browser console as a warning, with no associated message.

### Additional comments

While this issue only occurs when running the application using Element+ in development mode, it is still not ideal. It creates unnecessary noise in the browser console, with warning-level messages that are in reality debug-level, and that are useless to developers of consuming apps. Element+ should not be unnecessarily logging to the console in situations where functionality is working as expected (especially with warning-level messages).

The provided reproduction uses the `el-input` component as an example, but the console logging code exists in every component that supports `el-form-item`. It was added to all of them in PR #5401, however due to how the code was written at the time, this did not result in any console messages. When the form code was refactored in PR #6610, this triggered the previously dormant logging code to start firing.

Dsaquel:
close #19591
ref https://github.com/element-plus/element-plus/pull/5401#pullrequestreview-901033457

image

If a form validation failed warnings are thrown, it's actually only noises, not structured and they are disrupting the developer experience.

Also the ui feedback already show the error.

## Summary by CodeRabbit

* **Refactor**
  * Silenced debug-level warnings from form validation across multiple input and selection components. Validation behavior and UX remain the same, but console warnings for validation failures are no longer emitted, producing quieter runtime logs and less noisy developer consoles.


rzzf (approved):

keeplearning66 (approved):
👍

btea:
I also feel that printing this doesn't seem very useful. 

However, removing it directly feels too abrupt. Could we add a global configuration so users can choose whether to print this verification information, and then remove it in a later major version?

Dsaquel:
Adding an API entry to enable the warning is too overkill for a console message that nobody want.

We can merge the PR in minor if you want.

btea:
> Adding an API entry to enable the warning is too overkill for a console message that nobody want.
> 
> We can merge the PR in minor if you want.

That make sense.

btea (approved):

packages/components/form/src/form.vue
```diff
@@ -199,7 +199,7 @@ watch(
   () => props.rules,
   () => {
     if (props.validateOnRuleChange) {
-      validate().catch((err) => debugWarn(err))
```
thinkasany, line 242: In form.vue (line 202), validateOnRuleChange calls validate().catch(NOOP).
The issue is that this swallows both categories of rejections:

Expected, structured validation failures (ValidateFieldsError)
Real runtime exceptions (Error, e.g. TypeError / ReferenceError)

Image

btea: I'm not sure if I've understood your question correctly.

 Even if 'e' is a `TypeError`, it still satisfies the condition `e instanceof Error`. 🤔

PR #23670, opened 2026-02-18
fix(theme-chalk): [popper] don't cascade bg variable

Dsaquel:
Removed `--el-fill-color-blank` from `.el-popper` since it cascades to all popper-based components (tooltips, popovers, selects, etc.), making buttons in dark tooltips invisible. Now only popover and tree-select set it locally where needed.

fix #23555

## Summary by CodeRabbit

* **Style**
  * Updated CSS variable assignments in theme components to improve styling consistency across popover, popper, and tree-select elements. These changes refine the internal theme variable structure.


btea:
This PR was polluted by an action, so I decided to shut it down.

PR #23671, opened 2026-02-18
fix(style): [form-item] remove trailing inline margin

### LINKED ISSUE #5212: [Feature Request] the last `form-item` of inline-form should not have a `margin-right` value
### Existing Component
Yes

### Component Name
form-item

### Description
when we use multiple `form-item` inside inline-form like the code belows:
```
<el-form inline label-width="auto" size="small">
      <el-form-item label="教室:" >
        <el-input></el-input>
      </el-form-item>
            <el-form-item label="塾長:" >
        <el-input></el-input>
      </el-form-item>
    </el-form>
```
the last `form-item` will has `margin-right: 10px` from source code`.el-form--inline .el-form-item {
  margin-right: 10px;
}`.
image

### Expected
The last `form-item` will not have a `margin-right` rule. 
What about change it to below? I am glad to contribute it if u think it's a good idea.
```
.el-form--inline .el-form-item:not(:last-of-type) {
  margin-right: 10px ;
}
```

Dsaquel:
fix #5212

before
image

after

image

## Summary by CodeRabbit

* **Style**
  * Improved visual spacing in inline form layouts by removing excess margin from the final form item.

### DIFF
--- a/packages/theme-chalk/src/form.scss
+++ b/packages/theme-chalk/src/form.scss
@@ -10,6 +10,10 @@
       display: inline-flex;
       vertical-align: middle;
       margin-right: 32px;
+
+      &:last-child {
+        margin-right: 0;
+      }
     }
 
     &.#{$namespace}-form--label-top {


rzzf (approved):

btea:
Style changes can affect layout. For example, I often add query or reset buttons to forms in my projects. Therefore, I think this should be a breaking change.

Dsaquel:
>Style changes can affect layout. For example, I often add query or reset buttons to forms in my projects. Therefore, I think this should be a breaking change.

I understand, should we set it in minor too ?

btea:
> I understand, should we set it in minor too ?

Yes.

keeplearning66 (approved):
👍

warmthsea (approved):

PR #23694, opened 2026-02-27
feat(components): [popover] expose hide() through slot

ZacharyBear:

close #23690

## Summary by CodeRabbit

* **New Features**
  * The Popover's default slot now exposes a hide function, allowing custom slot content to close the popover programmatically.

* **Documentation**
  * Popover docs updated: Slots table adds a Type column and clarifies the default slot description to reflect the exposed slot properties.

### DIFF
--- a/docs/en-US/component/popover.md
+++ b/docs/en-US/component/popover.md
@@ -108,7 +108,7 @@ popover/directive-usage
 
 | Name              | Description                                                                | Type                          |
 | ----------------- | -------------------------------------------------------------------------- | ----------------------------- |
-| default ^(2.13.3) | content of popover                                                         | ^[object]`{hide: () => void}` |
+| default | content of popover, version ^(2.13.3) and later can receive the hide parameter.                                                         | ^[object]`{hide: () => void}` |
 | reference         | HTML element that triggers popover, only a single root element is accepted | -                             |
 
 ### Events
--- a/docs/en-US/component/popover.md
+++ b/docs/en-US/component/popover.md
@@ -108,7 +108,7 @@ popover/directive-usage
 
 | Name      | Description                                                                     | Type                          |
 | --------- | ------------------------------------------------------------------------------- | ----------------------------- |
-| default   | content of popover, version ^(2.13.3) and later can receive the hide parameter. | ^[object]`{hide: () => void}` |
+| default   | content of popover, version ^(2.13.4) and later can receive the hide parameter. | ^[object]`{hide: () => void}` |
 | reference | HTML element that triggers popover, only a single root element is accepted      | -                             |
 
 ### Events
--- a/docs/en-US/component/popover.md
+++ b/docs/en-US/component/popover.md
@@ -106,10 +106,10 @@ popover/directive-usage
 
 ### Slots
 
-| Name              | Description                                                                | Type                          |
-| ----------------- | -------------------------------------------------------------------------- | ----------------------------- |
-| default | content of popover, version ^(2.13.3) and later can receive the hide parameter.                                                         | ^[object]`{hide: () => void}` |
-| reference         | HTML element that triggers popover, only a single root element is accepted | -                             |
+| Name      | Description                                                                     | Type                          |
+| --------- | ------------------------------------------------------------------------------- | ----------------------------- |
+| default   | content of popover, version ^(2.13.3) and later can receive the hide parameter. | ^[object]`{hide: () => void}` |
+| reference | HTML element that triggers popover, only a single root element is accepted      | -                             |
 
 ### Events
 
--- a/docs/en-US/component/popover.md
+++ b/docs/en-US/component/popover.md
@@ -106,10 +106,10 @@ popover/directive-usage
 
 ### Slots
 
-| Name      | Description                                                                |
-| --------- | -------------------------------------------------------------------------- |
-| default   | text content of popover                                                    |
-| reference | HTML element that triggers popover, only a single root element is accepted |
+| Name              | Description                                                                | Type                          |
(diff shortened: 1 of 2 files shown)


docs/en-US/component/popover.md
```diff
@@ -106,10 +106,10 @@ popover/directive-usage
 
 ### Slots
 
-| Name      | Description                                                                |
-| --------- | -------------------------------------------------------------------------- |
-| default   | text content of popover                                                    |
-| reference | HTML element that triggers popover, only a single root element is accepted |
+| Name              | Description                                                                | Type                          |
+| ----------------- | -------------------------------------------------------------------------- | ----------------------------- |
+| default ^(2.13.3) | content of popover                                                         | ^[object]`{hide: () => void}` |
```
btea:
```suggestion
| default | content of popover, version ^(2.13.3) and later can receive the hide parameter.                                                         | ^[object]`{hide: () => void}` |
```
It would be more accurate to include the version number in the description.

ZacharyBear: Actually the parameter I've added in the first commit, so it will be released with this description.
Unnecessary to describe it as a feature will be published in the future.

btea (approved):

ZacharyBear:
Shall I add a test case to this slot

btea:
That would be better.

keeplearning66 (commented):
Thank you for your contribution! Based on the description in #23690, would directly using `Popconfirm` meet your needs?

ZacharyBear:
> Thank you for your contribution! Based on the description in #23690, would directly using `Popconfirm` meet your needs?

image

I think popconfirm couldn't achieve my needs, it does provided `cancel()` and `confirm()` prop to `actions` slot, but I can't write the content in <template>, I must to write vnode and pass it into `title` prop.

Another reason to choose [popover] is that it offers more APIs.

patienry:
> I think popconfirm couldn't achieve my needs, it does provided cancel() and confirm() prop to actions slot, but I can't write the content in , I must to write vnode and pass it into title prop.

Do you mean you want to have both `title` and `content` at the same time? The `title` prop only supports `string` type.

> Another reason to choose [popover] is that it offers more APIs.

Actually, some props in `Popconfirm` can be referenced from `Popover`.

00D2D8A4-E657-49ec-B14D-009A67A3AF01

ZacharyBear:
> > I think popconfirm couldn't achieve my needs, it does provided cancel() and confirm() prop to actions slot, but I can't write the content in , I must to write vnode and pass it into title prop.
> 
> Do you mean you want to have both `title` and `content` at the same time?
> 
> > Another reason to choose [popover] is that it offers more APIs.
> 
> Actually, some props in `Popconfirm` can be referenced from `Popover`.
> 
> 00D2D8A4-E657-49ec-B14D-009A67A3AF01

In this sample, popover is singleton and controlled by the `visible` prop.

In my situation, popover has been multiple duplicate rendered in a list. Using the `ref` value to controll their visibility will increase the logic complixety: each popover needs a value to control visibility. Or I have to package the Popover into a single component.

I just want to **hide** popover from inside, by adding this context prop to `default` slot is greatly helpful. And I believe it could be used in more situation.

patienry:
> > I think popconfirm couldn't achieve my needs, it does provided cancel() and confirm() prop to actions slot, but I can't write the content in , I must to write vnode and pass it into title prop.
> 
> Do you mean you want to have both `title` and `content` at the same time? The `title` prop only supports `string` type.

Thank you for your explanation. I just feel that the behavior of 'manually closing the popup through a button inside the popup' is more in line with `Popconfirm`. So based on my previous comment, I would like to confirm whether the existing `Popconfirm` truly cannot meet your needs. If it indeed cannot, then we can consider extending the component's functionality. 😃

ZacharyBear:
> > > I think popconfirm couldn't achieve my needs, it does provided cancel() and confirm() prop to actions slot, but I can't write the content in , I must to write vnode and pass it into title prop.
> > 
> > 
> > Do you mean you want to have both `title` and `content` at the same time? The `title` prop only supports `string` type.
> 
> Thank you for your explanation. I just feel that the behavior of 'manually closing the popup through a button inside the popup' is more in line with `Popconfirm`. So based on my previous comment, I would like to confirm whether the existing `Popconfirm` truly cannot meet your needs. If it indeed cannot, then we can consider extending the component's functionality. 😃

Here is my use-case, here is my ui looks like:
image

I'm implement custom [color-picker] by [color-picker-panel] and [popover]. For saving spaces, I put close button into `footer` slot of  [color-picker-panel], neither`footer` slot of [popconfirm], it occupies a whole line.

If use [popconfirm] the close button have to be moved to `footer` of [popconfirm].

docs/en-US/component/popover.md
```diff
@@ -106,10 +106,10 @@ popover/directive-usage
 
 ### Slots
 
-| Name      | Description                                                                |
-| --------- | -------------------------------------------------------------------------- |
-| default   | text content of popover                                                    |
-| reference | HTML element that triggers popover, only a single root element is accepted |
+| Name      | Description                                                                     | Type                          |
+| --------- | ------------------------------------------------------------------------------- | ----------------------------- |
+| default   | content of popover, version ^(2.13.3) and later can receive the hide parameter. | ^[object]`{hide: () => void}` |
```
patienry:
```suggestion
| default   | content of popover, version ^(2.13.4) and later can receive the hide parameter. | ^[object]`{hide: () => void}` |
```

keeplearning66 (commented):
Okay, thank you for your patient explanation. Just need to update the version number.

keeplearning66 (approved):
Thank you!

PR #23719, opened 2026-03-03
chore(project): remove `shamefullyhoist`

Dsaquel:
Context: I prepared a migration of template ref to be using only `useTemplateRef` because vue-tsc no longer support the old way after `3.1.5` resulting to typecheck errors.

The problem is that `useTemplateRef` is not being recognized at all in components directory.
In fact, in the docs vitepress install vue 3.4 which doesn't have `useTemplateRef`, then hoist vue at the root node_module level and override the vue 3.5, after debugging 
~~i found that replacing `shamefullyhoist` to `nodeLinker: hoisted` solve this problem.~~
We found with @btea that removing both is the ideal solution.

## Summary by CodeRabbit

* **Chores**
  * Raised minimum Node.js requirement to >=22, updated package manager hint to pnpm v11, adjusted workspace build/linker settings, and updated dev dependency entries for Vue tooling and testing.
* **CI / Tests**
  * Updated CI workflows and test matrix to use Node 22 (and Node 24 where applicable); removed Node 20 from matrices.
* **Bug Fixes**
  * Minor input height calculation robustness improvement in the Cascader component.


tsconfig.base.json
```diff
@@ -26,7 +26,6 @@
       "@element-plus/locale": ["packages/locale"],
       "@element-plus/locale/*": ["packages/locale/*"],
       "element-plus": ["packages/element-plus"]
-    },
-    "preserveSymlinks": true
```
Dsaquel, line 30: Followed this [documentation](https://pnpm.io/typescript#do-not-preserve-symlinks).

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "packageManager": "pnpm@10.18.2",
+  "packageManager": "pnpm@11.0.0-alpha.12",
```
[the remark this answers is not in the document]
Dsaquel: The documentation will be in our head.

tsconfig.base.json
btea: If I understand correctly, it seems that setting `node-linker: hoisted` should be used in conjunction with `"preserveSymlinks": true`. 🤔

Dsaquel: It's actually confusing but what i understand is that in both cases (`nodeLinker: hoisted`, or default pnpm settings) should not have `preserveSymlinks: true`.

pnpm-workspace.yaml
```diff
@@ -3,6 +3,11 @@ packages:
   - docs
   - play
   - internal/*
+allowBuilds:
+  '@parcel/watcher': true
+  esbuild: true
+  puppeteer: true
+  vue-demi: true
```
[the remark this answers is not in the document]
Dsaquel: After installing multiple times es5-ext and fsevents are not require anymore in onlyBuiltDependencies that's why i didn't migrated them.

pnpm-workspace.yaml
```diff
@@ -4,6 +4,12 @@ packages:
   - play
   - internal/*
 
+allowBuilds:
+  '@parcel/watcher': true
+  esbuild: true
+  puppeteer: true
+  vue-demi: true
```
btea, line 11: We've upgraded `vueuse`, so `vue-demi` should no longer be needed.

btea (commented):
Since pnpm11 is still a alpha version, is it enough for us to just upgrade pnpm to 10.31.0 for now?

pnpm-workspace.yaml
Dsaquel: I still need vue-demi when i do pnpm install an we already have it. https://github.com/element-plus/element-plus/blob/dev/pnpm-workspace.yaml#L29

Dsaquel:
>Since pnpm11 is still a alpha version, is it enough for us to just upgrade pnpm to 10.31.0 for now?

Upgrading to 10.31.0 doesn't solve this problem unfortunately:
>Upgrading pnpm because i had a really weird bug where every install made some packages to be deleted (likely the same issue https://github.com/pnpm/pnpm/issues/8871).

pnpm-workspace.yaml
btea: Are there any other packages that depend on vue-demi?

Dsaquel: `pnpm why vue-demi --recursive` retrieve it yes.

btea: There might be a version issue. I ran it with version 10.18.2, but it didn't output any information.

btea:
The link you provided seems to be for installing dependencies. Did you also download versions compatible with other platforms? When you use pnpm10, do you install other platform versions of Rolldown?

pnpm-workspace.yaml
Dsaquel: In this branch ? if i remove vue-demi:
image

Dsaquel:
>The link you provided seems to be for installing dependencies.

It's wanted, if i try to `pnpm install` the second times in `10.31.0` in this branch every runs will do remove packages for no reason:
image

>Did you also download versions compatible with other platforms? When you use pnpm10, do you install other platform versions of Rolldown?

Yes it install other versions in pnpm 10. Same for pnpm11.
image

btea:
I encountered the same situation as you when installing using 10.31.0 on the current branch.

Dsaquel:
Then it confirms pnpm 11 solve the issue.

btea:
The most reasonable solution is to remove `"preserveSymlinks": true` from the `tsconfig.base.json` file, and then configure `"preserveSymlinks": true` to override it when building and generating ts type files.

Dsaquel:
>The most reasonable solution is to remove "preserveSymlinks": true from the tsconfig.base.json file, and then configure "preserveSymlinks": true to override it when building and generating ts type files.

This was indeed the right fix, thanks.

btea (approved):
👍

packages/components/affix/src/affix.ts
```diff
@@ -7,14 +7,22 @@ import {
...
 import type Affix from './affix.vue'
 
+type ZindexType =
+  | 'inherit'
+  | 'initial'
+  | 'revert'
+  | 'unset'
+  | 'revert-layer'
+  | 'auto'
+  | number
```
rzzf: Just curious, why do we need to manually list the possible types?

btea: image

mark.

btea:
> The most reasonable solution is to remove `"preserveSymlinks": true` from the `tsconfig.base.json` file, and then configure `"preserveSymlinks": true` to override it when building and generating ts type files.

Inconsistencies between the external configuration and the `preserveSymlinks` configuration in `rolldown-plugin-dts` can cause type inference for `__VLS_base` in the generated *.vue.d.ts file to fail and become `any`.

Therefore, the latest commit seems more reasonable and healthier.

packages/utils/typescript.ts
```diff
@@ -92,3 +92,33 @@ type Path<T> =
...
+
+export type ObjectFit =
+  | 'fill'
+  | 'contain'
+  | 'cover'
+  | 'none'
+  | 'scale-down'
+  | Globals
+
+export type ZindexType = Globals | 'auto' | number
```
rzzf:
```suggestion
export type ZIndexType = Globals | 'auto' | number
```

This appears to be a typo.

rzzf (approved):
lgtm 👍

PR #23734, opened 2026-03-06
fix: issue numbers #23731 and #23705

Arcane-Cloud:


keeplearning66 (commented):
Thank you for your contribution! However, please avoid addressing multiple issues in a single PR. Also, there are some unrelated changes in the current PR, such as `popover`.

PR #23814, opened 2026-03-15
refactor(components): refs to useTemplateRef

Dsaquel:
https://github.com/vuejs/language-tools/issues/5815#issuecomment-3622337201

## Summary by CodeRabbit

* **Refactor**
  * Standardized template-ref usage across components and hooks, moving ownership of DOM/component refs into templates and passing them into composables to simplify ref management, focus/interaction, and sizing logic.

* **Chores**
  * Removed/relaxed vue-tsc dev dependency and updated package Vue peer requirement to a newer minor range.

### DIFF
--- a/packages/components/carousel/src/carousel-item.vue
+++ b/packages/components/carousel/src/carousel-item.vue
@@ -12,7 +12,7 @@
 </template>
 
 <script lang="ts" setup>
-import { computed, unref } from 'vue'
+import { computed, unref, useTemplateRef } from 'vue'
 import { useNamespace } from '@element-plus/hooks'
 import { useCarouselItem } from './use-carousel-item'
 import { CAROUSEL_ITEM_NAME } from './constants'
@@ -30,9 +30,10 @@ const props = withDefaults(defineProps<CarouselItemProps>(), {
 })
 const ns = useNamespace('carousel')
 
+const carouselItemRef = useTemplateRef<HTMLElement>('carouselItemRef')
+
 // inject
 const {
-  carouselItemRef,
   active,
   animating,
   hover,
@@ -43,7 +44,7 @@ const {
   scale,
   ready,
   handleItemClick,
-} = useCarouselItem(props)
+} = useCarouselItem(props, carouselItemRef)
 
 const itemKls = computed(() => [
   ns.e('item'),
--- a/packages/components/carousel/src/carousel-item.vue
+++ b/packages/components/carousel/src/carousel-item.vue
@@ -30,7 +30,7 @@ const props = withDefaults(defineProps<CarouselItemProps>(), {
 })
 const ns = useNamespace('carousel')
 
-const carouselItemRef = useTemplateRef<HTMLElement>('carouselItemRef')
+const carouselItemRef = useTemplateRef('carouselItemRef')
 
 // inject
 const {
--- a/packages/components/carousel/src/constants.ts
+++ b/packages/components/carousel/src/constants.ts
@@ -1,4 +1,4 @@
-import type { InjectionKey, Ref, VNode } from 'vue'
+import type { InjectionKey, Ref, TemplateRef, VNode } from 'vue'
 import type { CarouselItemProps } from './carousel-item'
 
 export type CarouselItemStates = {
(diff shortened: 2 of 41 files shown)


rzzf:
Since `useTemplateRef` is an API introduced in Vue 3.5.0, we should update the peerDependencies accordingly.

packages/components/carousel/src/carousel-item.vue
```diff
@@ -30,9 +30,10 @@ const props = withDefaults(defineProps<CarouselItemProps>(), {
 })
 const ns = useNamespace('carousel')
 
+const carouselItemRef = useTemplateRef<HTMLElement>('carouselItemRef')
```
rzzf: It’s worth noting that `useTemplateRef` can automatically infer types, including for components.
In my VS Code (Vue (Official) 3.2.5), manually declaring the type actually causes the "go to definition" feature to stop working. 
Do you also experience a loss of navigation when explicitly specifying types?

https://github.com/user-attachments/assets/7e894d62-baea-4b8d-8406-0d098e41602a

packages/components/carousel/src/constants.ts
```diff
@@ -20,7 +20,7 @@ export type CarouselItemContext = {
 }
 
 export type CarouselContext = {
-  root: Ref<HTMLElement | undefined>
+  root: Readonly<Ref<HTMLElement | null>>
```
rzzf: Perhaps we can just use `TemplateRef<T>`, which would be a bit simpler.

packages/components/cascader/src/cascader.vue
```diff
@@ -354,6 +355,8 @@ const { isComposing, handleComposition } = useComposition({
   },
 })
 
+const wrapperRef = useTemplateRef<HTMLElement>('wrapperRef')
+
 const tooltipRef = ref<TooltipInstance>()
```
micaiguai, line 360: Just curious. Why not refactor to use `useTemplateRef` at here? Also, `anchor`, `autocomplete` and `breadcrumb` etc. aren't refactored. Is this a preliminary submission with `refs` to be handled later?

Dsaquel: The purpose is only for dodging the errors that will be reported by `vue-tsc` above `3.5.0`.

I don't want to bloat the PR by updating all refs to useTemplateRef. It can be done separately i think.

packages/components/carousel/src/carousel-item.vue
Dsaquel: I'm experiencing the same issue, after the refactor you suggested above we don't have this anymore but it's maybe a bug from typescript or vue langue tools.

packages/components/carousel/src/constants.ts
Dsaquel: Didn't know about that, it's really convenient.

packages/components/cascader/src/cascader.vue
micaiguai: Thanks for your reply, I understand now.

packages/components/input-tag/src/composables/use-input-tag.ts
```diff
@@ -205,8 +214,9 @@ export function useInputTag({ props, emit, formItem }: UseInputTagOptions) {
     inputRef.value?.blur()
   }
 
-  const { wrapperRef, isFocused } = useFocusController(inputRef, {
+  const { isFocused } = useFocusController(inputRef, {
     disabled,
+    wrapperRef: externalWrapperRef,
```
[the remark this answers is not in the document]
Dsaquel: `TemplateRef` is just an alias to `Readonly<ShallowRef<T | null>>` [ref](https://github.com/vuejs/core/blob/81615d398a89beeccfe56a4a96fd2fba0c6fb37b/packages/runtime-core/src/helpers/useTemplateRef.ts#L8).

rzzf (approved):

packages/components/dialog/src/use-dialog.ts
```diff
@@ -137,7 +132,12 @@ export const useDialog = (
       return config
     }
 
-    return baseConfig
+    return {
+      name: transition,
+      onAfterEnter: afterEnter,
+      onBeforeLeave: beforeLeave,
+      onAfterLeave: afterLeave,
+    }
```
rzzf, line 140: I didn’t notice that this already resolves the issue in #23935. 😅

thinkasany:
@codex review

packages/components/dialog/src/use-dialog.ts
Dsaquel: I didn't know either :rofl:

rzzf:
This change requires devs to upgrade Vue to ^3.5.0, so I think it should be considered a breaking change.

PR #23851, opened 2026-03-19
fix(components): [popconfirm]修复气泡弹窗在TSX中属性提示问题_0319

Due07:

- [ x ] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).
- [ x ] Make sure you are merging your commits to `dev` branch.
- [ x ] Add some descriptions and refer to relative issues for your PR.

🫓 修复popconfirm 在tsx上 `placement` 属性不存在提示问题

Snipaste_2026-03-19_15-49-46

## Summary by CodeRabbit

* **New Features**
  * Popconfirm now supports additional tooltip-style options for finer control of appearance and behavior (excluding content settings), enabling more flexible positioning, interaction, and display tuning.
  * Added an optional placement option to customize where the confirmation popover appears relative to its trigger.


packages/components/popconfirm/src/popconfirm.ts
```diff
@@ -13,6 +13,7 @@ import type {
   ElTooltipTriggerProps,
 } from '@element-plus/components/tooltip'
 import type { IconPropType } from '@element-plus/utils'
+import type { Placement } from '@element-plus/components/popper'
 import type Popconfirm from './popconfirm.vue'
 
 export interface PopconfirmProps {
```
micaiguai:
```suggestion
export interface PopconfirmProps extends Omit<
  ElTooltipContentProps,
  'popper-class' | 'popper-style' | 'fallback-placements'
> {
```

micaiguai:
Perhaps we can directly extend `UseTooltipProps,` as props such as `append-to` and others also require TypeScript inline suggestions.
With this approach, props like `content` and `rawContent` need to be omitted (there may be additional props that require omission). For reference: https://element-plus.org/en-US/component/popconfirm#basic-usage.

Due07:
> 或许我们可以直接扩展`UseTooltipProps,`诸如 `popconfirm` 之类的属性，因为`append-to`其他属性也需要 TypeScript 内联建议。 采用这种方法，需要省略诸如 `popconfirm``content`和`popconfirm` 之类的属性（可能还有其他需要省略的属性） 。参考： https://element-plus.org/en-US/component/popconfirm#basic-usage`rawContent`[](https://element-plus.org/en-US/component/popconfirm#basic-usage)

您说的没错，是我没有仔细阅读其他组件文档内容😅，
我已修正使用 `UseTooltipProps`  方便后期基于Tooltip 迭代新属性的同步 🆒

rzzf:
I think declaring them in props alone is not enough — we still need to manually pass them to the Tooltip, and also explicitly define their default values.

<hr />

I've been following this RFC: https://github.com/vuejs/rfcs/discussions/479 

To be honest, when wrapping components, a natural best practice from a JavaScript perspective is to forward props to the child component via `$attrs`, while using the wrapper’s own props only for extension. 
However, in TypeScript, we often have to go against this intuition—the main purpose of defining props becomes getting proper type inference. The downside is that we have to manually maintain all props of the child component (including default values, etc.), which is essentially duplicating the entire props definition. 

For now, though, this seems to be an unavoidable trade-off.

One approach is to use [fallthroughAttributes](https://github.com/vuejs/language-tools/wiki/Vue-Compiler-Options#fallthroughattributes-v210), but it comes with a relatively high performance cost.

PR #23861, opened 2026-03-20
fix(components): #23858 fix dialog z-index when modal-penetrable enabled

LavenderDuskGlow:
场景：如果开启modalPenetrable，打开多个dialog 后段集那个dialog 那个dialog 的层级提升在最上面，如有不足请海涵，谢谢

## Summary by CodeRabbit

* **New Features**
  * Dialog component now features improved stacking management for better handling of overlapping and nested dialogs
  * Dynamic index modification system optimizes layering behavior and ensures smooth user interactions in complex scenarios
  * Enhanced support for penetrable modals with intelligent z-order management


packages/components/dialog/src/dialog.vue
```diff
@@ -56,6 +56,7 @@
               :title="title"
               :aria-level="headerAriaLevel"
               @close="handleClose"
+              @click="onModifyIndex"
```
snowbitx, line 59: Hi. I’d suggest using `mousedown` instead of `click` here, since bringing a dialog to front is more of an activation behavior and should happen as interaction starts

packages/components/dialog/src/use-dialog.ts
```diff
@@ -47,6 +47,7 @@ export const useDialog = (
   const rendered = ref(false) // when destroyOnClose is true, we initialize it as false vise versa
   const zIndex = ref(props.zIndex ?? nextZIndex())
   const closing = ref(false)
+  const historyIndex = ref(0)
```
snowbitx, line 50: It looks like historyIndex is not being used at the moment. Was some related code accidentally left out of this PR? If not, we could probably remove this unused variable.

packages/components/dialog/src/use-dialog.ts
```diff
@@ -229,6 +230,13 @@ export const useDialog = (
     }
   }
 
+  function onModifyIndex() {
+    if (props.modalPenetrable) {
```
snowbitx, line 234: I wonder if we should add a few extra guards here so this change stays scoped to the intended `modalPenetrable` scenario and does not unintentionally affect existing dialog behavior, especially for regular modal, fullscreen, or user-controlled `zIndex` cases.

Also, based on the documentation, I’d suggest restricting this with the `modal` prop as well.

Image

Issue #6796, opened 2022-03-24
[Feature Request] [New Component] [] Virtualized Cascader 


geekFeier:
Feature type: **`New Component`**

## Details

- New component name: `Cascader-v2`

## Description

Cascader has a large amount of data, which causes the page to be very stuck. I hope to support virtualization as soon as possible.

### Additional comments

_(empty)_

runbrick:
同求

Otto-J:
有规划吗？

msidolphin:
> 有规划吗？

已经完成大部分了，但是近期有些繁忙，后续有空继续跟进

geekFeier:
您好，这个问题还在跟进吗

shawDirectUser:
hello~ does this issue has any update?

troy351:
any update?

DavidKitano:
Is there any update?

PR #23877, opened 2026-03-23
feat(components): [cascader] support virtual-scroll & fitInputWidth

zhixiaotong:

rel: #23778 

This discussion was raised before, but it didn't receive much attention. Because some data in the project is displayed using Cascader, when the data volume is large, the excessive number of DOM elements causes the page to crash, preventing the testing process from passing.

## Summary by CodeRabbit

* **New Features**
  * Virtual scroll for Cascader panels and suggestion lists; new props to control virtualization and sizing (virtual-scroll, item-size, height) and fit-input-width support.
  * Programmatic panel APIs for scrolling/focus and a node style prop for custom inline styling.

* **Documentation**
  * Added docs and examples for virtual-scroll and fit-input-width, including large-data demo and slot guidance.

* **Tests**
  * Expanded tests covering virtual-scroll, keyboard navigation, selection modes, lazy loading, suggestion sizing, and edge cases.


rzzf:
Is it better to implement a `cascader-v2`?
Rel https://github.com/element-plus/element-plus/issues/6796#issuecomment-1104889253

zhixiaotong:
我认为这就是element的问题，一个组件使用非要分成两个来弄。table没问题，毕竟是实现不同，但select开始就不对劲。还有tree，如果后面要实现tree-select的虚拟化，还得配一个tree-select-v2。

另外分成两个做，两个刚开始可能是一样的，后面的走向只会越来越不同，对于用户使用来说，很痛苦。除非刚开始就选择虚拟化的select,tree，但很多功能都是缺失的。

zhixiaotong:
I still maintain my stance.

The initial code for `element-plus` was designed to replicate the functionality of `Element UI` in a 1:1 manner, leading to the creation of `select-v2`. Since `select-v2` and `table-v2` already existed within `Element UI`, I won't elaborate on that. Similar to `table-v2`, `select` and `select-v2` fundamentally differ in their calling methods, mainly in slots and configuration. However, the later `tree-v2` referenced select to create a separate component solution. Because `tree-select` internally uses tree instead of `tree-v2`, it cannot directly display large amounts of data. I believe this direction is incorrect, as both rely on configuration to pass data. Especially for components like tree and `cascader`, which are already quite functional, supporting virtualization is simply a matter of adding a property; there's no need to create a separate component.

Other UI libraries do not have independent components for each virtualization. Currently, element-plus is burdened with too much legacy data, and I don't think it's necessary to make it even heavier.

If you agree, I will continue to optimize the code; if you disagree, I will close this pull request.

我仍然坚持我的观点。

element-plus最初的代码是为1:1复刻`element ui`的功能，导致`select-v2`的出现，因为`select-v2`和`table-v2`本身`element ui`就存在，这个就不说了。和`table-v2`类似，`select`和`select-v2`本质调用方式不一样，都是插槽和配置的的区别。但后面的tree-v2，却参考了`select`做出了分离组件的方案，由于`tree-select`内部使用了`tree`而非`tree-v2`，导致无法直接使用`tree-select`实现大量数据的展示，我认为这个方向已经不对了，因为都是使用配置的方式传入数据。尤其是`tree`和`cascader`这种，本身功能就很完备的，支持虚拟化只是增加一个属性就可以解决的问题，没必要单独拎出来弄成又一个新的组件。

其他UI库并没有为每个虚拟化独立组件，目前element-plus的历史包袱很重了，我认为没必要更重。

如果你们认同，我将继续优化代码，不认同，我将关闭这个pr。

@element-plus/backers

kooriookami:
I fully agree with your point. Virtualization should not live as a separate component. In the long term, I believe the right direction is to merge existing virtualized components, such as select-v2, back into select itself. For components that are already mature, virtualization should be an optional capability rather than another split component. Perhaps we can also leverage AI to help speed up this process.

我完全赞同你的观点。虚拟化功能不应该作为一个独立组件存在。从长期来看，我认为更合适的方向，是把现有的虚拟化组件，比如  select-v2，逐步合并回 select 本身。对于已经比较成熟的组件来说，虚拟化更应该是一种可选能力，而不是再拆出一个新的组件。或许我们也可以借助 AI 来加快这个过程。

zhixiaotong:
> I fully agree with your point. Virtualization should not live as a separate component. In the long term, I believe the right direction is to merge existing virtualized components, such as select-v2, back into select itself. For components that are already mature, virtualization should be an optional capability rather than another split component. Perhaps we can also leverage AI to help speed up this process.
> 
> 我完全赞同你的观点。虚拟化功能不应该作为一个独立组件存在。从长期来看，我认为更合适的方向，是把现有的虚拟化组件，比如 select-v2，逐步合并回 select 本身。对于已经比较成熟的组件来说，虚拟化更应该是一种可选能力，而不是再拆出一个新的组件。或许我们也可以借助 AI 来加快这个过程。

Thank you for your reply~ I think so too. However, my opinion is to modify `tree` first instead of `select`. After all, there are differences in the calling between `select` and `select-v2`, but not with `tree`. Solving the virtual scrolling of `tree` will also solve the problem that `tree-select` cannot meet the big data requirements.

感谢你的回复~我也是这么认为的，不过我的看法是首先改动`tree`而非`select`，毕竟`select`与`select-v2`两者之间的调用存在差异，但`tree`则不然，解决了`tree`的虚拟滚动，就顺便也解决了`tree-select`无法满足海量数据的问题。

packages/components/cascader/src/cascader.vue
```diff
@@ -167,47 +167,97 @@
...
+          @keydown="handleSuggestionKeyDown"
+        >
+          <el-fixed-size-list
+            ref="suggestionVirtualListRef"
+            :height="height"
+            :item-size="itemSize"
+            :data="suggestions"
+            :total="suggestions.length"
+            inner-element="ul"
+            :inner-width="virtualSuggestionInnerWidth"
```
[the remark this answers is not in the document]
zhixiaotong: 这里尽量不要在上面直接添加v-if，会导致匹配的数据影响虚拟列表的频繁创建和销毁

zhixiaotong:
I encountered a problem: the width of the cascader's suggestion list expands with the content when not using virtual scrolling. To ensure consistency, the width of the text during virtual scrolling needs to be calculated by default. However, if the content is customized using suggestion-item slots, this flexible content cannot be calculated to achieve the desired result. Therefore, a `fit-input-width` similar to `select-v2` was introduced to achieve a similar constraint.

Another issue I discovered, which I believe is a bug, is that the default minimum width is calculated based on the width of the internal input DOM element, not the `el-input` width.

### Before

20260325160224_rec_

### After

20260325160304_rec_

zhixiaotong:
@coderabbitai review again

zhixiaotong:
Could you please review this for me? Thank you so much! @rzzf @kooriookami

kooriookami:
image
Missing vertical padding in virtual scroll mode.

packages/components/cascader-panel/src/menu.vue
```diff
@@ -1,51 +1,104 @@
...
+            :menu-id="menuId"
+            :style="style"
+            @expand="handleExpand"
+          />
+        </template>
+      </el-fixed-size-list>
+      <div v-if="isLoading" :class="ns.e('empty-text')">
+        <ElIcon :size="14" :class="ns.is('loading')">
+          <Loading />
+        </ElIcon>
```
kooriookami: Please using kebab-case.

packages/components/cascader-panel/src/menu.vue
```diff
@@ -1,51 +1,104 @@
...
+        v-for="node in nodes"
+        :key="node.uid"
+        :node="node"
+        :menu-id="menuId"
+        @expand="handleExpand"
+      />
+      <div v-if="isLoading" :class="ns.e('empty-text')">
+        <ElIcon :size="14" :class="ns.is('loading')">
+          <Loading />
+        </ElIcon>
```
kooriookami: Please using kebab-case.

zhixiaotong: got it

rzzf:
Has anyone encountered a similar issue? It can be reproduced in the play/src/App.vue. However, I’m unable to reproduce this issue in the playground or the documentation.

1. Select an option and refresh the page. The browser finishes reloading, but the Cascader remains unresponsive for a few seconds.
2. The scrollbar appears in an incorrect position.


https://github.com/user-attachments/assets/91bf4a0f-e313-42bf-8588-e7e8d97bfc2f

zhixiaotong:
> Has anyone encountered a similar issue? It can be reproduced in the play/src/App.vue. However, I’m unable to reproduce this issue in the playground or the documentation.
> 
> 1. Select an option and refresh the page. The browser finishes reloading, but the Cascader remains unresponsive for a few seconds.
> 2. The scrollbar appears in an incorrect position.
> 
>  20260326-114607.mp4

@rzzf The `main.ts` file in play references some basic scss files.

image

rzzf:
> @rzzf The `main.ts` file in play references some basic scss files.

Yes, let's add the relevant style imports.

packages/components/cascader-panel/style/index.ts
`import '@element-plus/components/virtual-list/style'`

packages/components/cascader-panel/style/css.ts
`import '@element-plus/components/virtual-list/style/css'`

zhixiaotong:
> > @rzzf The `main.ts` file in play references some basic scss files.
> 
> Yes, let's add the relevant style imports.
> 
> packages/components/cascader-panel/style/index.ts `import '@element-plus/components/virtual-list/style'`
> 
> packages/components/cascader-panel/style/css.ts `import '@element-plus/components/virtual-list/style/css'`

ok, i'll fix this.

rzzf:
However, I’m still seeing this issue: the width exceeds that of a single panel, which causes horizontal scrolling.


https://github.com/user-attachments/assets/f4f3a37c-d2d3-40ca-b304-cf8238fc4cf4

zhixiaotong:
> However, I’m still seeing this issue: the width exceeds that of a single panel, which causes horizontal scrolling.
> 
>  20260326-140338.mp4

This is because the `ul` element didn't have `padding: 0` added. I initially thought this was a default setting, but it isn't, and the `li` elements also lack a border-box.

I think it should be added in the SCSS.

zhixiaotong:
Since `virtual-list/style` includes `scrollbar/style`, keeping the former is sufficient. Additionally, the boundary issue when calculating the default width has been resolved. Furthermore, a style-related issue was discovered and resolved as well.

### Before
img_v3_02105_1d3ccfc6-7c87-4cf4-ae2c-4e8061d4c04g

### After
image

packages/components/cascader-panel/src/config.ts
```diff
@@ -130,6 +142,18 @@ export const cascaderPanelProps = buildProps({
   renderLabel: {
     type: Function as PropType<RenderLabel>,
   },
+  virtualScroll: {
+    type: Boolean,
+    default: false,
+  },
```
rzzf:
```suggestion
  virtualScroll: Boolean,
```

packages/components/cascader-panel/src/menu.vue
```diff
@@ -1,51 +1,105 @@
...
-      :node="node"
-      :menu-id="menuId"
-      @expand="handleExpand"
-    />
-    <div v-if="isLoading" :class="ns.e('empty-text')">
-      <el-icon size="14" :class="ns.is('loading')">
-        <loading />
-      </el-icon>
-      {{ t('el.cascader.loading') }}
+  <template v-if="virtualScroll">
```
rzzf, line 2: template is unnecessary

packages/components/cascader-panel/src/menu.vue
```diff
@@ -60,6 +114,18 @@ const props = defineProps({
     type: Number,
     required: true,
   },
+  virtualScroll: {
+    type: Boolean,
+    default: false,
+  },
```
rzzf:
```suggestion
  virtualScroll: Boolean,
```

packages/components/cascader-panel/src/menu.vue
```diff
@@ -73,11 +139,60 @@ let hoverTimer: number | undefined
...
+  } else if (
+    !panel.expandingNode &&
+    panel.checkedNodes.length > 0 &&
+    props.index < panel.checkedNodes[0].pathNodes.length
+  ) {
+    const activeNodeId = panel.checkedNodes[0]?.pathNodes[props.index]?.uid
+    return props.nodes.findIndex((node) => node.uid === activeNodeId)
+  }
+
+  return -1
```
rzzf:
```ts
let activeNodeId: number | undefined

...if / else if / else

return activeNodeId !== undefined
    ? props.nodes.findIndex((node) => node.uid === activeNodeId)
    : -1
```

packages/components/cascader-panel/src/node.vue
```diff
@@ -83,6 +84,10 @@ const props = defineProps({
     required: true,
   },
   menuId: String,
+  style: {
+    type: Object as PropType<CSSProperties>,
+    default: () => ({}),
+  },
```
rzzf: It seems unnecessary to explicitly declare style, as it will be fallthrough automatically.

packages/components/cascader/src/cascader.ts
```diff
@@ -314,6 +330,34 @@ export const cascaderProps = buildProps({
     type: Boolean,
     default: true,
   },
+  /**
+   * @description whether to enable virtual scrolling
+   */
+  virtualScroll: {
+    type: Boolean,
+    default: false,
+  },
```
rzzf:
```suggestion
  virtualScroll: Boolean,
```

packages/components/cascader-panel/src/menu.vue
zhixiaotong: @rzzf 这里有个问题，原来代码中有menuId，添加了if/else分支后为了避免两个重复的key，我在外面添加了template，去掉两个就会导致key重复有问题，我就只去掉了上面的template

rzzf:
This test case did not pass.

<details><summary>Test case</summary>

```ts
  test('virtual mode arrow navigation should skip disabled nodes', async () => {
    const options = [
      {
        value: 'a',
        label: 'Node A',
        children: [{ value: 'a-1', label: 'Node A-1' }],
      },
      {
        value: 'b',
        label: 'Node B',
        disabled: true,
      },
      {
        value: 'c',
        label: 'Node C',
        children: [{ value: 'c-1', label: 'Node C-1' }],
      },
    ]
    const value = ref<string[]>([])

    const wrapper = mount(() => (
      <CascaderPanel v-model={value.value} options={options} virtualScroll />
    ))
    await nextTick()

    const nodes = wrapper.findAll('.el-cascader-node')

    await nodes[0].trigger('keydown', {
      key: 'ArrowDown',
      code: EVENT_CODE.down,
    })
    await nextTick()

    expect(nodes[1].classes()).toContain('is-disabled')
    expect(nodes[2].classes()).toContain('in-active-path')
  })
```
</details>

zhixiaotong:
@rzzf 我想问一下这个目前进程是什么样的？如果短期内合并不了，我就只能把代码再弄到本地替换使用，产品对大数据量的支持需求提前了。。

rzzf:
> @rzzf 我想问一下这个目前进程是什么样的？如果短期内合并不了，我就只能把代码再弄到本地替换使用，产品对大数据量的支持需求提前了。。

我最近比较忙于工作，可能没那么快，你也可以先安装这个PR https://github.com/element-plus/element-plus/pull/23877#issuecomment-4109412598

packages/components/cascader-panel/src/index.vue
```diff
@@ -309,6 +326,27 @@ const handleKeyDown = (e: KeyboardEvent) => {
     case EVENT_CODE.down: {
       e.preventDefault()
       const distance = code === EVENT_CODE.up ? -1 : 1
+      const menuIndex = getMenuIndex(target)
```
btea: This logic can be placed inside the `props.virtualScroll` condition to prevent unnecessary calculations.

packages/components/cascader/src/cascader.vue
```diff
@@ -327,6 +388,8 @@ const props = withDefaults(defineProps<CascaderComponentProps>(), {
   showPrefix: true,
   popperStyle: undefined,
   valueOnClear: undefined,
+  itemSize: 34,
```
btea: These two values ​​are used in multiple places and can be saved and reused as constants.

packages/components/cascader/src/cascader.vue
```diff
@@ -608,48 +685,127 @@ const focusFirstNode = () => {
...
+  const ctx = canvas.getContext('2d')
+  if (!ctx) return 0
+
+  const suggestionPanelEl =
+    suggestionPanel.value instanceof HTMLElement
+      ? suggestionPanel.value
+      : suggestionPanel.value?.$el
+  const renderedSuggestion = suggestionPanelEl?.querySelector(
+    `.${nsCascader.e('suggestion-item')}`
+  ) as HTMLElement | null
```
btea: This code snippet can be extracted and placed inside a function that accepts an optional selector parameter and returns an element or null.

packages/components/cascader/src/cascader.vue
```diff
@@ -608,48 +685,127 @@ const focusFirstNode = () => {
...
+
+const getTagWrapperLeft = () => {
+  if (!slots.prefix) return 0
+
+  const prefix = inputRef.value?.$el.querySelector(
+    `.${nsInput.e('prefix')}`
+  ) as HTMLElement | null
+  if (!prefix) return 0
+
+  const prefixWidth = prefix.offsetWidth
```
btea: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetWidth#:~:text=The%20offsetWidth%20read%2Donly%20property%20of%20the%20HTMLElement%20interface%20returns%20the%20layout%20width%20of%20an%20element%20as%20an%20integer.

The `offsetWidth` returns an integer; perhaps we should use `getBoundingClientRect` for greater precision.

btea (commented):
I have left some suggestions for your reference.

zhixiaotong:
> I have left some suggestions for your reference.

Thanks for the review.I'll adjust the code when I'm free.

packages/components/cascader-panel/src/menu.vue
```diff
@@ -60,6 +112,15 @@ const props = defineProps({
     type: Number,
     required: true,
   },
+  virtualScroll: Boolean,
+  itemSize: {
+    type: Number,
+    default: 34,
```
btea: Here, the default value can be a saved constant.

packages/components/cascader/src/cascader.vue
```diff
@@ -586,16 +658,28 @@ const calculateSuggestions = () => {
...
+    }
+    updateStyle()
+  })
   updatePopperPosition()
 }
 
+const getSuggestionPanelEl = () => {
+  const el = suggestionPanel.value
+  return (el instanceof HTMLElement ? el : el?.$el) as HTMLElement | undefined
+}
```
btea:
```suggestion
const getSuggestionPanelEl = (selector?: string) => {
  const el = suggestionPanel.value
  const $el = (el instanceof HTMLElement ? el : el?.$el) as HTMLElement | undefined
  if (selector && $el) {
    return $el.querySelector(selector)
  }
  return $el
}
```
The logic for querying sub-elements can be further combined into one.

zhixiaotong: 谢谢你的review，改了。另外，如果以后其他现有组件v2均支持使用virtualScroll的时候，可以把这块统一抽取成hook，现在我暂时只放在cascader中。

Dsaquel (commented):
Overall looks great.

The filter is a bit slow if there are too many items but i think this can be improve later. I will review a bit more tomorrow 😃

zhixiaotong:
> Overall looks great.
> 
> The filter is a bit slow if there are too many items but i think this can be improve later. I will review a bit more tomorrow 😃

Yes, I tested it before, and there was a slight delay around 20,000 searches. In a real project, we were required to support a three-level cascading effect for 100,000 searches, so I abandoned this interaction and switched to a custom component. My personal feeling is that virtual scrolling is suitable for processing data of no more than 20,000 records; dumping massive amounts of data into the browser all at once is not a reasonable approach.

docs/en-US/component/cascader.md
```diff
@@ -193,6 +193,32 @@ cascader/custom-header-footer
 
 :::
 
+## Virtual Scroll ^(2.13.7)
```
btea:
```suggestion
## Virtual Scroll ^(2.14.0)
```

packages/theme-chalk/src/cascader-panel.scss
```diff
@@ -80,6 +90,7 @@
   height: 34px;
   line-height: 34px;
   outline: none;
+  box-sizing: border-box;
```
btea, line 93: Are these modifications necessary?

zhixiaotong: > Are these modifications necessary?

需要，在play中发现默认有横向滚动条，playground中没有是因为有了其他样式影响

docs/en-US/component/cascader.md
btea: Versions in other regions also need to be modified accordingly.

packages/theme-chalk/src/cascader-panel.scss
btea: https://github.com/element-plus/element-plus/pull/23877#issuecomment-4131436155 确实，解决了这里提到的问题。

docs/en-US/component/cascader.md
zhixiaotong: 哈，昨天忘了，昨天太忙了

docs/en-US/component/cascader.md
```diff
@@ -231,6 +257,10 @@ cascader/custom-header-footer
...
+| virtual-scroll ^(2.14.0)                   | whether to enable virtual scrolling for large data                                                                                                                               | ^[boolean]                                                                                                                                                                  | false        |
+| fit-input-width ^(2.14.0)                  | whether the width of the suggestion panel is the same as the input, if the value is `number`, then the width is fixed                                                            | ^[boolean] / ^[number]                                                                                                                                                      | false        |
```
btea, line 261: Based on the description and the implemented effect, `fit-input-width` only adapts to the suggested panels, not the default panel style. Was this intentional? It feels like it's only half done; perhaps all panels should be consistent?

wdyt? @rzzf @Dsaquel @kooriookami @zhixiaotong

rzzf: I share the same concern. From the perspective of prop naming, I tend to prefer keeping all panels consistent.

rzzf: However, when the input width is 100% and fit-input-width is set to true, this may cause all panels to become excessively large.

zhixiaotong: 我用手机回复吧，电脑现在vpn不管用了。

我是这么设计的，建议面板和select，treectselect之类的下拉框一致，如果是真实DOM会被拉宽，fit-input-width是为了虚拟化的时候可以设置一个合理的宽度。我个人觉得没有必要对cascader-panel设置这个属性，因为默认panel就是固定的宽度，多余的文本会被溢出省略号，不管用不用virtual都是统一视觉的。

zhixiaotong: 另外说一个问题，我看到@rzzf 去掉了defineProps的写法，因为最近这些会慢慢迁移到interface声明props，只是tree-v2之前不知道哪个版本去掉了treeV2Props的导出，2.13.1是正常的，由于现在element-plus不支持tree-select-v2，因为tree目前不支持virtual，我本地自己写了一个tree-select-v2供项目使用，之前可以直接导出tree-v2和select的props自己实现逻辑，但现在就不方便了。是否应该把那个导出先还原？过时不代表不能用。

rzzf: 似乎从来没有导出过`treeV2Props`的类型

image

btea: > 我用手机回复吧，电脑现在vpn不管用了。
> 
> 我是这么设计的，建议面板和select，treectselect之类的下拉框一致，如果是真实DOM会被拉宽，fit-input-width是为了虚拟化的时候可以设置一个合理的宽度。我个人觉得没有必要对cascader-panel设置这个属性，因为默认panel就是固定的宽度，多余的文本会被溢出省略号，不管用不用virtual都是统一视觉的。

看到你文档中添加了相关的提示说明，那就先保持这样吧。

btea (approved):
Overall, there seem to be no problems so far. Thank you for your contribution.

rzzf (approved):

docs/en-US/component/cascader.md
zhixiaotong: > 似乎从来没有导出过`treeV2Props`的类型
> 
> image
> 

有噢，我之前一直在用😳，我指的导出的是对象不是类型，可以通过增强defineProps({...treeProps})使用，但之前不是从element-plus引用，是从element-plus/es/components/tree-v2/src/virtual-tree引用，现在新版本不行了。

rzzf: 可能是构建产物生成的dts出了问题，源码中没有删除这个对象

rzzf: 应该与dts 插件的entry 有关

zhixiaotong:
thank you all!🫰

docs/en-US/component/cascader.md
rzzf: @zhixiaotong  除了tree-v2还有什么组件有类似的问题吗？

zhixiaotong: 我目前换到了最新版，暂时只发现了这一个问题，从内部源码引用的地方不是很多。

rzzf: > 另外说一个问题，我看到@rzzf 去掉了defineProps的写法，因为最近这些会慢慢迁移到interface声明props，只是tree-v2之前不知道哪个版本去掉了treeV2Props的导出，2.13.1是正常的，由于现在element-plus不支持tree-select-v2，因为tree目前不支持virtual，我本地自己写了一个tree-select-v2供项目使用，之前可以直接导出tree-v2和select的props自己实现逻辑，但现在就不方便了。是否应该把那个导出先还原？过时不代表不能用。

fixed by https://github.com/element-plus/element-plus/pull/24116

Issue #23862, opened 2026-03-20
[Style] [checkbox, checkbox-group] 【checkbox】更新后Checkbox 多选框无法通过css去修改选中图标了


liytingtt:
Bug Type: **`Style`**

## Environment

- Vue Version: `3.5.30`
- Element Plus Version: `2.13.6`
- Browser / OS: `Chrome`
- Build Tool: `Vite`

### Related Component

- `el-checkbox`
- `el-checkbox-group`

### Reproduction Link

[...]

PR #23882, opened 2026-03-23
feat(components): [checkbox] add custom checked icon slot

snowbitx:

ref #23862 

image

## Summary by CodeRabbit

* **New Features**
  * Checkbox component now supports customizable icons for the checked state, enabling users to replace the default checked icon with their own design.

* **Documentation**
  * Added comprehensive documentation for the custom icons feature, including guidance on implementation and practical examples showcasing how to customize checked icons in checkboxes.

### DIFF
--- a/docs/en-US/component/checkbox.md
+++ b/docs/en-US/component/checkbox.md
@@ -79,6 +79,16 @@ checkbox/intermediate
 
 :::
 
+## Custom icons
+
+Use `checked-icon` slot to customize the icon inside the checked checkbox.
+
+:::demo
+
+checkbox/custom-icon
+
+:::
+
 ## Minimum / Maximum items checked
 
 The `min` and `max` properties can help you to limit the number of checked items.
@@ -141,9 +151,10 @@ checkbox/with-border
 
 ### Checkbox Slots
 
-| Name    | Description               |
-| ------- | ------------------------- |
-| default | customize default content |
+| Name         | Description               |
+| ------------ | ------------------------- |
+| default      | customize default content |
+| checked-icon | customize checked icon    |
 
 ## CheckboxGroup API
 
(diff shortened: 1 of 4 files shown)


keeplearning66 (commented):
Thank you for your contribution! However, I personally feel that this feature isn’t very universal and can’t ensure consistency. For example, when using `checkbox` on its own, you can customize the checked icon, but components that use `checkbox` internally can’t be customized (e.g. `tree` component).

snowbitx:
> Thank you for your contribution! However, I personally feel that this feature isn’t very universal and can easily cause inconsistencies. For example, when using `checkbox` on its own, you can customize the checked icon, but components that use `checkbox` internally can’t be customized (e.g. `tree` component).

Thanks for the feedback.

That makes sense. If consistency is the main concern, this probably should not be limited to standalone `checkbox`, and we would need to consider other scenarios built on top of it as well.

My original motivation was that customizing the checked icon currently relies on CSS overrides, which feels a bit cumbersome and not very extensible.

If the team thinks this direction is worth exploring, I can try to extend it to other related components. We may also want to hear more opinions before deciding whether this API should be introduced.

rzzf:
Emmm, I don’t think icon customization for such a basic component needs to be supported by the library—it’s likely a relatively rare use case. Even if needed, overriding styles via CSS would be more flexible and controllable than using slots. 
As @keeplearning66  mentioned, if we introduce a slot for this, should other components that depend on checkbox also provide similar support? That starts to feel like over-engineering.

snowbitx:
> Emmm, I don’t think icon customization for such a basic component needs to be supported by the library—it’s likely a relatively rare use case. Even if needed, overriding styles via CSS would be more flexible and controllable than using slots. As @keeplearning66 mentioned, if we introduce a slot for this, should other components that depend on checkbox also provide similar support? That starts to feel like over-engineering.

Thank you both for the patient replies. I understand now😀

PR #23907, opened 2026-03-26
fix(hooks): [use-timeout] avoid window usage in SSR

### LINKED ISSUE #23903: [Build] [tooltip] fix(hooks): use-timeout breaks SSR when scope dispose runs (window.setTimeout / window.clearTimeout)
Bug Type: **`Build`**

## Environment

- Vue Version: `3.5.31`
- Element Plus Version: `2.11.2`
- Browser / OS: `Chrome`
- Build Tool: `Nuxt`

### Related Component

- `el-tooltip`

### Reproduction Link

[Link](https://www.seabell.com)

### Steps to reproduce

1. Nuxt 3 + `element-plus` + SSR enabled.
2. Render a page that includes `el-tooltip` (or any component that uses `useDelayedToggle` → `useTimeout`) in the server-rendered tree.
3. Use a **recent Vue 3.5.x** where SSR teardown runs dispose callbacks (e.g. compare 3.5.18 vs newer patches).
4. Production build + `renderToString` for that route → **500** / thrown error in server logs.

### What is Expected?

SSR should not throw when the effect scope is stopped: timer cleanup should use **environment-safe** APIs (e.g. global `setTimeout` / `clearTimeout`, which exist in Node and browsers), or guard with `typeof window !== 'undefined'` before touching `window`.

`useTimeout` in `packages/hooks/use-timeout` uses `window.setTimeout` and `window.clearTimeout` and registers `cancelTimeout` via `tryOnScopeDispose`.
When using SSR (e.g. Nuxt 3 + Vite):
1. After `renderToString`, Vue may run effect scope teardown and **execute** the dispose callback on the server (behavior depends on Vue 3.5.x patch version).
2. On the server there is no `window`, so cleanup throws: `Cannot read properties of undefined (reading 'clearTimeout')`.
3. Additionally, the SSR bundle may optimize `window` to `void 0`, producing code like `(void 0).clearTimeout(...)`, same runtime error.
This hook is pulled in by delayed open/close logic used from components such as **ElTooltip** (and similar), so it can break **any page** that renders those components during SSR.

### Additional comments

- Element Plus: 2.11.x (and current `dev` still uses `window` in this hook)
- Vue: 3.5.x (patch version affects whether dispose runs on server)
- Bundler: Vite SSR build
- Related consumer workaround: alias/replace `use-timeout` module in app build — works but maintenance-heavy; an upstream fix is preferred.

iml885203:

## 🐛 Bug Fixes

Fixes #23903

### Problem

`use-timeout` used `window.setTimeout` and `window.clearTimeout`. In SSR (Node) environment, `window` may be undefined, which can throw during scope dispose and break server rendering.

### Solution

- Replace `window.setTimeout` / `window.clearTimeout` with global `setTimeout` / `clearTimeout`
- Use `ReturnType<typeof setTimeout>` for cross-environment timer handle typing
- Add `undefined` guard in `cancelTimeout` and reset handle after cancel

### Tests

- Add SSR-like test case where `window` is stubbed as `undefined`
- Ensure cleanup in test always restores globals via `try/finally`
- Verified:
  - `pnpm vitest run packages/hooks/__tests__/use-timeout.test.ts`
  - `pnpm vitest run packages/hooks/__tests__`

## Summary by CodeRabbit

* **Tests**
  * Added tests covering Node/SSR and no-browser-global scenarios to ensure timeouts are not invoked during server rendering and are properly cleaned up.

* **Bug Fixes**
  * Made timeout registration and cancellation more robust so cancellations safely no-op when no timer exists and callbacks are prevented after scope teardown, including in non-browser environments.

### DIFF
--- a/packages/hooks/__tests__/use-timeout.ssr.test.ts
+++ b/packages/hooks/__tests__/use-timeout.ssr.test.ts
@@ -23,5 +23,6 @@ describe('use-timeout (SSR)', () => {
     })
 
     await expect(renderToString(createSSRApp(App))).resolves.toContain('ssr')
+    expect(fn).not.toHaveBeenCalled()
   })
 })
--- a/packages/hooks/__tests__/use-timeout.ssr.test.ts
+++ b/packages/hooks/__tests__/use-timeout.ssr.test.ts
@@ -1,0 +1,27 @@
+/**
+ * @vitest-environment node
+ */
+
+import { createSSRApp, defineComponent, h } from 'vue'
+import { renderToString } from '@vue/server-renderer'
+import { describe, expect, it, vi } from 'vitest'
+import { useTimeout } from '../use-timeout'
+
+describe('use-timeout (SSR)', () => {
+  it('should render without window in SSR environment', async () => {
+    const fn = vi.fn()
+
+    expect('window' in globalThis).toBe(false)
+
+    const App = defineComponent({
+      setup() {
+        const { registerTimeout } = useTimeout()
+        registerTimeout(fn, 0)
+
+        return () => h('div', 'ssr')
+      },
+    })
+
+    await expect(renderToString(createSSRApp(App))).resolves.toContain('ssr')
+  })
+})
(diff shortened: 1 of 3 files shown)


rzzf:
Please fix the errors reported by CI

rzzf (approved):

btea (approved):

PR #23952, opened 2026-03-31
feat(components): [calendar] add `select` event

chengyanqi123:

## Summary by CodeRabbit

* **New Features**
  * Calendar now emits a `select` event when a day is chosen, passing the selected date to handlers.

* **Tests**
  * Added test coverage verifying the `select` event is emitted and the handler receives a Dayjs date.

* **Documentation**
  * Added documentation for the Calendar `select` event, its callback type, and expected behavior.


keeplearning66 (commented):
This is similar to #19035. I feel that using `CHANGE_EVENT` as the event name would be better.

chengyanqi123:
> This is similar to #19035. I feel that using `CHANGE_EVENT` as the event name would be better.这与 #19035 类似。我认为使用 `CHANGE_EVENT` 作为事件名称会更好。



> This is similar to #19035. I feel that using `CHANGE_EVENT` as the event name would be better.这与 #19035 类似。我认为使用 `CHANGE_EVENT` 作为事件名称会更好。

Thank you for your review. I don’t think this follows the same logic as ‘#19035’. `change` should be triggered when the bound value changes (it would also be triggered when switching to the previous or next month), similar to `watch(model:value)`. `pickDay` is only triggered when a specific day is clicked, similar to `click`; perhaps it could be renamed to `clickDay`?


I’m sorry, my English isn’t very good, so this may make it difficult for you to read.

chengyanqi123:
> This is similar to #19035. I feel that using `CHANGE_EVENT` as the event name would be better.这与 #19035 类似。我认为使用 `CHANGE_EVENT` 作为事件名称会更好。



> This is similar to #19035. I feel that using `CHANGE_EVENT` as the event name would be better.这与 #19035 类似。我认为使用 `CHANGE_EVENT` 作为事件名称会更好。

Similar to the effect shown in this GIF:

keeplearning66 (commented):
Thank you for your explanation!
Directly using the `date-cell` slot should be able to achieve the effect shown in the GIF above. 🤔 
If we need to add a general event, I feel that adding `change` or `select` would be more appropriate. That way, it feels more consistent with other components. We can wait for others' opinions as well. 😃

keeplearning66:
> If we need to add a general event, I feel that adding change or select would be more appropriate. That way, it feels more consistent with other components.

Perhaps an additional parameter could be provided in the event to indicate whether a specific date was clicked. 🤔

chengyanqi123:
> > If we need to add a general event, I feel that adding change or select would be more appropriate. That way, it feels more consistent with other components.
> 
> Perhaps an additional parameter could be provided in the event to indicate whether a specific date was clicked. 🤔

Thank you for your suggestion. I have changed the event name to `select` :)

packages/components/calendar/src/calendar.vue
```diff
@@ -46,7 +50,7 @@
         :selected-day="realSelectedDay"
         :range="range_"
         :hide-header="index !== 0"
-        @pick="pickDay"
+        @pick="handlerPick"
```
btea: There should be no need to redefine a new function; you can use the original function logic and emit the relevant events internally.

chengyanqi123: Thank you for your review. I have removed `handlerPick` and implemented the `select` event based on the original code.
Could you please review this again?

docs/en-US/component/calendar.md
```diff
@@ -71,6 +71,12 @@ Note, date time locale (month name, first day of the week ...) are also configur
 | date-cell | `type` indicates which month the date belongs, optional values are prev-month, current-month, next-month; `isSelected` indicates whether the date is selected; `day` is the formatted date in the format `YYYY-MM-DD`; `date` is date the cell represents | ^[object]`{ data: { type: 'prev-month' \| 'current-month' \| 'next-month', isSelected: boolean, day: string, date: Date } }` |
 | header    | content of the Calendar header                                                                                                                                                                                                                            | ^[object]`{ date: string }`                                                                                                  |
 
+### Events
+
+| Name             | Description                      | Type                                    |
+| ---------------- | -------------------------------- | --------------------------------------- |
+| select ^(2.13.7) | Triggered when a day is selected | ^[Function]`(day: dayjs.Dayjs) => void` |
```
patienry:
```suggestion
| select ^(2.14.0) | Triggered when a day is selected | ^[Function]`(day: dayjs.Dayjs) => void` |
```

packages/components/calendar/src/calendar.ts
```diff
@@ -82,5 +84,6 @@ export type CalendarPropsPublic = ExtractPublicPropTypes<typeof calendarProps>
 export const calendarEmits = {
   [UPDATE_MODEL_EVENT]: (value: Date) => isDate(value),
   [INPUT_EVENT]: (value: Date) => isDate(value),
+  select: (value: Dayjs) => isObject(value),
```
patienry:
```suggestion
  select: (value: Date) => isDate(value),
```

patienry: It would be better to keep the parameter types consistent here.

chengyanqi123: Thank you for your guidance🙏

docs/en-US/component/calendar.md
```diff
@@ -71,6 +71,12 @@ Note, date time locale (month name, first day of the week ...) are also configur
 | date-cell | `type` indicates which month the date belongs, optional values are prev-month, current-month, next-month; `isSelected` indicates whether the date is selected; `day` is the formatted date in the format `YYYY-MM-DD`; `date` is date the cell represents | ^[object]`{ data: { type: 'prev-month' \| 'current-month' \| 'next-month', isSelected: boolean, day: string, date: Date } }` |
 | header    | content of the Calendar header                                                                                                                                                                                                                            | ^[object]`{ date: string }`                                                                                                  |
 
+### Events
+
+| Name             | Description                      | Type                                    |
+| ---------------- | -------------------------------- | --------------------------------------- |
+| select ^(2.14.0) | Triggered when a day is selected | ^[Function]`(day: dayjs.Dayjs) => void` |
```
patienry: The parameter types here also need to be updated.

keeplearning66 (commented):
Please fix the failing test case.
https://github.com/element-plus/element-plus/actions/runs/24285557162/job/70914904906?pr=23952

docs/en-US/component/calendar.md
chengyanqi123: Okay, I've made the changes and updated the test code. Thank you for your guidance :)

docs/en-US/component/calendar.md
```diff
@@ -71,6 +71,12 @@ Note, date time locale (month name, first day of the week ...) are also configur
 | date-cell | `type` indicates which month the date belongs, optional values are prev-month, current-month, next-month; `isSelected` indicates whether the date is selected; `day` is the formatted date in the format `YYYY-MM-DD`; `date` is date the cell represents | ^[object]`{ data: { type: 'prev-month' \| 'current-month' \| 'next-month', isSelected: boolean, day: string, date: Date } }` |
 | header    | content of the Calendar header                                                                                                                                                                                                                            | ^[object]`{ date: string }`                                                                                                  |
 
+### Events
+
+| Name             | Description                      | Type                             |
+| ---------------- | -------------------------------- | -------------------------------- |
+| select ^(2.14.0) | Triggered when a day is selected | ^[Function]`(day: Date) => void` |
```
patienry, line 78: For the `select` event, I have two ideas:

1. Trigger it **only** when a specific date is selected; do not trigger it when switching via the controller.
2. Trigger it **both** when a specific date is selected and when switching via the controller; additionally, **pass an extra parameter** in the `select` event to indicate the trigger type. (e.g. `(day: Date, triggerType: 'year' | 'month' | 'prevMonth' | 'nextMonth' | 'today' | 'day') => void`)

I'm not quite sure which option is better right now—we can wait for others' suggestions. 👀

chengyanqi123: Okay, It’s true that the second option is more flexible.🤔

PR #23984, opened 2026-04-07
ci: bump node-version to 24

thinkasany:
Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: [github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD infrastructure to use Node.js version 24 across all automated workflows, improving build environment compatibility and stability.

### DIFF
--- a/.github/workflows/lint-commit-message.yml
+++ b/.github/workflows/lint-commit-message.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v6
         with:
-          node-version: 20.19.0
+          node-version: 24
           cache: 'pnpm'
 
       - name: Install dependencies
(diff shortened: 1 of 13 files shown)


btea (approved):
There doesn't seem to be any problem.

thinkasany:
> There doesn't seem to be any problem.

In that case, I will proceed with the merge now. I will continue to monitor the results; should I encounter any problematic actions, I will investigate the root cause. If a fix proves unfeasible, I will roll back to the original version.

btea:
@thinkasany Thanks! 👍

PR #23989, opened 2026-04-07
docs(components): [tabs] supplement the props exposed by tab-bar

micaiguai:

## Summary by CodeRabbit

* **Documentation**
  * Added a "Tab-bar Exposes" section in Tabs docs clarifying the exposed ref and describing that the update member returns the updated tab-bar style.
* **Refactor**
  * Adjusted exposure wording for the tab-bar's update member to clarify its return semantics; no change to runtime behavior or public API.

### DIFF
--- a/docs/en-US/component/tabs.md
+++ b/docs/en-US/component/tabs.md
@@ -148,10 +148,10 @@ tabs/default-value
 
 ### Tab-bar Exposes
 
-| Name   | Description                             | Type                                        |
-| ------ | --------------------------------------- | ------------------------------------------- |
-| ref    | tab root html element                   | ^[object]`Ref<HTMLDivElement \| undefined>` |
-| update | method to manually update tab bar style | ^[Function]`() => void`                     |
+| Name   | Description                                                       | Type                                        |
+| ------ | ----------------------------------------------------------------- | ------------------------------------------- |
+| ref    | tab root html element                                             | ^[object]`Ref<HTMLDivElement \| undefined>` |
+| update | method to manually update tab bar style, return the updated style | ^[Function]`() => CSSProperties`            |
 
 ## Tab-pane API
 
--- a/docs/en-US/component/tabs.md
+++ b/docs/en-US/component/tabs.md
@@ -171,9 +171,9 @@ tabs/default-value
 ```ts
 type TabBarInstance = InstanceType<typeof TabBar> & {
   /** @description tab root html element */
-  ref: barRef
+  ref: Ref<HTMLDivElement | undefined>
   /** @description method to manually update tab bar style */
-  update
+  update: () => void
 }
 ```
 
--- a/docs/en-US/component/tabs.md
+++ b/docs/en-US/component/tabs.md
@@ -148,10 +148,10 @@ tabs/default-value
 
 ### Tab-bar Exposes
 
-| Name   | Description                                                       | Type                                        |
-| ------ | ----------------------------------------------------------------- | ------------------------------------------- |
-| ref    | tab root html element                                             | ^[object]`Ref<HTMLDivElement \| undefined>` |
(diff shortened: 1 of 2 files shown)


rzzf (approved):

keeplearning66 (commented):
Should this be consistent with the source code?

https://github.com/element-plus/element-plus/blob/f4e49a76b74fafe4d695d62ce77589c725d029ff/packages/components/tabs/src/tab-bar.ts#L30

micaiguai:
> Should this be consistent with the source code?
> 
> https://github.com/element-plus/element-plus/blob/f4e49a76b74fafe4d695d62ce77589c725d029ff/packages/components/tabs/src/tab-bar.ts#L30

https://github.com/element-plus/element-plus/blob/f4e49a76b74fafe4d695d62ce77589c725d029ff/packages/components/tabs/src/tab-bar.vue#L117-L122

Thanks for your suggestions. I think it would be helpful to more explicitly mention that developers can use `ref` and `update` directly. Otherwise, this would feel like a black box or hidden API for developers.

keeplearning66 (commented):
> I think it would be helpful to more explicitly mention that developers can use ref and update directly. Otherwise, this would feel like a black box or hidden API for developers.

Maybe we can directly specify the exposed properties in the document like `Tab-nav`, and make `TabBarInstance` consistent with the source code. 🤔 WDYT?

https://element-plus.org/en-US/component/tabs#tab-nav-exposes

micaiguai:
> Maybe we can directly specify the exposed properties in the document like `Tab-nav`, and make `TabBarInstance` consistent with the source code. 🤔 WDYT?
> 
> https://element-plus.org/en-US/component/tabs#tab-nav-exposes

Thanks for your guidance, I've updated the code. Since `type TabBarInstance = InstanceType<typeof TabBar> & unknown` is redundant and provides the same quantity of information as `TabBarInstance`, so i removed `TS declaration`.

packages/components/tabs/src/tab-bar.vue
```diff
@@ -84,7 +84,9 @@ const getBarStyle = (): CSSProperties => {
   }
 }
 
-const update = () => (barStyle.value = getBarStyle())
+const update = () => {
+  barStyle.value = getBarStyle()
+}
 
 const tabObservers = [] as ReturnType<typeof useResizeObserver>[]
```
micaiguai, line 89: Keep consistent to TS type `() => void`. I've checked the return value isn't used.

btea: Because this method is exposed, its return value could be obtained by the user. Removing the return value here might affect the user. Maintaining the status quo might be better.

micaiguai: Agreed.

docs/en-US/component/tabs.md
```diff
@@ -144,6 +144,15 @@ tabs/default-value
...
 | tabBarRef ^(2.9.10)  | el_tabs\_\_nav bar instance | ^[object]`Ref<TabBarInstance \| undefined>` |
 
+## Tab-bar API
+
+### Tab-bar Exposes
+
+| Name   | Description                             | Type                                        |
+| ------ | --------------------------------------- | ------------------------------------------- |
+| ref    | tab root html element                   | ^[object]`Ref<HTMLDivElement \| undefined>` |
+| update | method to manually update tab bar style | ^[Function]`() => void`                     |
```
btea: One thing was overlooked: the return values ​​in the documentation also need to be synchronized.

micaiguai: Thanks for your guidance. I’ve updated it.

btea (approved):

keeplearning66 (approved):
Thank you!

PR #24011, opened 2026-04-09
fix(components): [date-picker-panel] prevent early range fix

### LINKED ISSUE #24006: [Component] [date-picker] DateTimePicker 存在严重bug,选择的时间会回退一个月
Bug Type: **`Component`**

## Environment

- Vue Version: `3.6.0-beta.9`
- Browser / OS: `Chrome/145.0.0.0`
- Build Tool: `Vite`

### Related Component

- `el-date-picker`

### Reproduction Link

[Docs](https://element-plus.org/zh-CN/component/datetime-picker)

### Steps to reproduce

1、datetimerange模式下，通过鼠标选择任意时间区间(例如2026-02-10 00:00:00 ～ 2026-03-16 00:00:00)
2、输入框中按顺序依次使用复制或输入功能输入开始年月日，开始时分秒
3、输入框中继续复制或输入结束年月日，此时开始年月日发生变化自动变成了前一个月，开始时分秒变成了00:00:00 
4、点clear后 重复1～3步骤为必现

### What is Expected?

操作结束日期不应修改开始日期

操作结束日期后，修改了开始日期

### Additional comments

官网文档中也可复现

LostElkByte:
Fixes #24006

### Description

When using `type="datetimerange"`, manually editing the end date in the input field causes the start date to be corrupted. For example, changing the end date from `2026-05-01` to `2026-04-20`—when the month is changed from `05` to `04`, the intermediate value `2026-04-01` is momentarily before the start date `2026-04-01 01:00:00`, which prematurely triggers the range correction logic, resetting the start date to the previous month with time `00:00:00`.

### Root Cause

handleDateInput in [panel-date-range.vue] is bound to the `@input` event (fires on every keystroke) and contains range correction logic that adjusts the other date when `end < start`. Intermediate input states during typing can trigger this correction before the user finishes editing.

### Fix

- **Move the range correction logic** from [handleDateInput] (`@input`) to [handleDateChange](`@change`, fires on blur/enter), so correction only runs on the final committed value.
- **Keep the initialization logic** in [handleDateInput]: when the counterpart date doesn't exist, a default is still created during input.
- This aligns with how [handleTimeInput] / [handleTimeChange already work.
- Regression test added.

## Summary by CodeRabbit

* **Tests**
  * Added a test ensuring the date-range picker preserves the start date when an end date is edited via intermediate/incomplete input and only finalized on change.

* **Bug Fixes**
  * Improved date-range panel behavior to avoid corrupting the opposite bound during raw typing and to detect/repair inconsistent min/max ranges when a bound is changed.


LostElkByte:
@codex review

keeplearning66 (commented):
Could you use English for the comments in the test cases where possible? That would keep them consistent with the rest of the codebase.

keeplearning66 (approved):
Thank you!

LostElkByte:
Can this PR be merged? I plan to update PR #24085 on top of the latest changes once this one is landed. @Dsaquel

btea (approved):

Dsaquel (approved):
Looks good, this way the right date input have the same behavior.

Sorry for the late review 🙏

Issue #23486, opened 2026-01-20
[Component] [select] [Component] [select] bug - you need to use double click to select on [el-select]


eharat:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.26`
- Element Plus Version: `2.13.1`
- Browser / OS: `Safari iOS / Chrome iOS`
- Build Tool: `Other`

### Related Component

- `el-select`

### Reproduction Link

[Docs](https://element-plus.org/en-US/component/select#grouping)

### Steps to reproduce

You need to access the "Safari" or "Chrome" browser on your iOS phone, then try to select an item from [el-select], at this point you will only be able to select if you click twice when there are many options and scrolling occurs in the dropdown menu.
The same issue occurs in the "Grouping" section at the URL below.
https://element-plus.org/en-US/component/select#grouping

### What is Expected?

I hope it works on the first click.

### What is actually happening?

component [el-select] requires two clicks to select.

### Additional comments

In other Android browsers and computer browsers I have no problem.

rzzf:
I cannot reproduce this issue. Can you provide more information?

eharat:
[...]

PR #24033, opened 2026-04-12
fix(components): [select/v2] ios click listening

ixyzorg:

**Desc**
The cause of the issue is the same as in related: element-plus/element-plus/pull/16393. The difference is that the `<li>` above `el-option` does not handle the `mousemove` event.

By the way, removing the mousemove event on mobile devices does not affect the original functionality, and it also avoids the impact of events like mousemove and mouseenter on click.
image

related: element-plus/element-plus/issues/23486
related: element-plus/element-plus/pull/16393
realted: element-plus/element-plus/pull/23881

The following examples all occur occasionally.

1. In this case, it happens occasionally. Please note that when clicking `Option 4`, the click event is not triggered in the console.

https://github.com/user-attachments/assets/f349164e-f4af-408e-b4c7-82fc94588b7f

2. Note that I clicked the `Beijing` option, but it was not selected. Instead, only the mousemove hover highlight appeared, and a scrollbar showed up. https://github.com/element-plus/element-plus/issues/23486#issuecomment-4219667209

https://github.com/user-attachments/assets/e51652a4-5daf-4464-bd83-4ccd7bf0caac

3. I just tested select-v2 as well, and this issue occasionally happens there too.

https://github.com/user-attachments/assets/28fe8187-7124-42f0-aae1-6b1f956969c2

## Summary by CodeRabbit

* **Bug Fixes**
  * Disabled hover-based option highlighting on iOS to prevent spurious hover behavior while keeping hover interactions on other platforms.
  * No changes to public component APIs or visible props; non-iOS behavior remains unchanged.

### DIFF
--- a/packages/components/select-v2/src/option-item.vue
+++ b/packages/components/select-v2/src/option-item.vue
@@ -12,7 +12,7 @@
       ns.is('created', created),
       ns.is('hovering', hovering),
     ]"
-    @mousemove="hoverItem"
+    @[mouseMoveEventName]="hoverItem"
     @mousedown="handleMousedown"
     @click.stop="selectOptionClick"
   >
@@ -29,7 +29,7 @@ import { useOption } from './useOption'
 import { useProps } from './useProps'
 import { optionV2Emits, optionV2Props } from './defaults'
 import { selectV2InjectionKey } from './token'
-import { isFocusable } from '@element-plus/utils'
+import { isFocusable, isIOS } from '@element-plus/utils'
 
 export default defineComponent({
   props: optionV2Props,
@@ -37,6 +37,7 @@ export default defineComponent({
   setup(props, { emit }) {
     const select = inject(selectV2InjectionKey)!
     const ns = useNamespace('select')
+    const mouseMoveEventName = isIOS ? null : 'mousemove'
     const { hoverItem, selectOptionClick } = useOption(props, { emit })
     const { getLabel } = useProps(select.props)
     const contentId = select.contentId
@@ -58,6 +59,7 @@ export default defineComponent({
     return {
       ns,
       contentId,
+      mouseMoveEventName,
       hoverItem,
       handleMousedown,
       selectOptionClick,
--- a/packages/components/select/src/option.vue
+++ b/packages/components/select/src/option.vue
@@ -6,7 +6,7 @@
     role="option"
     :aria-disabled="isDisabled || undefined"
     :aria-selected="itemSelected"
-    @mousemove="hoverItem"
+    @[mouseMoveEventName]="hoverItem"
     @mousedown="handleMousedown"
     @click.stop="selectOptionClick"
   >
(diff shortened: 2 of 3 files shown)


packages/components/select-v2/src/option-item.vue
```diff
@@ -23,20 +23,21 @@
...
-import { isFocusable } from '@element-plus/utils'
+import { isFocusable, isIOS } from '@element-plus/utils'
 
 export default defineComponent({
   props: optionV2Props,
   emits: optionV2Emits,
   setup(props, { emit }) {
     const select = inject(selectV2InjectionKey)!
     const ns = useNamespace('select')
+    const mouseMoveEventName = computed(() => (isIOS ? null : 'mousemove'))
```
rzzf: I think `isIOS` isn’t a reactive value, so there’s no need to wrap it in a `computed`.

packages/components/select/src/option.vue
```diff
@@ -62,6 +62,7 @@ export default defineComponent({
       hover: false,
     })
 
+    const mouseMoveEventName = computed(() => (isIOS ? null : 'mousemove'))
```
rzzf: as above

ixyzorg:
@rzzf 
I have finished the fix. I found two more places with the same issue in the code. I will open a new PR to fix those two as well.

btea:
@ixyzorg Can you verify whether the problem is resolved by installing this pr locally? https://github.com/element-plus/element-plus/pull/24033#issuecomment-4231531454

ixyzorg:
hi there @warmthsea 
This issue still persists, and I don’t think it has been fixed. related: element-plus/element-plus/issues/19127

Here is my test based on the official documentation from last night. From the behavior we observed, after clicking, only the hover state change triggered by mousemove occurs, while the click event is not fired.

According to Apple’s developer documentation, once a mousemove event is triggered, subsequent events may not be dispatched as expected.

I tend to believe that when users try to tap on mobile devices, **maybe** their finger may slightly move (about 1–2px). The WebKit engine then interprets this as a mousemove, which triggers the hover state, and as a result, the click event is not fired.

https://ixyz.org:444/ScreenRecording_04-13-2026%2001-04-06_1.MP4
https://ixyz.org:444/ScreenRecording_04-13-2026%2001-11-50_1.MP4

ixyzorg:
@btea  
hi sir . 

I will run multiple tests because this issue occurs intermittently. After my changes, I have tested it many times and have not been able to reproduce the problem.

According to Apple’s developer documentation (as shown in the image above), once a mousemove event is triggered, it will suppress the subsequent click event.

From the observed behavior, we can see that mousemove is triggered (highlight appears), but click is not triggered (the value is not selected).

related: element-plus/element-plus/issues/19127 This suggests that the bug may not have been fixed. Previously, only the scrollbar was modified. 
https://ixyz.org:444/ScreenRecording_04-13-2026%2001-04-06_1.MP4
https://ixyz.org:444/ScreenRecording_04-13-2026%2001-11-50_1.MP4
This is based on my tests on the official website last night, and the behavior is consistent with what was described above.

as result , I tend to believe that when users try to tap on mobile devices, **maybe** their finger may slightly move (about 1–2px). Or WebKit may interpret it as a finger movement rather than a tap. This can explain all the issues.

btea:
Thank you for your detailed explanation and testing. Will the current fix completely resolve the issue you are experiencing?

btea (approved):

rzzf (approved):

PR #24057, opened 2026-04-13
fix(components): [color-picker] add focusOnShow prop to control focus

### LINKED ISSUE #24055: [Component] [color-picker] ColorPicker 颜色选择器 点击后期望不对输入框进行聚焦
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.32`
- Browser / OS: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/146.0.0.0 Safari/537.36`
- Build Tool: `Vite`

### Related Component

- `el-color-picker`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHNjcmlwdCBzZXR1cCBsYW5nPVwidHNcIj5cbmltcG9ydCB7IHJlZiB9IGZyb20gJ3Z1ZSdcblxuY29uc3QgY29sb3IgPSByZWYoJyNmZjQ1MDAnKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPGVsLWNvbG9yLXBpY2tlciB2LW1vZGVsPVwiY29sb3JcIiAvPlxuPC90ZW1wbGF0ZT5cbiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQGxhdGVzdC9kaXN0L2luZGV4LmNzcycsICdodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvdGhlbWUtY2hhbGsvZGFyay9jc3MtdmFycy5jc3MnXS5tYXAoKHN0eWxlKSA9PiB7XG4gICAgcmV0dXJuIG5ldyBQcm9taXNlKChyZXNvbHZlLCByZWplY3QpID0+IHtcbiAgICAgIGNvbnN0IGxpbmsgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCdsaW5rJylcbiAgICAgIGxpbmsucmVsID0gJ3N0eWxlc2hlZXQnXG4gICAgICBsaW5rLmhyZWYgPSBzdHlsZVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdsb2FkJywgcmVzb2x2ZSlcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignZXJyb3InLCByZWplY3QpXG4gICAgICBkb2N1bWVudC5ib2R5LmFwcGVuZChsaW5rKVxuICAgIH0pXG4gIH0pXG4gIHJldHVybiBQcm9taXNlLmFsbFNldHRsZWQoc3R5bGVzKVxufVxuIiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IFwiRVNOZXh0XCIsXG4gICAgXCJqc3hcIjogXCJwcmVzZXJ2ZVwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwidHlwZXNcIjogW1wiZWxlbWVudC1wbHVzL2dsb2JhbC5kLnRzXCJdLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcImFsbG93SnNcIjogdHJ1ZSxcbiAgICBcImNoZWNrSnNcIjogdHJ1ZVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9ydW50aW1lLWRvbUBsYXRlc3QvZGlzdC9ydW50aW1lLWRvbS5lc20tYnJvd3Nlci5qc1wiLFxuICAgIFwiQHZ1ZS9zaGFyZWRcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvc2hhcmVkQGxhdGVzdC9kaXN0L3NoYXJlZC5lc20tYnVuZGxlci5qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L1wiLFxuICAgIFwiQGVsZW1lbnQtcGx1cy9pY29ucy12dWVcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlQDIvZGlzdC9pbmRleC5taW4uanNcIlxuICB9LFxuICBcInNjb3Blc1wiOiB7fVxufSIsIl9vIjp7fX0=)

### Steps to reproduce

点击面板激活时

### What is Expected?

不对面板内部的输入框自动聚焦

对面板内部的输入框自动聚焦了

### Additional comments

在移动端下，因为面板内部的输入框自动聚焦了，导致自动调起输入法，比较影响体验，希望可以不自动聚焦，或者更改为配置来控制是否自动聚焦

LostElkByte:
#### What this PR does
 
Add a `focus-on-show` prop to control which element receives focus when the color picker panel opens.
 
| Value | Behavior |
|---|---|
| `'input'` (default) | Focus the text input (backward compatible) |
| `'sv-cursor'` | Focus the SV panel cursor (won't trigger mobile keyboard) |
| `false` | No focus |
 
Closes #24055
 
#### Why
 
The previous behavior always focused `el-input` on panel open, which triggered the virtual keyboard on mobile devices, covering the panel and breaking the UX.
 
Now users can set `focus-on-show="sv-cursor"` or `:focus-on-show="false"` to avoid this. The SV cursor is a good alternative because:
 
- **Not a form element** — won't trigger mobile virtual keyboard
- **Primary interaction area** — the SV panel is what users interact with most
- **A11y friendly** — already has `tabindex="0"`, `role="slider"`, and full ARIA attributes

## Summary by CodeRabbit

* **New Features**
  * Added a focus-on-show option for the color picker allowing selection of which element receives focus when the panel opens: input field, saturation/value cursor, or none (defaults to input).

* **Documentation**
  * API docs updated to describe the focus-on-show option and its default behavior.


rzzf:
No, autofocus is still meaningful on desktop

keeplearning66 (commented):
Thank you for your contribution, but we can’t make this change directly, as it would affect accessibility. Please refer to #22483.

rzzf:
Perhaps a better approach would be to move focus to the first focusable non-form element, such as `.el-color-svpanel__cursor`, or provide a prop so users can decide the focus behavior themselves.

LostElkByte:
I’ve been thinking about the autofocus behavior of the ColorPicker, and I believe it should be reconsidered for two main reasons:
First, on touch devices, autofocusing the input forces the system keyboard to pop up. This blocks the color palette and stops users from seeing the colors, forcing them to manually close the keyboard. It really breaks the user flow. Second, the primary goal of a color picker is visual selection. The focus ring caused by autofocus just distracts users from the colors.
At first, I thought about adding an autofocus-input prop so developers could turn it on or off. But after checking other libraries like Ant Design, I saw they don’t autofocus the input at all, so I decided to remove it directly in my initial plan. That said, if we want to keep the autofocus logic, maybe adding an autofocus-input prop is a better way to give developers flexible control on different devices.
From a design perspective, autofocusing here doesn't make much sense. Users usually look at the palette first, not type. If they want to type, clicking the input takes almost no effort. But if they just want to pick a color, having to close the keyboard and ignore the focus ring is a high-cost annoyance.

rzzf:
First, this is a PC-oriented UI library. More importantly, have you considered accessibility (a11y)? In some regions, providing accessibility support is mandatory, and it is even written into law.

LostElkByte:
> First, this is a PC-oriented UI library. More importantly, have you considered accessibility (a11y)? In some regions, providing accessibility support is mandatory, and it is even written into law.

Oh right, A11y! Totally agree! In that case, maybe we can handle focus based on the trigger: for keyboard users, focusing the palette handler is actually better since they can use arrow keys to tune colors immediately. For mouse or touch, we skip auto-focus to prevent the keyboard from blocking the panel. It’s a 'visual-first' approach—a win-win for both accessibility and mobile UX.

LostElkByte:
I feel that even though it's primarily a PC component library, it's often used on hybrid or touch devices, like touchscreen laptops and large tablets.

LostElkByte:
I'm not an expert on A11y, so this is just my perspective. I'd love to hear your thoughts and have you guys evaluate it!

rzzf:
> Perhaps a better approach would be to move focus to the first focusable non-form element, such as `.el-color-svpanel__cursor`

I think this is a good approach. Focusing on a non-form element avoids triggering the virtual keyboard while keeping accessibility unaffected.

btea:
I think we could add a configuration to keep the default settings, which would also help users resolve the issue of the pop-up keyboard on mobile devices. wdyt? @tolking

LostElkByte:
> > Perhaps a better approach would be to move focus to the first focusable non-form element, such as `.el-color-svpanel__cursor`
> 
> I think this is a good approach. Focusing on a non-form element avoids triggering the virtual keyboard while keeping accessibility unaffected.

The following shows the effect of changing the focus target from el-input to the SV panel cursor (cursorRef) when the panel opens:

image

rzzf (approved):

btea:
I believe the current implementation is also a breaking change, as it causes the input to lose focus by default.

LostElkByte:
> I believe the current implementation is also a breaking change, as it causes the input to lose focus by default.

add focusOnShow prop
```
focusOnShow?: 'input' | 'sv-cursor' | false
```
'input'（default）

LostElkByte:
> I believe the current implementation is also a breaking change, as it causes the input to lose focus by default.

different approach? We could add a focus-on-show prop to control which element gets focus when the panel opens.
We’ll keep the default as el-input, but since that pops the keyboard on mobile and blocks the panel, users can just set it to 'sv-cursor' or false to fix it. Does that sound like a good compromise?

btea:
cc @element-plus/backers

tolking:

Reasons for focusing on the input:

1. To facilitate user input;
2. To move the focus to the dropdown panel;

This modification may increase the difficulty for keyboard users to input colors. 

If `:focus-on-show="false"` is configured, the focus will not move when the dropdown panel opens, and keyboard users will not be able to input colors.

Adding such an attribute is probably not a good idea; I would recommend considering using the `ColorPickerPanel` component or creating your own custom implementation.

PR #24066, opened 2026-04-14
docs: implement creating new issues for a specifc component

micaiguai:
Here is the effect.

https://github.com/user-attachments/assets/4a242a9c-da73-4bae-98f8-64f819b92ea9

Related: https://github.com/element-plus/element-plus-issue-helper/pull/4

## Summary by CodeRabbit

* **New Features**
  * Added a "Create new issue" button to component documentation pages, enabling users to report issues directly with pre-filled component context.

### DIFF
--- a/docs/.vitepress/crowdin/en-US/component/component-meta.json
+++ b/docs/.vitepress/crowdin/en-US/component/component-meta.json
@@ -1,5 +1,6 @@
 {
   "changelog": "Changelog",
   "view-full-changelog": "View full changelog",
+  "create-new-issue": "Create New issue",
   "open-issues": "Open issues"
 }
--- a/docs/.vitepress/crowdin/en-US/component/component-meta.json
+++ b/docs/.vitepress/crowdin/en-US/component/component-meta.json
@@ -1,6 +1,6 @@
 {
   "changelog": "Changelog",
   "view-full-changelog": "View full changelog",
-  "create-new-issue": "Create New issue",
+  "create-new-issue": "Create new issue",
   "open-issues": "Open issues"
 }
--- a/docs/.vitepress/vitepress/components/globals/vp-component-meta.vue
+++ b/docs/.vitepress/vitepress/components/globals/vp-component-meta.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { computed, ref } from 'vue'
-import { Clock, Loading, Warning } from '@element-plus/icons-vue'
+import { CirclePlus, Clock, Loading, Warning } from '@element-plus/icons-vue'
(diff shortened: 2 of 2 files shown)


rzzf:
This depends on whether issue-help supports it first, right?

btea:
IMO, this is unnecessary. When you discover a problem with a component, the first thing you should do is search GitHub for similar issues, and only create an issue after confirming that no such issue exists. This avoids creating many duplicate issues.

micaiguai:
> This depends on whether issue-help supports it first, right?

Yes

rzzf:
Yes, searching for existing issues before deciding to create a new one is a best practice. Therefore, this feature is not necessary. Thanks for your contribution.

micaiguai:
> IMO, this is unnecessary. When you discover a problem with a component, the first thing you should do is search GitHub for similar issues, and only create an issue after confirming that no such issue exists. This avoids creating many duplicate issues.

The approach you've outlined makes sense. I'd like to propose an alternative set of steps.

1. When a developer finds an issue, they open the `element-plus` documentation to check its usage.
2. After confirming this is a bug, they can quickly filter and list issues via the `Open issues` section.
3. If no related issues are found, they can then create a new issue via `Create new issue`, with auto-filled fields.

I believe this more user-friendly approach will lower the barrier for developers to submit new issues, ultimately helping to strengthen the project's robustness.

If you're worried about developers skipping duplicate issue checks, I suggest adding a required confirmation checkbox in the `issues-helper` with the following label: **I have confirmed that no related issues already exist**

PS: This feature is implemented in the following UI libraries, and I believe this shortcut is reasonable.
- https://vuetifyjs.com/en/components/buttons
- https://ant.design/components/button

rzzf:
> 1. When a developer finds an issue, they open the `element-plus` documentation to check its usage.
> 2. After confirming this is a bug, they can quickly filter and list issues via the `Open issues` section.
> 3. If no related issues are found, they can then create a new issue via `Create new issue`, with auto-filled fields.

Since the user has already navigated to the GitHub Issues page via "Open issues", is it still necessary to return to the component documentation to create an issue?

micaiguai:
> > 1. When a developer finds an issue, they open the `element-plus` documentation to check its usage.
> > 2. After confirming this is a bug, they can quickly filter and list issues via the `Open issues` section.
> > 3. If no related issues are found, they can then create a new issue via `Create new issue`, with auto-filled fields.
> 
> Since the user has already navigated to the GitHub Issues page via "Open issues", is it still necessary to return to the component documentation to create an issue?

Thanks for your patient explanation. I understood.

PR #24179, opened 2026-04-30
fix(build): generate correct DTS for all non-entry files

rzzf:
resolves:
- https://github.com/element-plus/element-plus/pull/24178
- https://github.com/element-plus/element-plus/pull/24116
- https://github.com/element-plus/element-plus/issues/23736

Based on https://github.com/sxzz/rolldown-plugin-dts#entry

## Summary by CodeRabbit

* **Chores**
  * Updated build toolchain dependencies to newer pinned versions for more predictable, reproducible builds.
  * Refined type-definition generation: tightened which source files are included and updated plugin configuration to improve consistency and clearer exclusions.

### DIFF
--- a/internal/build/src/tasks/types-definitions.ts
+++ b/internal/build/src/tasks/types-definitions.ts
@@ -22,19 +22,11 @@ const external = [/^@floating-ui/, /^@vue/, /^vue/, /^csstype/, ...pkgExternal]
 
 export async function generateTypesDefinitions() {
   const input = excludeFiles(
-    await glob(
-      [
-        '**/index.ts',
-        'locale/lang/*.ts',
-        'components/tree-v2/src/virtual-tree.ts',
-        '!**/style/index.ts',
-      ],
-      {
-        cwd: pkgRoot,
-        absolute: true,
-        onlyFiles: true,
-      }
-    )
+    await glob(['**/*.{ts,tsx,vue}', '!**/style/*.ts'], {
+      cwd: pkgRoot,
+      absolute: true,
+      onlyFiles: true,
+    })
   )
   const options: BuildOptions = {
     input,
--- a/internal/build/package.json
+++ b/internal/build/package.json
@@ -22,8 +22,8 @@
     "lodash": "^4.18.1",
     "magic-string": "^0.30.21",
     "oxc-parser": "^0.110.0",
-    "rolldown": "^1.0.0-rc.6",
-    "rolldown-plugin-dts": "^0.22.2",
+    "rolldown": "^1.0.0",
+    "rolldown-plugin-dts": "^0.25.0",
     "tinyglobby": "^0.2.16"
   },
   "devDependencies": {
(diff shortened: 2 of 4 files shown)


rzzf:
@zhixiaotong 你可以试一下这个更改对你的项目有效吗？

rzzf:
The change in the full bundle size is caused by the upgrade of Rolldown

zhixiaotong:
我看到了，我稍后试一下~

zhixiaotong:
image
似乎不行，我用pnpm下载，可能是本地已经有2.13.7的版本，导致无法查看变化，于是用npm在另一个地方下载了，这个没看到里面有dts文件

rzzf:
@zhixiaotong 有删除node_modules之后重新安装吗？

zhixiaotong:
npm不会受pnpm缓存影响，但我打开npm下载的构建物，没有dts呢

internal/build/src/tasks/types-definitions.ts
```diff
@@ -44,6 +52,14 @@ export async function generateTypesDefinitions() {
       target,
     },
     plugins: dts({
+      cwd: pkgRoot,
+      entry: [
```
btea: Is this `entry` and `input` a duplicate?

rzzf: If we don’t need fine-grained control over which files should generate DTS, then this configuration is essentially redundant.

However, the issue is that when `input` includes too many files, the build becomes significantly slower.

So the approach we’re currently using is to keep `input` minimal (only `index.ts`) and rely on its module graph to generate DTS for other modules.

But this introduces another problem: if a module is not treated as an `entry` (or not explicitly marked as one), its default export may be rewritten as `export { _default }` instead of the expected `export { _default as default }`.

Given this, I think a “minimal input + full entry” strategy is a good balance. See: https://github.com/sxzz/rolldown-plugin-dts#entry

In theory, taking `components` as an example, we only need to provide `index.ts` as the `input`, together with a full set of `entry`.
In that case, all modules within the module graph of `index.ts` should have their corresponding DTS generated.
However, using “full `entry` + `emitDtsOnly`” seems to introduce another issue. I’m not sure whether this behavior is expected for `rolldown-plugin-dts`. See: https://github.com/sxzz/rolldown-plugin-dts/issues/231

rzzf:
Now, [rolldown 1.0](https://github.com/rolldown/rolldown/releases/tag/v1.0.0)
 was released. After upgrading the dependencies, builds with full `input` entries are no longer excessively slow.

btea:
Strange, I fetched the latest code from the current pull request locally, and the build process takes a very long time. 🤔 

It's not that the build process takes a long time; it's that the build process fails and gets stuck.

rzzf:
Based on the CI results, the build time after the upgrade does not differ much from the previous run. I'm not sure what exactly caused this, but on my machine, the build was almost impossible to complete before the upgrade. I once waited for more than 20 minutes and it still didn't finish, whereas now the result is similar to CI and completes in around 10-20 seconds.

- latest build: https://github.com/element-plus/element-plus/actions/runs/25533838024/job/74945456753?pr=24179
image


- last build: https://github.com/element-plus/element-plus/actions/runs/25250519279/job/74041782704?pr=24179
image

- local build
image

btea (approved):

PR #24249, opened 2026-05-15
fix(components): [menu] fix hover-bg-color calculation error when usi…

### LINKED ISSUE #4606: [Bug Report] 使用 css vars 代替props
### Existing Component
Yes

### Component Name
Menu

### Description
当我传入另一个变量的时候：
``` html
        <ElMenu
          background-color="var(--mk-aside-bg-color)"
          text-color="var(--mk-aside-text-color)"
          active-text-color="var(--mk-aside-active-color)"
        />
```
` --el-menu-hover-bg-color:rgb(204, 204, 204); ` 计算错误:
``` html
<ul 
  role="menubar" 
  class="el-menu" 
  style="--el-menu-text-color:var(--mk-aside-text-color); --el-menu-hover-text-color:var(--mk-aside-text-color); --el-menu-bg-color:var(--mk-aside-bg-color); --el-menu-hover-bg-color:rgb(204, 204, 204); --el-menu-active-color:var(--mk-aside-active-color);"
></ul>
```
源码是根据 `props.backgroundColor` 计算，建议改成 `window.getComputedStyle(elmenu).getPropertyValue("--el-menu-bg-color")`
https://github.com/element-plus/element-plus/blob/dev/packages/components/menu/src/use-menu-css-var.ts

Kinderi:
## Description

Fixes the issue where `hover-bg-color` is calculated incorrectly when using CSS variables for Menu component colors.

## Issue

When passing CSS variables to Menu component:

```vue
<ElMenu background-color="var(--mk-aside-bg-color)" ... />
```

The `hover-bg-color` was incorrectly calculated as `rgb(204, 204, 204)` because the source code used `TinyColor` to process the color value, which could not recognize CSS variables like `var(--mk-aside-bg-color)`.

**Before:**
```css
--el-menu-hover-bg-color: rgb(204, 204, 204); /* Incorrect */
```

## Solution

Added CSS variable detection in `use-menu-color.ts`. When the color value starts with `var(`, return an empty string to let CSS handle the hover color itself via CSS variable inheritance.

**After:**
```css
/* When background-color is a CSS var, hover-bg-color will be empty, 
   allowing CSS to compute the correct value or use inherited values */
--el-menu-hover-bg-color: ; 
```

## Files Changed

- `packages/components/menu/src/use-menu-color.ts`

## Additional Context

Fixes #4606

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved menu bar color computation for better handling of custom color values and edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/element-plus/element-plus/pull/24249)

### DIFF
--- a/packages/components/menu/src/use-menu-color.ts
+++ b/packages/components/menu/src/use-menu-color.ts
@@ -6,7 +6,9 @@ import type { MenuProps } from './menu'
 export default function useMenuColor(props: MenuProps) {
   const menuBarColor = computed(() => {
     const color = props.backgroundColor
-    return color ? new TinyColor(color).shade(20).toString() : ''
+    if (!color) return ''
+    if (typeof color === 'string' && color.startsWith('var(')) return ''
+    return new TinyColor(color).shade(20).toString()
   })
   return menuBarColor
 }


rzzf:
Thanks!

As far as I know, this API is being deprecated, so I don’t think it’s worth spending more effort maintaining it.

PR #24294, opened 2026-05-22
docs(components): update pagination examples to use v-model

### LINKED ISSUE #7383: [Feature Request] [Docs] [pagination] Pagination Events
Feature type: **`Docs`**

## Details

- Enhanced compone nt name: `pagination`

## Description

The **el-pagination**`s documentation has suggested that the Event of this component has been deprecated and will be deleted, but the examples in the documentation still use these Events, I hope that the related documentation can be modified without using these Events.

In addition, how to monitor only using **v-model**, using watchEffect?

### Additional comments

Events above are not recommended(but are still supported for compatible reason), better choice is to use the two-way data binding by v-model.

image
image

william-xue:
## Summary

Updated pagination documentation examples to use `v-model` for two-way data binding instead of the deprecated events (`@size-change`, `@current-change`).

## Changes

- **docs/examples/pagination/more-elements.vue**: Removed all `@size-change` and `@current-change` event handlers from the four pagination examples, and removed the unused `handleSizeChange` and `handleCurrentChange` functions.
- **docs/en-US/component/pagination.md**: Updated the description for the "More elements" example to reference `v-model` instead of the deprecated events.

## Why

The pagination documentation already warns that events are deprecated and recommends using `v-model`:
> "Events above are not recommended(but are still supported for compatible reason), better choice is to use the two-way data binding by v-model."

However, the examples were still using the deprecated events, which could confuse users (as noted in #7383).

## Test Plan

- Verify the documentation builds successfully
- Verify the examples render correctly in the docs site
- No component logic changes, only documentation updates

Fixes #7383

## Summary by CodeRabbit

* **Documentation**
  * Updated pagination documentation examples to demonstrate v-model-based two-way binding for page size and current page changes.

* **Refactor**
  * Simplified pagination examples by removing event handler callbacks in favor of direct binding approach.

[Review Change Stack](https://app.coderabbit.ai/change-stack/element-plus/element-plus/pull/24294?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/en-US/component/pagination.md
+++ b/docs/en-US/component/pagination.md
@@ -55,7 +55,7 @@ pagination/auto-hide-pagination
 
 Add more modules based on your scenario.
 
-:::demo This example is a complete use case. It uses `size-change` and `current-change` event to handle page size changes and current page changes. `page-sizes` accepts an array of integers, each of which represents a different page size in the `sizes` select options, e.g. `[100, 200, 300, 400]` indicates that the select will have four options: 100, 200, 300 or 400 items per page.
+:::demo This example is a complete use case. It uses `v-model` for two-way data binding to handle page size changes and current page changes. `page-sizes` accepts an array of integers, each of which represents a different page size in the `sizes` select options, e.g. `[100, 200, 300, 400]` indicates that the select will have four options: 100, 200, 300 or 400 items per page.
 
 pagination/more-elements
 
--- a/docs/examples/pagination/more-elements.vue
+++ b/docs/examples/pagination/more-elements.vue
@@ -27,8 +27,6 @@
       :background="background"
       layout="total, prev, pager, next"
       :total="1000"
-      @size-change="handleSizeChange"
-      @current-change="handleCurrentChange"
     />
   </div>
   <div class="demo-pagination-block">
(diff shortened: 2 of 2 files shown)


docs/en-US/component/pagination.md
```diff
@@ -55,7 +55,7 @@ pagination/auto-hide-pagination
 
 Add more modules based on your scenario.
 
-:::demo This example is a complete use case. It uses `size-change` and `current-change` event to handle page size changes and current page changes. `page-sizes` accepts an array of integers, each of which represents a different page size in the `sizes` select options, e.g. `[100, 200, 300, 400]` indicates that the select will have four options: 100, 200, 300 or 400 items per page.
+:::demo This example is a complete use case. It uses `v-model` for two-way data binding to handle page size changes and current page changes. `page-sizes` accepts an array of integers, each of which represents a different page size in the `sizes` select options, e.g. `[100, 200, 300, 400]` indicates that the select will have four options: 100, 200, 300 or 400 items per page.
```
patienry, line 58: It is reasonable to use the recommended approach in the description.
However, I don't think it's necessary to remove the usage of `size-change` and `current-change` in the code below. According to the documentation, event usage is currently just not recommended, but these events are not being deprecated. **This example is a complete use case**, so perhaps the usage of these events can be retained. 🤔

rzzf:
This PR appears to have been opened automatically by an agent, but it lacks human follow-up. In addition, the value of the proposed changes seems fairly limited, so I'm going to close it for now.

PR #24329, opened 2026-05-26
fix(components): [time-select] guard invalid time values

### LINKED ISSUE #24328: [Component] [time-select] TimeSelect can hang when start or end time is invalid
Bug Type: **`Component`**

## Environment

- Vue Version: `3.6.0-beta.12`
- Element Plus Version: `2.14.0`
- Browser / OS: `UserAgent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/148.0.0.0 Safari/537.36 Edg/148.0.0.0`
- Build Tool: `Vite`

### Related Component

- `el-time-select`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtdGltZS1zZWxlY3Qgc3RhcnQ9XCJhYmM6MDBcIiBlbmQ9XCIxODowMFwiIHN0ZXA9XCIwMDozMFwiIC8+XG48L3RlbXBsYXRlPiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQGxhdGVzdC9kaXN0L2luZGV4LmNzcycsICdodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvdGhlbWUtY2hhbGsvZGFyay9jc3MtdmFycy5jc3MnXS5tYXAoKHN0eWxlKSA9PiB7XG4gICAgcmV0dXJuIG5ldyBQcm9taXNlKChyZXNvbHZlLCByZWplY3QpID0+IHtcbiAgICAgIGNvbnN0IGxpbmsgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCdsaW5rJylcbiAgICAgIGxpbmsucmVsID0gJ3N0eWxlc2hlZXQnXG4gICAgICBsaW5rLmhyZWYgPSBzdHlsZVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdsb2FkJywgcmVzb2x2ZSlcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignZXJyb3InLCByZWplY3QpXG4gICAgICBkb2N1bWVudC5ib2R5LmFwcGVuZChsaW5rKVxuICAgIH0pXG4gIH0pXG4gIHJldHVybiBQcm9taXNlLmFsbFNldHRsZWQoc3R5bGVzKVxufVxuIiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IFwiRVNOZXh0XCIsXG4gICAgXCJqc3hcIjogXCJwcmVzZXJ2ZVwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwidHlwZXNcIjogW1wiZWxlbWVudC1wbHVzL2dsb2JhbC5kLnRzXCJdLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcImFsbG93SnNcIjogdHJ1ZSxcbiAgICBcImNoZWNrSnNcIjogdHJ1ZVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9ydW50aW1lLWRvbUBsYXRlc3QvZGlzdC9ydW50aW1lLWRvbS5lc20tYnJvd3Nlci5qc1wiLFxuICAgIFwiQHZ1ZS9zaGFyZWRcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvc2hhcmVkQGxhdGVzdC9kaXN0L3NoYXJlZC5lc20tYnVuZGxlci5qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L1wiLFxuICAgIFwiQGVsZW1lbnQtcGx1cy9pY29ucy12dWVcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0BlbGVtZW50LXBsdXMvaWNvbnMtdnVlQDIvZGlzdC9pbmRleC5taW4uanNcIlxuICB9LFxuICBcInNjb3Blc1wiOiB7fVxufSIsIl9vIjp7fX0=)

### Steps to reproduce

```vue
<template>
  <el-time-select start="abc:00" end="18:00" step="00:30" />
</template>
```

### What is Expected?

Invalid time strings should not cause option generation to hang. The component should safely treat invalid parsed times as invalid input and render no generated options for that invalid range.

The page can become unresponsive while TimeSelect is generating options.
﻿

### Additional comments

`packages/components/time-select/src/utils.ts` parses any string with at least two `:`-separated pieces into a `Time` object:

```ts
let hours = Number.parseInt(values[0], 10)
const minutes = Number.parseInt(values[1], 10)
return {
  hours,
  minutes,
}
```

For `abc:00`, this returns:

```ts
{ hours: NaN, minutes: 0 }
```

Then `compareTime('NaN:00', '18:00')` returns `-1`, so the option-generation loop keeps running:

```ts
while (current && end.value && compareTime(current, end.value) <= 0) {
  current = nextTime(current, step.value!)
}
```

`nextTime` keeps producing values such as `NaN:00` and `NaN:30`, so the loop never reaches the end time.

guozi9999:

## Summary
fix https://github.com/element-plus/element-plus/issues/24328

- Treat parsed `NaN` hour or minute values as invalid time input.
- Add a regression test for an invalid `start` value.

## Problem

`parseTime` accepted strings like `abc:00` because they contained at least two colon-separated pieces. This produced a `Time` object with `hours: NaN`.

That invalid value reached TimeSelect option generation, where `compareTime` kept returning `-1` and `nextTime` kept producing `NaN:*` values. As a result, the option-generation loop could run forever and freeze the page.

## Solution

Guard `parseTime` against invalid numeric parts:

```ts
if (Number.isNaN(hours) || Number.isNaN(minutes)) return null
```

This keeps invalid time strings out of the option-generation loop.

## Tests

```bash
pnpm exec vitest run packages/components/time-select/__tests__/time-select.test.tsx
```

## Notes

The regression test verifies that TimeSelect does not generate options when `start` is invalid.

## Summary by CodeRabbit

* **Bug Fixes**
  * TimeSelect now validates malformed start/end inputs, emits a debug warning, and reliably falls back to sensible default start (09:00) and end (18:00) times so displayed time options remain correct.

* **Tests**
  * Added tests covering fallback behavior and correct time-item generation when start/end props are invalid or negative.

### DIFF
--- a/packages/components/time-select/__tests__/time-select.test.tsx
+++ b/packages/components/time-select/__tests__/time-select.test.tsx
@@ -8,25 +8,12 @@ import Select from '@element-plus/components/select'
 import { ElForm, ElFormItem } from '@element-plus/components/form'
 import TimeSelect from '../src/time-select.vue'
 
-import type { VNode } from 'vue'
-import type { VueWrapper } from '@vue/test-utils'
-
 dayjs.extend(customParseFormat)
 
 const { Option } = Select
 
-const mountedWrappers: VueWrapper[] = []
-const _mount = (render: () => VNode, options?: Parameters<typeof mount>[1]) => {
-  const wrapper = mount(render, options)
-  mountedWrappers.push(wrapper)
-  return wrapper
-}
-
-afterEach(async () => {
-  mountedWrappers.forEach((wrapper) => wrapper.unmount())
-  mountedWrappers.length = 0
-  document.body.innerHTML = ''
-  await new Promise((resolve) => setTimeout(resolve, 0))
+afterEach(() => {
+  document.documentElement.innerHTML = ''
 })
 
 const WRAPPER_CLASS_NAME = 'el-select__wrapper'
@@ -34,7 +21,7 @@ const PLACEHOLDER_CLASS_NAME = 'el-select__placeholder'
 
 describe('TimeSelect', () => {
   it('create', async () => {
-    const wrapper = _mount(() => (
+    const wrapper = mount(() => (
       <TimeSelect style={{ color: 'red' }} class="customClass" />
     ))
 
@@ -44,7 +31,7 @@ describe('TimeSelect', () => {
   })
 
   it('should show clear btn on focus', async () => {
-    const wrapper = _mount(() => <TimeSelect modelValue="08:30" clearable />)
+    const wrapper = mount(() => <TimeSelect modelValue="08:30" clearable />)
     const input = wrapper.find('input')
     await input.trigger('blur')
     await input.trigger('focus')
@@ -53,7 +40,7 @@ describe('TimeSelect', () => {
 
   it('set default value', async () => {
     const value = ref('14:30')
-    const wrapper = _mount(() => <TimeSelect v-model={value.value} />)
+    const wrapper = mount(() => <TimeSelect v-model={value.value} />)
 
     const input = wrapper.find('input')
     input.trigger('blur')
@@ -64,7 +51,7 @@ describe('TimeSelect', () => {
   })
 
   it('set minTime', async () => {
-    const wrapper = _mount(() => <TimeSelect minTime="14:30" />)
+    const wrapper = mount(() => <TimeSelect minTime="14:30" />)
 
     const input = wrapper.find('input')
     input.trigger('blur')
@@ -76,7 +63,7 @@ describe('TimeSelect', () => {
   })
 
   it('set maxTime', async () => {
-    const wrapper = _mount(() => <TimeSelect maxTime="14:30" />)
+    const wrapper = mount(() => <TimeSelect maxTime="14:30" />)
 
     const input = wrapper.find('input')
     input.trigger('blur')
@@ -88,7 +75,7 @@ describe('TimeSelect', () => {
 
   it('set value update', async () => {
     const value = ref('10:00')
-    const wrapper = _mount(() => <TimeSelect v-model={value.value} />)
+    const wrapper = mount(() => <TimeSelect v-model={value.value} />)
 
     await nextTick()
     const input = wrapper.find('input')
@@ -107,7 +94,7 @@ describe('TimeSelect', () => {
 
   it('update value', async () => {
     const value = ref('10:00')
-    const wrapper = _mount(() => <TimeSelect v-model={value.value} />)
+    const wrapper = mount(() => <TimeSelect v-model={value.value} />)
 
     await nextTick()
     const vm = wrapper.findComponent({ name: 'ElTimeSelect' }).vm
@@ -128,7 +115,7 @@ describe('TimeSelect', () => {
   it('set disabled', async () => {
     const value = ref('10:00')
     const disabled = ref(false)
-    const wrapper = _mount(() => (
+    const wrapper = mount(() => (
       <TimeSelect v-model={value.value} disabled={disabled.value} />
     ))
 
@@ -143,7 +130,7 @@ describe('TimeSelect', () => {
   it('set editable', async () => {
     const value = ref('10:00')
     const editable = ref(false)
-    const wrapper = _mount(() => (
+    const wrapper = mount(() => (
       <TimeSelect v-model={value.value} editable={editable.value} />
     ))
 
@@ -156,7 +143,7 @@ describe('TimeSelect', () => {
   })
 
   it('should include end time', async () => {
-    const wrapper = _mount(() => (
+    const wrapper = mount(() => (
       <TimeSelect
         start="00:00"
         step="00:05"
@@ -175,7 +162,7 @@ describe('TimeSelect', () => {
   })
 
   it('should not duplicate end time when includeEndTime with custom format', async () => {
-    const wrapper = _mount(() => (
+    const wrapper = mount(() => (
       <TimeSelect
         start="08:30"
         step="00:15"
@@ -195,7 +182,7 @@ describe('TimeSelect', () => {
   })
 
   it('should not include end time', async () => {
-    const wrapper = _mount(() => (
+    const wrapper = mount(() => (
       <TimeSelect start="00:00" step="00:05" end="23:59" />
     ))
     const select = wrapper.findComponent({ name: 'ElTimeSelect' })
@@ -209,7 +196,7 @@ describe('TimeSelect', () => {
   })
 
   it('should include end whenever includeEndTime is false', async () => {
-    const wrapper = _mount(() => (
+    const wrapper = mount(() => (
       <TimeSelect start="00:10" end="00:20" step="00:02" />
     ))
     const select = wrapper.findComponent({ name: 'ElTimeSelect' })
@@ -223,7 +210,7 @@ describe('TimeSelect', () => {
   })
 
   it('ref focus', async () => {
-    const wrapper = _mount(() => <TimeSelect />, {
+    const wrapper = mount(() => <TimeSelect />, {
       attachTo: document.body,
     })
 
@@ -240,7 +227,7 @@ describe('TimeSelect', () => {
   })
 
   it('ref blur', async () => {
-    const wrapper = _mount(() => <TimeSelect />, {
+    const wrapper = mount(() => <TimeSelect />, {
       attachTo: document.body,
     })
 
@@ -259,7 +246,7 @@ describe('TimeSelect', () => {
 
   it('set format', async () => {
     const value = ref('')
-    const wrapper = _mount(() => (
+    const wrapper = mount(() => (
       <TimeSelect
         v-model={value.value}
         start="13:00"
@@ -277,13 +264,13 @@ describe('TimeSelect', () => {
   })
 
   it('should pass name to inner input', () => {
-    const wrapper = _mount(() => <TimeSelect name="timeSelectName" />)
+    const wrapper = mount(() => <TimeSelect name="timeSelectName" />)
     const input = wrapper.find('input')
     expect(input.attributes('name')).toBe('timeSelectName')
   })
 
   it('should fallback to default step when step is 00:00', async () => {
-    const wrapper = _mount(() => (
+    const wrapper = mount(() => (
       <TimeSelect start="09:00" end="10:00" step="00:00" />
     ))
     const input = wrapper.find('input')
@@ -294,7 +281,7 @@ describe('TimeSelect', () => {
   })
 
   it('should fallback to default start time when start time is invalid', async () => {
-    const wrapper = _mount(() => <TimeSelect start="abc:00" end="18:00" />)
+    const wrapper = mount(() => <TimeSelect start="abc:00" end="18:00" />)
 
     const input = wrapper.find('input')
     await input.trigger('click')
@@ -305,7 +292,7 @@ describe('TimeSelect', () => {
   })
 
   it('should fallback to default start time when start time is negative', async () => {
-    const wrapper = _mount(() => <TimeSelect start="-12:00" end="18:00" />)
+    const wrapper = mount(() => <TimeSelect start="-12:00" end="18:00" />)
 
     const input = wrapper.find('input')
     await input.trigger('click')
@@ -316,7 +303,7 @@ describe('TimeSelect', () => {
   })
 
   it('should fallback to default end time when end time is invalid', async () => {
-    const wrapper = _mount(() => <TimeSelect start="17:00" end="abc:00" />)
+    const wrapper = mount(() => <TimeSelect start="17:00" end="abc:00" />)
 
     const input = wrapper.find('input')
     await input.trigger('click')
@@ -327,7 +314,7 @@ describe('TimeSelect', () => {
   })
 
   it('should fallback to default end time when end time is negative', async () => {
-    const wrapper = _mount(() => <TimeSelect start="17:00" end="-12:00" />)
+    const wrapper = mount(() => <TimeSelect start="17:00" end="-12:00" />)
 
     const input = wrapper.find('input')
     await input.trigger('click')
@@ -339,7 +326,7 @@ describe('TimeSelect', () => {
 
   describe('form item accessibility integration', () => {
     it('automatic id attachment', async () => {
-      const wrapper = _mount(() => (
+      const wrapper = mount(() => (
         <ElFormItem label="Foobar" data-test-ref="item">
           <TimeSelect />
         </ElFormItem>
@@ -356,7 +343,7 @@ describe('TimeSelect', () => {
     })
 
     it('specified id attachment', async () => {
-      const wrapper = _mount(() => (
+      const wrapper = mount(() => (
         <ElFormItem label="Foobar" data-test-ref="item">
           <TimeSelect
             // type checking failed as `id` is a fallthrough attribute
@@ -378,7 +365,7 @@ describe('TimeSelect', () => {
     })
 
     it('form item role is group when multiple inputs', async () => {
-      const wrapper = _mount(() => (
+      const wrapper = mount(() => (
         <ElFormItem label="Foobar" data-test-ref="item">
           <TimeSelect />
           <TimeSelect />
@@ -391,7 +378,7 @@ describe('TimeSelect', () => {
     })
 
     it('The disabled state of a component has higher priority than that of a form', async () => {
-      const wrapper = _mount(() => (
+      const wrapper = mount(() => (
         <ElForm disabled>
           <TimeSelect disabled={false} />
         </ElForm>
--- a/packages/components/time-select/__tests__/time-select.test.tsx
+++ b/packages/components/time-select/__tests__/time-select.test.tsx
@@ -8,12 +8,25 @@ import Select from '@element-plus/components/select'
 import { ElForm, ElFormItem } from '@element-plus/components/form'
 import TimeSelect from '../src/time-select.vue'
 
+import type { VNode } from 'vue'
+import type { VueWrapper } from '@vue/test-utils'
+
 dayjs.extend(customParseFormat)
 
 const { Option } = Select
 
-afterEach(() => {
-  document.documentElement.innerHTML = ''
+const mountedWrappers: VueWrapper[] = []
+const _mount = (render: () => VNode, options?: Parameters<typeof mount>[1]) => {
+  const wrapper = mount(render, options)
+  mountedWrappers.push(wrapper)
+  return wrapper
+}
+
+afterEach(async () => {
+  mountedWrappers.forEach((wrapper) => wrapper.unmount())
+  mountedWrappers.length = 0
+  document.body.innerHTML = ''
+  await new Promise((resolve) => setTimeout(resolve, 0))
 })
 
 const WRAPPER_CLASS_NAME = 'el-select__wrapper'
@@ -21,7 +34,7 @@ const PLACEHOLDER_CLASS_NAME = 'el-select__placeholder'
 
 describe('TimeSelect', () => {
   it('create', async () => {
-    const wrapper = mount(() => (
+    const wrapper = _mount(() => (
       <TimeSelect style={{ color: 'red' }} class="customClass" />
     ))
 
@@ -31,7 +44,7 @@ describe('TimeSelect', () => {
   })
 
   it('should show clear btn on focus', async () => {
-    const wrapper = mount(() => <TimeSelect modelValue="08:30" clearable />)
+    const wrapper = _mount(() => <TimeSelect modelValue="08:30" clearable />)
     const input = wrapper.find('input')
     await input.trigger('blur')
     await input.trigger('focus')
@@ -40,7 +53,7 @@ describe('TimeSelect', () => {
 
   it('set default value', async () => {
     const value = ref('14:30')
-    const wrapper = mount(() => <TimeSelect v-model={value.value} />)
+    const wrapper = _mount(() => <TimeSelect v-model={value.value} />)
 
     const input = wrapper.find('input')
     input.trigger('blur')
@@ -51,7 +64,7 @@ describe('TimeSelect', () => {
   })
 
   it('set minTime', async () => {
-    const wrapper = mount(() => <TimeSelect minTime="14:30" />)
+    const wrapper = _mount(() => <TimeSelect minTime="14:30" />)
 
     const input = wrapper.find('input')
     input.trigger('blur')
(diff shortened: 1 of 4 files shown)


keeplearning66 (commented):
Perhaps we can refer to [#23635 (comment)](https://github.com/element-plus/element-plus/pull/23635#pullrequestreview-3773373506) and use the default time when an invalid time is passed in. 🤔 WDYT?

packages/components/time-select/src/time-select.vue
```diff
@@ -87,12 +87,12 @@ const { lang } = useLocale()
 const value = computed(() => props.modelValue)
 const start = computed(() => {
   const time = parseTime(props.start)
-  return time ? formatTime(time) : null
+  return time ? formatTime(time) : DEFAULT_START
```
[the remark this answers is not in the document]
guozi9999: In this case we intentionally fall back to the default time whenever the provided bound cannot be parsed, including an empty string.

guozi9999:
Here is a demo showing the behavior: https://stackblitz.com/edit/nhdidhmk-rdo8lpq3?file=package.json

rzzf:
Maybe we need a warning here, similar to the one used in the `step` prop.

guozi9999:
> Maybe we need a warning here, similar to the one used in the `step` prop.

Thanks for the suggestion. I updated the implementation to follow #23635: invalid `start` and `end` now fall back to `DEFAULT_START` / `DEFAULT_END`, and emit `debugWarn` messages consistent with the invalid `step` fallback.

keeplearning66 (commented):
It looks like there's also an issue when the time is negative. 🤔 [demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtdGltZS1zZWxlY3Qgc3RhcnQ9XCItMTI6MDBcIiBlbmQ9XCIxODowMFwiIHN0ZXA9XCIwMDozMFwiIC8+XG48L3RlbXBsYXRlPiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vcmF3LmVzbS5zaC9wci9lbGVtZW50LXBsdXNAMjQzMjkvZGlzdC9pbmRleC5jc3MnLCAnaHR0cHM6Ly9yYXcuZXNtLnNoL3ByL2VsZW1lbnQtcGx1c0AyNDMyOS90aGVtZS1jaGFsay9kYXJrL2Nzcy12YXJzLmNzcyddLm1hcCgoc3R5bGUpID0+IHtcbiAgICByZXR1cm4gbmV3IFByb21pc2UoKHJlc29sdmUsIHJlamVjdCkgPT4ge1xuICAgICAgY29uc3QgbGluayA9IGRvY3VtZW50LmNyZWF0ZUVsZW1lbnQoJ2xpbmsnKVxuICAgICAgbGluay5yZWwgPSAnc3R5bGVzaGVldCdcbiAgICAgIGxpbmsuaHJlZiA9IHN0eWxlXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2xvYWQnLCByZXNvbHZlKVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdlcnJvcicsIHJlamVjdClcbiAgICAgIGRvY3VtZW50LmJvZHkuYXBwZW5kKGxpbmspXG4gICAgfSlcbiAgfSlcbiAgcmV0dXJuIFByb21pc2UuYWxsU2V0dGxlZChzdHlsZXMpXG59XG4iLCJ0c2NvbmZpZy5qc29uIjoie1xuICBcImNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogXCJFU05leHRcIixcbiAgICBcImpzeFwiOiBcInByZXNlcnZlXCIsXG4gICAgXCJtb2R1bGVcIjogXCJFU05leHRcIixcbiAgICBcIm1vZHVsZVJlc29sdXRpb25cIjogXCJCdW5kbGVyXCIsXG4gICAgXCJ0eXBlc1wiOiBbXCJlbGVtZW50LXBsdXMvZ2xvYmFsLmQudHNcIl0sXG4gICAgXCJhbGxvd0ltcG9ydGluZ1RzRXh0ZW5zaW9uc1wiOiB0cnVlLFxuICAgIFwiYWxsb3dKc1wiOiB0cnVlLFxuICAgIFwiY2hlY2tKc1wiOiB0cnVlXG4gIH0sXG4gIFwidnVlQ29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiAzLjNcbiAgfVxufVxuIiwiUGxheWdyb3VuZE1haW4udnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCBBcHAgZnJvbSAnLi9BcHAudnVlJ1xuaW1wb3J0IHsgc2V0dXBFbGVtZW50UGx1cyB9IGZyb20gJy4vZWxlbWVudC1wbHVzLmpzJ1xuc2V0dXBFbGVtZW50UGx1cygpXG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8QXBwIC8+XG48L3RlbXBsYXRlPlxuIiwiaW1wb3J0LW1hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge1xuICAgIFwidnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AdnVlL3J1bnRpbWUtZG9tQGxhdGVzdC9kaXN0L3J1bnRpbWUtZG9tLmVzbS1icm93c2VyLmpzXCIsXG4gICAgXCJAdnVlL3NoYXJlZFwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9zaGFyZWRAbGF0ZXN0L2Rpc3Qvc2hhcmVkLmVzbS1idW5kbGVyLmpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXNcIjogXCJodHRwczovL3Jhdy5lc20uc2gvcHIvZWxlbWVudC1wbHVzQDI0MzI5L2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwiaHR0cHM6Ly9yYXcuZXNtLnNoL3ByL2VsZW1lbnQtcGx1c0AyNDMyOS9cIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJfbyI6eyJzaG93SGlkZGVuIjp0cnVlLCJzdHlsZVNvdXJjZSI6Imh0dHBzOi8vcmF3LmVzbS5zaC9wci9lbGVtZW50LXBsdXNAMjQzMjkvZGlzdC9pbmRleC5jc3MifX0=)
Maybe we can refer to the validation approach below.

https://github.com/element-plus/element-plus/blob/c8ecbe4b404abe3efb33e95fe7caff36589724f3/packages/components/time-select/src/time-select.vue#L110-L122

guozi9999:
> It looks like there's also an issue when the time is negative. 🤔 [demo](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtdGltZS1zZWxlY3Qgc3RhcnQ9XCItMTI6MDBcIiBlbmQ9XCIxODowMFwiIHN0ZXA9XCIwMDozMFwiIC8+XG48L3RlbXBsYXRlPiIsImVsZW1lbnQtcGx1cy5qcyI6ImltcG9ydCBFbGVtZW50UGx1cyBmcm9tICdlbGVtZW50LXBsdXMnXG5pbXBvcnQgeyBnZXRDdXJyZW50SW5zdGFuY2UgfSBmcm9tICd2dWUnXG5cbmxldCBpbnN0YWxsZWQgPSBmYWxzZVxuYXdhaXQgbG9hZFN0eWxlKClcblxuZXhwb3J0IGZ1bmN0aW9uIHNldHVwRWxlbWVudFBsdXMoKSB7XG4gIGlmIChpbnN0YWxsZWQpIHJldHVyblxuICBjb25zdCBpbnN0YW5jZSA9IGdldEN1cnJlbnRJbnN0YW5jZSgpXG4gIGluc3RhbmNlLmFwcENvbnRleHQuYXBwLnVzZShFbGVtZW50UGx1cylcbiAgaW5zdGFsbGVkID0gdHJ1ZVxufVxuXG5leHBvcnQgZnVuY3Rpb24gbG9hZFN0eWxlKCkge1xuICBjb25zdCBzdHlsZXMgPSBbJ2h0dHBzOi8vcmF3LmVzbS5zaC9wci9lbGVtZW50LXBsdXNAMjQzMjkvZGlzdC9pbmRleC5jc3MnLCAnaHR0cHM6Ly9yYXcuZXNtLnNoL3ByL2VsZW1lbnQtcGx1c0AyNDMyOS90aGVtZS1jaGFsay9kYXJrL2Nzcy12YXJzLmNzcyddLm1hcCgoc3R5bGUpID0+IHtcbiAgICByZXR1cm4gbmV3IFByb21pc2UoKHJlc29sdmUsIHJlamVjdCkgPT4ge1xuICAgICAgY29uc3QgbGluayA9IGRvY3VtZW50LmNyZWF0ZUVsZW1lbnQoJ2xpbmsnKVxuICAgICAgbGluay5yZWwgPSAnc3R5bGVzaGVldCdcbiAgICAgIGxpbmsuaHJlZiA9IHN0eWxlXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2xvYWQnLCByZXNvbHZlKVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdlcnJvcicsIHJlamVjdClcbiAgICAgIGRvY3VtZW50LmJvZHkuYXBwZW5kKGxpbmspXG4gICAgfSlcbiAgfSlcbiAgcmV0dXJuIFByb21pc2UuYWxsU2V0dGxlZChzdHlsZXMpXG59XG4iLCJ0c2NvbmZpZy5qc29uIjoie1xuICBcImNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogXCJFU05leHRcIixcbiAgICBcImpzeFwiOiBcInByZXNlcnZlXCIsXG4gICAgXCJtb2R1bGVcIjogXCJFU05leHRcIixcbiAgICBcIm1vZHVsZVJlc29sdXRpb25cIjogXCJCdW5kbGVyXCIsXG4gICAgXCJ0eXBlc1wiOiBbXCJlbGVtZW50LXBsdXMvZ2xvYmFsLmQudHNcIl0sXG4gICAgXCJhbGxvd0ltcG9ydGluZ1RzRXh0ZW5zaW9uc1wiOiB0cnVlLFxuICAgIFwiYWxsb3dKc1wiOiB0cnVlLFxuICAgIFwiY2hlY2tKc1wiOiB0cnVlXG4gIH0sXG4gIFwidnVlQ29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiAzLjNcbiAgfVxufVxuIiwiUGxheWdyb3VuZE1haW4udnVlIjoiPHNjcmlwdCBzZXR1cD5cbmltcG9ydCBBcHAgZnJvbSAnLi9BcHAudnVlJ1xuaW1wb3J0IHsgc2V0dXBFbGVtZW50UGx1cyB9IGZyb20gJy4vZWxlbWVudC1wbHVzLmpzJ1xuc2V0dXBFbGVtZW50UGx1cygpXG48L3NjcmlwdD5cblxuPHRlbXBsYXRlPlxuICA8QXBwIC8+XG48L3RlbXBsYXRlPlxuIiwiaW1wb3J0LW1hcC5qc29uIjoie1xuICBcImltcG9ydHNcIjoge1xuICAgIFwidnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AdnVlL3J1bnRpbWUtZG9tQGxhdGVzdC9kaXN0L3J1bnRpbWUtZG9tLmVzbS1icm93c2VyLmpzXCIsXG4gICAgXCJAdnVlL3NoYXJlZFwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9zaGFyZWRAbGF0ZXN0L2Rpc3Qvc2hhcmVkLmVzbS1idW5kbGVyLmpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXNcIjogXCJodHRwczovL3Jhdy5lc20uc2gvcHIvZWxlbWVudC1wbHVzQDI0MzI5L2Rpc3QvaW5kZXguZnVsbC5taW4ubWpzXCIsXG4gICAgXCJlbGVtZW50LXBsdXMvXCI6IFwiaHR0cHM6Ly9yYXcuZXNtLnNoL3ByL2VsZW1lbnQtcGx1c0AyNDMyOS9cIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJfbyI6eyJzaG93SGlkZGVuIjp0cnVlLCJzdHlsZVNvdXJjZSI6Imh0dHBzOi8vcmF3LmVzbS5zaC9wci9lbGVtZW50LXBsdXNAMjQzMjkvZGlzdC9pbmRleC5jc3MifX0=) Maybe we can refer to the validation approach below.
> 
> https://github.com/element-plus/element-plus/blob/c8ecbe4b404abe3efb33e95fe7caff36589724f3/packages/components/time-select/src/time-select.vue#L110-L122

Hi @keeplearning66, thanks for pointing this out.

I've updated the validation for `start` / `end` to also treat negative hour or minute values as invalid. Negative bounds now fall back to `DEFAULT_START` / `DEFAULT_END` through the same warning path.

packages/components/time-select/__tests__/time-select.test.tsx
```diff
@@ -8,20 +8,33 @@ import Select from '@element-plus/components/select'
...
+  const wrapper = mount(render, options)
+  mountedWrappers.push(wrapper)
+  return wrapper
+}
+
+afterEach(async () => {
+  mountedWrappers.forEach((wrapper) => wrapper.unmount())
+  mountedWrappers.length = 0
+  document.body.innerHTML = ''
+  await new Promise((resolve) => setTimeout(resolve, 0))
```
patienry: The changes here are not related to the topic of this PR. If you believe there's indeed a problem here, it might be better to create a new PR for it.

packages/components/time-select/src/time-select.vue
```diff
@@ -87,12 +87,38 @@ const { lang } = useLocale()
 const value = computed(() => props.modelValue)
 const start = computed(() => {
   const time = parseTime(props.start)
-  return time ? formatTime(time) : null
+  const isInvalidStart =
+    !time ||
+    time.hours < 0 ||
+    time.minutes < 0 ||
+    Number.isNaN(time.hours) ||
+    Number.isNaN(time.minutes)
```
patienry: Consider extracting this into a method in `utils.ts` to minimize code duplication.

guozi9999: Thanks for the review. I've updated the PR to address both points:

- Removed the unrelated test cleanup changes and kept only the regression cases for invalid/negative `start` and `end`.
- Extracted the repeated time validation logic into `isValidTime` in `utils.ts`, and reused it for `start`, `end`, and `step`.

packages/components/time-select/__tests__/time-select.test.tsx
guozi9999: Thanks for the review. I've updated the PR to address both points:

- Removed the unrelated test cleanup changes and kept only the regression cases for invalid/negative `start` and `end`.
- Extracted the repeated time validation logic into `isValidTime` in `utils.ts`, and reused it for `start`, `end`, and `step`.

packages/components/time-select/src/time-select.vue
```diff
@@ -87,12 +93,26 @@ const { lang } = useLocale()
...
-  return time ? formatTime(time) : null
+  if (!isValidTime(time)) {
+    debugWarn(
+      'ElTimeSelect',
+      `invalid start, fallback to default start (${DEFAULT_START}).`
+    )
+    return DEFAULT_START
+  }
+  return formatTime(time)
 })
```
rzzf: Since `isValidTime` has already been extracted, could we take it a step further and extract the logic here as well? The handling of `start`, `end`, and `step` appears to be identical.

rzzf (approved):

packages/components/time-select/src/time-select.vue
guozi9999: adjusted

guozi9999:
Regarding extracting the validation logic into `utils.ts`: I intentionally kept it in the component because I believe `utils.ts` should contain pure, side-effect-free functions. The validation here involves `debugWarn`, which is a side effect, so it feels more appropriate to keep it at the component level.

Happy to adjust if you think otherwise!

packages/components/time-select/src/time-select.vue
```diff
@@ -135,7 +143,7 @@ const items = computed(() => {
     })
   }
 
-  if (props.start && props.end && props.step) {
+  if (start.value && end.value && step.value) {
```
patienry: Is this check necessary? Since `start`, `end`, and `step` will always return computable values. 🤔

guozi9999: Good point. Since start/end/step now always fall back to valid defaults, this guard is redundant. I'll simplify the items generation logic.

keeplearning66 (approved):
Thank you!

btea (approved):

PR #24348, opened 2026-05-29
feat(components): [table] add `getHalfSelectionRows` method

### LINKED ISSUE #24345: [Component] [table] el-table树形数据选择勾选了部分子节点，没有提供对应的API去获取半选中的父节点
Bug Type: **`Component`**

## Environment

- Vue Version: `3.3.5`
- Element Plus Version: `2.14.0`
- Browser / OS: `5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/148.0.0.0 Safari/537.36"`
- Build Tool: `Vite`

### Related Component

- `el-table`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZGl2PuW9k+WJjemAieS4reeahOmhue+8mnt7c2VsZWN0VmFsdWV9fTwvZGl2PlxuICA8ZGl2PuW9k+WJjemAieS4reeahOmhue+8mnt7c2VsZWN0aW9uVmFsdWV9fTwvZGl2PlxuICA8ZGl2PuS7peS4i+aYr+WumOaWueaWh+aho+S7o+eggTwvZGl2PlxuICA8ZWwtcmFkaW8tZ3JvdXAgdi1tb2RlbD1cInRyZWVQcm9wcy5jaGVja1N0cmljdGx5XCIgY2xhc3M9XCJtYi0yXCI+XG4gICAgPGVsLXJhZGlvLWJ1dHRvbiA6dmFsdWU9XCJ0cnVlXCIgbGFiZWw9XCJ0cnVlXCIgLz5cbiAgICA8ZWwtcmFkaW8tYnV0dG9uIDp2YWx1ZT1cImZhbHNlXCIgbGFiZWw9XCJmYWxzZVwiIC8+XG4gIDwvZWwtcmFkaW8tZ3JvdXA+XG4gIDxlbC10YWJsZVxuICAgIDpkYXRhPVwidGFibGVEYXRhXCJcbiAgICA6dHJlZS1wcm9wcz1cInRyZWVQcm9wc1wiXG4gICAgcm93LWtleT1cImlkXCJcbiAgICBkZWZhdWx0LWV4cGFuZC1hbGxcbiAgICBAc2VsZWN0PVwic2VsZWN0RnVuY1wiXG4gICAgQHNlbGVjdGlvbi1jaGFuZ2U9XCJzZWxlY3Rpb25GdW5jXCJcbiAgPlxuICAgIDxlbC10YWJsZS1jb2x1bW4gdHlwZT1cInNlbGVjdGlvblwiIHdpZHRoPVwiNTVcIiA6c2VsZWN0YWJsZT1cInNlbGVjdGFibGVcIiAvPlxuICAgIDxlbC10YWJsZS1jb2x1bW4gcHJvcD1cImRhdGVcIiBsYWJlbD1cIkRhdGVcIiAvPlxuICAgIDxlbC10YWJsZS1jb2x1bW4gcHJvcD1cIm5hbWVcIiBsYWJlbD1cIk5hbWVcIiAvPlxuICAgIDxlbC10YWJsZS1jb2x1bW4gcHJvcD1cImFkZHJlc3NcIiBsYWJlbD1cIkFkZHJlc3NcIiAvPlxuICA8L2VsLXRhYmxlPlxuPC90ZW1wbGF0ZT5cblxuPHNjcmlwdCBsYW5nPVwidHNcIiBzZXR1cD5cbmltcG9ydCB7IHJlZixyZWFjdGl2ZSB9IGZyb20gJ3Z1ZSdcblxuaW50ZXJmYWNlIFVzZXIge1xuICBpZDogbnVtYmVyXG4gIGRhdGU6IHN0cmluZ1xuICBuYW1lOiBzdHJpbmdcbiAgYWRkcmVzczogc3RyaW5nXG4gIGhhc0NoaWxkcmVuPzogYm9vbGVhblxuICBjaGlsZHJlbj86IFVzZXJbXVxufVxuXG5jb25zdCBzZWxlY3RWYWx1ZSA9IHJlZigpXG5jb25zdCBzZWxlY3Rpb25WYWx1ZT0gcmVmKClcblxuY29uc3QgdHJlZVByb3BzID0gcmVhY3RpdmUoe1xuICBjaGVja1N0cmljdGx5OiBmYWxzZSxcbn0pXG5cbmNvbnN0IHNlbGVjdGFibGUgPSAocm93OiBVc2VyKSA9PiAhWzEsIDMxXS5pbmNsdWRlcyhyb3cuaWQpXG5cbmNvbnN0IHRhYmxlRGF0YTogVXNlcltdID0gW1xuICB7XG4gICAgaWQ6IDEsXG4gICAgZGF0ZTogJzIwMTYtMDUtMDInLFxuICAgIG5hbWU6ICd3YW5neGlhb2h1JyxcbiAgICBhZGRyZXNzOiAnTm8uIDE4OSwgR3JvdmUgU3QsIExvcyBBbmdlbGVzJyxcbiAgfSxcbiAge1xuICAgIGlkOiAyLFxuICAgIGRhdGU6ICcyMDE2LTA1LTA0JyxcbiAgICBuYW1lOiAnd2FuZ3hpYW9odScsXG4gICAgYWRkcmVzczogJ05vLiAxODksIEdyb3ZlIFN0LCBMb3MgQW5nZWxlcycsXG4gIH0sXG4gIHtcbiAgICBpZDogMyxcbiAgICBkYXRlOiAnMjAxNi0wNS0wMScsXG4gICAgbmFtZTogJ3dhbmd4aWFvaHUnLFxuICAgIGFkZHJlc3M6ICdOby4gMTg5LCBHcm92ZSBTdCwgTG9zIEFuZ2VsZXMnLFxuICAgIGNoaWxkcmVuOiBbXG4gICAgICB7XG4gICAgICAgIGlkOiAzMSxcbiAgICAgICAgZGF0ZTogJzIwMTYtMDUtMDEnLFxuICAgICAgICBuYW1lOiAnd2FuZ3hpYW9odScsXG4gICAgICAgIGFkZHJlc3M6ICdOby4gMTg5LCBHcm92ZSBTdCwgTG9zIEFuZ2VsZXMnLFxuICAgICAgfSxcbiAgICAgIHtcbiAgICAgICAgaWQ6IDMyLFxuICAgICAgICBkYXRlOiAnMjAxNi0wNS0wMScsXG4gICAgICAgIG5hbWU6ICd3YW5neGlhb2h1JyxcbiAgICAgICAgYWRkcmVzczogJ05vLiAxODksIEdyb3ZlIFN0LCBMb3MgQW5nZWxlcycsXG4gICAgICB9LFxuICAgICAge1xuICAgICAgICBpZDogMzMsXG4gICAgICAgIGRhdGU6ICcyMDE2LTA1LTAxJyxcbiAgICAgICAgbmFtZTogJ3dhbmd4aWFvaHUnLFxuICAgICAgICBhZGRyZXNzOiAnTm8uIDE4OSwgR3JvdmUgU3QsIExvcyBBbmdlbGVzJyxcbiAgICAgIH0sXG4gICAgXSxcbiAgfSxcbiAge1xuICAgIGlkOiA0LFxuICAgIGRhdGU6ICcyMDE2LTA1LTAzJyxcbiAgICBuYW1lOiAnd2FuZ3hpYW9odScsXG4gICAgYWRkcmVzczogJ05vLiAxODksIEdyb3ZlIFN0LCBMb3MgQW5nZWxlcycsXG4gIH0sXG5dXG5cbmNvbnN0IHNlbGVjdEZ1bmMgPSAodmFsOiBhbnkpID0+IHtcbiAgc2VsZWN0VmFsdWUudmFsdWUgPSB2YWxcbn1cblxuY29uc3Qgc2VsZWN0aW9uRnVuYyA9ICh2YWwpID0+IHtcbiAgc2VsZWN0aW9uVmFsdWUudmFsdWUgPSB2YWxcbn1cblxuPC9zY3JpcHQ+IiwiZWxlbWVudC1wbHVzLmpzIjoiaW1wb3J0IEVsZW1lbnRQbHVzIGZyb20gJ2VsZW1lbnQtcGx1cydcbmltcG9ydCB7IGdldEN1cnJlbnRJbnN0YW5jZSB9IGZyb20gJ3Z1ZSdcblxubGV0IGluc3RhbGxlZCA9IGZhbHNlXG5hd2FpdCBsb2FkU3R5bGUoKVxuXG5leHBvcnQgZnVuY3Rpb24gc2V0dXBFbGVtZW50UGx1cygpIHtcbiAgaWYgKGluc3RhbGxlZCkgcmV0dXJuXG4gIGNvbnN0IGluc3RhbmNlID0gZ2V0Q3VycmVudEluc3RhbmNlKClcbiAgaW5zdGFuY2UuYXBwQ29udGV4dC5hcHAudXNlKEVsZW1lbnRQbHVzKVxuICBpbnN0YWxsZWQgPSB0cnVlXG59XG5cbmV4cG9ydCBmdW5jdGlvbiBsb2FkU3R5bGUoKSB7XG4gIGNvbnN0IHN0eWxlcyA9IFsnaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAMi4xNC4wL2Rpc3QvaW5kZXguY3NzJywgJ2h0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQDIuMTQuMC90aGVtZS1jaGFsay9kYXJrL2Nzcy12YXJzLmNzcyddLm1hcCgoc3R5bGUpID0+IHtcbiAgICByZXR1cm4gbmV3IFByb21pc2UoKHJlc29sdmUsIHJlamVjdCkgPT4ge1xuICAgICAgY29uc3QgbGluayA9IGRvY3VtZW50LmNyZWF0ZUVsZW1lbnQoJ2xpbmsnKVxuICAgICAgbGluay5yZWwgPSAnc3R5bGVzaGVldCdcbiAgICAgIGxpbmsuaHJlZiA9IHN0eWxlXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2xvYWQnLCByZXNvbHZlKVxuICAgICAgbGluay5hZGRFdmVudExpc3RlbmVyKCdlcnJvcicsIHJlamVjdClcbiAgICAgIGRvY3VtZW50LmJvZHkuYXBwZW5kKGxpbmspXG4gICAgfSlcbiAgfSlcbiAgcmV0dXJuIFByb21pc2UuYWxsU2V0dGxlZChzdHlsZXMpXG59IiwidHNjb25maWcuanNvbiI6IntcbiAgXCJjb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IFwiRVNOZXh0XCIsXG4gICAgXCJqc3hcIjogXCJwcmVzZXJ2ZVwiLFxuICAgIFwibW9kdWxlXCI6IFwiRVNOZXh0XCIsXG4gICAgXCJtb2R1bGVSZXNvbHV0aW9uXCI6IFwiQnVuZGxlclwiLFxuICAgIFwidHlwZXNcIjogW1wiZWxlbWVudC1wbHVzL2dsb2JhbC5kLnRzXCJdLFxuICAgIFwiYWxsb3dJbXBvcnRpbmdUc0V4dGVuc2lvbnNcIjogdHJ1ZSxcbiAgICBcImFsbG93SnNcIjogdHJ1ZSxcbiAgICBcImNoZWNrSnNcIjogdHJ1ZVxuICB9LFxuICBcInZ1ZUNvbXBpbGVyT3B0aW9uc1wiOiB7XG4gICAgXCJ0YXJnZXRcIjogMy4zXG4gIH1cbn1cbiIsIlBsYXlncm91bmRNYWluLnZ1ZSI6IjxzY3JpcHQgc2V0dXA+XG5pbXBvcnQgQXBwIGZyb20gJy4vQXBwLnZ1ZSdcbmltcG9ydCB7IHNldHVwRWxlbWVudFBsdXMgfSBmcm9tICcuL2VsZW1lbnQtcGx1cy5qcydcbnNldHVwRWxlbWVudFBsdXMoKVxuPC9zY3JpcHQ+XG5cbjx0ZW1wbGF0ZT5cbiAgPEFwcCAvPlxuPC90ZW1wbGF0ZT5cbiIsImltcG9ydC1tYXAuanNvbiI6IntcbiAgXCJpbXBvcnRzXCI6IHtcbiAgICBcInZ1ZVwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9ydW50aW1lLWRvbUAzLjMuNS9kaXN0L3J1bnRpbWUtZG9tLmVzbS1icm93c2VyLmpzXCIsXG4gICAgXCJAdnVlL3NoYXJlZFwiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vQHZ1ZS9zaGFyZWRAMy4zLjUvZGlzdC9zaGFyZWQuZXNtLWJ1bmRsZXIuanNcIixcbiAgICBcImVsZW1lbnQtcGx1c1wiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQDIuMTQuMC9kaXN0L2luZGV4LmZ1bGwubWluLm1qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzL1wiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQDIuMTQuMC9cIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJfbyI6eyJ2dWVWZXJzaW9uIjoiMy4zLjUiLCJlcFZlcnNpb24iOiIyLjE0LjAifX0=)

### Steps to reproduce

选中某个子节点，无法获取半选中的父节点

### What is Expected?

期望能通过配置参数来获取半选中的父节点，类似Tree组件

没有获取半选中节点的API可用

### Additional comments

_(empty)_

qeTM:
### What is this PR for?

Closes #24345

This PR adds a new `getHalfSelectionRows()` method to the `el-table` component expose.

Users of tree-structured selectable tables had no native API to get parent rows that are in an indeterminate (half-selected) state. This change exposes that internal state directly via a new method.

### Changes

- `packages/components/table/src/store/watcher.ts`: implement `getHalfSelectionRows` logic using internal `selectionIndeterminate` state
- `packages/components/table/src/table/utils-helper.ts`: expose the method from store to hook
- `packages/components/table/src/table.vue`: declare in `defineExpose`
- `packages/components/table/__tests__/table.test.ts`: add unit test
- `docs/en-US/component/table.md`: update API docs

### What kind of change does this PR introduce?

- [x] New feature

### PR Checklist

- [x] I have read the [Contributing Guide](https://github.com/element-plus/element-plus/blob/dev/CONTRIBUTING.md)
- [x] My commit messages follow the [Commit Convention](https://github.com/element-plus/element-plus/blob/dev/COMMIT_CONVENTION.md)
- [x] Tests are added/updated
- [x] Documentation is added/updated

## Summary by CodeRabbit

* **New Features**
  * Table exposes getHalfSelectionRows() to retrieve rows currently in an indeterminate (half-selected) state.

* **Documentation**
  * Table API docs updated to document getHalfSelectionRows() and its availability (version marker added).

* **Tests**
  * Added unit tests validating half-selection behavior and verifying getHalfSelectionRows() returns expected results when rows become partially selected.


qeTM:
I'm aware that this can currently be worked around by directly accessing the component's internal store state, but that approach relies on implementation details and isn't stable or clean. I think this functionality is worth exposing as an official API, so I've submitted a PR that adds a `getHalfSelectionRows()` method, making it straightforward for users to retrieve parent rows in an indeterminate (half-selected) state from a tree table.

docs/en-US/component/table.md
```diff
@@ -341,23 +341,24 @@ table/tooltip-formatter
...
+| ------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------- |
+| clearSelection                 | used in multiple selection Table, clear user selection                                                                                                            | ^[Function]`() => void`                                                      |
+| getSelectionRows               | returns the currently selected rows                                                                                                                               | ^[Function]`() => any[]`                                                     |
+| getHalfSelectionRows ^(2.13.3) | returns the currently half-selected rows                                                                                                                          | ^[Function]`() => any[]`                                                     |
```
patienry:
```suggestion
| getHalfSelectionRows ^(2.14.2) | returns the currently half-selected rows                                                                                                                          | ^[Function]`() => any[]`                                                     |
```

keeplearning66 (commented):
Could we save `halfSelection` when setting `indeterminateMap`, and then directly return `halfSelection` when `getHalfSelectionRows` is called? This way, we could **avoid recursively processing the entire table data every time** `getHalfSelectionRows` is called. 🤔 WDYT?

qeTM:
That's a great suggestion! I totally agree.

Saving halfSelection simultaneously while updating the indeterminateMap avoids the expensive $O(N)$ tree traversal every time getHalfSelectionRows is called. This provides a significant performance improvement, especially when dealing with large amounts of tree data.

I've just implemented this optimization and added the necessary edge-case handling (such as clearing halfSelection when selectionIndeterminate resets during checkStrictly toggles). All unit tests have passed and the changes are committed.

Thanks for pointing this out!

keeplearning66 (approved):
Thank you!

packages/components/table/src/store/watcher.ts
```diff
@@ -80,6 +80,7 @@ function useWatcher<T extends DefaultRow>() {
   const isAllSelected = ref(false)
   const selection: Ref<T[]> = ref([])
   const selectionIndeterminate = ref<Record<string, boolean>>({})
+  const halfSelection: Ref<T[]> = ref([])
```
btea: Does this value seem unnecessary to initialize as reactive?

qeTM: You're right that it could work without ref since we always do full replacement. But I kept it consistent with the other store state fields (selection, selectionIndeterminate), and                                              
     │ since it's exported via states, staying reactive avoids surprises if someone watches it later.

packages/components/table/src/store/watcher.ts
```diff
@@ -849,6 +861,7 @@ function useWatcher<T extends DefaultRow>() {
       isAllSelected,
       selection,
       selectionIndeterminate,
+      halfSelection,
```
btea: It seems unnecessary to add `halfSelection` to the `states` property here; exporting the `getHalfSelectionRows` method should be sufficient. 🤔

packages/components/table/src/store/watcher.ts
btea: We do not export the relevant variables for users to access directly, but obtain them through an export method. Therefore, it seems that this situation can be disregarded for now.
https://github.com/element-plus/element-plus/pull/24348#discussion_r3346344046

packages/components/table/src/store/watcher.ts
qeTM: Thanks for the review! You're right that halfSelection doesn't need to be exposed in states since external access goes through getHalfSelectionRows(). I've removed it. Please take another look, thanks!

btea (approved):

packages/components/table/src/store/watcher.ts
patienry: Based on the discussion above, `halfSelection` shouldn’t need to be reactive, but it is still reactive in the code. 🤔

qeTM: first, all other internal states (selection, selectionIndeterminate, etc.) are ref — halfSelection should be consistent. Second, without caching it we'd need recursive tree traversal on every getHalfSelectionRows() call — we already cache it via nextHalfSelection, so keeping it as ref is the natural choice.

patienry: We only make a variable reactive if it needs to be reactive. For example, in the code below, we need `selection` to be reactive. Since we have determined that `halfSelection` does not need to be reactive, there’s no need to make it reactive just to keep it consistent with other variables.

https://github.com/element-plus/element-plus/blob/47a34a70deb3cfa34232ff4a342073bf9779d9b0/packages/components/table/src/config.ts#L52-L56

qeTM: Already updated as you suggested — halfSelection is now a plain array without ref, only selection keeps its reactivity. No unnecessary reactive wrappers just for consistency with other variables. Thanks for the review!

keeplearning66 (approved):

PR #24327, opened 2026-05-26
fix(components): [calendar] correctly validate date ranges that span across years

guozi9999:
Please make sure these boxes are checked before submitting your PR, thank you!

- [ ] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/master/.github/CONTRIBUTING.fr-FR.md)).
- [ ] Make sure you are merging your commits to `dev` branch.
- [ ] Add some descriptions and refer to relative issues for your PR.

## Summary

fix: https://github.com/element-plus/element-plus/issues/24326

- Validate Calendar range month boundaries using Dayjs month-unit comparison.
- Add a regression test for a cross-year range longer than two months.

## Problem

`el-calendar` previously checked whether a range exceeded two months by comparing only the numeric month value:

```ts
startDayjs.add(1, 'month').month() !== endDayjs.month()
```

This ignores the year, so a range like `2024-12-02` to `2026-01-04` could be treated as valid because both dates resolve to January after the month-only comparison.

## Solution

Use Dayjs `isSame(..., 'month')` so the comparison includes both year and month:

```ts
!startDayjs.add(1, 'month').isSame(endDayjs, 'month')
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed date range validation in the Calendar component to properly handle ranges spanning more than two months across different years. The component now correctly falls back to non-range calendar behavior when appropriate.

[Review Change Stack](https://app.coderabbit.ai/change-stack/element-plus/element-plus/pull/24327?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #24372, opened 2026-06-02
fix(components): [calendar] reject invalid cross-year range

sijie-Z:
### Description

  `validatedRange` compared only month numbers without checking year, so cross-year ranges exceeding two months (e.g. `Dec 2024 to Jan 2026`) were incorrectly accepted. Use dayjs `isSame('month')`to compare both year and month.

  ### Related

## Bug Fixes
- Improved the calendar component's date range validation to more accurately handle month boundary edge cases when validating two-month intervals.


rzzf:
Thanks for contribution.

This is a duplicate of https://github.com/element-plus/element-plus/pull/24327.

PR #24413, opened 2026-06-12
feat(components): add vertical track gap for scrollbar

jiajie-yang:
### What problem does this PR solve?

In some layouts, scrollable content is visually covered by a header, footer, or gradient mask. The content can continue scrolling behind the translucent or masked area, but the custom scrollbar thumb should stay within the visible track area instead of moving into that covered region.

This PR adds `vertical-start-gap` and `vertical-end-gap` to `ElScrollbar` so the vertical scrollbar track can be inset from the viewport edges without limiting the actual scroll range.

### What is changed and how does it work?

- Add `vertical-start-gap` and `vertical-end-gap` props.
- The props only affect the vertical custom scrollbar track.
- The scroll content can still be fully scrolled to the top and bottom.
- Negative gap values are normalized to `0`.
- Overlarge gap values are clamped to keep the thumb draggable.
- Update docs for the new props.

### Tests

- Added unit tests for normal top and bottom vertical track gaps.
- Added unit tests for one-sided gap values.
- Added unit tests for negative gap normalization.
- Added unit tests for overlarge gap clamping.

### Verification

- `pnpm test packages/components/scrollbar/__tests__/scrollbar.test.tsx`
- `pnpm test`
- `pnpm lint`
- `pnpm typecheck`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Configurable vertical scrollbar track gaps with `vertical-start-gap` and `vertical-end-gap` to control top/bottom spacing for vertical scrollbars (default `0`).
* **Bug Fixes**
  * Improved vertical gap-to-scroll thumb mapping to keep dragging consistent, including edge cases where markup should not be rendered.
* **Tests**
  * Expanded coverage for vertical gap behavior: negative normalization, one-/both-sided gaps, clamping, and non-rendering scenarios.
* **Documentation**
  * Updated the Scrollbar API “Attributes” table with the new gap options and defaults.


keeplearning66 (commented):
> In some layouts, scrollable content is visually covered by a header, footer, or gradient mask. The content can continue scrolling behind the translucent or masked area, but the custom scrollbar thumb should stay within the visible track area instead of moving into that covered region.

Thank you! Could you provide an example to illustrate a specific application scenario? 🤔

jiajie-yang:
A concrete use case is a scrollable panel with a translucent header/footer or a gradient mask.

For example, in a chat panel, notification list, command palette, or data preview drawer, the content can visually scroll underneath a fixed translucent header/footer so rows gradually fade out behind the mask. In that layout, the scrollbar thumb should remain within the fully visible track area; otherwise the content appears to move behind the overlay, but the thumb still moves into the covered area, which creates a visual mismatch.

`vertical-start-gap` and `vertical-end-gap` allow the thumb track to be inset from those covered regions while keeping the content scroll range unchanged.

jiajie-yang:
> > In some layouts, scrollable content is visually covered by a header, footer, or gradient mask. The content can continue scrolling behind the translucent or masked area, but the custom scrollbar thumb should stay within the visible track area instead of moving into that covered region.
> 
> Thank you! Could you provide an example to illustrate a specific application scenario? 🤔

image

Here is a concrete example with a notification panel.

The header and footer are translucent masks, and the list content can scroll behind them to create a fade-out effect. In this layout, the scrollbar thumb should stay within the fully visible track area instead of moving into the masked regions.

`vertical-start-gap` and `vertical-end-gap` make that possible without changing the content scroll range.

jiajie-yang:
> > In some layouts, scrollable content is visually covered by a header, footer, or gradient mask. The content can continue scrolling behind the translucent or masked area, but the custom scrollbar thumb should stay within the visible track area instead of moving into that covered region.
> 
> Thank you! Could you provide an example to illustrate a specific application scenario? 🤔

https://github.com/user-attachments/assets/4373ac01-cae7-4afa-92b7-158b885cca40

@keeplearning66  Here is a short interaction demo.

The content scrolls behind the translucent header/footer masks. Without the gap, the scrollbar thumb also moves into the masked area. With `vertical-start-gap` and `vertical-end-gap`, the content scroll range remains unchanged, while the thumb stays within the fully visible track area.

jiajie-yang:
Updated the PR with a small cleanup commit:

- Avoid adding inline `top` / `bottom` styles when no vertical track gap is configured, so the default scrollbar path keeps relying on the existing CSS.
- Made the clamped-gap test less brittle by comparing the `translateY` value with tolerance.
- Updated the docs version tag from `2.14.2` to `2.14.3` since `2.14.2` has already been released.

keeplearning66 (commented):
> The header and footer are translucent masks, and the list content can scroll behind them to create a fade-out effect. In this layout, the scrollbar thumb should stay within the fully visible track area instead of moving into the masked regions.

I feel that the effect shown in the screen recording is somewhat customized. This kind of functionality is only needed when other content overlays the scroll container. 😕

jiajie-yang:
> > The header and footer are translucent masks, and the list content can scroll behind them to create a fade-out effect. In this layout, the scrollbar thumb should stay within the fully visible track area instead of moving into the masked regions.
> 
> I feel that the effect shown in the screen recording is somewhat customized. This kind of functionality is only needed when other content overlays the scroll container. 😕

Thanks for taking another look.

I agree that the demo itself is a specific UI composition, but the props are not intended to support only that notification-panel example.

The underlying pattern is more general: in modern message/list UIs, content is often allowed to scroll behind a sticky or translucent header/footer, or fade into a top/bottom mask. In those cases, the content viewport and the visible scrollbar control area are not always the same. The content should still use the full scroll range, while the thumb track may need to stay inside the fully visible area.

Without these props, users would need to solve this outside of `ElScrollbar`. A simple CSS override may adjust the visual track position, but it cannot keep the thumb size, drag distance, and track-click mapping in sync with that shorter track. Achieving a fully consistent interaction would require a custom wrapper or custom scrollbar logic, which is a much higher maintenance cost.

So the goal is to expose a small generic track-inset capability:

- it only affects the vertical thumb track bounds;
- it does not change the content scroll range;
- the default value is `0`, so existing behavior remains unchanged.

The demo is just one example of this overlay/masked scroll layout pattern, not the only target use case.

patienry:
Thank you for the detailed explanation.

> The content should still use the full scroll range, while the thumb track may need to stay inside the fully visible area.

Is this type of layout common on desktop websites? Personally, I haven't come across many websites that use it. Could you share some examples of websites where this pattern is used?

Or is this type of layout actually more common on mobile devices?

jiajie-yang:
> Thank you for the detailed explanation.
> 
> > The content should still use the full scroll range, while the thumb track may need to stay inside the fully visible area.
> 
> Is this type of layout common on desktop websites? Personally, I haven't come across many websites that use it. Could you share some examples of websites where this pattern is used?
> 
> Or is this type of layout actually more common on mobile devices?

@keeplearning66 Thank you for the follow-up.

I agree that this pattern is probably more common on mobile, and it is not something every traditional desktop website needs.

My point is that it also appears in app-like desktop web UIs, especially in layouts with sticky headers, drawers, side panels, command palettes, notification/message lists, or documentation/admin pages where content can scroll underneath a persistent top area. In those cases, the visual content viewport and the visible scrollbar control area may not be exactly the same.

A public desktop example of the layout pattern is the Element Plus component overview page itself:
https://element-plus.org/zh-CN/component/overview

When scrolling the page, the content continues moving behind the sticky top navigation area. The official site uses the native page scrollbar, so it is not the exact same implementation case. But if a similar layout is built inside an `ElScrollbar` container, the custom thumb track would need a way to stay below the covered header area while still controlling the full content scroll range.

Here is a screenshot of that layout pattern on the Element Plus website:

iShot_2026-06-24_15 09 57

So I agree this is not a common need for every scrollbar, but I think it is a reasonable capability for application-style scroll containers. The props keep the scope small: they only inset the vertical thumb track, do not change the content scroll range, and default to `0`, so existing behavior remains unchanged.

patienry (commented):
> My point is that it also appears in app-like desktop web UIs, especially in layouts with sticky headers, drawers, side panels, command palettes, notification/message lists, or documentation/admin pages where content can scroll underneath a persistent top area. In those cases, the visual content viewport and the visible scrollbar control area may not be exactly the same.

This is a very common **affix** effect.

I think the behavior shown in the recording is more of a customization-specific feature, which developers can implement themselves.

jiajie-yang:
Thank you @patienry for taking the time to review this proposal and discuss the use case.

I understand the concern that, although affix and overlay layouts are common, adjusting the scrollbar track for a particular visual composition may still be considered application-level customization rather than something that should become part of the core ElScrollbar API.

I also revisited the current implementation before wrapping this up. With the existing public API, shortening the built-in track cannot be implemented correctly through CSS alone, because the thumb size, scroll-to-thumb positioning, track-click behavior, and drag-to-scroll mapping all depend on the original track geometry. An application-level alternative is possible, but it would require hiding the built-in bar and implementing a separate custom bar around the exposed scroll container.

That implementation cost was the main reason I proposed these two small, backward-compatible props. However, I understand that this alone does not necessarily justify expanding the public API for a relatively specific presentation requirement.

I’ll close this PR for now. The implementation and tests can remain as a reference in case a broader scrollbar customization API or similar use cases are considered in the future.

Thanks again for the thoughtful feedback and for maintaining Element Plus.

PR #24434, opened 2026-06-17
fix(components): [table] fixed column shadow when table-layout is auto

### LINKED ISSUE #24431: [Component] [table] table-layout: auto，使单元格内容不换行展示，列宽度自适应内容宽度，右侧固定列没有显示阴影
Bug Type: **`Component`**

## Environment

- Vue Version: `3.6.0-beta.8`
- Browser / OS: `Chrome/145.0.7632.117，Mac OS/14.6.1`
- Build Tool: `Vite`

### Related Component

- `el-table`

### Reproduction Link

[Element Plus Playground](https://element-plus.run/#eyJBcHAudnVlIjoiPHRlbXBsYXRlPlxuICA8ZWwtdGFibGUgOmRhdGE9XCJ0YWJsZURhdGFcIiB0YWJsZS1sYXlvdXQ9XCJhdXRvXCI+XG4gICAgPGVsLXRhYmxlLWNvbHVtbiBwcm9wPVwiZGF0ZVwiIGxhYmVsPVwiRGF0ZVwiIC8+XG4gICAgPGVsLXRhYmxlLWNvbHVtbiBwcm9wPVwibmFtZVwiIGxhYmVsPVwiTmFtZVwiIC8+XG4gICAgPGVsLXRhYmxlLWNvbHVtbiBwcm9wPVwiYWRkcmVzc1wiIGxhYmVsPVwiQWRkcmVzc1wiIC8+XG4gICAgPGVsLXRhYmxlLWNvbHVtbiBwcm9wPVwiYWN0aW9uXCIgbGFiZWw9XCLmk43kvZxcIiBmaXhlZD1cInJpZ2h0XCIgPlxuICAgICAgPGVsLWJ1dHRvbj50ZXN0PC9lbC1idXR0b24+XG4gICAgPC9lbC10YWJsZS1jb2x1bW4+XG4gIDwvZWwtdGFibGU+XG48L3RlbXBsYXRlPlxuXG48c2NyaXB0IHNldHVwPlxuaW1wb3J0IHsgcmVmIH0gZnJvbSAndnVlJ1xuXG5jb25zdCB0YWJsZUxheW91dCA9IHJlZignZml4ZWQnKVxuXG5jb25zdCB0YWJsZURhdGEgPSBbXG4gIHtcbiAgICBkYXRlOiAnMjAxNi0wNS0wMycsXG4gICAgbmFtZTogJ1RvbScsXG4gICAgYWRkcmVzczogJ05vLiAxODksIEdyb3ZlIFN0LCBMb3MgQW5nZWxlc05vLiAxODksIEdyb3ZlIFN0LCBMb3MgQW5nZWxlc05vLiAxODksIEdyb3ZlIFN0LCBMb3MgQW5nZWxlc05vLiAxODksIEdyb3ZlIFN0LCBMb3MgQW5nZWxlc05vLiAxODksIEdyb3ZlIFN0LCBMb3MgQW5nZWxlcycsXG4gIH0sXG4gIHtcbiAgICBkYXRlOiAnMjAxNi0wNS0wMicsXG4gICAgbmFtZTogJ1RvbScsXG4gICAgYWRkcmVzczogJ05vLiAxODksIEdyb3ZlIFN0LCBMb3MgQW5nZWxlc05vLiAxODksIEdyb3ZlIFN0LCBMb3MgQW5nZWxlcycsXG4gIH0sXG4gIHtcbiAgICBkYXRlOiAnMjAxNi0wNS0wNCcsXG4gICAgbmFtZTogJ1RvbScsXG4gICAgYWRkcmVzczogJ05vLiAxODksIEdyb3ZlIFN0LCBMb3MgQW5nZWxlcycsXG4gIH0sXG4gIHtcbiAgICBkYXRlOiAnMjAxNi0wNS0wMScsXG4gICAgbmFtZTogJ1RvbScsXG4gICAgYWRkcmVzczogJ05vLiAxODksIEdyb3ZlIFN0LCBMb3MgQW5nZWxlcycsXG4gIH0sXG5dXG48L3NjcmlwdD5cblxuPHN0eWxlPlxuLmVsLXRhYmxlIC5jZWxsIHtcbiAgd2hpdGUtc3BhY2U6IG5vd3JhcCAhaW1wb3J0YW50O1xuICB3b3JkLWJyZWFrOiBrZWVwLWFsbDtcbn1cbjwvc3R5bGU+XG4iLCJlbGVtZW50LXBsdXMuanMiOiJpbXBvcnQgRWxlbWVudFBsdXMgZnJvbSAnZWxlbWVudC1wbHVzJ1xuaW1wb3J0IHsgZ2V0Q3VycmVudEluc3RhbmNlIH0gZnJvbSAndnVlJ1xuXG5sZXQgaW5zdGFsbGVkID0gZmFsc2VcbmF3YWl0IGxvYWRTdHlsZSgpXG5cbmV4cG9ydCBmdW5jdGlvbiBzZXR1cEVsZW1lbnRQbHVzKCkge1xuICBpZiAoaW5zdGFsbGVkKSByZXR1cm5cbiAgY29uc3QgaW5zdGFuY2UgPSBnZXRDdXJyZW50SW5zdGFuY2UoKVxuICBpbnN0YW5jZS5hcHBDb250ZXh0LmFwcC51c2UoRWxlbWVudFBsdXMpXG4gIGluc3RhbGxlZCA9IHRydWVcbn1cblxuZXhwb3J0IGZ1bmN0aW9uIGxvYWRTdHlsZSgpIHtcbiAgY29uc3Qgc3R5bGVzID0gWydodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL2VsZW1lbnQtcGx1c0BsYXRlc3QvZGlzdC9pbmRleC5jc3MnLCAnaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lbGVtZW50LXBsdXNAbGF0ZXN0L3RoZW1lLWNoYWxrL2RhcmsvY3NzLXZhcnMuY3NzJ10ubWFwKChzdHlsZSkgPT4ge1xuICAgIHJldHVybiBuZXcgUHJvbWlzZSgocmVzb2x2ZSwgcmVqZWN0KSA9PiB7XG4gICAgICBjb25zdCBsaW5rID0gZG9jdW1lbnQuY3JlYXRlRWxlbWVudCgnbGluaycpXG4gICAgICBsaW5rLnJlbCA9ICdzdHlsZXNoZWV0J1xuICAgICAgbGluay5ocmVmID0gc3R5bGVcbiAgICAgIGxpbmsuYWRkRXZlbnRMaXN0ZW5lcignbG9hZCcsIHJlc29sdmUpXG4gICAgICBsaW5rLmFkZEV2ZW50TGlzdGVuZXIoJ2Vycm9yJywgcmVqZWN0KVxuICAgICAgZG9jdW1lbnQuYm9keS5hcHBlbmQobGluaylcbiAgICB9KVxuICB9KVxuICByZXR1cm4gUHJvbWlzZS5hbGxTZXR0bGVkKHN0eWxlcylcbn1cbiIsInRzY29uZmlnLmpzb24iOiJ7XG4gIFwiY29tcGlsZXJPcHRpb25zXCI6IHtcbiAgICBcInRhcmdldFwiOiBcIkVTTmV4dFwiLFxuICAgIFwianN4XCI6IFwicHJlc2VydmVcIixcbiAgICBcIm1vZHVsZVwiOiBcIkVTTmV4dFwiLFxuICAgIFwibW9kdWxlUmVzb2x1dGlvblwiOiBcIkJ1bmRsZXJcIixcbiAgICBcInR5cGVzXCI6IFtcImVsZW1lbnQtcGx1cy9nbG9iYWwuZC50c1wiXSxcbiAgICBcImFsbG93SW1wb3J0aW5nVHNFeHRlbnNpb25zXCI6IHRydWUsXG4gICAgXCJhbGxvd0pzXCI6IHRydWUsXG4gICAgXCJjaGVja0pzXCI6IHRydWVcbiAgfSxcbiAgXCJ2dWVDb21waWxlck9wdGlvbnNcIjoge1xuICAgIFwidGFyZ2V0XCI6IDMuM1xuICB9XG59XG4iLCJQbGF5Z3JvdW5kTWFpbi52dWUiOiI8c2NyaXB0IHNldHVwPlxuaW1wb3J0IEFwcCBmcm9tICcuL0FwcC52dWUnXG5pbXBvcnQgeyBzZXR1cEVsZW1lbnRQbHVzIH0gZnJvbSAnLi9lbGVtZW50LXBsdXMuanMnXG5zZXR1cEVsZW1lbnRQbHVzKClcbjwvc2NyaXB0PlxuXG48dGVtcGxhdGU+XG4gIDxBcHAgLz5cbjwvdGVtcGxhdGU+XG4iLCJpbXBvcnQtbWFwLmpzb24iOiJ7XG4gIFwiaW1wb3J0c1wiOiB7XG4gICAgXCJ2dWVcIjogXCJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL0B2dWUvcnVudGltZS1kb21AbGF0ZXN0L2Rpc3QvcnVudGltZS1kb20uZXNtLWJyb3dzZXIuanNcIixcbiAgICBcIkB2dWUvc2hhcmVkXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AdnVlL3NoYXJlZEBsYXRlc3QvZGlzdC9zaGFyZWQuZXNtLWJ1bmRsZXIuanNcIixcbiAgICBcImVsZW1lbnQtcGx1c1wiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQGxhdGVzdC9kaXN0L2luZGV4LmZ1bGwubWluLm1qc1wiLFxuICAgIFwiZWxlbWVudC1wbHVzL1wiOiBcImh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vZWxlbWVudC1wbHVzQGxhdGVzdC9cIixcbiAgICBcIkBlbGVtZW50LXBsdXMvaWNvbnMtdnVlXCI6IFwiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9AZWxlbWVudC1wbHVzL2ljb25zLXZ1ZUAyL2Rpc3QvaW5kZXgubWluLmpzXCJcbiAgfSxcbiAgXCJzY29wZXNcIjoge31cbn0iLCJfbyI6e319)

### Steps to reproduce

设定 table 单元格内容不允许换行，使单元格宽度自适应内容宽度，调整页面宽度使 table 出现横向滚动条，最右侧的固定列没有显示阴影

### What is Expected?

出现滚动条时被固定的操作列有阴影效果

出现滚动条时被固定的操作列无阴影效果

### Additional comments

_(empty)_

LostElkByte:
## Summary

- Fix missing shadow on fixed columns when `table-layout="auto"` and cell content causes real horizontal overflow
- Detect horizontal scroll from the scrollbar container when `layout.scrollX` is incorrect
- Add a regression test for the `is-scrolling-left` state

## Problem

fix #24431 

With `table-layout="auto"` and non-wrapping cell content, the table can overflow horizontally while `layout.scrollX` stays `false` (width is estimated from `minWidth`/defaults). `syncPosition` then always applies `is-scrolling-none`, so fixed column shadows never appear.

## Solution

- In `table-layout.ts`, extend `updateColumnsWidth()` to detect real horizontal overflow for `table-layout="auto"` tables (without `maxHeight`): when column-width math sets `scrollX` to `false`, fall back to `wrapRef.scrollWidth > wrapRef.clientWidth`.
- In `style-helper.ts`, use `clientWidth` instead of `offsetWidth` when computing `maxScrollLeftPosition`, so scrollbar gutter does not misclassify `scrollLeft === 0` as `is-scrolling-right`.

## Test plan

- [x] `pnpm test packages/components/table/__tests__/table.test.ts`
- [x] Regression: `scrollX` is corrected via `updateColumnsWidth()` when DOM overflows
- [x] Regression: scrollbar gutter case uses `clientWidth` for scroll position classification
- [ ] Manually verified in playground: `table-layout="auto"`, left/right fixed columns, `white-space: nowrap`, narrow container

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved horizontal scrolling detection for tables with fixed columns and `table-layout="auto"`, so overflow/shadow indicators update correctly under constrained-width containers, including scrollbar-gutter scenarios.
* **Tests**
  * Added new coverage to verify fixed-column shadow behavior for auto layout, including cases where scrollability is determined via mocked scrollbar wrapper metrics and correct scrolling-state class updates.


keeplearning66 (commented):
The newly added test case seems to pass even before the changes. 🤔

LostElkByte:
> The newly added test case seems to pass even before the changes. 🤔

Indeed! I'm done revising it.

packages/components/table/__tests__/table.test.ts
```diff
@@ -2892,6 +2892,63 @@ describe('Table.vue', () => {
...
+
+    tableRef.layout.scrollX.value = false
+    Object.defineProperty(wrapRef, 'scrollWidth', {
+      configurable: true,
+      value: 1000,
+    })
+    Object.defineProperty(wrapRef, 'offsetWidth', {
+      configurable: true,
+      value: 400,
+    })
```
patienry: We could use `defineGetter` here.

keeplearning66 (commented):
Could we make this change in `updateColumnsWidth` instead? That way, all the logic related to the horizontal scrollbar would be centralized in a single file. 🤔 

https://github.com/element-plus/element-plus/blob/2089381ed8217746b7735b05fd3dfcd575ea599f/packages/components/table/src/table-layout.ts#L162-L164

packages/components/table/src/table-layout.ts
```diff
@@ -214,6 +214,12 @@ class TableLayout<T extends DefaultRow> {
       this.bodyWidth.value = bodyMinWidth
     }
 
+    const wrapRef = this.table.refs.scrollBarRef?.wrapRef
+    if (wrapRef) {
+      this.scrollX.value =
+        this.scrollX.value || wrapRef.scrollWidth > wrapRef.clientWidth
```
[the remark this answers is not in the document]
LostElkByte: 已经把 syncPosition 里的 offsetWidth 改成了 clientWidth

packages/components/table/__tests__/table.test.ts
patienry: My point is that we can use `defineGetter` from `test-utils`.
`import defineGetter from '@element-plus/test-utils/define-getter'`

packages/components/table/__tests__/table.test.ts
```diff
@@ -2892,6 +2892,134 @@ describe('Table.vue', () => {
...
+
+    tableRef.layout.updateColumnsWidth()
+    expect(tableRef.layout.scrollX.value).toBe(true)
+
+    tableRef.$el.classList.remove(
+      'is-scrolling-left',
+      'is-scrolling-middle',
+      'is-scrolling-right'
+    )
+    tableRef.$el.classList.add('is-scrolling-none')
```
patienry: Would it be possible to avoid writing it this way? Ideally, component tests should avoid relying on internal implementation details.

patienry: `should show fixed column shadow when scrollX is false but DOM overflows`
After removing it, the test case succeeds even without the change, which is a bit strange. 🤔

LostElkByte: > `should show fixed column shadow when scrollX is false but DOM overflows` After removing it, the test case succeeds even without the change, which is a bit strange. 🤔在进行了移除操作之后，即使没有进行任何修改，测试案例仍然能够成功运行，这有点奇怪。🤔

```
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL   unit  packages/components/table/__tests__/table.test.ts > Table.vue > should detect overflow against client width when scrollbar gutter is present
AssertionError: expected false to be true // Object.is equality

- Expected
+ Received

- true
+ false

 ❯ packages/components/table/__tests__/table.test.ts:2997:66
    2995|     // clientWidth so scrollLeft 0 is classified as is-scrolling-left …
    2996|     // than is-scrolling-right, keeping fixed-column shadows correct.
    2997|     expect(tableRef.$el.classList.contains('is-scrolling-left')).toBe(…
       |                                                                  ^
    2998|     expect(tableRef.$el.classList.contains('is-scrolling-none')).toBe(…
    2999|

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯


 Test Files  1 failed (1)
      Tests  1 failed | 84 passed (85)
   Start at  13:16:48
   Duration  3.78s (transform 848ms, setup 59ms, import 1.11s, tests 2.14s, environment 402ms)
```

Reverting to the version after `git commit 8a0114ac4a4d2ad9ecc493d24a82c02a950bd952` resulted in an error in the test cases. This seems fine.

The first use case covers a basic overflow scenario, which the old code can handle, so it doesn't throw an error.
The second use case is specifically written for gutter-only boundary scenarios, and the old code cannot pass the test.

LostElkByte: The two cases cover different scenarios. The first tests basic DOM overflow detection; the second specifically targets the scrollbar-gutter edge case where the old code fails. Merging them would blur the intent and reduce coverage clarity.

LostElkByte: I can merge them into a single test that covers both scenarios sequentially:

Case 1: normal overflow (scrollWidth >> offsetWidth)
Case 2: gutter-only overflow (scrollWidth > clientWidth but <= offsetWidth)
Would that be acceptable to you?

patienry: `should show fixed column shadow when scrollX is false but DOM overflows`
The test case above should fail when run on the **dev** branch, because it is specifically intended to verify the issue described in #24431.

packages/components/table/src/table-layout.ts
```diff
@@ -214,6 +214,12 @@ class TableLayout<T extends DefaultRow> {
...
+    }
```
btea:
```suggestion
    if (wrapRef && !this.scrollX.value) {
      this.scrollX.value = wrapRef.scrollWidth > wrapRef.clientWidth
    }
```
Based on the code logic, the value should be recalculated and reassigned only when the value of `this.scrollX.value` is `false`.

packages/components/table/__tests__/table.test.ts
LostElkByte: > `should show fixed column shadow when scrollX is false but DOM overflows` The test case above should fail when run on the **dev** branch, because it is specifically intended to verify the issue described in #24431.在开发分支上运行上述测试案例应该会失败，因为该测试案例正是为了验证 #24431 中描述的问题而设计的。

Thanks for the feedback! I've updated the test case.

Setting tableRef.$el.clientWidth to a large value (2000) ensures the column-width calculation path sets scrollX = false (columns fit within the container), so only the DOM fallback (wrapRef.scrollWidth > wrapRef.clientWidth) can flip it to true.

On the unpatched dev branch, scrollX stays false and the test fails as expected. On the patched branch, the DOM fallback sets scrollX = true and the test passes.

packages/components/table/src/table-layout.ts
LostElkByte: Thanks for the feedback! I've updated the.

packages/components/table/src/table-layout.ts
```diff
@@ -214,6 +214,13 @@ class TableLayout<T extends DefaultRow> {
       this.bodyWidth.value = bodyMinWidth
     }
 
+    const isAutoLayout =
+      this.table.props.tableLayout === 'auto' && !this.table.props.maxHeight
+    const wrapRef = this.table.refs.scrollBarRef?.wrapRef
+    if (isAutoLayout && wrapRef && !this.scrollX.value) {
+      this.scrollX.value = wrapRef.scrollWidth > wrapRef.clientWidth
```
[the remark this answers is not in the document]
LostElkByte: This scenario is relatively rare, and there’s a self-repair mechanism in place. You could use `nextTick` to thoroughly check the DOM, ensuring that it’s read after Vue’s patching process. But I don’t think it’s worth fixing.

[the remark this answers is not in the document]
LostElkByte: @codex review

packages/components/table/src/table-layout.ts
```diff
@@ -214,6 +214,11 @@ class TableLayout<T extends DefaultRow> {
       this.bodyWidth.value = bodyMinWidth
     }
 
+    const wrapRef = this.table.refs.scrollBarRef?.wrapRef
+    if (wrapRef && !this.scrollX.value) {
+      this.scrollX.value = wrapRef.scrollWidth > wrapRef.clientWidth
```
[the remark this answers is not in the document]
LostElkByte: @codex review

packages/components/table/src/table-layout.ts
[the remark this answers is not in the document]
LostElkByte: I agree this is theoretically possible when resizing from an overflowing width to a fitting width, because `bodyWidth` is updated before Vue patches the new table width into the DOM.

However, I don't plan to fix this in the current PR for a few reasons:

1. It is a different failure mode from #24431. The linked issue is about missing shadows when the table actually overflows (`scrollX = false` while DOM overflows). This comment is about a transient false positive after resizing wider.
2. The repro conditions are narrow: `table-layout="auto"`, fixed columns, and a resize that lands exactly in the stale-DOM window.
3. The table already self-corrects on the next layout/resize pass via `ResizeObserver` → `doLayout()` → `syncPosition()`.
4. Fixing it with `nextTick` would make `updateColumnsWidth()` asynchronous, which is a broader behavioral change than this bug fix should introduce.

I'll leave this as a known edge case for now. Happy to revisit if a maintainer thinks it should be handled separately.

patienry (approved):
Rel [#24474 (comment)](https://github.com/element-plus/element-plus/pull/24474#pullrequestreview-4912352685)

I'm not sure why `table-layout="auto"` was restricted when `max-height` is set. 😕 Let's keep it consistent for now.

Issue #6696, opened 2022-03-17
[Bug Report] [Component] [table] table-layout is not work when max-height is set


caidongting:
Bug Type: **`Component`**

## Environment

- Vue Version: `3.2.25`
- Element Plus Version: `2.1.2`
- Browser / OS: `Chrome 99.0.4844.51`
- Build Tool: `Vite`

### Related Component

- `el-table`

### Reproduction Link

[...]

PR #24474, opened 2026-06-26
docs(components): [table] add warning for max-height forcing fixed layout

LostElkByte:
## Summary

Document a known limitation: when `max-height` is set, the table layout is
internally forced to `fixed` regardless of the `table-layout` prop value.

## Problem

Users setting `table-layout="auto"` with `max-height` may find horizontal
scrolling broken when columns lack explicit `width` or `min-width`. This
happens because the forced `fixed` layout falls back to `80px` per column
for overflow calculation, which may underestimate the actual content width.

## Solution

Add a warning block to the Table Layout section in the docs explaining
this behavior and advising users to set `width` or `min-width` on all
columns when using `max-height`.

## Summary by CodeRabbit

* **Documentation**
  * Added a warning to the Table layout docs explaining how `max-height` affects table behavior.
  * Clarified that columns without a set `width` or `min-width` use a default value during overflow calculations.
  * Recommended setting `width` or `min-width` on all columns to keep horizontal scrolling working as expected.


docs/en-US/component/table.md
```diff
@@ -239,6 +239,10 @@ table/custom-index
 
 The [table-layout](https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout) property sets the algorithm used to lay out table cells, rows, and columns.
 
+:::warning
+When `max-height` is set, the layout is internally forced to `fixed`, regardless of the `table-layout` prop. In this case, columns without an explicit `width` or `min-width` will fall back to `80px` for overflow calculation. To ensure horizontal scrolling works correctly, specify `width` or `min-width` on all columns.
+:::
```
rzzf, line 244: First, I need to clarify whether this is by design or a bug.

Releted https://github.com/element-plus/element-plus/issues/6696#issuecomment-1073170684

LostElkByte: The root cause is the column width calculation path in fixed layout:

```
flattenColumns.forEach((column) => {
      if (!column.width && !column.minWidth) {
        column.realWidth = 80
      } else {
          column.realWidth = Number(column.width || column.minWidth)
      }
      bodyMinWidth += column.realWidth
})
```
Columns without width/minWidth default to 80px. If the actual content is wider but the user has not set the width, bodyMinWidth will be underestimated and scrollX will not be set to true.

I think the risk of modifying this issue is too high. it’s impossible to determine the actual width of content using JavaScript in a fixed layout, since the DOM hasn’t yet been fully rendered.

keeplearning66 (commented):
I tried making some changes to support both `table-layout="auto"` and `max-height`, and I don't see any obvious issues so far. 🤔 

```diff
const tableLayout = computed(() => {
- if (props.maxHeight) return 'fixed'
  return props.tableLayout!
})

const scrollbarStyle = computed(() => {
  if (props.height) {
    return {
      height: '100%',
    }
  }
  if (props.maxHeight) {
+   if (props.tableLayout === 'auto') {
+     return !Number.isNaN(Number(props.maxHeight))
+       ? { maxHeight: `${props.maxHeight}px` }
+       : { maxHeight: props.maxHeight }
+   }
    if (!Number.isNaN(Number(props.maxHeight))) {
```

### auto & max-height

https://github.com/user-attachments/assets/03f2dfee-fe76-4c96-ba85-fc01b73ffee2

### auto & height

https://github.com/user-attachments/assets/9496395d-7283-4a34-878b-5c6776e21e2a

PR #24496, opened 2026-06-29
fix(components): [time-picker] skip key handling during IME

mahirhir:
## What

`common/picker.vue` is the keydown handler shared by `el-time-picker` and `el-date-picker`. `handleKeydownInput` runs its esc / arrow / enter logic for every keydown, including the ones that fire while an IME composition is active.

When a CJK user presses Enter to confirm a conversion candidate, `event.code` is still `Enter` (the physical key), so the handler enters the Enter branch, calls `event.preventDefault()` / `event.stopPropagation()` and closes the panel. The Enter that was meant to commit the composition instead closes the picker in the middle of typing.

Steps to reproduce (Japanese, Chinese or any other IME):

1. Use an editable `el-date-picker` or `el-time-picker`.
2. Focus the input and start typing with the IME on, so a candidate is being composed.
3. Press Enter to confirm the candidate.

Expected: Enter confirms the IME candidate and the picker stays open.
Actual: the picker closes and `preventDefault()` is called on the composing Enter.

## How

Return early from `handleKeydownInput` when `event.isComposing` is true, before the internal esc / arrow / enter handling. This is the same guard the other editable components already use (`input-tag`, `select`, `mention`, `cascader`, `autocomplete`). The public `keydown` event is still emitted exactly as before, so only the picker's own key handling is skipped while composing.

## Test

Added a regression test in `time-picker.test.tsx`: it dispatches an `Enter` keydown with `isComposing: true` and asserts the panel stays open and the event is not default-prevented, while a normal Enter still closes the panel. Reverting the guard makes the new test fail because the composing Enter closes the panel. The `time-picker`, `date-picker` and `date-picker-panel` suites pass.

I did not find an existing issue for this, so there is no linked issue to reference.

---

- [x] Make sure you follow contributing guide [English](https://github.com/element-plus/element-plus/blob/dev/.github/CONTRIBUTING.en-US.md) | ([中文](https://github.com/element-plus/element-plus/blob/dev/.github/CONTRIBUTING.zh-CN.md) | [Español](https://github.com/element-plus/element-plus/blob/dev/.github/CONTRIBUTING.es.md) | [Français](https://github.com/element-plus/element-plus/blob/dev/.github/CONTRIBUTING.fr-FR.md)).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard handling in the time picker so Enter won’t accidentally close or confirm while IME text input is in progress.
  * Kept the picker panel open during composition, preventing unwanted interruption of non-Latin text entry.
  * Added coverage to verify Enter behaves normally once composition has ended.

### DIFF
--- a/packages/components/time-picker/src/common/picker.vue
+++ b/packages/components/time-picker/src/common/picker.vue
@@ -567,6 +567,10 @@ const handleKeydownInput = async (event: Event | KeyboardEvent) => {
 
   const code = getEventCode(event as KeyboardEvent)
   emitKeydown(event as KeyboardEvent)
+  // Skip internal key handling while an IME composition is active so that
+  // confirming/closing a candidate with Enter is not swallowed by
+  // preventDefault. Matches the guard in input-tag, select, mention, etc.
+  if ((event as KeyboardEvent).isComposing) return
   if (code === EVENT_CODE.esc) {
     if (pickerVisible.value === true) {
       pickerVisible.value = false
--- a/packages/components/time-picker/__tests__/time-picker.test.tsx
+++ b/packages/components/time-picker/__tests__/time-picker.test.tsx
@@ -297,6 +297,41 @@ describe('TimePicker', () => {
     expect(handleVisibleChange).toHaveBeenCalledTimes(3)
   })
 
+  it('does not handle Enter while an IME composition is active', async () => {
+    const handleVisibleChange = vi.fn()
+    const value = ref(new Date(2016, 9, 10, 18, 40))
+
+    const wrapper = mount(() => (
+      <TimePicker v-model={value.value} onVisibleChange={handleVisibleChange} />
+    ))
+
+    const input = wrapper.find('input')
+    input.trigger('focus')
+    await nextTick()
+    await rAF()
+    expect(handleVisibleChange).toHaveBeenCalledTimes(1)
+
+    // Pressing Enter to confirm an IME candidate must not close the panel or
+    // swallow the keystroke that commits the composition.
+    const composingEnter = new KeyboardEvent('keydown', {
+      code: EVENT_CODE.enter,
+      isComposing: true,
+      bubbles: true,
+      cancelable: true,
+    })
+    input.element.dispatchEvent(composingEnter)
+    await nextTick()
+    await rAF()
+    expect(handleVisibleChange).toHaveBeenCalledTimes(1)
+    expect(composingEnter.defaultPrevented).toBe(false)
+
+    // A normal Enter (composition already finished) still closes the panel.
+    input.trigger('keydown', { code: EVENT_CODE.enter })
+    await nextTick()
+    await rAF()
+    expect(handleVisibleChange).toHaveBeenCalledTimes(2)
+  })
+
   it('should cancel correctly the right value after manual input', async () => {
     const value = ref<[Date, Date]>([
       new Date(2016, 9, 10, 8, 40),


packages/components/time-picker/src/common/picker.vue
```diff
@@ -567,6 +567,10 @@ const handleKeydownInput = async (event: Event | KeyboardEvent) => {
 
   const code = getEventCode(event as KeyboardEvent)
   emitKeydown(event as KeyboardEvent)
+  // Skip internal key handling while an IME composition is active so that
+  // confirming/closing a candidate with Enter is not swallowed by
+  // preventDefault. Matches the guard in input-tag, select, mention, etc.
+  if ((event as KeyboardEvent).isComposing) return
```
patienry, line 573:
```suggestion

  if ((event as KeyboardEvent).isComposing) return

```

patienry: This code is fairly clear and doesn’t require comments.

rzzf:
Could you provide a specific reproduction scenario? I’m unable to reproduce the issue you mentioned on my side. https://element-plus.org/en-US/component/date-picker#enter-date


https://github.com/user-attachments/assets/d9b1cea6-c845-4cad-b1fa-9324873d5686

mahirhir:
It's browser-specific, which is probably why it doesn't reproduce on your setup. The confirming Enter only reaches the keydown handler with event.code === 'Enter' on Safari/WebKit. On Chromium and Firefox that keystroke is reported as keyCode 229 (or with isComposing already cleared), so it never enters the Enter branch. To reproduce on Safari (macOS) with a Japanese/Chinese IME: open an editable el-date-picker, type a reading (e.g. きょう) so candidates show, press Enter to confirm. Without the guard the panel closes mid-composition; with event.isComposing checked first the Enter is left to the IME. The added test simulates the Safari case. Happy to also rename the PR title to the commitlint format if that's the failing check.

rzzf:
I also can’t reproduce this in Safari with a Chinese IME. Could you provide a video? Or could you confirm whether the issue is limited to Japanese IMEs?

mahirhir:
Sorry for the 62 days of silence. That was mine to answer and I left it.

I cannot give you the video. I have no macOS and no Safari, so every Safari claim I made on 3 July came from reading, not from running it. I should have said that at the time instead of describing the keydown behaviour as if I had watched it.

So the honest state of this PR:

- The change is 4 lines in `picker.vue` and a test that simulates the composing Enter. It is not exotic; `isComposing` is already used in `input-tag`, `mention`, `select` and `message-box`, so it is not a new pattern for this codebase.
- Neither of us can reproduce it. You could not in Safari with a Chinese IME, and I cannot check Safari at all.
- I do not know whether it is Japanese-IME-specific. You asked that on 6 July and I still do not have a way to find out.

Two things I would rather flag than leave implied. WebKit bug 165004 describes the confirming Enter arriving with `isComposing` already false on Safari; if that is what happens here, this guard would not fix it, and the test passing only shows the guard works when `isComposing` is true. And the other components use tracked `compositionstart`/`compositionend` state alongside `isComposing` rather than the flag alone, which may be the more reliable shape if this is worth doing at all.

Given nobody can reproduce it and I cannot supply what you asked for, closing this is reasonable and I will not argue with it. If you would rather keep it open on the chance a Safari user turns up, that works too. Your call, and either way I should have answered in July.

rzzf:
I hope each fix has reproducible issue tracking. When someone can provide a stable reproduction scenario in the future, feel free to submit a PR anytime. Thank you for your contribution regardless.

PR #24610, opened 2026-07-24
fix(components): [progress] use theme css variables for status colors

zsqsmart:
Replace hardcoded status color hex values with theme CSS variables (with hex fallbacks) so progress status colors follow the design tokens.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated circular progress styling so status-related SVG stroke colors are derived from theme-aware CSS variables (with fallbacks), ensuring customized theme success tones render correctly.

* **Tests**
  * Added a unit test confirming that for `type="circle"` and `status="success"`, the progress circle path uses the expected CSS variable–based `stroke` value (including the fallback).


btea:
This appears to be a **breaking change**; any user with a different theme color will have their status color changed if the latest modified code is applied.

PR #24637, opened 2026-07-30
fix(components): [table-v2, carousel] correct element attributes

thlovey:

## Summary

This PR fixes invalid accessibility attributes in Table V2 and prevents Carousel indicator buttons from submitting a parent form.

It also improves related test coverage and removes several sources of test noise.

### Table V2

- Replace the invalid `arialabel` attribute with `aria-label`.
- Replace the invalid `ariaexpanded` attribute with `aria-expanded`.
- Replace the invalid `ariasort` attribute with `aria-sort`.
- Update accessibility tests to assert the standard ARIA attributes.
- Remove a leftover `console.log` from the test suite.

Previously, camel-cased ARIA property names passed through object spreads were rendered as non-standard lowercase attributes, which could not be reliably recognized by assistive technologies.

### Carousel

- Add `type="button"` to indicator buttons.
- Add a regression test covering the button type.

Without an explicit button type, clicking a Carousel indicator inside a `<form>` could trigger an unintended form submission.

### Test improvements

- Replace the ineffective Tag `disableTransitions` assertion with a test that verifies the Transition wrapper is actually disabled.
- Run `useThrottleRender` tests inside a Vue component lifecycle.
- Remove the Vue `onMounted` warnings caused by invoking the composable outside `setup()`.
- Ensure mounted test components are properly unmounted.

## Testing

The following checks passed:

- `pnpm test --run`
- `pnpm typecheck`
- `pnpm lint`
- `git diff --check`

Focused tests also passed for:

- Table V2
- Carousel
- Tag
- `useThrottleRender`

## Summary by CodeRabbit

* **Bug Fixes**
  * Carousel indicator buttons no longer trigger unintended form submissions inside forms.
  * Table V2 accessibility was improved with standards-compliant expand/collapse and sorting ARIA attributes, including accurate multi-sort behavior.

* **Tests**
  * Expanded coverage for Carousel, Table V2, Tag transition disabling, and `useThrottleRender`, including mounting and cleanup scenarios.

### DIFF
--- a/packages/components/table-v2/src/components/expand-icon.tsx
+++ b/packages/components/table-v2/src/components/expand-icon.tsx
@@ -18,8 +18,8 @@ const ExpandIcon = (
 
   const expandIconProps = {
     onClick: expandable ? () => onExpand(!expanded) : undefined,
-    ariaLabel,
-    ariaExpanded: expanded,
+    'aria-label': ariaLabel,
+    'aria-expanded': expanded,
     class: props.class,
   } as any
 
--- a/packages/components/carousel/__tests__/carousel.test.tsx
+++ b/packages/components/carousel/__tests__/carousel.test.tsx
@@ -130,6 +130,14 @@ describe('Carousel', () => {
     expect(wrapper.find('.el-carousel__button span').text()).toBe('1')
   })
 
+  it('indicator buttons should not submit a parent form', () => {
+    wrapper = createComponent({ autoplay: false })
+
+    wrapper.findAll('.el-carousel__button').forEach((button) => {
+      expect(button.attributes('type')).toBe('button')
+    })
+  })
+
   describe('manual control', () => {
     it('hover', async () => {
       vi.useFakeTimers()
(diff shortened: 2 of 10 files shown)


packages/components/table-v2/src/components/expand-icon.tsx
```diff
@@ -18,8 +18,8 @@ const ExpandIcon = (
 
   const expandIconProps = {
     onClick: expandable ? () => onExpand(!expanded) : undefined,
-    ariaLabel,
-    ariaExpanded: expanded,
+    'aria-label': ariaLabel,
```
btea, line 21: The previous syntax was valid and does not need to be modified.

thlovey: `ariaLabel` is a valid reflected DOM property in modern browsers, so it works
in recent CSR environments. However, ARIA reflection only became broadly
available in 2023, while Element Plus still supports older browser versions.

More importantly, Vue SSR currently serializes these camel-cased keys as
`arialabel` / `ariaexpanded`, which are not valid ARIA attributes. Vue also
declares native JSX ARIA attributes using their hyphenated names.

Using `aria-label` and `aria-expanded` therefore keeps the output correct
across CSR, SSR, and the documented browser support range.

PR #24679, opened 2026-08-07
fix(components): [input] expose word limit count to screen readers

### LINKED ISSUE #24674: [Component] [input] [accessibility] Missing role on input limit length
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.41`
- Element Plus Version: `2.14.3`
- Browser / OS: `Firefox / Ubuntu`
- Build Tool: `Vite`

### Related Component

- `el-input`

### Reproduction Link

[Docs](https://element-plus.org/en-US/component/input#limit-length)

### Steps to reproduce

Enter some text with vocal assistant

### What is Expected?

Remaining characters count must be read by vocal assistant

Nothing

### Additional comments

Correction: Add a role="status" attribute around the label so that screen readers read it out
whenever it is updated (+ add hidden text “characters” within the same component so that the user
knows that #/120, for example, corresponds to the number of characters)

lazerg:

Fixes #24674

The `show-word-limit` counter renders as a plain `span`, so a screen reader never picks up the character count while you type. Marking it `role="status"` turns it into a polite live region, and the updated `5 / 10` gets read out on each change. Both the input and the textarea counter are covered.

The same live region now also holds a visually hidden `characters` label, so the announcement is "5 / 10 characters" rather than two bare numbers. That needed a new `el.input.characters` key across all 67 files in `packages/locale/lang` (generated with `pnpm locale:sync`, then translated) and a `utils-visually-hidden` mixin in `theme-chalk`, since the library had no screen-reader-only rule of its own.

## Summary by CodeRabbit

* **New Features**
  * Input and textarea character counters now display localized labels across supported languages.
  * Counters show the current character count alongside the maximum limit.

* **Accessibility**
  * Character counters now provide live status updates and accessible labels with current and maximum counts.

* **Tests**
  * Added coverage for localized character-count text and accessibility announcements.

### DIFF
--- a/packages/components/input/src/input.vue
+++ b/packages/components/input/src/input.vue
@@ -99,6 +99,9 @@
               <span :class="nsInput.e('count-inner')">
                 {{ textLength }} / {{ maxlength }}
               </span>
+              <span :class="nsInput.e('count-unit')">
+                {{ t('el.input.characters') }}
+              </span>
             </span>
             <el-icon
               v-if="validateState && validateIcon && needStatusIcon"
@@ -174,6 +177,9 @@
         role="status"
       >
         {{ textLength }} / {{ maxlength }}
+        <span :class="nsInput.e('count-unit')">
+          {{ t('el.input.characters') }}
+        </span>
       </span>
     </template>
   </div>
@@ -216,6 +222,7 @@ import {
   useComposition,
   useCursor,
   useFocusController,
+  useLocale,
   useNamespace,
 } from '@element-plus/hooks'
 import {
@@ -275,6 +282,7 @@ const inputSize = useFormSize()
 const inputDisabled = useFormDisabled()
 const nsInput = useNamespace('input')
 const nsTextarea = useNamespace('textarea')
+const { t } = useLocale()
 
 const input = shallowRef<HTMLInputElement>()
 const textarea = shallowRef<HTMLTextAreaElement>()
--- a/packages/components/input/src/input.vue
+++ b/packages/components/input/src/input.vue
@@ -94,13 +94,11 @@
                 nsInput.e('count'),
                 nsInput.is('outside', wordLimitPosition === 'outside'),
               ]"
+              :aria-label="wordLimitLabel"
               role="status"
             >
               <span :class="nsInput.e('count-inner')">
                 {{ textLength }} / {{ maxlength }}
-              </span>
-              <span :class="nsInput.e('count-unit')">
-                {{ t('el.input.characters') }}
               </span>
             </span>
             <el-icon
@@ -174,12 +172,10 @@
           nsInput.e('count'),
           nsInput.is('outside', wordLimitPosition === 'outside'),
         ]"
+        :aria-label="wordLimitLabel"
         role="status"
       >
         {{ textLength }} / {{ maxlength }}
-        <span :class="nsInput.e('count-unit')">
-          {{ t('el.input.characters') }}
-        </span>
       </span>
     </template>
   </div>
@@ -383,6 +379,12 @@ const textLength = computed(() => {
   }
   return nativeInputValue.value.length
 })
+const wordLimitLabel = computed(() =>
+  t('el.input.characters', {
+    count: textLength.value,
+    max: maxlength.value ?? '',
+  })
+)
 const inputExceed = computed(
   () =>
     // show exceed style if length of initial value greater then maxlength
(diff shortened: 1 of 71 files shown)


packages/components/input/src/input.vue
```diff
@@ -90,6 +90,7 @@
             </el-icon>
             <span
               v-if="isWordLimitVisible"
+              role="status"
```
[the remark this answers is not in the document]
lazerg: Left out on purpose, not an oversight.

Element Plus ships no screen-reader-only class. The only `.visually-hidden` rule lives in the docs site (`docs/.vitepress/vitepress/styles/base.scss`), not in `theme-chalk`, so hidden text here needs a new public utility class in the library itself.

The word cannot be hardcoded either. Every user-facing string in `packages/components` goes through `t()`, and `translate()` falls back to the literal key path when a key is missing. A new key therefore has to land in all 67 files under `packages/locale/lang`.

Both of those are much larger than the role attribute, so I kept them out and left #24674 open for the unit text. Happy to add it here if a maintainer prefers one PR.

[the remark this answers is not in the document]
lazerg: Same answer as on the thread above. The library has no screen-reader-only class of its own, and the word cannot be hardcoded because every user-facing string goes through `t()`, which needs the key in all 67 locale files. Both are out of scope here, so #24674 stays open for the unit text.

packages/components/input/src/input.vue
```diff
@@ -94,6 +94,7 @@
                 nsInput.e('count'),
                 nsInput.is('outside', wordLimitPosition === 'outside'),
               ]"
+              role="status"
```
btea, line 98: A single role doesn't seem very useful; perhaps adding `aria-label` together would be better.

lazerg: Same issue CodeRabbit/Codex flagged above: `aria-label` is user-facing text, and `packages/components` routes everything through `t()`, so it needs a key in all 67 `packages/locale/lang` files. No screen-reader-only class exists outside the docs site either, so a hidden label isn't a shortcut. Kept this PR to `role="status"` and left #24674 open for that. Can add the label and locale entries here if you'd rather it be one PR.

btea: The features and issues are basically divided by components. Therefore, regarding the a11y issue, I also hope that each component can be handled separately. The text labels added here are for the input component, so I think that adding the label text at the same time would make it a complete feature.

lazerg: Added it. The counter now carries a visually hidden `characters` label inside the same `role="status"` span, so the announcement reads "5 / 10 characters"

Two supporting pieces were needed: a new `el.input.characters` key in all 67 `packages/locale/lang` files (generated with `pnpm locale:sync`, then translated, `mg.ts` still carries the `to be translated` marker), and a `utils-visually-hidden` mixin in `theme-chalk` behind `.el-input__count-unit`, since the library had no screen-reader-only rule of its own.

The two textarea inline snapshots that asserted the counter text were updated for the added word.

packages/locale/lang/mg.ts
```diff
@@ -86,6 +86,9 @@ export default {
         dec: 'Des',
       },
     },
+    input: {
+      characters: 'characters', // to be translated
+    },
```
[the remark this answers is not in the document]
lazerg: This follows the file's existing convention, not something new: 53 of the 205 lines in `mg.ts` already fall back to English with the same `// to be translated` marker on keys nobody's translated to Malagasy yet. I added `characters` the same way rather than guessing a translation myself. Real translations do exist for the other locales I could verify (de, fr, ja, es).

packages/theme-chalk/src/input.scss
```diff
@@ -331,7 +331,7 @@
       align-items: center;
       justify-content: center;
 
-      @if $slot == prefix {
+      @if $slot ==prefix {
```
btea: The result of the automatic formatting looks a bit strange, but it seems like it can't be changed back.

btea (approved):

packages/theme-chalk/src/input.scss
lazerg: It was safe to change back. I tested it: `prettier --check` and `prettier --write` both keep `$slot == prefix`, and eslint has no config for `.scss`, so lint-staged leaves the line alone. So the repo tooling did not remove that space, some other formatter did. I pushed the space back.

btea: Well done.  👍 

It seems there's something wrong with the local script configuration.

rzzf (approved):

PR #24715, opened 2026-08-12
fix(components): [tree-select] reset checked keys when data changes

ValentinYoushkevich:

With `show-checkbox`, clearing `v-model` and `data` in the same tick and then putting the same data back leaves every node checked while the model is empty (#23207). `ElTree`'s store remembers the last keys handed to `setCheckedKeys` as its `defaultCheckedKeys` and re-applies them from `setData` whenever the data instance changes. When both are cleared at once, the tree-select watcher runs while the tree is already empty, sees `getCheckedKeys()` and `modelValue` both equal to `[]`, and skips the `setCheckedKeys` call — so the stale keys survive in the store, and since `modelValue` never changes again, nothing re-syncs the tree once the data returns.

The watcher in `useTree` only tracked `modelValue` and the tree ref; this adds `data` as a source, so the tree is re-synced with `modelValue` on every data change too. Everything else, including the `isEqual` guard, is unchanged, so the extra work is a no-op whenever the tree already matches the model. Covered by a new test in `tree-select.test.tsx` that reproduces the issue's sequence and asserts both `getCheckedKeys()` and the rendered checkboxes end up empty; it fails on the unpatched source with `expected [ '1-1' ] to deeply equal []`.

closes #23207

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tree selection synchronization when data is cleared and later restored.
  * Preserved checked node states after in-place edits or full data refreshes.
  * Preserved programmatically checked nodes during tree updates.
  * Prevented selected states from being lost when the bound value is empty.

* **Tests**
  * Added regression coverage for clearing, restoring, refreshing, and editing tree data without incorrect selection states.

### DIFF
--- a/packages/components/tree-select/__tests__/tree-select.test.tsx
+++ b/packages/components/tree-select/__tests__/tree-select.test.tsx
@@ -1088,6 +1088,332 @@ describe('TreeSelect.vue', () => {
     expect(spy2).toBeCalledWith(2)
   })
 
+  test('reset checked keys when data is restored after clearing both data and v-model', async () => {
+    const treeData = [
+      {
+        value: '1',
+        label: '1',
+        children: [
+          { value: '1-1', label: '1-1' },
+          { value: '1-2', label: '1-2' },
+        ],
+      },
+    ]
+    const modelValue = ref<string[]>(['1-1'])
+    const data = ref<any[]>(treeData)
+
+    const { tree } = createComponent({
+      props: {
+        modelValue,
+        data,
+        multiple: true,
+        showCheckbox: true,
+        defaultExpandAll: true,
+      },
+    })
+
+    await nextTick()
+    expect(tree.vm.getCheckedKeys()).toEqual(['1-1'])
+
+    // clear the model value and the data in the same tick
+    modelValue.value = []
+    data.value = []
+    await nextTick()
+    await nextTick()
+
+    // then restore the same data
+    data.value = treeData
+    await nextTick()
+    await nextTick()
+    await nextTick()
+
+    expect(tree.vm.getCheckedKeys()).toEqual([])
+    expect(tree.findAll('.el-checkbox.is-checked')).toHaveLength(0)
+    expect(modelValue.value).toEqual([])
+  })
+
+  test('editing data in place keeps imperatively checked keys', async () => {
+    const treeData = [
+      {
+        value: '1',
+        label: '1',
+        children: [
+          { value: '1-1', label: '1-1' },
+          { value: '1-2', label: '1-2' },
+        ],
+      },
+    ]
+    const modelValue = ref<string[]>([])
+    const data = ref<any[]>(treeData)
+
+    const { getWrapperRef, tree } = createComponent({
+      props: {
+        modelValue,
+        data,
+        multiple: true,
+        showCheckbox: true,
+        defaultExpandAll: true,
+      },
+    })
+
+    const wrapperRef = await getWrapperRef()
+    wrapperRef.setCheckedKeys(['1-2'])
+    await nextTick()
+    expect(tree.vm.getCheckedKeys()).toEqual(['1-2'])
+
+    // editing the nodes in place must not reconcile the tree back to `modelValue`
+    data.value[0].children[0].label = 'renamed'
+    await nextTick()
+    await nextTick()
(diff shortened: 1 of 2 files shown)


ValentinYoushkevich:
Thanks — the Codex point about `deep: true` was right, and it turned out to be reproducible: with `data` as a source of the deep watcher, editing a node label in place reconciled the tree back to `modelValue` and wiped keys set through the exposed `setCheckedKeys`.

So the sync body is now a small function used by two watchers: the original deep one on `modelValue` and the tree ref, plus a separate identity-only watcher on `data`. That is enough for this bug, because the store only re-applies its retained keys when the data instance is replaced, and in-place edits no longer touch imperative state. Added a test for exactly that case — it fails on the previous version of this patch with `expected [] to deeply equal [ '1-2' ]` — and the `modelValue` assertion CodeRabbit asked for. 41 tests in the tree-select suite pass, typecheck and lint are clean.

ValentinYoushkevich:
Right again, and this one I could reproduce: on `dev`, calling the exposed `setCheckedKeys(['1-2'])` and then replacing `data` with a refreshed array keeps the checks, because the store re-applies its own retained keys; with the previous revision my data watcher then reconciled them away to `[]`.

The resync is now gated. tree-select remembers the keys it last wrote into the tree itself, and on a data change it only re-asserts when what the store restored is exactly that — otherwise the checked state came from somewhere else and is left alone. That still fixes the reported case, where the store restores the keys tree-select wrote before the model was emptied, and it also covers the case where the user checked a node through the UI before the refresh. Two tests now pin the imperative behaviour, in-place edits and a full data refresh, and both pass on unpatched `dev` too, so they describe existing behaviour rather than inventing it. 42 tests in the tree-select suite, 400 across tree-select, tree and select, typecheck and lint clean.

ValentinYoushkevich:
Confirmed and fixed: `ownCheckedKeys` now stores what the tree reports back after the write, so the cascaded parent is included and the gate matches. Added a test for the all-children-checked case; it fails on `dev` like the first one. 43 tests in the suite, 401 across tree-select, tree and select.

ValentinYoushkevich:
Both imperative tests now go through the exposed `TreeSelectInstance.setCheckedKeys` instead of the inner tree, and assert `modelValue` stays empty. 43 tests in the suite, 401 across tree-select, tree and select.

ValentinYoushkevich:
Confirmed and fixed — that path was the common one and the patch missed it. `ownCheckedKeys` is now recorded after every reconcile, not only when a write happened, and skipped while the tree has no nodes so an emptied `data` cannot erase the snapshot. Added a test that checks a node through the checkbox UI before the clear; it fails on `dev` like the other two. I also checked `check-strictly`, single mode and a data-only clear by hand — all behave. 44 tests in the suite, 402 across tree-select, tree and select.

ValentinYoushkevich:
Confirmed, and it made me rethink the approach rather than patch it again. Comparing key arrays to decide ownership was fragile by construction — order, cascaded parents and partial data all break it.

The patch no longer watches `data` at all. It just stops trusting `getCheckedKeys()` while the tree has no nodes: an empty tree always reports nothing checked, so the equality guard silently skipped the write and left the store holding the previous keys, which is what came back with the data. Now the model keys are written in that case, the store stays in step and nothing is restored later. Two lines in the end, and imperative `setCheckedKeys` is untouched because data changes are never acted on.

Swept the cases by hand before pushing: same data, reordered siblings, partial restore, cascading parents, check-strictly, single mode, data-only clear, and imperative checks across a refresh. Four tests in the suite now, all failing on `dev`; 45 in tree-select, 403 across tree-select, tree and select.

ValentinYoushkevich:
Confirmed — that one was a real regression. The empty-data write is gone; instead a `modelValue` change that lands while the tree has no nodes is remembered and applied once the data is back, so `default-checked-keys` is untouched. Added a test for it. 46 tests in the suite, 404 across tree-select, tree and select.

ValentinYoushkevich:
Confirmed. The flag is gone — the data watcher now compares `modelValue` against the last value actually synced into the tree, so a value-equivalent reset is a no-op and `default-checked-keys` survives. Test added. 47 tests in the suite, 405 across tree-select, tree and select.

PR #24716, opened 2026-08-12
fix(components): [dropdown, menu] don't trigger :focus-visible on hover

ValentinYoushkevich:

Hovering an `el-dropdown` trigger or an `el-sub-menu` programmatically calls `.focus()` on the hovered element so keyboard navigation can continue from there, and the browser then paints the `:focus-visible` ring even though the focus was pointer-initiated (#23253). It shows up all the more often because `focusElement` temporarily sets `tabindex="-1"` on elements that are not focusable on their own. This passes `focusVisible: false` at those two hover call sites. It is a standard `FocusOptions` member (Firefox 104+, Chromium 145+, and the reporter confirmed Chromium shipped it in 145), and browsers without support ignore the unknown key, so it degrades safely.

One thing worth a maintainer's opinion: I kept this at the two pointer-driven call sites rather than making `focusVisible: false` the default inside `focusElement` itself. There is a decent argument for the other way round — `focusElement` is what adds the temporary `tabindex="-1"` that provokes the ring, so suppressing it in the same place would also cover any future hover-driven caller. The reason I did not is that `focusElement` is used on keyboard paths too (`focus-trap`, restoring focus in tooltip content), where the ring must be drawn, and those would then have to ask for `focusVisible: true` explicitly, which inverts the responsibility. The helper keeps forwarding options verbatim, and a test in `aria.test.ts` pins that neutral contract so it fails if the default is ever changed. Happy to flip it if you prefer.

closes #23253

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pointer-initiated focus behavior in submenus and tooltips.
  * Prevented keyboard-style focus rings from appearing when elements receive focus through mouse interactions.
  * Preserved scroll position when focusing interactive elements.

* **Tests**
  * Added coverage verifying focus behavior and option handling for menus, tooltips, and accessibility utilities.

### DIFF
--- a/packages/components/menu/__tests__/menu.test.ts
+++ b/packages/components/menu/__tests__/menu.test.ts
@@ -304,6 +304,33 @@ describe('submenu', () => {
     await nextTick()
     expect(submenu.classes()).toContain('is-opened')
   })
+  test('should not trigger :focus-visible when focusing on mouseenter', async () => {
+    const wrapper = _mount(
+      `<el-menu mode="horizontal">
+        <el-sub-menu index="1" ref="submenu">
+          <template #title>导航一</template>
+          <el-menu-item index="1-1">选项1</el-menu-item>
+        </el-sub-menu>
+      </el-menu>`
+    )
+    await nextTick()
+    const submenu = wrapper.findComponent({ ref: 'submenu' })
+    const focusSpy = vi.spyOn(submenu.element as HTMLElement, 'focus')
+
+    vi.useFakeTimers()
+    await submenu.trigger('mouseenter')
+    vi.runAllTimers()
+    vi.useRealTimers()
+    await rAF()
+
+    // pointer-initiated focus must not paint the keyboard focus ring
+    expect(focusSpy).toHaveBeenCalledTimes(1)
+    expect(focusSpy).toHaveBeenCalledWith({
+      preventScroll: true,
+      focusVisible: false,
+    })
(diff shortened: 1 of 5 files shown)


ValentinYoushkevich:
Reworked both hover tests as CodeRabbit suggested: the spies now sit on the actual hovered element (`submenu.element` / the trigger element) instead of `HTMLElement.prototype`, and each asserts exactly one call with `{ preventScroll: true, focusVisible: false }`. They still fail when the two source changes are reverted, so they are pinning the real behaviour. 60 tests across menu, tooltip and the aria utils pass, typecheck and lint clean.

On the Safari point I would rather not add a fallback, and I would like a maintainer's view before doing anything. `focusVisible` is Safari 18.4+, so on 14.1-18.3 the option is simply ignored and those versions behave exactly as they do today — this is a progressive enhancement, not a regression for them. The only fallbacks I can think of mean suppressing the ring ourselves, either by overriding `:focus-visible` styling for these elements, which also hides it for genuine keyboard focus, or by tracking pointer-versus-keyboard recency globally. Both are considerably more invasive than the problem, and the second is essentially reimplementing the heuristic the browser already owns. Happy to look at it if you disagree.

patienry (commented):
Thank you! This still seems to be an issue (Chrome 151.0.7922.170).

https://github.com/user-attachments/assets/42f01154-e8eb-46fb-9e2c-191465e7200e

PR #24740, opened 2026-08-16
fix(components): [message, notification] call nextZIndex eagerly to prevent z-index conflicts with overlays

### LINKED ISSUE #24738: [Component] [message, drawer] 在 Nuxt 4 中ElMessage 在 el-drawer 中首次显示被遮罩层遮挡
Bug Type: **`Component`**

## Environment

- Vue Version: `3.5.4`
- Element Plus Version: `2.14.4`
- Browser / OS: `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36`
- Build Tool: `Nuxt`

### Related Component

- `el-message`
- `el-drawer`

### Reproduction Link

[Github Repo](https://github.com/wou101771/element-nuxt.git)

### Steps to reproduce

1. 在 Nuxt 4 项目中引入 Element Plus
2. 页面中放置一个 `el-drawer`，设置 `append-to-body`
3. Drawer 内部放一个按钮，点击调用 `ElMessage.success()`
4. 打开 Drawer，点击按钮
5. ❌ 首次消息被遮罩层遮挡，不可见
6. 再次点击按钮
7. ✅ 消息正常显示在顶层

### What is Expected?

每次调用 ElMessage 都应显示在页面最顶层，不被任何遮罩层遮挡。

首次调用时，消息被 el-drawer 的遮罩层遮挡，用户无法看到消息内容。

### Additional comments

_(empty)_

Image

waterWang:
## Summary

Fixes #24738 — `ElMessage` (and `ElNotification`) are covered by the `el-drawer` overlay mask on their **first** call.

## Root cause

In `message.vue` / `notification.vue`, `nextZIndex()` is called inside `onMounted`, but the rendered style (`zIndex`) is computed **before** that:

```ts
const customStyle = computed(() => ({
  zIndex: currentZIndex.value,  // evaluated at first render
}))

onMounted(() => {
  startTimer()
  nextZIndex()  // too late — first paint already happened
  visible.value = true
})
```

On the first call, `currentZIndex` is still the initial value (`2000`), which is **lower** than the z-index already claimed by the `el-drawer` overlay (drawer claims its z-index via `useDialog` → `nextZIndex()` **eagerly** when it opens). As a result the first message is painted underneath the overlay. The second call works because `zIndex` was already incremented by the first `onMounted`.

## Fix

Move `nextZIndex()` to the setup body (before any computed is evaluated), and remove it from `onMounted`. This guarantees the very first render carries the correct, already-incremented z-index.

## Changes

- `packages/components/message/src/message.vue` — call `nextZIndex()` eagerly in setup; remove from `onMounted`
- `packages/components/notification/src/notification.vue` — same fix (identical pattern, same bug)

## Test plan

1. In a Nuxt (or plain Vue) app, open an `el-drawer` with `append-to-body`
2. Click a button inside that calls `ElMessage.success()` — the message must appear **above** the drawer overlay on the first call
3. Repeat — still correct
4. Same for `ElNotification`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message and notification layering during initial display.
  * Prevented visual stacking issues by assigning the correct z-index before components become visible.
  * Preserved notification timing and visibility behavior.

### DIFF
--- a/packages/components/message/src/message.vue
+++ b/packages/components/message/src/message.vue
@@ -90,6 +90,11 @@ const isStartTransition = ref(false)
 const { ns, zIndex } = useGlobalComponentSettings('message')
 const { currentZIndex, nextZIndex } = zIndex
 
+// Call nextZIndex eagerly so the first render gets the correct z-index,
+// avoiding the message being covered by overlay elements (e.g. el-drawer
+// with appendToBody) on the first call.
+nextZIndex()
+
 const messageRef = ref<HTMLDivElement>()
 const visible = ref(false)
 const height = ref(0)
@@ -166,7 +171,6 @@ function keydown(event: KeyboardEvent) {
 
 onMounted(() => {
   startTimer()
-  nextZIndex()
   visible.value = true
 })
 
--- a/packages/components/notification/src/notification.vue
+++ b/packages/components/notification/src/notification.vue
@@ -79,6 +79,10 @@ defineEmits(notificationEmits)
 const { ns, zIndex } = useGlobalComponentSettings('notification')
 const { nextZIndex, currentZIndex } = zIndex
 
+// Call nextZIndex eagerly so the first render gets the correct z-index,
+// avoiding the notification being covered by overlay elements on the first call.
+nextZIndex()
+
 const visible = ref(false)
 let timer: (() => void) | undefined = undefined
 const percentage = ref(100)
@@ -219,7 +223,6 @@ function onKeydown(event: KeyboardEvent) {
 // lifecycle
 onMounted(() => {
   startTimer()
-  nextZIndex()
   visible.value = true
 })
 


lazerg:
Thanks for taking this on. I don't think this fixes the reported Nuxt case, though.

`@element-plus/nuxt` explicitly provides `ZINDEX_INJECTION_KEY` on the app (`src/core/injection.ts`), so `el-drawer`'s `useZIndex()` reads from that injected counter. `ElMessage`/`ElNotification` are mounted via `render()` with no app context, since the Nuxt module registers components individually instead of through `app.use(ElementPlus)`, so they fall back to the module-scoped `initial` object, a separate counter the drawer never touches. Moving `nextZIndex()` earlier in setup changes when the fallback counter increments, not which counter it uses, so the message can still land under the mask depending on call order.

There's also a side effect: calling `nextZIndex()` directly in setup now runs during SSR too, since `setup()` executes server-side while `onMounted` doesn't. That mutates the shared module-level `zIndex` ref across requests, which is likely why it was in `onMounted` to begin with.

I opened #24741, which realigns the fallback counter to the injected one inside `nextZIndex()` itself, guarded by `isClient`, with a test that reproduces the divergence.

lazerg:
@rzzf @thinkasany flagging this account for a look, separate from the technical point above. [waterWang](https://github.com/waterWang) has opened `1925` PRs across GitHub, dozens today alone across large, unrelated projects (`saltstack`, `ant-design`, `moby`, `grafana`, `TanStack`, `radix-ui`, `envoy`, and more), including several duplicate PRs on the same repos opened and closed minutes apart. The profile bio references an airdrop wallet address (`@OSSDAO-ORG•AIRDROP-0x3570bcfcb556F68E4452498C8793E78a3d549EC6`), which matches known contribution-farming patterns where PR count is farmed for token rewards rather than real fixes. Worth checking whether this account has other open PRs here.

PR #24837, opened 2026-09-06
ci: pin actions-awesome/pr-helper version

thinkasany:
https://www.github.com/actions-awesome/pr-helper/tree/48f7641b8b2feebab7d6a2fbc22812fd0bfee829

image

image

## Summary by CodeRabbit

* **Chores**
  * Updated workflow version annotations while preserving the existing pinned revisions and behavior.

### DIFF
--- a/.github/workflows/pr-conflict.yml
+++ b/.github/workflows/pr-conflict.yml
@@ -17,7 +17,7 @@ jobs:
         if: |
           contains(github.event.pull_request.labels.*.name, 'conflict pending') &&
           github.repository_owner == 'element-plus'
-        uses: actions-awesome/pr-helper@48f7641b8b2feebab7d6a2fbc22812fd0bfee829 # v1.1.1
+        uses: actions-awesome/pr-helper@48f7641b8b2feebab7d6a2fbc22812fd0bfee829 # 1.1.1
         with:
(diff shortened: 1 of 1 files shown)


rzzf:
This repository seems to be no longer maintained, can we migrate to element-plus? @btea

btea:
I think it's fine, but I guess we'll have to communicate with the author.

https://github.com/element-plus/element-plus/pull/24791. It seems that the migration was from `actions-cool/maintain-one-comment-backup` to `actions-awesome/pr-helper`?What is the difference between these two?

rzzf:
> #24791. It seems that the migration was from `actions-cool/maintain-one-comment-backup` to `actions-awesome/pr-helper`?What is the difference between these two?

maintain-one-comment-backup can only operate on comments, but we need to delete both the labels and the outdated comments at the same time

Switching to actions-awesome/pr-helper is the most concise method.

btea:
Oh, I see. 

`actions-cool/maintain-one-comment-backup` also seems to support deleting comments.
https://github.com/actions-cool/maintain-one-comment-backup#:~:text=Will%20delete%20all%20filter%20comments.

rzzf:
Of course, but this requires two steps

PR #24860, opened 2026-09-10
fix(components): prevent duplicate keyboard handling in dates picker

ixyzorg:

## What is fixed

Fixes an error triggered when pressing `Enter` or `Space` on a date cell in a DatePicker with `type="dates"`.

The keyboard event previously bubbled to the parent panel handler, which did not handle the multiple-date selection correctly.

## Changes

- Handle `Enter` and `Space` directly on date cells in `dates` mode.
- Prevent the handled keyboard event from propagating to the parent panel.
- Add a regression test covering keyboard selection and deselection.

## Steps to reproduce

1. Render a DatePicker with `type="dates"`.
2. Open the picker panel.
3. Click a date to select it and focus the date cell.
4. Press `Enter` or `Space`.
5. An error is thrown because the keyboard event bubbles to the parent panel handler.

## Implementation

The fix follows the existing keyboard interaction behavior of the
`quarters` picker.

`Enter` and `Space` are handled directly by the focused date cell in
`dates` mode. After the date is selected or deselected, the event's
default behavior and propagation are prevented so that it is not
processed again by the parent panel.

https://github.com/user-attachments/assets/057f1f77-434c-4c4d-8c9e-def9ad3a773d

## Summary by CodeRabbit

- **New Features**
  - Date pickers in dates-selection mode now support selecting and deselecting dates using the Enter and Space keys.
  - Keyboard interactions prevent unintended browser behavior and integrate with existing date selection.

- **Tests**
  - Added coverage verifying that Enter selects a date and Space toggles it off.

### DIFF
--- a/packages/components/date-picker-panel/src/composables/use-basic-date-table.ts
+++ b/packages/components/date-picker-panel/src/composables/use-basic-date-table.ts
@@ -326,7 +326,7 @@ export const useBasicDateTable = (
   }
 
   const handlePickDate = (
-    event: FocusEvent | MouseEvent,
+    event: FocusEvent | KeyboardEvent | MouseEvent,
     isKeyboardMovement = false
   ) => {
     if (props.disabled) return
@@ -387,6 +387,12 @@ export const useBasicDateTable = (
     return false
   }
 
+  const handleKeydown = (event: KeyboardEvent) => {
+    if (props.selectionMode !== 'dates') return
+    event.preventDefault()
+    event.stopPropagation()
+    handlePickDate(event)
+  }
   return {
     WEEKS,
     rows,
@@ -405,6 +411,7 @@ export const useBasicDateTable = (
     handleMouseDown,
     handleMouseMove,
     handleFocus,
+    handleKeydown,
   }
 }
 
--- a/packages/components/date-picker-panel/src/date-picker-com/basic-date-table.vue
+++ b/packages/components/date-picker-panel/src/date-picker-com/basic-date-table.vue
@@ -42,6 +42,8 @@
           :tabindex="disabled ? undefined : isSelectedCell(cell) ? 0 : -1"
           :aria-disabled="disabled"
           @focus="handleFocus"
+          @keydown.enter="handleKeydown"
+          @keydown.space="handleKeydown"
         >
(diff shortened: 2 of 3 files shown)


btea (approved):

thinkasany (approved):

packages/components/date-picker-panel/src/composables/use-basic-date-table.ts
```diff
@@ -387,6 +387,12 @@ export const useBasicDateTable = (
     return false
   }
 
+  const handleKeydown = (event: KeyboardEvent) => {
+    if (props.selectionMode !== 'dates') return
+    event.preventDefault()
+    event.stopPropagation()
+    handlePickDate(event)
```
patienry, line 394: The multi-select picker does not support full keyboard navigation. With this change, pressing `Enter` or `Space` would trigger an action, which seems to be conflicting behavior. Keyboard navigation that doesn't allow users to navigate between dates with the arrow keys doesn't seem particularly useful either.

ixyzorg: Good morning! Imagine accidentally pressing the Space or Enter key when the component’s type is set to dates, causing a bug—obviously, that’s not the behavior we want.
Based on the behavior of type="quarters", I’ve fixed the issue. Also, type="years" and type="months" behave differently.

patienry: I think stopping propagation should be enough here; no need to call `handlePickDate`.

btea: The `dates` panel appears to behave differently from `years` and `months`, while the newly added `quaters` panel behaves identically to the latest `dates` panel.

`dates`/`quaters`: Pressing the key will deselect the selected dates and select the unselected dates.

`years`/`months`: Pressing the key will close the panel.

Perhaps we should unify the alignment to `years`/`months`?

ixyzorg: I previously opened an issue about this inconsistency ( Related to #24862 ) . If we agree that the years/months behavior should be the standard, I’ll open a new PR to align the dates/quarters panels.

